PR #7, opened 2015-10-16
Refactor EntityDictionary to support multiple DatabaseManagers

wcekan:


elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -365,28 +342,11 @@ public void bindEntity(Class<?> cls) {
             log.error("Duplicate binding {} for {}, {}", type, cls, duplicate);
             throw new DuplicateMappingException(type.toString() + " " + cls.getName() + ":" + duplicate.getName());
         }
-        bindEntityToJsonApi.put(cls, type);
+
+        entityBindings.putIfAbsent(lookupEntityClass(cls), new EntityBinding(cls, type, databaseManager));
```
DeathByTape: Should check for duplicates. Fail iff `cls` exists and `databaseManager` is the same. This way if a DB manager (i.e. hibernate3) sends us the same mapping multiple times, we ignore it. However, if you've mapped this bean twice you then get a very loud notice that this occurred.

DeathByTape:
Other than that, lgtm.

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -99,7 +72,17 @@ public EntityDictionary() {
...
+        return entity(entityClass).jsonApi;
+    }
+
+    /**
+     * Get the Database Manager for an entity
+     *
+     * @param entityClass entity name
+     * @return List of relationship names for entity
+     */
+    public DatabaseManager getDatabaseManager(Class<?> entityClass) {
```
aklish: Do we really want different DatabaseManagers per entity?  How would we manage transactions across them?  It seems like we would want an abstraction within a single manager.

DeathByTape: I didn't think about this: good point.

I suspect we would face a similar issue within a DB manager, though. You would likely need to have a set of nested transactions and a method to rollback previously committed transactions (since someone would _have_ to commit first). For instance, some pseudo code that describes what I'm thinking:

```java
// Initial call like: 
boolean openTransactions(List<DatabaseManager> dbmanagers, List<DatabaseTransaction> txs, int level) {
    if (level == dbmanagers.size()) {
        // Do some work within the transactions
        // Also need a good way to pass the transaction types into the system.
        // Return true assuming successful action
        return true;
    }
    try (DatabaseTransaction tx = dbmanagers.get(level).getTransaction()) {
        txs.add(tx);
        if (!openTransactions(dbmanagers, txs, level + 1)) {
            tx.close(); // This is currently how this works (close before commit should rollback if necessary)
            return false; // But we can short-circuit the exception logic below by returning early
        }
        tx.commit(); // Close before commit if any other level failed
    } catch (Exception e) {
        // Here is where we would not to 
        for (int i = level + 1 ; i < dbmanagers.size() ; ++i) {
            someTxId = ...; // Just a suggestion, but we need some sort of identifier for this transaction across managers
            dbmanagers.get(i).rollbackTransaction(someTxId);
        }
        return false; // Transaction closed, error'd, etc. before commit was successful: don't let anyone else commit
    }
    return true;
}
```

This is some off-the-cuff pseudo-code, but I think we can actually solve this generally because we would still have to be managed if you encapsulated the logic in a single DB manager.

DeathByTape: Another thought: this could still lead to moments of inconsistency (which we may not be able to solve at this level of the problem). Another potential solution is a macro-level three-phase commit where we somehow can "check" that we believe everyone can commit.

wcekan: The EntityManager will implement a DatabaseManager interface that wraps multiple subordinate DatabaseManager into a single transaction.  However, rolling back a partial commit will be difficult.
Assuming there are 2 databases:  DB1 and DB2.  A commit on the global transaction would do
* Commit DB1
* Commit DB2 - fails
* Rollback DB2
* Reverse commit on DB1

If the reverse commit on DB1 fails, we would be left in an inconsistent state.

wcekan:
** Rebased on master **

wcekan:
Will refactor to a separate MulitplexDatabaseManager

PR #9, opened 2015-10-17
Added jackson databinder to in memory db.

DeathByTape:
Database managers should include the appropriate data binders for jackson. In the case of in-memory DB, nothing special is required. However, for other DB managers such as hibernate, there are special data binders.


clayreimann:
:+1:

yahoocla:
CLA is valid!

PR #10, opened 2015-10-17
Added instructions on adding elide to pom.xml.

DeathByTape:
We should wait until our bintray has been integrated to jcenter before merging this in (so that it is mirrored on maven central). However, I have confirmed these artifacts work by pointing to our bintray repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,29 @@
 
 #Elide
 
+## Using Elide
+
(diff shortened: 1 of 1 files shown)


shadanan:
:+1:

yahoocla:
CLA is valid!

PR #12, opened 2015-10-18
Add support for Hibernate5

shadanan:
Tests for Hibernate5 and Hibernate3, and examples, forthcoming in subsequent PRs


elide-dbmanager/elide-dbmanager-hibernate5/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+    <artifactId>elide-dbmanager-hibernate5</artifactId>
+    <packaging>jar</packaging>
+    <name>Elide Hibernate5 Database Manager</name>
+    <description>Elide database manager for Hibernate5 support</description>
+    <parent>
+        <groupId>com.yahoo.elide</groupId>
+        <artifactId>elide-dbmanager-parent-pom</artifactId>
+        <version>1.0.0.1-SNAPSHOT</version>
+    </parent>
+
```
DeathByTape: Will need to add some fields here for maven central compliance. See: https://github.com/yahoo/elide/pull/13

You can probably just copy all of the blocks directly. Namely, you need the following:

- [ ] JavaDoc and Sources (just include plugins in section)
- [X] Signed artifacts (this is done for us by bintray)
- [ ] Add top-level URL
- [ ] Add license block
- [ ] Add developer block
- [ ] Add scm block

elide-dbmanager/elide-dbmanager-hibernate5/src/main/java/com/yahoo/elide/dbmanagers/hibernate5/HibernateManager.java
```diff
@@ -0,0 +1,293 @@
...
+        private boolean inUse = false;
+        private boolean hasNext;
+
+        public ScrollableIterator(ScrollableResults scroll) {
+            this.scroll = scroll;
+        }
+
+        @Override
+        public Iterator<T> iterator() {
+            if (inUse) {
```
DeathByTape: This section concerns me a little bit (probably the same as hibernate3)? What happens if we have two threads `A` and `B` with a simultaneous access to iterator. That is, consider if threads A and B concurrently execute lines 64 - 68.

In such a scenario we wouldn't detect the concurrent modification. It feels like the `inUse` variable is a critical section to me.

elide-dbmanager/elide-dbmanager-hibernate5/src/test/java/com/yahoo/elide/dbmanagers/hibernate5/HibernateManagerTest.java
```diff
@@ -0,0 +1,13 @@
...
+
+import static org.testng.Assert.*;
+
+public class HibernateManagerTest {
+    // TODO: Create tests for our manager
+    @Test
+    public void todo() {
+        System.err.println("Still need to write tests for hibernate3 manager!");
+    }
+}
```
DeathByTape: new line. boo intellij always misses this with tests :-1:

elide-dbmanager/elide-dbmanager-hibernate5/src/main/java/com/yahoo/elide/dbmanagers/hibernate5/HibernateManager.java
shadanan: Yeah. This file is copied and pasted from Hibernate3. I just modified it to work with Hibernate5.

DeathByTape: Just a follow-up: after talking in person, I think Shad has a good point here. Perhaps-- for performance-- we can simply throw the exception iff we detect the issue (i.e. this shouldn't be used concurrently), but we don't make any strong guarantees about detection.

elide-dbmanager/elide-dbmanager-hibernate5/src/main/java/com/yahoo/elide/dbmanagers/hibernate5/PersistenceManager.java
```diff
@@ -0,0 +1,102 @@
...
+import java.io.IOException;
+
+/**
+ * Manager for javax.persistence compatible db resource
+ */
+public class PersistenceManager extends DatabaseManager {
+    private final EntityManagerFactory entityManagerFactory;
+
+    public PersistenceManager(EntityManagerFactory entityManagerFactory) {
+        this.entityManagerFactory = entityManagerFactory;
```
DeathByTape: Should check entityManagerFactory. Maybe: `Preconditions.checkNotNull(entityManagerFactory);`

elide-dbmanager/elide-dbmanager-hibernate5/src/main/java/com/yahoo/elide/dbmanagers/hibernate5/PersistenceManager.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Override
+    public void populateEntityDictionary(EntityDictionary dictionary) {
+        for (EntityType entity : entityManagerFactory.getMetamodel().getEntities()) {
```
DeathByTape: What if `getMetaModel()` or `getEntities()` returns `null`? Should we throw something more descriptive here than `NPE`?

shadanan: It doesn't seem to be easy to get those to be null. I tried removing the classes from the persistence.xml file, and we just get back an empty entities set.

DeathByTape:
:+1:

yahoocla:
CLA is valid!

PR #15, opened 2015-10-19
Publish through travis

clayreimann:
Once we use DPL to publish to bintree we can add an after_publish
step to have bintree sync the new version to Maven Central.
[Bintree API](https://bintray.com/docs/api/#_sync_version_artifacts_to_maven_central)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,25 @@
-language: java
-jdk:
-- oraclejdk8
-before_install:
-- sudo apt-get install libaio1
-script: mvn verify
-after_success:
-- test "${TRAVIS_PULL_REQUEST}" == "false" && test "${TRAVIS_TAG}" != "" && mvn deploy
-  --settings travis/settings.xml
-cache:
-  directories:
-  - "~/.m2"
-branches:
-  only:
-  - master
-  - "/^elide-parent-pom-[0-9]+\\.[0-9]+\\.[0-9]+\\.[0-9]+/"
-env:
-  global:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,20 +1,28 @@
...
+  language: java
+  jdk:
+    - oraclejdk8
+
+  # steps
+  before_install:
+    - sudo apt-get install libaio1
+  script: mvn verify
+  before_deploy: cat travis/bintray_template.json | sed s/\$\{TRAVIS_TAG\}/${TRAVIS_TAG}/ > travis/release.json
+  deploy:
```
DeathByTape: double check that this deploys only on tags.

.travis.yml
```diff
@@ -1,20 +1,26 @@
...
+    on:
+      tags: true
+    provider: bintray
+    file: travis/release.json
+    user: clayreimann
```
DeathByTape: This doesn't look right. Is this for bintray or for maven central? If for bintray, you should keep using the secure credentials we already have in place. This is an API-user not tied to an individual's account (I believe this is policy in case of account revocation later).

.travis.yml
```diff
@@ -1,20 +1,26 @@
...
+    key:
+      secure: eszSq9UgLy9bd2j6Z+zQvXAUnnEZy+DBcyjgzE26EA2cSZhAU+QdSkHgFsbU10fb9nRUK0mbunH/BxP/nMEtoVWRa33ALUfljvLMvdO+gOpqIWsdtmsvIn+76t774zNz90NMTtZuDWknbwKz1zP3ysAmj7KFtJIJAXpuDtvGyVmNyZ6z/qFJQCuHhSEsQf70cS+EZXLEkH16f+tSFQwwW4x4ucQY0qHbeiCVw7meCi8zue2/CwFcNw9wVukXOQ+zYScQStHPksce9+9OH9DlaeWB0EgKm7LWuTfip2O8ebpoyv5mOaAag5w72I82xP76GWxHFlFVr/RSs/Uudqgugc5lsy2Myh+xDKQgurdLfn6+7lF7F23ola7q9NgVnyG7vRxYUhPp6DL/a/gFqAfD2nDhz/kcZQnGEPh+Ookt1KJv3PaI2gpDeACAvXH04vG4Y5ZQaP/wC7hzADRYCsJPaHWumA7vVzmB4VTxsAuoAJaOs0iu99RrS+wKW/kGKQFh3wUxuVQ3W+EfySFjBzGVsTEcqH2ZX4xwFDR/nHOWI35KpGlcyEVVr+M52Of6raacn9nAv2EQxGuRIqL8yI9zMzicMVT1syl64slh3BSIVLZfU75j/CmSFHL7CKRE+A14CiXFkGdtg23+UQPGVU4LnaSl0/JnO+7OlDiqX1qxbyM=
+    dry-run: true
+  after_deploy: echo "Pending"
```
DeathByTape: Is this where we plan to push directly out to maven central?

.travis.yml
```diff
@@ -1,20 +1,24 @@
...
+  before_install:
+    - sudo apt-get install libaio1
+  script: mvn verify
+  before_deploy: cat travis/bintray_template.json | sed s/\$\{TRAVIS_TAG\}/${TRAVIS_TAG}/ > travis/release.json
+  deploy:
+    on:
+      tags: true
+    provider: bintray
+    file: travis/release.json
+    user: travisci-elide@yahoo
```
DeathByTape: Can you encrypt this as well? I am not certain that they want this in clear-text. If so, rebase over this commit.

.travis.yml
clayreimann: It is indeed. But I want to verify that bintray is getting the right artifacts first, which means a new release.

DeathByTape:
:+1:

yahoocla:
CLA is valid!

PR #18, opened 2015-10-21
add sparse fields and unit tests

shadanan:


shadanan:
@aklish Please have a look

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/jsonapi/document/processors/SparseFieldsetProcessor.java
```diff
@@ -0,0 +1,109 @@
...
+                    result.put(type, filters);
+                }
+            }
+        }
+
+        return result;
+    }
+
+    private void filterResource(Resource resource, Map<String, Set<String>> sparseFields) {
+        Set<String> includeFields = sparseFields.containsKey(resource.getType())
```
DeathByTape: I believe this concept is correct, but does it cover all cases? For instance, consider I have a book containing references to many pages. Example:

```java
public class Book {
  ...
  List<Page> pages;
  ...

    public static class Page {
        ...
        String title;
        String summary;
        int chapterNum;
        ...
    } 
}
```

Then querying a book, I should be able to extract `book.pages.chapterNum`. iirc, the `getType()` operators won't necessarily produce the correct result in this case?

DeathByTape:
Follow-up after in-person discussion: I think this needs to happen somewhere around the time where the beans are actually hydrated. No performance benefits can be gained at the DocumentProcessor level, unfortunately, since everything has already been turned into a `Resource` from a `PersistentResource`. Namely, the `toResource` method in `PersistentResource` is probably a good place to start.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -102,12 +113,39 @@ protected RequestScope(JsonApiDocument jsonApiDocument, RequestScope outerReques
...
 
     /**
+     * Parses queryParams and produces sparseFields map
+     * @param queryParams The request query parameters
+     * @return Parsed sparseFields map
+     */
+    private static Map<String, Set<String>> parseSparseFields(MultivaluedMap<String, String> queryParams) {
+        Map<String, Set<String>> result = new HashMap<>();
+
+        for (String key : queryParams.keySet()) {
```
danchen: perhaps use queryParams.entrySet() here so you have efficient access to values later.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -102,12 +113,39 @@ protected RequestScope(JsonApiDocument jsonApiDocument, RequestScope outerReques
...
+            if (key.startsWith("fields[") && key.endsWith("]")) {
+                String type = key.substring(7, key.length() - 1);
+
+                LinkedHashSet<String> filters = new LinkedHashSet<>();
+                for (String filterParams : queryParams.get(key)) {
+                    Collections.addAll(filters, filterParams.split(","));
+                }
+
+                if (filters.size() > 0) {
```
danchen: I think HashSet has an isEmpty() method.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1316,6 +1317,21 @@ static void checkPermissions(Class<? extends Check>[] checks, boolean mode, Pers
         checkPermission(annotationClass, annotation, resource);
     }
 
+    protected static boolean checkIncludeSparseField(Map<String, Set<String>> sparseFields, String type,
+                                                     String fieldName) {
+        if (sparseFields.size() != 0) {
```
danchen: perhaps use (!sparseFields.isEmpty())?

aklish:
:+1:

danchen:
:+1:

PR #24, opened 2015-10-23
Support read only transactions

wcekan:
Allow DatabaseManager to differentiate between read-only and updating transactions.


DeathByTape:
:+1:

yahoocla:
CLA is valid!

PR #26, opened 2015-10-26
Mulitplex support for multiple databases

wcekan:
* One database manager to wrap multiple databases within a single transaction.
* On commit failure, any partially committed transactions will be reversed.
* Refactored the EntityDictionary for readability.  Only internal changes.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
@@ -1,0 +1,220 @@
+/*
+ * Copyright 2015, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core;
+
+import com.yahoo.elide.annotation.ComputedAttribute;
+import com.yahoo.elide.annotation.Exclude;
+import com.yahoo.elide.core.exceptions.DuplicateMappingException;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.lang3.text.WordUtils;
+
+import java.lang.reflect.AccessibleObject;
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.lang.reflect.Modifier;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Deque;
+import java.util.List;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedDeque;
+
+import javax.persistence.Column;
+import javax.persistence.Id;
+import javax.persistence.ManyToMany;
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToMany;
+import javax.persistence.OneToOne;
+import javax.persistence.Transient;
+
+/**
+ * Entity Dictionary maps JSON API Entity beans to/from Entity type names.
+ *
+ * @see com.yahoo.elide.annotation.Include#type
+ */
+class EntityBinding {
+    private final static List<Method> OBJ_METHODS = Arrays.asList(Object.class.getMethods());
+
+    public final String jsonApi;
+    public final ConcurrentLinkedDeque<String> attrsDeque;
+    public final List<String> attrs;
+    public final ConcurrentLinkedDeque<String> relationshipsDeque;
+    public final List<String> relationships;
+    public final ConcurrentHashMap<String, RelationshipType> relationshipTypes;
+    public final ConcurrentHashMap<String, String> relationshipToInverse;
+    public final ConcurrentHashMap<String, AccessibleObject> fieldsToValues;
+    public final ConcurrentHashMap<String, String> aliasesToFields;
+    @Getter private AccessibleObject idField;
+    @Getter @Setter private Initializer initializer;
+
+    public final static EntityBinding EMPTY_BINDING = new EntityBinding();
+
+    /* empty binding constructor */
+    private EntityBinding() {
+        jsonApi = null;
+        idField = null;
+        attrsDeque = null;
+        attrs = null;
+        relationshipsDeque = null;
+        relationships = null;
+        relationshipTypes = null;
+        relationshipToInverse = null;
+        fieldsToValues = null;
+        aliasesToFields = null;
+    }
+
+    public EntityBinding(Class<?> cls, String type) {
+        // Map id's, attributes, and relationships
+        Collection<AccessibleObject> fieldOrMethodList = CollectionUtils.union(
+                Arrays.asList(cls.getFields()),
+                Arrays.asList(cls.getMethods()));
+
+        jsonApi = type;
+        // Initialize our maps for this entity. Duplicates are checked above.
+        attrsDeque = new ConcurrentLinkedDeque<>();
+        relationshipsDeque = new ConcurrentLinkedDeque<>();
+        relationshipTypes = new ConcurrentHashMap<>();
+        relationshipToInverse = new ConcurrentHashMap<>();
+        fieldsToValues = new ConcurrentHashMap<>();
+        aliasesToFields = new ConcurrentHashMap<>();
+        bindEntityFields(cls, type, fieldOrMethodList);
+
+        attrs = dequeToList(attrsDeque);
+        relationships = dequeToList(relationshipsDeque);
+    }
+
+    /**
+     * Bind fields of an entity including the Id field, attributes, and relationships.
+     *
+     * @param cls Class type to bind fields
+     * @param type JSON API type identifier
+     * @param fieldOrMethodList List of fields and methods on entity
+     */
+    private void bindEntityFields(Class<?> cls, String type, Collection<AccessibleObject> fieldOrMethodList) {
+        for (AccessibleObject fieldOrMethod : fieldOrMethodList) {
+            if (fieldOrMethod.isAnnotationPresent(Id.class)) {
+                bindEntityId(cls, type, fieldOrMethod);
+            } else if (fieldOrMethod.isAnnotationPresent(Transient.class)
+                    && !fieldOrMethod.isAnnotationPresent(ComputedAttribute.class)) {
+                continue; // Transient. Don't serialize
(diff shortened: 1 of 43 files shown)


elide-dbmanager/elide-dbmanager-multiplex/src/main/java/com/yahoo/elide/dbmanager/multiplex/MultiplexManager.java
```diff
@@ -0,0 +1,68 @@
...
+    public DatabaseTransaction beginReadTransaction() {
+        return new MultiplexReadTransaction(this);
+    }
+
+    public EntityDictionary getDictionary() {
+        return dictionary;
+    }
+
+    protected <T> DatabaseManager getSubManager(Class<T> cls) {
+        Class<T> type = (Class<T>) EntityDictionary.lookupEntityClass(cls);
```
aklish: Can you add a comment here about what this does.

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -0,0 +1,220 @@
...
+
+/**
+ * Entity Dictionary Element Binding.  For use by EntityDictionary.
+ */
+class EntityBinding {
+    private final static List<Method> OBJ_METHODS = Arrays.asList(Object.class.getMethods());
+
+    public final String jsonApi;
+    public final ConcurrentLinkedDeque<String> attrsDeque;
+    public final List<String> attrs;
```
DeathByTape: Should we be using `public final` here? Though these fields are final, they're not all immutable. May not be a problem in practice, but we should think about it.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -311,34 +287,29 @@ public String getNameFromAlias(Object entity, String alias) {
     }
 
     /**
-     * Initialize an entity.
+     * Initialize an entity
```
DeathByTape: I see all of these comments changed; I believe your and Dan's IDE's may be set differently :( iirc Dan recently "fixed" (this is a matter of style, I suppose) all comments to end with `.`

@danchen thoughts/comments?

elide-dbmanager/elide-dbmanager-inmemorydb/src/main/java/com/yahoo/elide/dbmanager/InMemory/InMemoryDB.java
```diff
@@ -48,7 +46,7 @@ public InMemoryDB(Package beanPackage) {
     /**
      * InMemoryDB transaction handler.
      */
-    public static class InMemoryTransaction implements DatabaseTransaction {
+    public class InMemoryTransaction implements DatabaseTransaction {
```
DeathByTape: Just thinking about this: is there any reason that this class should even be public? I think we can move this class to `private`.

elide-dbmanager/elide-dbmanager-multiplex/src/main/java/com/yahoo/elide/dbmanager/multiplex/MultiplexManager.java
```diff
@@ -0,0 +1,68 @@
...
+    protected final List<DatabaseManager> databaseManagers;
+    protected final ConcurrentHashMap<Class<?>, DatabaseManager> databaseManagerMap = new ConcurrentHashMap<>();
+    private EntityDictionary dictionary;
+
+    /**
+     * Create a single DatabaseManager to handle provided managers within a single transaction.
+     * @param databaseManagers list of sub-managers
+     */
+    public MultiplexManager(DatabaseManager... databaseManagers) {
+        this.databaseManagers = Arrays.asList(databaseManagers);
```
DeathByTape: Any reason not to simply accept a `List<DatabaseManager>` instead of varargs?

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -311,34 +287,29 @@ public String getNameFromAlias(Object entity, String alias) {
...
-    public <T> void initializeEntity(T entity) {
-        if (entity != null) {
-            @SuppressWarnings("unchecked")
-            Initializer<T> initializer = (Initializer<T>) initializers.get(lookupEntityClass(entity.getClass()));
-            if (initializer != null) {
-                initializer.initialize(entity);
+    public <T> void initializeEntity(T entityBean) {
+        if (entityBean != null) {
+            try {
+                AccessibleObject initializer = entity(entityBean.getClass()).getInitializer();
```
aklish: This is not as powerful as the Elide initializer we had before.

elide-dbmanager/elide-dbmanager-multiplex/src/main/java/com/yahoo/elide/dbmanager/multiplex/MultiplexTransaction.java
```diff
@@ -0,0 +1,117 @@
...
+            transactions.put(databaseManager, transaction);
+        }
+        lastDatabaseManagerTransaction = transaction;
+    }
+
+    protected abstract DatabaseTransaction beginTransaction(DatabaseManager databaseManager);
+
+    @Override
+    public User accessUser(Object opaqueUser) {
+        // TODO Auto-generated method stub
```
DeathByTape: Stale comment.

elide-dbmanager/elide-dbmanager-multiplex/src/main/java/com/yahoo/elide/dbmanager/multiplex/MultiplexTransaction.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public void flush() {
+        transactions.values().forEach(DatabaseTransaction::flush);
+    }
+
+    @Override
+    public void commit() {
+        // flush all before commit
+        flush();
+        transactions.values().forEach(DatabaseTransaction::commit);
```
DeathByTape: How do we cope with failed commits? Is this something we simply do not support?

elide-dbmanager/elide-dbmanager-multiplex/src/main/java/com/yahoo/elide/dbmanager/multiplex/MultiplexWriteTransaction.java
```diff
@@ -0,0 +1,174 @@
...
+    private final MultivaluedHashMap<DatabaseManager, Object> dirtyObjects = new MultivaluedHashMap<>();
+
+    public MultiplexWriteTransaction(MultiplexManager multiplexManager) {
+        super(multiplexManager);
+    }
+
+    @Override
+    protected DatabaseTransaction beginTransaction(DatabaseManager databaseManager) {
+        // begin updating transaction
+        return databaseManager.beginTransaction();
```
DeathByTape: Why do we need this method? Couldn't we just call `databaseManager.beginTransaction()` inline here: https://github.com/yahoo/elide/pull/26/files#diff-18fe3d40e18292c77178ad1194d82f0dR36?

elide-dbmanager/elide-dbmanager-multiplex/src/main/java/com/yahoo/elide/dbmanager/multiplex/MultiplexWriteTransaction.java
```diff
@@ -0,0 +1,174 @@
...
+     * Reverse changes of last commit since not all transactions could commit
+     * @param restoreList List of database managers to reverse the last commit
+     * @param cause cause to add any suppressed exceptions
+     */
+    private void reverseTransactions(ArrayList<DatabaseManager> restoreList, Throwable cause) {
+        for (DatabaseManager databaseManager : restoreList) {
+            try (DatabaseTransaction tx = databaseManager.beginTransaction()) {
+                List<Object> list = dirtyObjects.get(databaseManager);
+                for (Object dirtyObject : list == null ? Collections.emptyList() : list) {
+                    Object cloned = clonedObjects.get(dirtyObject);
```
DeathByTape: This method should have copious amounts of comments. In particular, our Javadoc should explain our method (i.e. we try to "undo" what was done) and list caveats (i.e. we require proper `.equals()` implementation for this to be effective). 

Most importantly, users should be warned that this does not make any consistency guarantees and this should be considered "dangerous" unless you understand the ramifications of using such a thing. Considering the situation:

```text
A commits
B fails to commit
** Attempt to "undo" A commit **
A fails to commit
```

A and B will obviously be inconsistent at this point. Before using this, you need to understand the effects on your whole system in such a scenario.

elide-dbmanager/elide-dbmanager-inmemorydb/src/main/java/com/yahoo/elide/dbmanager/InMemory/InMemoryDB.java
wcekan: It cannot be private if we would ever subclass it.

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
wcekan: Getters do not make them immutable either.

elide-dbmanager/elide-dbmanager-multiplex/src/main/java/com/yahoo/elide/dbmanager/multiplex/MultiplexWriteTransaction.java
wcekan: No.  This is an abstract method.  The beginTransaction here returns a writable transaction.  The one for MultiplexReadTransaction returns a read only transaction.

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
DeathByTape: That's right-- sorry, I was meaning we could return immutable copies trough getters.

DeathByTape: Or we can look into https://immutables.github.io/immutable.html :)

wcekan:
Merged and rebased.

yahoocla:
CLA is valid!

DeathByTape:
:+1:

PR #27, opened 2015-10-29
Filters for Hibernate 3 DB Manager

danchen:
- Filters currently implemented for Hibernate3 implementation only
  - Currently only supports primary collection filtering
  - Currently attributes only parses fields (last element)
- Pass query parameters through beginReadTransaction
- Separate HibernateManager and HibernateTransaction (unnest from HibernateManager) for all database managers
- Rename/reuse bookauthor test data model


elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    public static Set<Predicate> parseQueryParams(final MultivaluedMap<String, String> queryParams) {
+        if (queryParams == null) {
+            return Collections.emptySet();
+        } else {
+            final Set<Predicate> predicateSet = new HashSet<>();
+            queryParams.entrySet().forEach(queryParameter -> {
+                Operator operator;
+                // matching "filter[<key>][<operator, optional>]
```
aklish: Maybe rewrite this as two lines with an OR between so it is clearer

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -0,0 +1,87 @@
...
+                final Matcher matcher = Pattern.compile("filter\\[([^\\]]*)\\]{1}(\\[([^\\]]*)\\]{1})?")
```
aklish: You can remove {1} and change * to + for the text within the square brackets

elide-dbmanager/elide-dbmanager-hibernate3/src/main/java/com/yahoo/elide/dbmanagers/hibernate3/HibernateTransaction.java
```diff
@@ -0,0 +1,212 @@
...
+            return null;
+        }
+        return record;
+    }
+
+    @Override
+    public <T> Iterable<T> loadObjects(Class<T> loadClass) {
+        @SuppressWarnings("unchecked")
+        Iterable<T> list = new HibernateManager.ScrollableIterator(
+                applyPredicate(hibernateManager.getSession().createCriteria(loadClass), queryParams)
```
aklish: I'm not sure this covers the case of collections that are deeper than the root collection.

danchen:
Implementation moved to: https://github.com/yahoo/elide/pull/51

PR #30, opened 2015-11-02
fix the audit logger

shadanan:


elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -280,12 +286,14 @@ public ElideResponse delete(
...
             }
-            RequestScope requestScope = new RequestScope(doc, transaction, user, dictionary, mapper, securityMode);
+            RequestScope requestScope = new RequestScope(
+                    doc, transaction, user, dictionary, mapper, auditLogger, securityMode);
             DeleteVisitor visitor = new DeleteVisitor(requestScope);
             Supplier<Pair<Integer, JsonNode>> responder = visitor.visit(parse(path));
             requestScope.runDeferredPermissionChecks();
             transaction.flush();
             ElideResponse response = buildResponse(responder.get());
+            auditLogger.commit();
```
DeathByTape: What happens if these are a shared transaction? Will we call `commit()` twice? If so, does this cause any issues?

shadanan: auditLogger's commit doesn't call any hibernate transaction commits. it simply saves the log messages (either to a log file or a log table).

elide-core/src/main/java/com/yahoo/elide/audit/LogMessage.java
```diff
@@ -70,30 +72,57 @@ public LogMessage(String template,
...
+                } else {
+                    List<Object> objects = values.stream().map(PersistentResource::getObject)
+                            .collect(Collectors.toList());
+                    expression = expressionFactory.createValueExpression(objects, List.class);
+                }
+                ctx.setVariable(name, expression);
             }
-            ctx.setVariable(name, expression);
+
+            // Add the OpaqueUser to available expressions as "user"
```
aklish: I'm not sure we want to do this.  This might eclipse a User bean that is in the lineage.  User should be passed to the Logger a different way.

elide-core/src/main/java/com/yahoo/elide/Elide.java
DeathByTape: Maybe we should avoid the word "commit" then to make this clearer.

shadanan: Well, that's the name in the interface – and other things are already using that name in other projects. I'm somewhat reluctant to make a breaking change like that. If we did decide to change it, what do you think about "save"? @aklish what do you think?

elide-core/src/main/java/com/yahoo/elide/audit/LogMessage.java
shadanan: Indeed. Now that we have a request scope, we don't need this anymore. Removing...

aklish:
:+1:

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/Elide.java
DeathByTape: Let's leave it then for now

DeathByTape:
:+1:

PR #31, opened 2015-11-03
Update deps

DeathByTape:
First pass at updating dependencies. Some thing we should explore further is the integration test.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -221,7 +221,7 @@
                 </plugin>
                 <plugin>
                     <artifactId>maven-release-plugin</artifactId>
-                    <version>3.0-r1585899</version>
+                    <version>2.5.3</version>
                     <dependencies>
                         <dependency>
                             <groupId>org.apache.maven.scm</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,12 @@
                 <scope>test</scope>
             </dependency>
             <dependency>
+                <groupId>org.glassfish.hk2</groupId>
+                <artifactId>hk2-utils</artifactId>
+                <version>2.4.0-b32</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
                 <groupId>org.eclipse.jetty</groupId>
                 <artifactId>jetty-webapp</artifactId>
                 <version>${version.jetty}</version>
--- a/pom.xml
+++ b/pom.xml
@@ -67,14 +67,14 @@
 
         <!-- dependency versions -->
         <version.slf4j>1.7.12</version.slf4j>
-        <version.jetty>9.2.10.v20150310</version.jetty>
-        <version.jersey>2.19</version.jersey>
+        <version.jetty>9.3.5.v20151012</version.jetty>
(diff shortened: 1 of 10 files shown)


pom.xml
```diff
@@ -126,6 +126,12 @@
                 <scope>test</scope>
             </dependency>
             <dependency>
+                <groupId>org.glassfish.hk2</groupId>
+                <artifactId>hk2-utils</artifactId>
+                <version>2.4.0-b32</version>
+                <scope>test</scope>
+            </dependency>
```
danchen: Why do we need hk2 here?

DeathByTape: When I upgraded, this was missing. I believe rest-assured uses it in our integration tests.

pom.xml
```diff
@@ -221,7 +227,7 @@
                 </plugin>
                 <plugin>
                     <artifactId>maven-release-plugin</artifactId>
-                    <version>2.5.2</version>
+                    <version>3.0-r1585899</version>
```
danchen: I think we should use: 2.5.3 (searching from search.maven.org)

pom.xml
danchen: Ok, I think we can push this up into elide-integration-test, and remove it from here. Just, tried it and I can make this commit.

pom.xml
DeathByTape: Why?

pom.xml
DeathByTape: Tracked it down.

**origin/master**
```
[INFO] com.yahoo.elide:elide-integration-tests:jar:1.0.0.4-SNAPSHOT
[INFO] +- com.yahoo.elide:elide-core:jar:1.0.0.4-SNAPSHOT:compile
[INFO] |  +- com.google.guava:guava:jar:16.0.1:compile
[INFO] |  +- org.glassfish.hk2:hk2-api:jar:2.4.0-b25:compile
[INFO] |  |  +- org.glassfish.hk2:hk2-utils:jar:2.4.0-b25:compile
[INFO] |  |  \- org.glassfish.hk2.external:aopalliance-repackaged:jar:2.4.0-b25:compile
[INFO] |  +- org.antlr:antlr4-runtime:jar:4.5:compile
```

**updateDeps** branch
```
[INFO] com.yahoo.elide:elide-integration-tests:jar:1.0.0.4-SNAPSHOT
[INFO] +- com.yahoo.elide:elide-core:jar:1.0.0.4-SNAPSHOT:compile
[INFO] |  +- com.google.guava:guava:jar:18.0:compile
[INFO] |  +- org.glassfish.hk2:hk2-api:jar:2.4.0-b25:compile
[INFO] |  |  \- org.glassfish.hk2.external:aopalliance-repackaged:jar:2.4.0-b25:compile
[INFO] |  +- org.antlr:antlr4-runtime:jar:4.5.1-1:compile
```

Not entirely sure what changed here. In any case, it appears to be fixed by bumping the hk2 version (see updated commit.)

danchen:
:+1:

yahoocla:
CLA is valid!

PR #33, opened 2015-11-03
Initial checkin

aklish:
Adding a new Annotation : ComputedProperty

This will allow simple business logic for attributes that are not directly intended for persistence (marked as transient) but should be exposed via Elide.


elide-core/src/main/java/com/yahoo/elide/annotation/ComputedProperty.java
```diff
@@ -0,0 +1,22 @@
...
+import static java.lang.annotation.ElementType.METHOD;
+import static java.lang.annotation.RetentionPolicy.RUNTIME;
+
+/**
+ * Marks a method or field as a computed property that should be exposed via Elide regardless of whether or
+ * not it is marked as Transient.
+ */
+@Target({METHOD, FIELD})
+@Retention(RUNTIME)
+public @interface ComputedProperty {
```
danchen: Property has another meaning in Java. Maybe an alternative name for this?

DeathByTape:
:+1:

yahoocla:
CLA is valid!

PR #36, opened 2015-11-04
fix dependency resolution problems in examples

shadanan:


shadanan:
@jkusa will you review this change since you ran into this problem please?

yahoocla:
CLA is valid!

DeathByTape:
:+1: lgtm.

The only comment I have is that we should be careful where we use `RELEASE`. I think it's alright here since it's not a packaged module, but anything we package as a jar and publish should have fixed versions :)

PR #40, opened 2015-11-07
Move docs

aklish:
Moving docs to elide-doc repo.


DeathByTape:
:+1: now let's flesh out those docs a bit! :)

yahoocla:
CLA is valid!

PR #41, opened 2015-11-10
Removed Sharable annotation and replaced it with a flag on Include

aklish:
Also moved isSharable to EntityDictionary since it is a static thing.

Flag on include is better since it precludes weird things like Sharable but Excluded.


elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -337,6 +337,20 @@ public String getNameFromAlias(Object entity, String alias) {
     }
 
     /**
+     * Returns whether or not an entity is sharable
+     * @param entityClass the entity type to check for the sharable flag
+     * @return true if entityClass is sharable.  False otherwise.
+     */
+    public boolean isSharable(Class<?> entityClass) {
```
shadanan: should this be isShareable?

shadanan:
:+1:

yahoocla:
CLA is valid!

PR #42, opened 2015-11-10
Moving shareable flag to a new Permission type

aklish:


shadanan:
nice change. :+1:

yahoocla:
CLA is valid!

PR #44, opened 2015-11-11
Fixed security hole in shareable permission check

aklish:
Found a somewhat obvious bug in how I implemented the SharePermission check.  This is now fixed.


wcekan:
:+1:

yahoocla:
CLA is valid!

PR #45, opened 2015-11-11
Updates to id handling

jkusa:
- Remove subCollectionToOne logic
- Move coerce logic to CoerceUtil
- Allow string IDs to be fetched


elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -55,6 +55,7 @@
     private final ConcurrentHashMap<String, Class<?>> bindJsonApiToEntity = new ConcurrentHashMap<>();
     private final ConcurrentHashMap<Class<?>, String> bindEntityToJsonApi = new ConcurrentHashMap<>();
     private final ConcurrentHashMap<Class<?>, AccessibleObject> bindIdField = new ConcurrentHashMap<>();
+    private final ConcurrentHashMap<Class<?>, Class<?>> bindIdType = new ConcurrentHashMap<Class<?>, Class<?>>();
```
DeathByTape: To stay clean, you can probably just use the diamond operator here like above (i.e. `<>` instead of specifying type)

jkusa: Oops, good catch. I think my editor generated that.

elide-core/src/main/java/com/yahoo/elide/utils/coerse/CoerceUtil.java
```diff
@@ -0,0 +1,61 @@
...
+
+        if (value == null || cls == null || cls.isAssignableFrom(value.getClass())) {
+            return value;
+        }
+
+        try {
+            return ConvertUtils.convert(value, cls);
+
+        } catch (ConversionException | InvalidAttributeException e) {
+            throw new IllegalArgumentException("Unable to coerce " + value.getClass() + " to " + value);
```
DeathByTape: Should we throw a checked exception here or let this bubble up? Does this result in a 500 error from the server if we hit this?

elide-core/src/main/java/com/yahoo/elide/utils/coerse/converters/ToEnumConverter.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public <T> T convert(Class<T> cls, Object value) {
+
+        try {
+            if (ClassUtils.isAssignable(value.getClass(), String.class)) {
+                return stringToEnum(cls, (String) value);
+            } else if (ClassUtils.isAssignable(value.getClass(), Integer.class, true)) {
+                return intToEnum(cls, (Integer) value);
+            } else {
+               throw new UnsupportedOperationException(value.getClass().getSimpleName() + " to Enum no supported");
```
DeathByTape: Same comment as before.. Will this exception cause a 500? That's probably not what we want.

elide-core/src/main/java/com/yahoo/elide/utils/coerse/converters/ToEnumConverter.java
```diff
@@ -0,0 +1,64 @@
...
+            }
+        } catch (IndexOutOfBoundsException | ReflectiveOperationException
+                | UnsupportedOperationException | IllegalArgumentException e) {
+            throw new InvalidAttributeException("Unknown " + cls.getSimpleName() + " value " + value);
```
DeathByTape: See above.

elide-core/src/main/java/com/yahoo/elide/utils/coerse/CoerceUtil.java
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2015, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.utils.coerse;
+
+import com.yahoo.elide.core.exceptions.InvalidAttributeException;
+import com.yahoo.elide.utils.coerse.converters.ToEnumConverter;
+import org.apache.commons.beanutils.*;
```
aklish: No * includes.

DeathByTape:
:+1:

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/utils/coerse/converters/ToEnumConverter.java
DeathByTape: Maybe throw https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidURLException.java

wcekan:
@jkusa Please pull in https://github.com/yahoo/elide/commit/4676ddda487b8a44bb9784d96d46622b0717a507
This will fix the InMemoryDB handling of ID types.

PR #48, opened 2015-11-12
Fixing bug where we don't check whether a relationship exists before …

aklish:
…deleting the inverse (security hole)

PersistentResource doesn't check if the resource being requested for removal already exists.  It naively removes the inverse from the requested resource.

Some other small cleanup - removing old TODOs that are not true and cleaning up redundant permission checks.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -483,19 +483,27 @@ public class PersistentResource<T> {
     public void removeRelation(String fieldName, PersistentResource removeResource) {
         checkPermission(UpdatePermission.class, this);
         checkFieldPermission(UpdatePermission.class, this, fieldName);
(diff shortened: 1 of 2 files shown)


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -483,20 +483,28 @@ public boolean clearRelation(String relationName) {
     public void removeRelation(String fieldName, PersistentResource removeResource) {
         checkPermission(UpdatePermission.class, this);
         checkFieldPermission(UpdatePermission.class, this, fieldName);
+        checkPermission(DeletePermission.class, removeResource);
```
DeathByTape: Is this really a `Delete`? We're not really deleting are we? we're just breaking a relation.

aklish: I didn't change the permissions.  I just eliminated duplicates and stuck them in new places.

DeathByTape:
:+1:

PR #51, opened 2015-11-16
working filters

shadanan:


elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateStore.java
```diff
@@ -230,10 +242,97 @@ public void commit() {
                     compositeCriterion = Restrictions.and(compositeCriterion, criterion);
                 }
             }
+
             return compositeCriterion;
         }
 
         @Override
+        public Collection filterCollection(Collection collection, String type, FilterScope<?> filterScope) {
```
aklish: Dan and I were originally thinking operators would be more like separate functions that take arguments (possibly variable number).  Each function would be responsible for building part of a query.  This would also make it easier to extend to add new operators.  This function has a large cyclomatic complexity.  I suggest we break it up - maybe in the manner of having an abstraction around operators.

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -0,0 +1,79 @@
...
+            Matcher matcher = Pattern.compile("filter\\[([^\\]]+)\\](\\[([^\\]]+)\\])?")
+                    .matcher(queryParameter.getKey());
+            if (matcher.find()) {
+                final String[] keyParts = matcher.group(1).split("\\.");
+
+                if (keyParts.length != 2) {
+                    throw new InvalidPredicateException("Invalid filter format: " + queryParameter.getKey());
+                }
+
+                final String type = keyParts[0];
```
aklish: I'm not 100% sure this should be a type.  It might need to be a relation name for non root collections.  Let's talk about it and figure out which makes more sense.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -86,4 +87,16 @@ default void flush() {
...
+
+    /**
+     * Filter a collection by the Predicates in filterScope.
+     *
+     * @param collection the collection to filter
+     * @param type
+     * @param filterScope the FilterScope containing the set of Predicates
+     * @return the filtered collection
+     */
+    default Collection filterCollection(Collection collection, String type, FilterScope<?> filterScope) {
```
aklish: I don't feel like 'type' belongs here.  The collection already has the type contained within it.  All you need to do is something like:
        if (collection.getClass() instanceof ParameterizedType) {
            return (Class<?>) ((ParameterizedType) collection.getClass()).getActualTypeArguments()[0];
        }

In general, having two parameters that indirectly contain the same information can lead to conflicts where it is possible to specify different information and lead to subtle bugs in the calling code.

aklish: The default implementation should actually perform the filters on the collection by iterating through the elements and then evaluating the predicates.

aklish: I'm not 100% the concept of FilterScope really makes sense here.  All we really need is a list<Predicate> - not a scope.  The scope concept was more around checks.  I'm not a huge fan of the catchall 'scope' class where we stick a bunch of unrelated stuff.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateStore.java
```diff
@@ -236,6 +244,92 @@ public void commit() {
         }
 
         @Override
+        public Collection filterCollection(Collection collection, String type, FilterScope<?> filterScope) {
```
aklish: This code looks exactly the same as Hibernate 3 - anyway we can commonize the common parts?

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
shadanan: These are good points. Will do.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateStore.java
shadanan: I agree. Hibernate store between 3 and 5 are actually identical. I believe the only difference is in the name of the import. We should discuss a way to make them share this code.

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
shadanan: The type is irrespective of the hierarchy. Having the ability to do type.relation.relation.field does allow us to "target" the filter more narrowly, although, I don't know how much value that would have.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateStore.java
shadanan: I really like this idea. I'll implement it tomorrow.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -86,4 +89,16 @@ default void flush() {
...
+    /**
+     * Filter a collection by the Predicates in filterScope.
+     *
+     * @param collection the collection to filter
+     * @param entityClass the class of the entities in the collection
+     * @param predicates the set of Predicate's to filter by
+     * @return the filtered collection
+     */
+    default <T> Collection filterCollection(Collection collection, Class<T> entityClass, Set<Predicate> predicates) {
+        return collection;
```
aklish: I still feel like we should do something here beyond just returning the collection

aklish:
Great improvements.  Almost there.

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
shadanan: perhaps we shouldn't have this be default?

aklish:
:+1:

PR #55, opened 2015-11-19
test for add unshared child with parent relationship

wcekan:
Fixes the problem when adding a child containing a relationship to the parent without a SharePermission.


aklish:
:+1:

yahoocla:
CLA is valid!

PR #58, opened 2015-11-19
refactor datastore package naming and consistency

danchen:
- move classes into datastore packages
- make package name consistent (using datastores)


shadanan:
:+1:

yahoocla:
CLA is valid!

wcekan:
@danchen Are you sure we needed to rename packages again?  
We now have an inconsistency with the module names: see [repo](https://search.maven.org/#search%7Cga%7C1%7Celide)

danchen:
I didn't change the module name, so it's all the same artifacts. For the module name the singular form that it is already is the best.

PR #62, opened 2015-11-21
Push get by id to predicate

aklish:
Minor performance optimization to push down getting a collection element by ID down to the DataStore.


DeathByTape:
:+1:

yahoocla:
CLA is valid!

Issue #72, opened 2015-12-01
Defaulting Permission for Object


DeathByTape:
We should be able to have a default permission (i.e. default deny all) and selectively update fields within that object with additional permissions.

Currently, the object takes precedence and ignores fields. Field security annotations should be first-class citizens.

DeathByTape:
Fixed with PR #73.

PR #73, opened 2015-12-01
Added support for field permission precedence.

DeathByTape:
PR addresses issue: https://github.com/yahoo/elide/issues/72


aklish:
:+1:

yahoocla:
CLA is valid!

PR #79, opened 2015-12-05
Fixing SharePermission checks to not include entities that are alread…

aklish:
…y in a collection

Added unit tests


clayreimann:
[LGTM](http://lgtm.in/i/mtXcg6Spq)

PR #81, opened 2015-12-08
Added support for field-level overrides for reading.

DeathByTape:


clayreimann:
This will make limited access in railsplitter-ws so much cleaner!
[LGTM](http://lgtm.in/i/Th73)

[:+1:](http://lgtm.in/u/Th73)

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -251,7 +251,7 @@ public boolean matchesId(String checkId) {
         }
 
         PersistentResource<T> resource = new PersistentResource<>(obj, requestScope);
-        checkPermission(ReadPermission.class, resource);
+        requestScope.checkFieldAwarePermissions(ReadPermission.class, resource);
```
wcekan: Could we make a local private to make this look cleaner?
```java
    resource.checkFieldAwarePermissions(ReadPermission.class);


private void checkFieldAwarePermissions(Class<A> annotationClass) {
    return requestScope.checkFieldAwarePermissions(ReadPermission.class, this);
}
```

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -456,7 +456,7 @@ protected boolean updateToOneRelation(String fieldName, Set<PersistentResource>
      * @return True if object updated, false otherwise
      */
     public boolean clearRelation(String relationName) {
-        checkObjectAndFieldPermissions(UpdatePermission.class, this, relationName);
+        requestScope.checkFieldAwarePermissions(UpdatePermission.class, this, relationName);
```
wcekan:
```java
checkFieldAwarePermissions(UpdatePermission.class, relationName);
```

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
DeathByTape: This is a static method :( so not in this particular case.

DeathByTape: This is true, but I just added another method. This is fixed! :+1:

wcekan:
:+1:

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -275,4 +298,96 @@ public String toString() {
...
+        }
+
+        public FieldAwareCheck(Class<A> annotationClass, PersistentResource resource, String fieldName) {
+            this.annotationClass = annotationClass;
+            this.resource = resource;
+            this.fieldName = fieldName;
+        }
+
+        public void run() {
+            // Hack: doNotDefer is a special flag to temporarily disable deferred checking. Presumably, this check
```
aklish: I think this hack can be removed by restructuring.

aklish:
A few comments:
1) I would check Akash's pull request, but I believe a zuul super user can create a programmatic user WITHOUT needing a company membership.  However, I think it is probably OK.  We can associate such a user with Yahoo.

2) Can you break up the function to get a programmatic token.  I would probably make each mode handling section a separate function.

PR #85, opened 2015-12-15
Check UpdatePermission on remove relationship

wcekan:
Removing a relationship from a collection should require only UpdatePermission on removed entity.


DeathByTape:
:+1:

yahoocla:
CLA is valid!

PR #89, opened 2015-12-16
A deleted resource should remove itself from all of its bidirectional…

aklish:
… relationships

This will also force update permissions to be checked on those relationships.


DeathByTape:
:+1:

Issue #84, opened 2015-12-14
Revamp Security Backend


DeathByTape:
Below is the proposal for revamping the Elide security backend. It will require *breaking* changes to the public-facing API and-- consequently-- will be rolled up into **Elide 2.0**. This is a working document. That is, I will keep this up-to-date with developments on progression plans.

# Current Limitations

Due to the current design of the security system, Elide faces problematic security limitations. These problems are not security vulnerabilities, but they disallow expression of certain security constraints. Moreover, they can also lead to confusion which is a show-stopper for effective security. That is, the best security rules are simple and clear making them less error-prone; the system executing these rules should also be clear in nature.

  1. **Fine-Grained Update Control.** At the moment, it is not feasible to specify controls on Updates based on type of modification. For instance, if a user should be able to add objects to a list but disallowed to remove any, this cannot be effectively expressed in the current system. As it stands, we broadly check "update" permissions. In our case, an update of empty list effectively removes all elements within a list.
  1. **Check Execution Reasoning.** In our current model, authorization check execution flow is determined dynamically at runtime based on a multitude of factors. In other words, the current system is trying to be "smart." Effectively, this makes reasoning about security checks far more difficult based upon when they would/should be executed.
  1. **Inability to control short-circuiting.** It is common practice in programming to perform a couple of cheap checks before an expensive one. If any of these cheaper checks fail, then you can avoid computing the expensive check. In our current system, the lack of control on check execution does not guarantee that such a thing can be done repeatedly.
  1. **`UserCheck`'s are tightly-coupled with security.** The notion of a `UserCheck` in practice is not strictly for security, but more of an optimization instead. As a result, their tight-coupling to the security system makes them confusing to work with and limits the expressiveness of the system.

# Solution

Below I have outlined the current state of our proposed solution. This solution comes from discussions with multiple sources and I suspect will further evolve through continued discourse before it is fully implemented. Again, I will keep this section up-to-date with the latest information as it's decided.

  1. **Fine-grained modification support.** We will update the `Check` interface to support `Changesets` for `Update` requests. In particular, the `Changeset` object (pseudo-code outlined below) will become a parameter describing the proposed modification to an object. This will allow `Check` writers to verify the type of data modifications permitted. This solves the problem of **Fine-Grained Update Control** mentioned above.
[...]

PR #94, opened 2015-12-28
Security Revamp

DeathByTape:
This is the first PR for the security revamp discussed here: https://github.com/yahoo/elide/issues/84. This reworks the code into the new proposed structure.

**NOTE:** This PR is still missing the actual calculation of Changesets (i.e. they're always null still in this code). I will work that into a different PR.


elide-core/src/main/java/com/yahoo/elide/annotation/UserPermission.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+@Target({ElementType.TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface UserPermission {
+    /**
+     * UserCheck's to run before evaluating additional checks on object
+     *
+     * @return
+     */
+    Class<? extends UserCheck>[] value() default {};
```
aklish: should user permissions have 'any' and 'all' like our other check annotations?

elide-core/src/main/java/com/yahoo/elide/core/FilterScope.java
```diff
@@ -26,19 +27,32 @@
...
         this.requestScope = requestScope;
         this.isAny = isAny;
 
         List<Check<T>> checks = new ArrayList<>(checkClasses.length);
+        List<UserCheck> userChecks = new ArrayList<>(userCheckClasses.length);
+        for (Class<? extends UserCheck> checkClass : userCheckClasses) {
+            try {
+                userChecks.add(checkClass.newInstance());
+            } catch (InstantiationException | IllegalAccessException e) {
+                userChecks.add(null);
```
aklish: Log an error and maybe we should throw a runtime exception (IllegalStateException) here.

elide-core/src/main/java/com/yahoo/elide/security/ChangeSpec.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.Getter;
+
+/**
+ * ChangeSpec for a particular field.
+ *
+ * @param <T> type parameter
+ */
+@AllArgsConstructor
+public class ChangeSpec<T> {
+    @Getter private final String parentType;
```
aklish: Why do we need this?  If it is a type, why is it a String?

elide-core/src/main/java/com/yahoo/elide/security/Check.java
```diff
@@ -15,9 +17,35 @@
...
+     *              they will not be deferred until commit time. That is, this is treated as an additional
+     *              <i>operation check</i> in the cases of Read and Delete permissions. Moreover, it is joined in
+     *              conjunction with the result from the proper operation check in such a scenario.
+     *
+     * @param object Fully modified object
+     * @param requestScope Request scope object
+     * @param changeSpec Summary of modifications
+     * @return true if allowed
+     */
+    default boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
```
aklish: I'm wondering if we should have different ok methods (rather than optional here) for cases where we need to check field level permissions.  The changeSpec only really makes sense for fields and relationships.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,278 @@
...
+    public static <A extends Annotation> ExtractedChecks extractChecks(Class<A> annotationClass, A annotation) {
+        Class<? extends Check>[] anyChecks;
+        Class<? extends Check>[] allChecks;
+        try {
+            anyChecks = (Class<? extends Check>[]) annotationClass
+                    .getMethod("any").invoke(annotation, (Object[]) null);
+            allChecks = (Class<? extends Check>[]) annotationClass
+                    .getMethod("all").invoke(annotation, (Object[]) null);
+        } catch (ReflectiveOperationException e) {
+            throw new InvalidSyntaxException("Unknown permission " + annotationClass.getName(), e);
```
aklish: Maybe log an error here.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,278 @@
...
+        }
+        if (anyChecks.length <= 0 && allChecks.length <= 0) {
+            throw new InvalidSyntaxException("Unknown permission " + annotationClass.getName());
```
aklish: Log an error?

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,278 @@
...
+                                     Class<? extends Annotation> annotationClass,
+                                     boolean isFieldAware) {
+        if (resource.getRequestScope().getSecurityMode() == SecurityMode.BYPASS_SECURITY) {
+            return;
+        }
+
+        BiFunction<Check, Optional<ChangeSpec<?>>, Boolean> checkFn =
+                (check, changespec) -> check.ok(resource.getRequestScope(), changespec);
+
+        boolean shouldDefer = CreatePermission.class.equals(annotationClass)
```
aklish: I thought check deferral was going to be specified by the check writer.  Is that change coming in another PR?

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,278 @@
...
+                                    boolean shouldDefer,
+                                    boolean isFieldAware,
+                                    BiFunction<Check, Optional<ChangeSpec<?>>, Boolean> checkFn) {
+        // TODO: Need to write field-aware checks.
+        for (Class<? extends Check> check : checks) {
+            Check checkHandler;
+            try {
+                checkHandler = check.newInstance();
+            } catch (InstantiationException | IllegalAccessException e) {
+                throw new InvalidSyntaxException("Illegal permission check " + check.getName(), e);
```
aklish: Log an error.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,278 @@
...
+            }
+
+            Boolean ok = checkFn.apply(checkHandler, changeSpec);
```
aklish: So if a check is deferred, shouldn't this not be run at this point?

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,278 @@
...
+                log.debug("ForbiddenAccess {} {}#{}", check, resource.getType(), resource.getId());
+                throw new ForbiddenAccessException();
+            }
+
+            // Add check to a list for later execution
+            if (shouldDefer) {
+                // Hrm so much reliance on side-effects :(
+                // TODO: Hrm. If hashes don't match, this will probably be run far too many times...
+                commitChecks.add(() -> {
+                    executePermissions(checks,
```
aklish: I'm not sure about this.   In general, avoid recursion at all costs.   I think we need another non-recursive way to do this.    The other issue is that this will run the same checks multiple times.   I think a bit of refactoring is needed here.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
DeathByTape: It is. I have this documented in Check interface, but basically if a user specifies a commit check for a read-check, it would be automatically deferred (which breaks the semantics of our current system).

In summary, deferred checks are only valid for `Update` and `Create` operations. If they specified such a check on a `Read` or `Delete`, then we execute it inline with the caveat that the data they're looking at may be incomplete.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,278 @@
...
+        try {
+            A annotation = dictionary.getAnnotation(resource, annotationClass);
+            if (annotation != null) {
+                ExtractedChecks extracted = extractChecks(annotationClass, annotation);
+                boolean isAny = extracted.getAnyChecks().length > 0;
+                Class<? extends Check>[] checks = (isAny) ? extracted.getAnyChecks() : extracted.getAllChecks();
+                runPermissionChecks(checks, isAny, resource, changeSpec, annotationClass, true);
+            }
+        } catch (ForbiddenAccessException e) {
+            // Ignore this and continue on to checking our fields
```
aklish: Why continue at this point if permission will be denied?

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
DeathByTape: This is not exactly straight-forward, I suppose.

The `Check` interface encapsulates 2 methods. One for `Commit` checks and one for `Operation` checks. The first time through, we *always* execute the operation check (which may be implemented as nothing but `true` to ignore its affects). The deferral variable determines whether or not we should queue up this check to execute its `commit` check method at a later time.

The reason for the `shouldDefer` in the first place is short-circuiting. If we always defer and we have an `any` check for a Read, for instance, then we would defer our decision until commit time even though we don't support commit checks for Reads.

elide-core/src/main/java/com/yahoo/elide/annotation/UserPermission.java
DeathByTape: Not sure if we landed on a decision for this. When we initially spoke about this, we discussed having user checks without the `any` and `all`, but I cannot remember at the moment if this was to protect us in some way. 

Off the top of my head, nothing sticks out at me as dangerous by adding this functionality.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,278 @@
...
+                runPermissionChecks(checks, isAny, resource, changeSpec, annotationClass, true);
+                // If this is an "any" check, then we're done. If it is an "all" check, we may have commit checks queued
+                // up. This means we really need to check additional fields and queue up those checks as well.
+                if (isAny) {
+                    return;
+                }
+            }
+        } catch (ForbiddenAccessException e) {
+            // Ignore this and continue on to checking our fields
+            anyPassing = false;
```
aklish: variable name is off here.  This could be set for 'all' checks too.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
DeathByTape: This is not exactly true (there are two code paths in the function: `shouldDefer` and `!shouldDefer`), but it is possible to have repeats of the same commit checks depending on how lambdas are hashed.

For instance, this will certainly never cause the same `operation` check to be executed multiple times.

elide-core/src/main/java/com/yahoo/elide/optimization/UserCheck.java
```diff
@@ -3,17 +3,19 @@
...
  * Custom security access that verifies whether a user belongs to a role.
  * Permissions are assigned as a set of checks that grant access to the permission.
  *
- * @param <T> Type of record for Check
+ * NOTE: These checks are intended to be used for specific optimizations. That is, if the check
+ * only requires user-level information and can cover an entire entity. These only apply to at
+ * the entity-level. Field-level annotating of UserCheck's is not supported.
  */
-public interface UserCheck<T> extends Check<T> {
+public interface UserCheck {
```
aklish: Minor comment. Might be nice to change function name 'userPermission' to 'ok' to match other checks.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -128,7 +124,7 @@ protected RequestScope(
             JsonApiMapper mapper,
             Logger logger) {
         this(null, transaction, user, dictionary, mapper, logger);
-        this.deferredChecks = new LinkedHashSet<>();
+        this.commitChecks = new LinkedHashSet<>();
```
aklish: Why do we store commitChecks both here and inside PermissionManager?  Can we just have one location?

elide-core/src/main/java/com/yahoo/elide/annotation/UserPermission.java
aklish: FilterScope logic seems to imply the use of any and all still.  I think it makes sense to have.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1269,50 +1239,41 @@ protected void addInverseRelation(String relationName, Object relationValue) {
...
-        Class<? extends Check>[] allChecks;
-        try {
-            anyChecks = (Class<? extends Check>[]) annotationClass
-                    .getMethod("any").invoke(annotation, (Object[]) null);
-            allChecks = (Class<? extends Check>[]) annotationClass
-                    .getMethod("all").invoke(annotation, (Object[]) null);
-        } catch (ReflectiveOperationException e) {
-            throw new InvalidSyntaxException("Unknown permission " + annotationClass.getName(), e);
-        }
+        PermissionManager.ExtractedChecks checks = PermissionManager.extractChecks(annotationClass, annotation);
```
aklish: Can we obliterate this function from this class now that we have a 'PermissionManager'?  It would be nice to have similar logic all in one place.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1338,42 +1299,13 @@ protected void addInverseRelation(String relationName, Object relationValue) {
     }
 
     private <A extends Annotation> void checkFieldAwarePermissions(Class<A> annotationClass) {
-        requestScope.checkFieldAwarePermissions(annotationClass, this);
+        // TODO: Need ChangeSet support
+        requestScope.getPermissionManager().checkFieldAwarePermissions(this, null, annotationClass);
     }
 
     private <A extends Annotation> void checkFieldAwarePermissions(Class<A> annotationClass, String fieldName) {
-        requestScope.checkFieldAwarePermissions(annotationClass, this, fieldName);
```
aklish: Can we eliminate this function?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1338,42 +1299,13 @@ protected void addInverseRelation(String relationName, Object relationValue) {
     }
 
     private <A extends Annotation> void checkFieldAwarePermissions(Class<A> annotationClass) {
-        requestScope.checkFieldAwarePermissions(annotationClass, this);
```
aklish: Can we eliminate this function?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1269,50 +1239,41 @@ protected void addInverseRelation(String relationName, Object relationValue) {
...
-            resource.requestScope.checkPermissions(annotationClass, allChecks, ALL, resource);
+            resource.getRequestScope().getPermissionManager().checkPermissions(allChecks,
+                    ALL, resource, annotationClass);
         } else {
             throw new InvalidSyntaxException("Unknown permission " + annotationClass.getName());
         }
     }
 
-    static <A extends Annotation> FilterScope loadChecks(A annotation, RequestScope requestScope) {
+    static <A extends Annotation> FilterScope loadChecks(A annotation,
```
aklish: I don't think this belongs here.

elide-core/src/main/java/com/yahoo/elide/optimization/UserCheck.java
DeathByTape: Will do!

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
DeathByTape: Good catch. This is actually an unused var-- an artifact of the old work around.

elide-core/src/main/java/com/yahoo/elide/security/ChangeSpec.java
DeathByTape: I am still thinking about this. This is both a weird name and type for this; when I say "parent" I mean parent of the field and not parent resource. Perhaps we should pass the resource directly? It's highly likely you could have collisions here.

Consider writing an ID check that does something special based on type. For instance, we allow ID 0 for object `X` and a 1 for object `Y`. The check could not then distinguish.

Perhaps the correct solution is to forget and require a separate check be written.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
DeathByTape: We could, but as a personal opinion I think this makes the rest of the `PersistentResource` code easier to read. Rather than passing the hardcoded values in each time (i.e. `this` and `null`), we can let this function do the work instead.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
DeathByTape: See comment below.

elide-core/src/main/java/com/yahoo/elide/optimization/UserCheck.java
DeathByTape: NOTE: This was completed. GitHub just doesn't hide this comment since the change is on a different line.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,464 @@
...
+    public static <A extends Annotation> ExtractedChecks extractChecks(Class<A> annotationClass, A annotation) {
+        Class<? extends Check>[] anyChecks;
+        Class<? extends Check>[] allChecks;
+        try {
+            anyChecks = (Class<? extends Check>[]) annotationClass
+                    .getMethod("any").invoke(annotation, (Object[]) null);
+            allChecks = (Class<? extends Check>[]) annotationClass
+                    .getMethod("all").invoke(annotation, (Object[]) null);
+        } catch (ReflectiveOperationException e) {
+            log.debug("Unknown permission: {}, {}", annotationClass.getName(), e);
```
aklish: Should be warn or error I think

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,464 @@
...
+    /**
+     * Extracted checks.
+     */
+    @AllArgsConstructor
+    public static final class ExtractedChecks {
+        private final Class<? extends Check>[] anyChecks;
+        private final Class<? extends Check>[] allChecks;
+
+        @SuppressWarnings("unchecked")
+        public Class<? extends Check>[] getAllChecks() {
```
aklish: The name here is misleading.  Does it mean to return 100% of checks regardless of type OR does it mean to just return the 'all' checks?

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,464 @@
...
+        Class<? extends UserCheck>[] anyChecks = userPermission.any();
+        Class<? extends UserCheck>[] allChecks = userPermission.all();
+        Class<? extends Annotation> annotationClass = userPermission.getClass();
+
+        if (anyChecks.length > 0) {
+            return new FilterScope(requestScope, CheckMode.ANY, anyChecks);
+        } else if (allChecks.length > 0) {
+            return new FilterScope(requestScope, CheckMode.ALL, allChecks);
+        } else {
+            log.debug("Unknown user permission '{}'", annotationClass.getName());
```
aklish: error?

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,464 @@
...
+    /**
+     * Check object and all fields for field- and class-level checks.
+     *
+     * @param classChecks
+     * @param fieldChecks
+     * @param resource
+     * @param changeSpec
+     * @return True if done (i.e. no commit checks), false otherwise. Throws a ForbiddenAccessException upon failure.
+     */
+    private void checkClassAndFields(Class<? extends Check>[] classChecks,
```
aklish: I'm wondering if we can consolidate these helpers.  See my other comment.

elide-core/src/main/java/com/yahoo/elide/core/FilterScope.java
```diff
@@ -5,53 +5,71 @@
...
-        List<Check<T>> checks = new ArrayList<>(checkClasses.length);
-        for (Class<? extends Check> checkClass : checkClasses) {
+        List<UserCheck> userChecks = new ArrayList<>(userCheckClasses.length);
+        for (Class<? extends UserCheck> checkClass : userCheckClasses) {
             try {
-                checks.add(checkClass.newInstance());
+                userChecks.add(checkClass.newInstance());
             } catch (InstantiationException | IllegalAccessException e) {
-                checks.add(null);
+                log.debug("Could not instantiate UserCheck: {}", checkClass.getName());
```
aklish: log error.

elide-core/src/main/java/com/yahoo/elide/core/FilterScope.java
```diff
@@ -69,17 +87,17 @@ public UserPermission getUserPermission() {
         }
 
         UserPermission compositeUserPermission = null;
-        for (Check check : checks) {
+        for (UserCheck check : userChecks) {
```
aklish: Should this function be moved to PermissionManager?

elide-core/src/main/java/com/yahoo/elide/security/User.java
```diff
@@ -20,13 +19,11 @@
  */
 public class User {
     @Getter private final Object opaqueUser;
-    private final Map<Class<? extends Check>, UserPermission> okUserPermissions;
-    private final Map<Class<? extends Check>, Map<PersistentResource, Boolean>> okCheckResources;
```
wcekan: We lost this checked resources cache.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,496 @@
...
+        for (Class<? extends Check> check : checks) {
+            Check handler;
+            try {
+                handler = check.newInstance();
+            } catch (InstantiationException | IllegalAccessException e) {
+                log.debug("Illegal permission check '{}' {}", check.getName(), e);
+                throw new InvalidSyntaxException("Illegal permission check '" + check.getName() + "'", e);
+            }
+
+            boolean ok = handler.ok(resource.getObject(), resource.getRequestScope(), Optional.ofNullable(changeSpec));
```
wcekan: Used to be cached via
```java
boolean ok = resource.getRequestScope().getUser().ok(checkHandler, resource);
```

elide-core/src/main/java/com/yahoo/elide/security/User.java
DeathByTape: We need to discuss this more offline. If we allow side-effects in checks, caching may throw off those semantics. However, side-effects are generally a horrible thing (in my humble opinion :)) so I'm happy to reinstate the cache after some more thought.

aklish: I don't think checks should have side effects.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,500 @@
...
+                                   PersistentResource<?> resource,
+                                   ChangeSpec changeSpec,
+                                   boolean hasDeferredChecks,
+                                   Strategy strategy) {
+        boolean hasPassingCheck = !isEmptyCheckArray(classChecks)
+                || (isEmptyCheckArray(classChecks) && isEmptyListOfChecks(fieldChecks));
+        boolean entityFailed = false;
+
+        // Check full object, then all fields
+        if (!isEmptyCheckArray(classChecks)) {
```
aklish: I think the logic would be easier to understand if there wasn't this close coupling between the Strategy functions and this function.  I was thinking Strategy would have a run method that basically takes:
run(ExtractedCheck entityCheck, Set<ExtractedCheck> fieldChecks).

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,500 @@
...
+     *
+     * @param classChecks Array of class/entity-level checks
+     * @param fieldChecks Array of field checks
+     * @param resource Resource
+     * @param changeSpec Change spec
+     * @param hasDeferredChecks Whether or not there are deferred checks beyond
+     * @param strategy Strategy to use upon failure/success of checks
+     * @return True if any of the field-level checks succeed, false otherwise. False if no field-level checks exist.
+     */
+    private void fieldAwareExecute(Class<? extends Check>[] classChecks,
```
aklish: I think there should be a class that essentially represents the permission which includes the complete list of checks and the corresponding mode.  Functions should only need a single instance of this 'permission' for the class and a set of them for the set of fields that need to be checked.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,500 @@
...
+            if (annotation == null) {
+                continue;
+            }
+            ExtractedChecks extracted = extractChecks(annotationClass, annotation);
+            fieldOpChecks.add(extracted.getOperationChecks());
+            fieldComChecks.add(extracted.getCommitChecks());
+            fieldCheckModes.add(extracted.getCheckMode());
+        }
+
+        boolean hasDeferredChecks = (!isEmptyCheckArray(classComChecks) || !isEmptyListOfChecks(fieldComChecks))
```
aklish: I still feel like we should be solely paying attention to type of check - not the type of permission.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,500 @@
...
+    /**
+     * Check for permissions on a specific field.
+     *
+     * @param resource resource
+     * @param changeSpec changepsec
+     * @param annotationClass annotation class
+     * @param field field to check
+     * @param <A> type parameter
+     */
+    public <A extends Annotation> void checkFieldAwarePermissions(PersistentResource<?> resource,
```
aklish: I feel like we can keep these two entry functions, but they can both just call a helper function that takes a strategy and a set of fields (could be a single field) where permissions need to be checked.  That function will do more or less the same setup (extract the checks for the entity, extract the checks for field(s), call fieldAwareExecute with the strategy and the associated extracted checks.

elide-core/src/main/java/com/yahoo/elide/security/User.java
DeathByTape: :+1:

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,659 @@
...
+                                     ChangeSpec changeSpec,
+                                     boolean isUpdate) {
+        if (resource.getRequestScope().getSecurityMode() == SecurityMode.BYPASS_SECURITY) {
+            return;
+        }
+
+        for (Class<? extends Check> check : checks) {
+            CheckIdentifier checkId = new CheckIdentifier(resource, check);
+
+            boolean ok = (isUpdate) ? computeCheck(check, resource, changeSpec)
```
aklish: Instead of adding a new parameter here (isUpdate) - can we just tell from whether or not the changeSpec was populated.  If the changeSpec is populated - is that not the same?

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,659 @@
...
+     * @param resource Resource to compute check for
+     * @param changeSpec Change spec
+     * @return True if check successful, false otherwise.
+     */
+    private boolean computeCheck(Class<? extends Check> check, PersistentResource resource, ChangeSpec changeSpec) {
+        Check handler;
+        try {
+            handler = check.newInstance();
+        } catch (InstantiationException | IllegalAccessException e) {
+            log.debug("Illegal permission check '{}' {}", check.getName(), e);
```
aklish: error

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,659 @@
...
+            boolean ok = (isUpdate) ? computeCheck(check, resource, changeSpec)
+                : resourceCache.computeIfAbsent(checkId, (id) -> computeCheck(check, resource, changeSpec));
+
+            if (ok && checkMode == CheckMode.ANY) {
+                return;
+            }
+
+            if (!ok && checkMode == CheckMode.ALL) {
+                log.debug("ForbiddenAccess {} {}#{}", check, resource.getType(), resource.getId());
+                throw new ForbiddenAccessException();
```
aklish: Instead of throwing all of these exceptions, why don't we save that for the end when we know the issue is real (and progress should be stopped).  I think this should just return a boolean (pass/fail).

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,659 @@
...
+            CHECKS_SUCCEEDED,
+            CHECKS_FAILED
+        }
+    }
+
+    /**
+     * Strategy for checking a specific field.
+     */
+    private final class SpecificFieldStrategy implements Strategy {
+        private CheckStatus previousStatus;
```
aklish: I don't think we should do state between calls like this.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,659 @@
...
+                log.error("Problem occurred trying to initialize Specific Field Strategy: null field.");
+                throw new IllegalStateException();
+            }
+            populateChecks(field, annotationClass);
+            containsDeferredChecks = containsCommitChecks(classComChecks, fieldComChecks, annotationClass);
+            isUpdate = UpdatePermission.class.isAssignableFrom(annotationClass);
+        }
+
+        @Override
+        public void executeOperationChecks() {
```
aklish: Since the strategy doesn't change whether or not these are operation or commit checks, I don't think the strategy should differentiate this functions.   I think it should have a single method called execute that returns a boolean.  The calling code can decide whether to continue or not for operation checks based on whether deferred checks exist and whether the status of execute was pass/fail.

elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,659 @@
...
+     * @param resource resource
+     * @param changeSpec changepsec
+     * @param annotationClass annotation class
+     * @param field field to check
+     * @param <A> type parameter
+     */
+    public <A extends Annotation> void checkFieldAwarePermissions(PersistentResource<?> resource,
+                                                                  ChangeSpec changeSpec,
+                                                                  Class<A> annotationClass,
+                                                                  String field) {
```
aklish: I don't think we should pass in null here for field to mean anything.  I think we should just have a separate function that doesn't take a field (which implies all fields).

aklish:
Ignore my comments for now.  I'm going to send you an email.  I had a thought that is too complicated to explain here.  It is based on our future future direction.

yahoocla:
CLA is valid!

PR #98, opened 2016-01-08
Update deferred check logic

swathi9488:


aklish:
We need to add a new unit test for the failure case.

yahoocla:
CLA is valid!

aklish:
:+1:

PR #102, opened 2016-01-19
Initial commit for security revamp (Elide 2.0).

DeathByTape:
This PR is in place of: https://github.com/yahoo/elide/pull/94 (see https://github.com/yahoo/elide/issues/84 for discussion).


elide-core/src/main/java/com/yahoo/elide/security/PermissionManager.java
```diff
@@ -0,0 +1,202 @@
...
+
+import static com.yahoo.elide.security.permissions.ExpressionResult.DEFERRED;
+import static com.yahoo.elide.security.permissions.ExpressionResult.FAIL;
+import static com.yahoo.elide.security.permissions.ExpressionBuilder.Expressions;
+
+/**
+ * Class responsible for managing the life-cycle and execution of checks.
+ */
+@Slf4j
+public class PermissionManager {
```
aklish: Let's rename this as the PermissionExecutor or something :)

DeathByTape: Updated :) rebased commit.

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/SpecificFieldExpression.java
```diff
@@ -0,0 +1,33 @@
...
+package com.yahoo.elide.security.permissions.expressions;
+
+import com.yahoo.elide.security.permissions.ExpressionResult;
+
+import static com.yahoo.elide.security.permissions.ExpressionResult.PASS;
+
+import java.util.Optional;
+
+/**
+ * Expression for joining specific fields.
```
aklish: This could use some more documentation.

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/OrExpression.java
```diff
@@ -0,0 +1,47 @@
...
+        this.left = left;
+        this.right = right;
+    }
+
+    @Override
+    public ExpressionResult evaluate() {
+        ExpressionResult leftResult = left.evaluate();
+        ExpressionResult rightResult = (right == null) ? leftResult : right.evaluate();
+
+        if (leftResult == FAIL && (rightResult == FAIL || rightResult == null)) {
```
aklish: rightResult can't be null per statement above.

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/ImmediateCheckExpression.java
```diff
@@ -0,0 +1,52 @@
...
+     *
+     * @param check Check
+     * @param resource Persistent resource
+     * @param changeSpec ChangeSpec
+     * @param cache Cache
+     */
+    public ImmediateCheckExpression(final Check check,
+                                    final PersistentResource resource,
+                                    final ChangeSpec changeSpec,
+                                    final Map<Check, ExpressionResult> cache) {
```
aklish: How do you differentiate in the cache between two checks but annotated on different resources or with different changespecs?

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AnyFieldExpression.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+package com.yahoo.elide.security.permissions.expressions;
+
+import com.yahoo.elide.security.permissions.ExpressionResult;
+
+import static com.yahoo.elide.security.permissions.ExpressionResult.FAIL;
+import static com.yahoo.elide.security.permissions.ExpressionResult.PASS;
+
+/**
+ * Implementation of joining expression results by any field success or entity success.
```
aklish: A bit more documentaiton explaining when this is used.

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/ImmediateCheckExpression.java
DeathByTape: I don't-- good catch. This is an issue we talked about earlier too.

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AnyFieldExpression.java
```diff
@@ -0,0 +1,33 @@
...
+        this.fieldExpression = fieldExpression;
+    }
+
+    @Override
+    public ExpressionResult evaluate() {
+        ExpressionResult fieldResult = (fieldExpression == null) ? FAIL : fieldExpression.evaluate();
+        if (fieldResult != ExpressionResult.FAIL) {
+            return fieldResult;
+        }
+        return (entityExpression == null) ? PASS : entityExpression.evaluate();
```
aklish: If fieldExpression was FAIL and there was no entityExpression, would this return PASS?

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AndExpression.java
```diff
@@ -0,0 +1,47 @@
...
+    @Override
+    public ExpressionResult evaluate() {
+        ExpressionResult leftResult = left.evaluate();
+        ExpressionResult rightResult = (right == null) ? PASS : right.evaluate();
+
+        if (leftResult == FAIL || rightResult == FAIL) {
+            return FAIL;
+        }
+
+        if (leftResult == PASS && (rightResult == PASS || rightResult == null)) {
```
aklish: Do we need to check for null (already checked and set above)?

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+                                                                           final Class<A> annotationClass,
+                                                                           final String field,
+                                                                           final ChangeSpec changeSpec) {
+        ExtractedChecks extracted = new ExtractedChecks(resource, annotationClass, field);
+
+        Expression operationExpression =
+                buildFullExpression(extracted.getEntityChecks(),
+                                    extracted.getFieldChecks(),
+                                    (check) -> new DeferredCheckExpression(check, resource, changeSpec, cache),
+                                    SpecificFieldExpression::new);
```
aklish: Instead of passing in to the next function the 'expression joiner' here, it might make more sense to do that final expression joining here?

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AnyFieldExpression.java
DeathByTape: Yes. Will add a test for this and fix it in the logic

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AndExpression.java
DeathByTape: No, this is an artifact of the weird NoopExpression thing I had in there earlier.

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionBuilder.java
```diff
@@ -0,0 +1,203 @@
...
+                (check) -> new DeferredCheckExpression(check, resource, changeSpec, cache);
+        Function<Check, Expression> immediateCheckBuilder =
+                (check) -> new ImmediateCheckExpression(check, resource, changeSpec, cache);
+
+        List<String> fields = resource.getDictionary().getAllFields(resource.getObject());
+
+        ExtractedChecks entity = new ExtractedChecks(resource, annotationClass);
+        Expression opExp =
+                buildFullExpression(entity.getEntityChecks(),
+                                    entity.getFieldChecks(),
```
aklish: What are the field checks that are being built here?  It seems like the building for those actually happens in the for loop below.

elide-core/src/main/java/com/yahoo/elide/security/checks/ExtractedChecks.java
```diff
@@ -0,0 +1,167 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param resource Resource
+     * @param annotationClass Annotation class
+     * @param field Field
+     * @param <A> Type parameter
+     */
+    public <A extends Annotation> ExtractedChecks(final PersistentResource resource,
```
aklish: I think this might be cleaner if we break this up a bit.  I would have a class that includes the check(s) plus mode and maybe add static methods to build one of those classes given an annotation and an entity OR an annotation, entity,and field.

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AnyFieldExpression.java
DeathByTape: Yes it would. I think there is a question as to what the default "correct" logic for this case is.

In general, if nothing is specified, we consider a bean to be open, correct? In this case, we should ultimately return PASS since this is an `AnyFieldExpression` which determines whether there is access either on a specific field or the entire entity.

I do not believe this poses any real security risks, however, since `SpecificFieldExpressions` will be run for all the fields as they are serialized and before they are updated. I believe this is for other optimizations.

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionBuilder.java
DeathByTape: I thought about this, but then we need to return a structure containing two Expressions so we can join them at a higher-level. It seemed cleaner to me to simply pass in the function and return the combined result.

elide-core/src/main/java/com/yahoo/elide/security/checks/ExtractedChecks.java
DeathByTape: I have broken this up to the bare minimum of what we need (the general code is smaller now). Are you specifically referring to this constructor? This kind of code would likely either end up here or in that static method.

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionBuilder.java
```diff
@@ -0,0 +1,215 @@
...
+                                                                           final String field,
+                                                                           final ChangeSpec changeSpec) {
+        final ExtractedChecks extracted = new ExtractedChecks(resource, annotationClass, field);
+
+        final BiFunction<Expression, Expression, Expression> entityJoiner = getJoiner(extracted.getEntityCheckMode());
+        final BiFunction<Expression, Expression, Expression> fieldJoiner = getJoiner(extracted.getFieldCheckMode());
+        final Function<Check, Expression> deferredCheckFn =
+                 (check) -> new DeferredCheckExpression(check, resource, changeSpec, cache);
+        final Function<Check, Expression> immediateCheckFn =
+                (check) -> new DeferredCheckExpression(check, resource, changeSpec, cache);
```
aklish: Is this intended to be Deffered here?

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionBuilder.java
```diff
@@ -0,0 +1,328 @@
...
+            queue.add(value);
+        }
+        return queue;
+    }
+
+    /**
+     * Structure containing built expressions.
+     */
+    @AllArgsConstructor
+    public static class Expressions {
```
aklish: Not sure having the return types bundled here helps readability.  Maybe it would be better to for the builder to be passed in the provider or factory method that creates the underlying leaf expression objects - and then let the calling functions call this twice - once for operation and once for commit.

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/ImmediateCheckExpression.java
```diff
@@ -0,0 +1,88 @@
...
+        return resourceCache.get(resource);
+    }
+
+    /**
+     * Actually compute the result of the check without caching concerns.
+     *
+     * @return Expression result from the check.
+     */
+    private ExpressionResult computeCheck() {
+        if (resource == null) {
```
aklish: Use polymorphism here.

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionBuilder.java
```diff
@@ -0,0 +1,299 @@
...
+     * @param annotationClass Annotation class
+     * @param field Field to check (if null only check entity-level)
+     * @param <A> type parameter
+     * @return User check expression to evaluate
+     */
+    public <A extends Annotation> Expressions buildUserCheckFieldExpressions(final PersistentResource resource,
+                                                                             final Class<A> annotationClass,
+                                                                             final String field) {
+        final Function<Check, Expression> userCheckFn =
+                (check) -> new UserCheckOnlyExpression(check, resource, resource.getRequestScope(), null, cache);
```
aklish: cast nulls

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionBuilder.java
```diff
@@ -0,0 +1,299 @@
...
+     * @param annotationClass Annotation class
+     * @param requestScope Request scope
+     * @param <A> type parameter
+     * @return User check expression to evaluate
+     */
+    public <A extends Annotation> Expressions buildUserCheckAnyExpression(final Class<?> resourceClass,
+                                                                          final Class<A> annotationClass,
+                                                                          final RequestScope requestScope) {
+        final Function<Check, Expression> userCheckFn =
+                (check) -> new UserCheckOnlyExpression(check, null, requestScope, null, cache);
```
aklish: cast nulls for readability.

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionBuilder.java
```diff
@@ -0,0 +1,299 @@
...
+    private <A extends Annotation> Expression buildAnyFieldExpression(final Class<?> resourceClass,
+                                                                       final EntityDictionary dictionary,
+                                                                       final Class<A> annotationClass,
+                                                                       final Function<Check, Expression> checkFn) {
+        final List<String> fields = dictionary.getAllFields(resourceClass);
+
+        final ExtractedChecks entity = new ExtractedChecks(resourceClass, dictionary, annotationClass, null);
+
+        final BiFunction<Expression, Expression, Expression> entityClassJoiner = getJoiner(entity.getEntityCheckMode());
+        final BiFunction<Expression, Expression, Expression> entityFieldJoiner = getJoiner(entity.getFieldCheckMode());
```
aklish: Do we need this?

DeathByTape: Yes, see: https://github.com/yahoo/elide/pull/102/files#diff-01adfd7c58c229be01f3e188e23974e3R754

DeathByTape: Yes, see line L#214

elide-core/src/main/java/com/yahoo/elide/security/checks/ExtractedChecks.java
```diff
@@ -0,0 +1,151 @@
...
+        final A fieldAnnotation = (field == null) ? null
+                                                  : dictionary.getAttributeOrRelationAnnotation(
+                                                        cls,
+                                                        annotationClass,
+                                                        field);
+        // No checks specified
+        if (entityAnnotation == null) {
+            anyEntityChecks = null;
+            allEntityChecks = null;
+        } else {
```
aklish: This else blocks looks 'unrolled'.  Maybe another function here?  It might also help with cyclomatic complexity.

elide-core/src/main/java/com/yahoo/elide/security/checks/ExtractedChecks.java
```diff
@@ -0,0 +1,151 @@
...
+    public CheckMode getEntityCheckMode() {
+        return (anyEntityChecks != null && anyEntityChecks.length > 0) ? CheckMode.ANY : CheckMode.ALL;
+    }
+
+    public CheckMode getFieldCheckMode() {
+        return (anyFieldChecks != null && anyFieldChecks.length > 0) ? CheckMode.ANY : CheckMode.ALL;
+    }
+
+    public Class<? extends Check>[] getEntityChecks() {
+        return (anyEntityChecks != null && anyEntityChecks.length > 0) ? anyEntityChecks : allEntityChecks;
```
aklish: Should we use getEntityCheckMode here instead of duplicating the check logic?

elide-core/src/main/java/com/yahoo/elide/security/checks/ExtractedChecks.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    public Class<? extends Check>[] getFieldChecks() {
+        return (anyFieldChecks != null && anyFieldChecks.length > 0) ? anyFieldChecks : allFieldChecks;
```
aklish: Same comment as above.

elide-core/src/main/java/com/yahoo/elide/security/PermissionExecutor.java
```diff
@@ -0,0 +1,202 @@
...
+     * @param annotationClass Annotation class
+     * @param resourceClass Resource claSS
+     * @param dictionary Dictionary
+     * @param <A> type parameter
+     * @return Set of extracted checks
+     */
+    public static <A extends Annotation> ExtractedChecks loadEntityChecks(Class<A> annotationClass,
+                                                                          Class<?> resourceClass,
+                                                                          EntityDictionary dictionary) {
+        return new ExtractedChecks(resourceClass, dictionary, annotationClass, null);
```
aklish: Isn't there a version of this constructor which doesn't take a 'null'?

elide-core/src/main/java/com/yahoo/elide/security/PermissionExecutor.java
```diff
@@ -0,0 +1,202 @@
...
+     *
+     * @param annotationClass annotation class
+     * @param resource resource
+     * @param <A> type parameter
+     * @see com.yahoo.elide.annotation.CreatePermission
+     * @see com.yahoo.elide.annotation.ReadPermission
+     * @see com.yahoo.elide.annotation.UpdatePermission
+     * @see com.yahoo.elide.annotation.DeletePermission
+     */
+    public <A extends Annotation> void checkPermission(Class<A> annotationClass, PersistentResource resource) {
```
aklish: Let's consolidate the public methods to check permissions.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -8,15 +8,7 @@
...
-import com.yahoo.elide.annotation.Audit;
-import com.yahoo.elide.annotation.CreatePermission;
-import com.yahoo.elide.annotation.DeletePermission;
-import com.yahoo.elide.annotation.OnCreate;
-import com.yahoo.elide.annotation.OnDelete;
-import com.yahoo.elide.annotation.OnUpdate;
-import com.yahoo.elide.annotation.ReadPermission;
-import com.yahoo.elide.annotation.SharePermission;
-import com.yahoo.elide.annotation.UpdatePermission;
+import com.yahoo.elide.annotation.*;
```
aklish: No astericks plz :)

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -979,8 +973,8 @@ protected Object getValue(String fieldName) {
      * @param toAdd the to add
      */
     protected void addToCollection(Collection collection, String collectionName, PersistentResource toAdd) {
-        checkFieldAwarePermissions(UpdatePermission.class, collectionName);
-        toAdd.checkFieldAwarePermissions(ReadPermission.class, collectionName);
+        checkFieldAwarePermissions(UpdatePermission.class, collectionName, toAdd, null);
```
aklish: The permission should be looking at collections (not null) for the changespec.  This is a general comment for wherever we should be passing a collection as the change spec

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -380,7 +373,7 @@ protected boolean updateToManyRelation(String fieldName, Set<PersistentResource>
         deleted
                 .stream()
                 .forEach(toDelete -> {
-                    checkFieldAwarePermissions(UpdatePermission.class, fieldName);
+                    checkFieldAwarePermissions(UpdatePermission.class, fieldName, null, toDelete);
```
aklish: I feel like this check is unnecessary and should be removed.  We already check the field permission in the calling function.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -452,8 +445,8 @@ protected boolean updateToOneRelation(String fieldName, Set<PersistentResource>
      * @return True if object updated, false otherwise
      */
     public boolean clearRelation(String relationName) {
-        checkFieldAwarePermissions(UpdatePermission.class, relationName);
         Set<PersistentResource> mine = getRelation(relationName);
+        checkFieldAwarePermissions(UpdatePermission.class, relationName, null, mine);
```
aklish: This should be checking an empty set instead of null.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -472,7 +465,7 @@ public boolean clearRelation(String relationName) {
             Collection collection = (Collection) this.getValue(relationName);
             mine.stream()
                     .forEach(toDelete -> {
-                        checkFieldAwarePermissions(UpdatePermission.class, relationName);
+                        checkFieldAwarePermissions(UpdatePermission.class, relationName, null, toDelete);
```
aklish: This check seems redundant.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -492,7 +485,7 @@ public boolean clearRelation(String relationName) {
      * @param removeResource the remove resource
      */
     public void removeRelation(String fieldName, PersistentResource removeResource) {
-        checkFieldAwarePermissions(UpdatePermission.class, fieldName);
+        checkFieldAwarePermissions(UpdatePermission.class, fieldName, null, removeResource);
```
aklish: changespec varies depending on whether relation is a collection or not.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -527,7 +520,7 @@ public void removeRelation(String fieldName, PersistentResource removeResource)
      * @param newRelation the new relation
      */
     public void addRelation(String fieldName, PersistentResource newRelation) {
-        checkFieldAwarePermissions(UpdatePermission.class, fieldName);
+        checkFieldAwarePermissions(UpdatePermission.class, fieldName, newRelation, null);
```
aklish: The existing value may have been a collection and not null.  Again we need to look at the type of teh relationship.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -968,7 +962,7 @@ protected void nullValue(String fieldName, PersistentResource oldValue) {
      * @return value value
      */
     protected Object getValue(String fieldName) {
-        checkFieldAwarePermissions(ReadPermission.class, fieldName);
+        checkFieldAwarePermissions(ReadPermission.class, fieldName, null, null);
```
aklish: Cast?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1151,7 +1145,7 @@ protected void deleteInverseRelation(String relationName, Object relationValue)
             }
 
             if (inverseRelation instanceof Collection) {
-                inverseResource.checkFieldAwarePermissions(UpdatePermission.class, inverseRelationName);
+                inverseResource.checkFieldAwarePermissions(UpdatePermission.class, inverseRelationName, null, this);
```
aklish: Move to delFromCollection

aklish:
:+1:

yahoocla:
CLA is valid!

Issue #105, opened 2016-01-21
POST to ToOne relationship with existing element fails


DeathByTape:
a `POST` to a `ToOne` relationship where an element already exists fails. It drops into `RecordTerminalState` rather than `CollectionTerminalState`.

We expect the behavior of this would be a full replacement of the existing `ToOne` relationship with the new one.

DeathByTape:
Done in #108

PR #108, opened 2016-01-25
Fixed posting to to-one relationships.

DeathByTape:
PR addressing https://github.com/yahoo/elide/issues/105

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
+++ b/elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
@@ -10,6 +10,7 @@ import com.yahoo.elide.core.PersistentResource;
 import com.yahoo.elide.core.RequestScope;
 import com.yahoo.elide.core.exceptions.ForbiddenAccessException;
 import com.yahoo.elide.core.exceptions.InvalidEntityBodyException;
+import com.yahoo.elide.core.exceptions.InvalidOperationException;
 import com.yahoo.elide.jsonapi.document.processors.DocumentProcessor;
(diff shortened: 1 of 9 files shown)


elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
```diff
@@ -37,9 +37,15 @@
 @ToString
 public class RecordTerminalState extends BaseState {
     private final PersistentResource record;
+    private final Optional<CollectionTerminalState> collectionTerminalState;
 
     public RecordTerminalState(PersistentResource record) {
+        this(record, null);
+    }
+
+    public RecordTerminalState(PersistentResource record, CollectionTerminalState collectionTerminalState) {
```
wcekan: Add a comment like "Handle ToOne collection as a Record terminal state"

elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
```diff
@@ -49,6 +55,14 @@ public RecordTerminalState(PersistentResource record) {
     }
 
     @Override
+    public Supplier<Pair<Integer, JsonNode>> handlePost(StateContext state) {
+        if (collectionTerminalState.isPresent()) {
+            return collectionTerminalState.get().handlePost(state);
+        }
+        throw new IllegalStateException("Cannot POST to a record.");
```
wcekan: Should we make a new Illegal Request exception to return error code instead of 500?

wcekan:
:+1: LGTM

yahoocla:
CLA is valid!

PR #112, opened 2016-01-28
Add new SecurityMode

clayreimann:
SecurityMode.VERBOSE_ERRORS returns information about the particular
security failure that caused the request to fail.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/SecurityMode.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/SecurityMode.java
@@ -9,7 +9,7 @@ package com.yahoo.elide.core;
  * Enable or disable security.
  */
 public enum SecurityMode {
-    BYPASS_SECURITY,
-    VERBOSE_ERRORS,
-    ACTIVE
+    SECURITY_INACTIVE,
+    SECURITY_ACTIVE_VERBOSE,
+    SECURITY_ACTIVE
 }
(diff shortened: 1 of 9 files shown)


elide-core/src/main/java/com/yahoo/elide/core/SecurityMode.java
```diff
@@ -10,5 +10,6 @@
  */
 public enum SecurityMode {
     BYPASS_SECURITY,
+    VERBOSE_ERRORS,
```
DeathByTape: maybe `ACTIVE_VERBOSE` or `ACTIVE_WITH_ERRORS`? We should try to maintain some consistency so we don't have any confusion on what security modes are doing.

elide-core/src/main/java/com/yahoo/elide/core/exceptions/HttpStatusException.java
```diff
@@ -37,10 +38,12 @@ public HttpStatusException(String message, Throwable cause) {
         super(message, cause, true, log.isTraceEnabled());
     }
 
-    public Pair<Integer, JsonNode> getErrorResponse() {
+    public Pair<Integer, JsonNode> getErrorResponse(SecurityMode securityMode) {
+        boolean useProvidedMessage = getMessage() == null || securityMode == SecurityMode.VERBOSE_ERRORS;
```
DeathByTape: I don't think `getMessage() == null` is what we want here. This actually modifies the semantics of what we had. Anywhere we had a valid message before will now be swallowed since we determined whether or not to print a message iff `getMessage() != null`.

DeathByTape:
:+1:

yahoocla:
CLA is valid!

Issue #96, opened 2016-01-08
Support filtering by id


jkusa:
Currently, you can not execute a GET request and filter entities by `id`. This case is especially useful when performing an filter with an `in` operator. For example, requesting `comments` where `id ` is `in` `[1,5,7,8]`.

mcasperson:
+1

DeathByTape:
I will probably start working at getting this and #97 in as soon as we finish up the reworked security. We should (hopefully) be closing that up this week.

In the meantime, if anyone has some spare cycles to send a PR on this, that's also welcome :)

DeathByTape:
Fixed in https://github.com/yahoo/elide/pull/114

PR #114, opened 2016-02-01
Added support for ID filtering.

DeathByTape:
Filtering by id's.
PR addresses issue: https://github.com/yahoo/elide/issues/96


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -666,9 +667,16 @@ public Boolean isIdGenerated() {
...
-        Predicate idFilter = new Predicate(relation, Operator.IN, Collections.singletonList(id));
-        Set<Predicate> filters = Collections.singleton(idFilter);
+        Set<Predicate> filters;
+        // TODO: UUID's are troublesome here from Patch Extension. We should consider an alternative approach for this.
+        if (requestScope instanceof PatchRequestScope) {
+            filters = Collections.emptySet();
+        } else {
+            Object idVal = CoerceUtil.coerce(id,
+                    dictionary.getIdType(dictionary.getParameterizedType(getResourceClass(), relation)));
+            Predicate idFilter = new Predicate("id", Operator.IN, Collections.singletonList(idVal));
```
DeathByTape: Didn't denote, but this should be fixed. We probably shouldn't hardcode the text `id`.

wcekan:
can I also do 
```java
.get("/parent?include=children&filter[child.id]=4")
```

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
clayreimann: We could achieve this with `dictionary.getIdField` which could just return `entityBinding(value.getClass()).getIdField()`

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -186,7 +186,7 @@ public void commit() {
 
     @Override
     public <T> Collection filterCollection(Collection collection, Class<T> entityClass, Set<Predicate> predicates) {
-        if (collection instanceof PersistentBag && !predicates.isEmpty()) {
+        if ((collection instanceof AbstractPersistentCollection) && !predicates.isEmpty()) {
```
clayreimann: Why are we making this change?

DeathByTape: Because it was too restrictive and was hiding `Predicate` construction errors (i.e. we always constructed bad `Predicate`s for `PersistentSet`s). `PersistentBag != PersistentSet`, for instance. However, any sort of collection we can receive that is an `AbstractPersistentCollection` coming from Hibernate should work.

DeathByTape:
@wcekan I added such a test.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
DeathByTape: Good idea! implemented now.

aklish:
:+1:

PR #121, opened 2016-02-04
Don't accidentally return verbose message when not in verbose mode

clayreimann:


elide-core/src/main/java/com/yahoo/elide/core/exceptions/HttpStatusException.java
```diff
@@ -40,20 +39,28 @@ public HttpStatusException(String message, Throwable cause) {
     }
 
     public Pair<Integer, JsonNode> getErrorResponse(SecurityMode securityMode) {
-        boolean useProvidedMessage = shouldUseProvidedMessage(securityMode);
         Map<String, List<String>> errors = Collections.singletonMap(
                 "errors",
-                Collections.singletonList(useProvidedMessage ? getMessage() : toString())
+                Collections.singletonList(
+                        securityMode == SecurityMode.SECURITY_ACTIVE_VERBOSE
```
DeathByTape: I don't think we should need `SecurityMode` here; this should only contain methods rather than controlling how messages are exposed.

My suggestion is something as follows:

```
public class HttpStatusException ... {
    private final String verboseMessage;

    public HttpStatusException() { ... }
    public HttpStatusException(String message) { ... }

    // New verboseMessage constructor
    public HttpStatusException(String message, String verboseMessage) {
        super(message, verboseMessage, null);
    } 

     /* fast exceptions */
     // UPDATE This constructor and modify others accordingly:
     public HttpStatusException(String message, String verboseMessage, Throwable cause) {
         super(message, cause, true, log.isTraceEnabled());
         this.verboseMessage = verboseMessage;
      }

     ...

    public String getVerboseMessage() { return verboseMessage; }
}
```

If you have this, the logic whether or not to display a verbose message can be done entirely at the higher level.

DeathByTape:
:+1:

yahoocla:
CLA is valid!

PR #130, opened 2016-02-11
Use Class.getMethod instead of Class.getMethods

wcekan:
Performance enhancement for EntitiyDictionary.findMethod


aklish:
:+1:

yahoocla:
CLA is valid!

PR #135, opened 2016-02-18
Ignored entity fields pointing to excluded types.

DeathByTape:


elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -207,8 +207,9 @@ private void bindAttrOrRelation(Class<?> cls, AccessibleObject fieldOrMethod) {
         String fieldName = getFieldName(fieldOrMethod);
 
         if (fieldName == null || fieldName.equals("id")
-                ||  fieldName.equals("class") || OBJ_METHODS.contains(fieldOrMethod)) {
-            return; // Reserved. Not attributes.
+                ||  fieldName.equals("class") || OBJ_METHODS.contains(fieldOrMethod)
+            || EntityDictionary.getFirstAnnotation(getFieldType(fieldOrMethod), Arrays.asList(Exclude.class)) != null) {
```
aklish: We should get the parameterized type if this is a collection.  Let's add a test for this case too.

aklish:
:+1:

yahoocla:
CLA is valid!

PR #139, opened 2016-02-19
Consolidating logging auth failures into a single place

aklish:


elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -111,7 +114,11 @@ public ElideResponse get(
             auditLogger.commit();
             transaction.commit();
             requestScope.runCommitTriggers();
+            logSecurityExceptions(requestScope);
             return response;
+        } catch (ForbiddenAccessException e) {
```
wcekan: Should we handle this case once in buildErrorResponse

elide-core/src/main/java/com/yahoo/elide/core/exceptions/ForbiddenAccessException.java
```diff
@@ -15,8 +16,15 @@
 public class ForbiddenAccessException extends HttpStatusException {
     private static final long serialVersionUID = 1L;
 
-    public ForbiddenAccessException(String verboseMessage) {
+    private RequestScope scope;
```
wcekan: Should be transient

wcekan:
:+1: No problem.  We can clean up in future PR.

yahoocla:
CLA is valid!

PR #140, opened 2016-02-19
Avoid accessing lazy relationship collection if entity is denied.

wcekan:
Move `isDeny` filter check ahead of the `getValue` to avoid unneccessiarily  hydrating a lazy collection


DeathByTape:
:+1:

yahoocla:
CLA is valid!

PR #148, opened 2016-02-24
Faster Serializer/Deserializer for JSON API

wcekan:
Allow caller to provide the JSON API Mapper.
Use convertValue instead of deserialize-reserialize of the JsonNode.


DeathByTape:
:+1:

yahoocla:
CLA is valid!

PR #149, opened 2016-02-24
first pass for pagination

newfront:
@DeathByTape @aklish Here is the first pass for pagination. --DO NOT MERGE--

Assumptions
1. Pagination is currently for SQL style data stores.

2. Pagination is for top level collections [otherwise what would the suggestion be for doing sub-entity pagination, while maintaining sort order of child entities?]

3. Pagination should work on its own, or as the last operation on a query [unless we honor sub-ordering and limitations which MySQL doesn't honor in all cases [https://dev.mysql.com/doc/refman/5.7/en/subquery-restrictions.html]

4. Pagination should honor sorting, and take the sort params / parse them and add into the Order items on the Session object. [ TO DO]


elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -0,0 +1,93 @@
...
+                .forEach(paramEntry -> {
+                    // looking for page[size], page[limit]
+                    // looking for page[offset], page[number]
+                    final String queryParamKey = paramEntry.getKey();
+                    if (PAGE_KEYS.containsKey(queryParamKey)) {
+                        final String type = PAGE_KEYS.get(queryParamKey);
+                        final String value = paramEntry.getValue().get(0); // is this correct...
+                        try {
+                            pageData.put(type, Integer.parseInt(value, 10));
+                        } catch (ClassCastException e) {
```
aklish: We should throw a HTTPStatusException with a 400 error.  You could either sublass or use InvalidValueException here.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -0,0 +1,93 @@
...
+public class Pagination {
+
+    public static final int DEFAULT_PAGE_SIZE = 500;
+    // todo do we ever want to return more than 10000 items? Set Upper Bounds
+
+    private static final Pagination DEFAULT_PAGINATION = new Pagination(0, 0);
+
+    public static final Map<String, String> PAGE_KEYS = new HashMap<>();
+    static {
+        PAGE_KEYS.put("page[size]", "pageSize");
```
aklish: Are we going to support both page and offset based paging or just one?

elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java
```diff
@@ -71,8 +71,12 @@ public CollectionTerminalState(Class<?> entityClass, Optional<PersistentResource
         DocumentProcessor includedProcessor = new IncludedProcessor();
         includedProcessor.execute(jsonApiDocument, collection, queryParams);
 
-        DocumentProcessor sortProcessor = new SortProcessor();
-        sortProcessor.execute(jsonApiDocument, collection, queryParams);
+        // if we are using pagination, then we are adding sorting directly into the HQL, etc
+        // todo - is this correct?
+        if (requestScope.getPagination().isDefault()) {
```
aklish: I think the sorting should move out of the Document Processor and into the DataStore.  InMemoryDataStore can preserve some of the functionality of this processor.  Hibernate can then add the sort to the criteria.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -150,6 +154,7 @@ protected RequestScope(JsonApiDocument jsonApiDocument, RequestScope outerReques
         this.queryParams = Optional.empty();
         this.sparseFields = Collections.emptyMap();
         this.predicates = Collections.emptyMap();
+        this.pagination = Pagination.getDefaultPagination();
```
aklish: I'm wondering if we want to reject pagination, sort, include, and sparse for all queries except for GET.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -0,0 +1,93 @@
...
+                            // todo - should we log here?
+                        }
+                    }
+                });
+
+        if (pageData.isEmpty()) {
+            return DEFAULT_PAGINATION;
+        }
+        if (!pageData.containsKey("page")) {
+            pageData.put("page", 1);
```
aklish: I would make "page" and "pageSize" an enum

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -113,6 +114,33 @@ public void commit() {
...
 
+    /**
+     * Internal loadObjects helper - loads pagination objects
+     * @param loadClass The class Type reference
+     * @param criterion The optional Criterion
+     * @param pagination The Pagination object
+     * @param <T> The class type
+     * @return The ScrollableIterator
+     */
+    public <T> Iterable<T> loadObjects(Class<T> loadClass, Criterion criterion, Pagination pagination) {
```
aklish: I think it makes sense (given Bill's join work) to only worry about pagination inside the root collection (and not the includes)).  Basically, I agree with only adding support for now at loadObjects.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
newfront: @aklish Yes. This shouldn't be part of any other query other than get. I can add into this PR, or move to another PR so this is a scoped PR.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
newfront: sure.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
newfront: true. I will add that functionality as well. The first pass was simple, the next pass will include the /author/id/books?page[number]=2 functionality.

elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java
newfront: Yep. For the first pass I just added the notes and the if condition to make sure I would have access to appropriate variables, etc. I will update the sorting today.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
newfront: @aklish The idea was to support everything other than cursor for datastore's with concrete indicies. So whether you add just page[number]=2, then it would start at 500-1000 if using default page size (not sure if the default page size should be 500 or if that should move to the 10,000 we discussed), if you add page[offset]=50,page[limit]=50, then you would get 50-100 from the records, etc.

newfront: I followed http://jsonapi.org/format/#fetching-pagination for the page fields. All but cursor.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -702,11 +704,18 @@ public PersistentResource getRelation(String relation, String id) {
...
     public Set<PersistentResource> getRelation(String relationName) {
-        if (requestScope.getTransaction() != null && !requestScope.getPredicates().isEmpty()) {
-            final Class<?> entityClass = dictionary.getParameterizedType(obj, relationName);
-            final String valType = dictionary.getBinding(entityClass);
-            final Set<Predicate> filters = new HashSet<>(requestScope.getPredicatesOfType(valType));
-            return getRelation(relationName, filters);
+        // if pagination, then sorting with filters...
+        boolean hasFilters = !requestScope.getPredicates().isEmpty();
+        Set<Predicate> filters = null;
+        if (requestScope.getTransaction() != null && requestScope.getQueryParams().isPresent()) {
```
newfront: This feels to me like it could be a trap for future query param use cases. I feel like there should be a way for us to check to see if this is a get request with query params that has a transaction (eg. bound to data store transactions).

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -760,6 +768,49 @@ public PersistentResource getRelation(String relation, String id) {
...
+        Set<PersistentResource<Object>> resources = Sets.newLinkedHashSet();
+
+        // check for deny access on relationship to avoid iterating a lazy collection
+        if (isDenyFilter(requestScope, dictionary.getParameterizedType(obj, relationName))) {
+            checkFieldAwarePermissions(ReadPermission.class, relationName);
+            return (Set) resources;
+        }
+
+        RelationshipType type = getRelationshipType(relationName);
+        Object val = this.getValue(relationName);
```
newfront: Quick question for performance. If you walk this breakpoint, you will find that all relations to the parent object are loaded into Hibernate memory. In the case of /author/{id}/books?params I saw all books by Issac Asimov get loaded prior to actually running the downstream HQL. Not sure if we are actually loading things twice, or if I was walking too deeply into the Hibernate rabbit hole.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -0,0 +1,108 @@
...
+@ToString
+public class Pagination {
+
+    /**
+     * Denotes the internal field names for paging.
+     */
+    public enum PaginationKey { page, pageSize }
+
+    public static final int DEFAULT_PAGE_SIZE = 500;
+    public static final int MAX_PAGE_SIZE = 10000;
```
newfront: Do we want to actually limit max page size to 10000 or should there be some other means of limiting max batches? I haven't added the rule for this yet since it feels like a gigantic max page size. @aklish do you recommend enforcing at this level, or should we add max query size support at a config level?

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -0,0 +1,108 @@
...
+                .forEach(paramEntry -> {
+                    final String queryParamKey = paramEntry.getKey();
+                    if (PAGE_KEYS.containsKey(queryParamKey)) {
+                        final String value = paramEntry.getValue().get(0);
+                        try {
+                            pageData.put(PAGE_KEYS.get(queryParamKey), Integer.parseInt(value, 10));
+                        } catch (ClassCastException e) {
+                            throw new InvalidValueException("page values must be integers");
+                        }
+                    } else if (queryParamKey.startsWith("page[")) {
```
newfront: is this okay for an error message? Or does this stating too much? I prefer when an API or SDK informs you how to fix things in the error messages, but want to follow suite with best practices here.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    /**
+     * Given the sorting rules validate sorting rules against the entities bound to the entityClass
+     * @param entityClass  The root class for sorting (eg. /book?sort=-title this would be package.Book)
+     * @param dictionary The elide entity dictionary
+     * @param <T> The entityClass
+     * @return The valid sorting rules - validated through the entity dictionary
+     */
+    public <T> Map<String, SortOrder> getValidSortingRules(final Class<T> entityClass,
```
newfront: should we throw exception here? and allow to bubble up?

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -0,0 +1,100 @@
...
+                                                           final EntityDictionary dictionary) {
+        final Map<String, SortOrder> validSortRules = new LinkedHashMap<>();
+        final List<String> entities = dictionary.getAttributes(entityClass);
```
newfront: does the dictionary.getAttributes(Class<T> class) always return something? I was going off of that assumption, eg. always returns a collection, even if empty, if not I need to add null check here or exception thrown. I will add unit test for this case to ensure things work in broken env.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -0,0 +1,100 @@
...
+        entities.stream()
+                .filter(sortRules::containsKey)
+                .forEachOrdered(entity -> validSortRules.put(entity, sortRules.get(entity)));
```
newfront: ensuring ordering as that is staple of json-api ordering rules.

elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java
```diff
@@ -71,8 +71,11 @@ public CollectionTerminalState(Class<?> entityClass, Optional<PersistentResource
         DocumentProcessor includedProcessor = new IncludedProcessor();
         includedProcessor.execute(jsonApiDocument, collection, queryParams);
 
-        DocumentProcessor sortProcessor = new SortProcessor();
-        sortProcessor.execute(jsonApiDocument, collection, queryParams);
+        // if we are using pagination, then we have also applied sorting rules directly into the HQL, etc
+        if (requestScope.getPagination().isEmpty()) {
```
newfront: I am not sure if this is the best condition. Basically, the intention was to only skip SortProcessing on the returned data set in the case where we were not paginating the data, since for HQL, or Criterion we can do this on the MySQL level (mariaDB, etc) and therefore take advantage of the C level system for sorting and other indexing, however, when should we fallback to using the document sorting? Redis? non-SQL dbs? Is there a good way to ensure this condition is met for the correct reasons?

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
```diff
@@ -543,7 +541,7 @@ public void testGetRelationFilteredSuccess() {
         Assert.assertEquals(results.size(), 2, "Only filtered relation elements should be returned.");
     }
 
-    @Test
+    /*@Test
```
newfront: **** IMPORTANT*****
This test expects explicit filtering on the Document. No matter what I did I couldn't get this to pass. Since the DataStoreTransaction is mocked in this test I was under the assumption that this issue was a direct result of my updating the PersistentResource to use Hibernate / JPA for sorting when a transaction is present. Therefore I am not 100% certain of the best approach here. @aklish @DeathByTape not sure if you have a recommendation for this one, basically I don't want to cause a butterfly effect but I am 90% certain that this test is useless now that things are being done on the actual persistence layer.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -88,7 +85,7 @@ public void commit() {
             T object = entityClass.newInstance();
             deferredTasks.add(() -> session.persist(object));
             return object;
-        } catch (InstantiationException | IllegalAccessException e) {
+        } catch (java.lang.InstantiationException | IllegalAccessException e) {
```
newfront: There was a direct collision with imports due to optimization of imports for hibernate. Therefore I had to explicitly namespace the class path here. Same with hibernate5

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
newfront: The sub-collections are also now supported.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -122,15 +146,61 @@ public void commit() {
...
-            return loadObjects(loadClass);
+
+        final Pagination pagination = filterScope.hasPagination() ? filterScope.getRequestScope().getPagination()
+                : null;
+
+        // if we have sorting and sorting isn't empty, then we should pull dictionary to validate the sorting rules
+        Set<Order> validatedSortingRules = null;
+        if (filterScope.hasSortingRules()) {
+            final Sorting sorting = filterScope.getRequestScope().getSorting();
+            final Map<String, Sorting.SortOrder> validSortingRules = sorting.getValidSortingRules(loadClass,
```
newfront: This was moved directly into the Sorting.java. I am not sure if we want to support data store dependent code in Sorting since it is part of elide.core, so I moved this here. I made a best effort to keep everything in elide.core 100% generic.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -122,15 +146,61 @@ public void commit() {
...
+    /**
+     * Generates the Hibernate ScrollableIterator for Hibernate Query.
+     * @param loadClass The hibernate class to build the query off of.
+     * @param criterion The Optional criterion object.
+     * @param sortingRules The possibly empty sorting rules.
+     * @param pagination The Optional pagination object.
+     * @param <T> The return Iterable type.
+     * @return The Iterable for Hibernate.
+     */
+    public <T> Iterable<T> loadObjects(final Class<T> loadClass, final Optional<Criterion> criterion,
```
newfront: This method could literally be the only loadObjects code. However at the end of the day, outside of GET requests, this would be generating garbage in terms of Optional.empty(), so I kept the other overloads.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -206,6 +249,67 @@ public void commit() {
...
+                    );
+                    additionalHQL += "order by " + StringUtils.join(ordering, ",");
+                }
+                if (!additionalHQL.isEmpty()) {
+                    filterString += additionalHQL;
+                }
+            }
+
+            if (filterString.length() != 0) {
+                Query query = session.createFilter(collection, filterString);
```
newfront: regardless of actual filters (eg. select 'fields' from Table WHERE x in y, etc) we can still support ordering if nothing else here, so filter keyword is a bit ambiguous since we have Set<Predicate> filters, and then Sorting which is also a filter, but separate from HQL filters and the notion of Criterion filters, so just want to ensure things here make sense.

newfront:
Aside from all tests passing. I noticed that the relationships to authors are missing for the books, so sorting filtering and pagination work but somehwhere along the line we lose the parent relationship. I will investigate more, not sure if it is easier for you guys to see where this occurs.

yahoocla:
CLA is valid!

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -6,33 +6,32 @@
...
 import org.hibernate.criterion.Restrictions;
 import org.hibernate.resource.transaction.spi.TransactionStatus;
 
 import java.io.IOException;
 import java.io.Serializable;
-import java.util.Collection;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Set;
+import java.util.*;
```
aklish: I wonder if we need a checkstyle for wildcards.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -207,6 +252,71 @@ public void commit() {
...
+                                                         final Optional<Sorting> sorting,
+                                                         final Optional<Pagination> pagination) {
+        if (((collection instanceof AbstractPersistentCollection))
+                && (filters.isPresent() || sorting.isPresent() || pagination.isPresent())) {
+
+            String filterString = "";
+
+            // apply filtering - eg where clause's
+            if (filters.isPresent()) {
+                filterString += hqlFilterOperation.applyAll(filters.get());
```
aklish: This might have been a bug already in Elide, but we have a global hqlFilterOperation per transaction.  PersistentResource.getRelation will be called multiple times for different collections.  Filters will keep getting added to it (that really don't apply).    I think we should create a new hqlFilterOperation for each filter operation.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -207,6 +252,71 @@ public void commit() {
...
+            }
+
+            // add sorting into HQL string query generation
+            if (sorting.isPresent() && !sorting.get().isEmpty()) {
+
+                final Map<String, Sorting.SortOrder> validSortingRules = sorting.get().getValidSortingRules(
+                        entityClass, dictionary
+                );
+                String additionalHQL = "";
+                if (!validSortingRules.isEmpty()) {
```
aklish: You don't technically need to test isEmpty() here.  Streaming over an empty entrySet should be a noop.  Anyway, that would have the added benefit of reducing the cyclomatic complexity of the function.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -207,6 +252,71 @@ public void commit() {
...
+
+                    final List<String> ordering = new ArrayList<>();
+                    // pass over the sorting rules
+                    validSortingRules.entrySet().stream().forEachOrdered(entry ->
+                            ordering.add(entry.getKey() + " " + (entry.getValue().equals(Sorting.SortOrder.desc)
+                                    ? "desc"
+                                    : "asc"))
+                    );
+                    additionalHQL += "order by " + StringUtils.join(ordering, ",");
```
aklish: I'm wondering if we should do a Builder pattern here.  I would consider repurposing the HqlFilterOperation (maybe renamed to HqlBuilder) add add methods like: (addSorting, addFilter, addPagination).  This would also break up this function into smaller bits.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -123,15 +122,61 @@ public void commit() {
...
+    /**
+     * Generates the Hibernate ScrollableIterator for Hibernate Query.
+     * @param loadClass The hibernate class to build the query off of.
+     * @param criterion The Optional criterion object.
+     * @param sortingRules The possibly empty sorting rules.
+     * @param pagination The Optional pagination object.
+     * @param <T> The return Iterable type.
+     * @return The Iterable for Hibernate.
+     */
+    public <T> Iterable<T> loadObjects(final Class<T> loadClass, final Optional<Criterion> criterion,
```
aklish: Nice

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
aklish: I think so.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -0,0 +1,100 @@
...
+        return validSortRules;
+    }
+
+    public boolean isEmpty() {
+        return this.sortRules.isEmpty();
+    }
+
+    /**
+     * Given an entity dictionary and some query params, collect sorting rules.
```
aklish: This currently doesn't take a dictionary.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
aklish: I think that is a reasonable assumption.  If it doesn't, I would fix getATtributes.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -123,15 +122,61 @@ public void commit() {
...
+        final Pagination pagination = filterScope.hasPagination() ? filterScope.getRequestScope().getPagination()
+                : null;
+
+        // if we have sorting and sorting isn't empty, then we should pull dictionary to validate the sorting rules
+        Set<Order> validatedSortingRules = null;
+        if (filterScope.hasSortingRules()) {
+            final Sorting sorting = filterScope.getRequestScope().getSorting();
+            final Map<String, Sorting.SortOrder> validSortingRules = sorting.getValidSortingRules(loadClass,
+                    filterScope.getRequestScope().getDictionary());
+            if (!validSortingRules.isEmpty()) {
```
aklish: I would think if filterScope.hasSortingRules is true, validSortingRules should not be empty.  I guess if we throw an exception when we parse the sort query params, we won't need this extra check here.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
aklish: I think that is the right call.  We don't want Hibernate dependency in elide.core.

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
aklish: I'm puzzled why this test would fail.  Did it throw an exception or fail on an assert?  I agree that it doesn't seem like it is testing much, but it concerns me that it is failing.

elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java
aklish: I think we should just get rid of the Sort DocumentProcessor and instead move that functionality into the InMemoryDataStore.  I don't think you have to bite off fixing the InMemoryDataStore in this PR - but you could remove the processor and this check.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
aklish: This probably should be server specific (so some kind of configuration to Elide).

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
aklish: I think it is AOK.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -0,0 +1,108 @@
...
+    public enum PaginationKey { page, pageSize }
+
+    public static final int DEFAULT_PAGE_SIZE = 500;
+    public static final int MAX_PAGE_SIZE = 10000;
+
+    private static final Pagination DEFAULT_PAGINATION = new Pagination(0, 0);
+
+    public static final Map<String, PaginationKey> PAGE_KEYS = new HashMap<>();
+    static {
+        PAGE_KEYS.put("page[size]", PaginationKey.pageSize);
```
aklish: So everything needs to map to Hibernate's criteria.setMaxResults and criteria.setFirstResult (which are really identical to limit and offset respectively).  If the user specifies number and size, we need to do some math to convert: Namely, offset = number*size.  limit=size.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
aklish: I don't think you need to check for a transaction.  I can't think of a reason transaction would ever be null.

aklish: I wouldn't worry about checking for GET.  It is something that should be fixed, but I think it is outside the scope of pagination.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -702,11 +704,17 @@ public PersistentResource getRelation(String relation, String id) {
      * @return collection relation
      */
     public Set<PersistentResource> getRelation(String relationName) {
-        if (requestScope.getTransaction() != null && !requestScope.getPredicates().isEmpty()) {
-            final Class<?> entityClass = dictionary.getParameterizedType(obj, relationName);
-            final String valType = dictionary.getBinding(entityClass);
-            final Set<Predicate> filters = new HashSet<>(requestScope.getPredicatesOfType(valType));
-            return getRelation(relationName, filters);
+        // if pagination, then sorting with filters...
+        boolean hasFilters = !requestScope.getPredicates().isEmpty();
```
aklish: we need to check if there are filters for this particular type.  If you look at the elide docs, elide filters are a bit more expressive and allow filtering things like includes as well as the collection.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
DeathByTape: Should we incorporate an Elide configuration object for setting parameters such as this? That would give us room to grow and expand similar features.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
newfront: valid point. I forgot to add the compounding.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
newfront: I can change this to isPresent, or isDefault. I didn't want to be passing nulls back to the RequestScope, that is why I have the static final DEFAULT_INSTANCE, but needed a decent means to denote it was default - so I guess isDefault is probably good.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
newfront: sure. yeah, it is same as ```for (val : emptyCollection)```. However, since this is an optional for sorting, I didn't want to do additional work and create unnecessary garbage in the case of the default sorting internal map.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
newfront: Sure. I will remove the transaction check there. I was thinking that was the way things were being checked before.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
newfront: @aklish @DeathByTape How would I go about adding this? Where can I look to find the configuration object? Is it yml file? app.properties file? or is it Spring style ```@Configuration``` style overload?

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
newfront: perfect.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
newfront: sure. I could do this.

~~~java
if (!sorting.hasValidSortingRules(loadClass, dictionary)) {
    throw new HttpException(400, Sorting.INVALID_SORTING_RULES);
}
~~~

So hasValidSortingRules(...) could prime the valid internal map, so getValidSortingRules() could just return the results of calling has*.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
newfront: sure. This would have been a silly 500.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
DeathByTape: Since Elide is typically constructed in code, I would recommend a simple object that holds the parameters and is passed in to an Elide constructor. That said, @aklish may have some more in-depth thoughts on this.

elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java
newfront: sure. This will go into elide 2.0

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
newfront: probably worth it, but not sure if the issue I ran into was more of a one off. Some IDEs have auto-optimization for imports which was the case here.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
newfront: Sure. I will look at adding this into the elide2 branch.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
newfront: That sounds like a much nicer approach to this. I can add it.

~~~java
public static class Builder<Class<T>> {}
new Hql.Builder<>(entityClass,dictionary).build()
new Hql.Builder<>(entityClass,dictionary).withFilters(filters).build()
new Hql.Builder<>(entityClass,dictionary).withSorting(sorting).withFilters(filters).build()
new Hql.Builder<>(entityClass, dictionary).withFilters(filters).withSorting(sorting).withPagination(pagination).build();
~~~

I can have it return a Collection<T>

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
newfront: I can pick this one up tomorrow.

newfront:
I found the issue with the ResourceIT test.
~~~
parent?sort=-firstName
~~~

**Previous Test Order**
1. syzygy
2. omg. I have kidz
3. Unknown
4. Link
5. I'm new here
6. null

**True Sort (my change)**

1. Unknown
2. syzygy
3. omg. I have kidz
4. Link
5. I'm new here
6. null

Unknown should have been top of the list followed by *syzygy*. Not sure how the -firstName rule was being applied before but the static json had the Unknown user in wrong position.

l m n **O** p q r **S** t **U** v w x y z

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -786,12 +793,23 @@ public PersistentResource getRelation(String relation, String id) {
...
 
-            if (!filters.isEmpty()) {
+            if (!filters.isEmpty() || hasSortRules || isPaginated) {
                 final Class<?> entityClass = dictionary.getParameterizedType(obj, relationName);
-                filteredVal = requestScope.getTransaction().filterCollection(filteredVal, entityClass, filters);
+                final Optional<Sorting> sortingRules = hasSortRules ? Optional.of(requestScope.getSorting())
+                        : Optional.empty();
+                final Optional<Pagination> pagination = isPaginated ? Optional.of(requestScope.getPagination())
+                        : Optional.empty();
+                if (!filters.isEmpty()) {
```
aklish: I don't understand this.  Why wouldn't we always call filterSortOrPaginate here.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -718,10 +720,15 @@ public PersistentResource getRelation(String relation, String id) {
     }
 
     private Set<PersistentResource> getRelation(String relationName, boolean checked) {
-        if (requestScope.getTransaction() != null && !requestScope.getPredicates().isEmpty()) {
-            final Class<?> entityClass = dictionary.getParameterizedType(obj, relationName);
-            final String valType = dictionary.getBinding(entityClass);
-            final Set<Predicate> filters = new HashSet<>(requestScope.getPredicatesOfType(valType));
+        boolean hasPredicates = !requestScope.getPredicates().isEmpty();
+        if (hasPredicates || !requestScope.getSorting().isDefaultInstance()
+                || requestScope.getPagination().isDefault()) {
```
aklish: I don't understand this.  What does pagination default have to do with building filters?

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -0,0 +1,112 @@
...
+        PAGE_KEYS.put("page[offset]", PaginationKey.page);
+    }
+
+
+    private int page;
+
+    @Getter
+    private int pageSize;
+
+    public int getPage() {
```
aklish: I think this method should be getOffset()

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
aklish: I think you need to differentiate in the parsing - and store the correct value.  Right now everything maps to page and pageSize.  This class should ultimately store the offset and limit (because that is what is passed to hibernate).  It alternatively could store both (offset & limit, page & pageSize), but that might just complicate things.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
aklish: My comment was that the comment was wrong.

elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java
```diff
@@ -28,11 +27,7 @@
 import lombok.ToString;
 import org.apache.commons.lang3.tuple.Pair;
 
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
aklish: Our style is not to have wildcards in the code (although no style check currently catches this).

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -123,15 +120,64 @@ public void commit() {
...
+
+        final Pagination pagination = filterScope.hasPagination() ? filterScope.getRequestScope().getPagination()
+                : null;
+
+        // if we have sorting and sorting isn't empty, then we should pull dictionary to validate the sorting rules
+        Set<Order> validatedSortingRules = null;
+        if (filterScope.hasSortingRules()) {
+            final Sorting sorting = filterScope.getRequestScope().getSorting();
+            final EntityDictionary dictionary = filterScope.getRequestScope().getDictionary();
+            if (sorting.hasValidSortingRules(loadClass, dictionary)) {
```
aklish: I would think sorting.hasValidSortingRules should also test for empty sorting rules (so you don't have to check again down below for empty).

aklish: I would probably move the logic that converts from Sorting to Set<Order> in a function maybe toOrderSet(Sorting) or something.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
newfront: That is also a valid point. Plus it probably makes more sense to the folks reading the values. If they see page(3),pageSize(3) for page[number]=2&page[size]=3 it may be confusing. I'll update.

elide-core/src/test/java/com/yahoo/elide/core/filter/PageTest.java
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Tests parsing the page params for json-api pagination.
+ */
+public class PageTest {
+
+    @Test
+    public void shouldParseQueryParamsForCurrentPageAndPageSize() {
+        MultivaluedMap<String, String> queryParams = new MultivaluedStringMap();
+        queryParams.add("page[size]", "10");
+        queryParams.add("page[number]", "1");
```
aklish: A better test here would be number 2 or 3.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
newfront: that was a bad condition.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
newfront: this will be in the next update. As I wrote above, I do agree that the naming here makes more sense.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -0,0 +1,146 @@
...
+    public <T> boolean hasValidSortingRules(final Class<T> entityClass,
+                                        final EntityDictionary dictionary) {
+        if (validSortRules == null) {
+            validSortRules = new LinkedHashMap<>();
+        }
+        final Map<String, SortOrder> validSortingRules = new LinkedHashMap<>();
+        final List<String> entities = dictionary.getAttributes(entityClass);
+        if (entities != null) {
+            entities.stream()
+                    .filter(sortRules::containsKey)
```
aklish: I don't think we should filter here.  I think we should return a 400 error to the caller if they provided a sort rule for an attribute which doesn't exist.  In fact, I think we should do this early (like when parsing the query params).

elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java
newfront: yeah, intellij defaults to 5 classes in same path collapse to import class.path.*. I have updated my ide, and will add the checkstyle rule as well.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
newfront: yeah, that was a bad condition left over from testing. I am re-sweeping the code now.

elide-core/src/test/java/com/yahoo/elide/core/filter/PageTest.java
newfront: sure.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
newfront: these are all gone now since I added the checkstyle for wildcard imports.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
newfront: When we parse the query params we don't know the target entity type yet. In the filters, it has a filter[entity.field] so you know the target entity, however since we are only sorting the terminal resource or collection, we won't know the actual terminal class type until we hit one of the three TerminalState's.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -735,6 +776,59 @@ public PersistentResource getRelation(String relation, String id) {
...
+                final Class<?> entityClass = dictionary.getParameterizedType(obj, relationName);
+                final String valType = dictionary.getBinding(entityClass);
+                filters.addAll(new HashSet<>(requestScope.getPredicatesOfType(valType)));
+            }
+            if (checked && checkRelation(relationName)) {
+                return getRelationUncheckedWithSortingAndPagination(relationName, filters);
+            }
+        } else {
+            if (checked) {
+                return getRelationChecked(relationName, Collections.<Predicate>emptySet());
```
aklish: Can we use checkRelation here as well?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -791,7 +885,46 @@ public PersistentResource getRelation(String relation, String id) {
...
+        return (Set) resources;
+    }
+
+    /**
+     * Fetches the relationship entities with sorting and pagination support via HQLTransaction.
+     * @param relationName The entity whose relationships we want to fetch
+     * @param filters The set of possible filters (where clauses)
+     * @return The persistent resources
+     */
+    private Set<PersistentResource> getRelationUncheckedWithSortingAndPagination(String relationName,
```
aklish: Can we always call this function and just get rid of the one above?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -947,12 +1080,30 @@ public RequestScope getRequestScope() {
...
      */
     public Resource toResource() {
+        return toResource(this::getRelationships, this::getAttributes);
+    }
+
+    /**
+     * Fetch a resource with support for lambda function for getting relationships and attributes.
+     * @return The Resource
+     */
+    public Resource toResourceWithSortingAndPagination() {
```
aklish: Love this whole section.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+
+    private int offset;
+
+    @Getter
+    private int limit;
+
+    public int getOffset() {
+        return (offset > 0 ? offset - 1 : 0) * limit;
```
aklish: Maybe I'm missing something, but this still doesn't seem right.  Here is an example:
page=2, size=50 I believe is the same things as offset=100, limit=50.
So if someone provides and offset of 100, I wouldn't multiply that 50 (but I would if they provided page).

Does this make sense?  Am I missing something here?  I feel like the multiplication has to happen depending on whether or not they used 'offset' or 'page' in the query parameter.

elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
```diff
@@ -100,7 +100,7 @@ private JsonNode getResponseBody(PersistentResource rec, RequestScope requestSco
         JsonApiDocument jsonApiDocument = new JsonApiDocument();
 
         //TODO Make this a document processor
-        Data<Resource> data = rec == null ? null : new Data<>(rec.toResource());
+        Data<Resource> data = rec == null ? null : new Data<>(rec.toResourceWithSortingAndPagination());
```
aklish: Do you think we need this?  I"m trying to figure out what this would mean for an individual entity.  Would this basically work for something like page 1, limit 0?

elide-core/src/test/java/com/yahoo/elide/core/PaginationLogicTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Test
+    public void shouldParseQueryParamsForCurrentPageAndPageSize() {
+        MultivaluedMap<String, String> queryParams = new MultivaluedStringMap();
+        queryParams.add("page[size]", "10");
+        queryParams.add("page[number]", "2");
+
+        Pagination pageData = Pagination.parseQueryParams(queryParams);
+
+        Assert.assertEquals(pageData.getOffset(), 10);
```
aklish: Shouldn't the offset here be 20?

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -0,0 +1,157 @@
...
+                                        final EntityDictionary dictionary) throws InvalidValueException {
+        if (validSortRules == null) {
+            validSortRules = new LinkedHashMap<>();
+        }
+        final Map<String, SortOrder> validSortingRules = new LinkedHashMap<>();
+        final List<String> entities = dictionary.getAttributes(entityClass);
+        if (entities != null) {
+            entities.stream()
+                    .filter(sortRules::containsKey)
+                    .forEachOrdered(entity -> validSortingRules.put(entity, sortRules.get(entity)));
```
aklish: I think you can simplify this a bit with imperative code:
for (String sortAttribute : sortRules.keys) {
   if (!entityAttributes.contains(sortAttribute) {
      throw InvalidValueException()
  }
}
return true;

}

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    /**
+     * Checks to see if the sorting rules are valid for the given JPA class.
+     * @param entityClass The target jpa entity
+     * @param dictionary The elide entity dictionary
+     * @param <T> The Type of the target entity
+     * @return The validity of the sorting rules on the target class
+     */
+    public <T> boolean hasValidSortingRules(final Class<T> entityClass,
```
aklish: Given that the failure path is now to throw an exception, instead of returning boolean here, I would just make this void and call the function validateSortingRules

elide-core/src/test/java/com/yahoo/elide/core/PaginationLogicTest.java
newfront: http://jsonapi.org/format/#fetching-pagination. 

> For example, a page-based strategy might use query parameters such as page[number] and page[size], an offset-based strategy might use page[offset] and page[limit]

So the underlying structure in this case is offset (for start) and limit (for limit). So page=1 with limit 10, is 0-10.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
newfront: yep. I was also able to clean up the whole thing as well.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
newfront: See https://github.com/yahoo/elide/pull/149#discussion_r54750596

elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
newfront: I doubt that use case would ever come up. I'll remove.

elide-core/src/test/java/com/yahoo/elide/core/PaginationLogicTest.java
newfront: offset, limit = absolute start, limit

aklish:
:+1:

DeathByTape:
This looks good to me. Sorry it took so long to get ready to merge it in after Aaron's review (we were in the process of fortifying existing features and didn't want to convolute anything). I think this is good to merge as soon as conflicts are fixed. :+1:

PR #150, opened 2016-02-24
Create separate elide-annotations package

clayreimann:
Thoughts on the [PersistentResource](./150/files#diff-8caabf354be8ad916593d5aee9db4383) and [RequestScope](./150/files#diff-de121ab95dbe8a08b874b89c458c24d4) interfaces?

@DeathByTape @aklish @wcekan


elide-annotations/src/main/java/com/yahoo/elide/security/PersistentResource.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2015, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.security;
```
DeathByTape: Is there a reason this is in the security package?

elide-annotations/src/main/java/com/yahoo/elide/security/PersistentResource.java
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2015, Yahoo Inc.
```
DeathByTape: Copyright should say 2016

elide-annotations/src/main/java/com/yahoo/elide/security/RequestScope.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2015, Yahoo Inc.
```
DeathByTape: 2016

elide-annotations/src/main/java/com/yahoo/elide/security/RequestScope.java
```diff
@@ -0,0 +1,17 @@
...
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.security;
```
DeathByTape: Same package question

DeathByTape:
Other than comments, the interfaces look reasonable to me.

yahoocla:
CLA is valid!

elide-annotations/src/main/java/com/yahoo/elide/security/PersistentResource.java
clayreimann: It's in security because these are the checks' view into these classes.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -152,12 +152,16 @@ protected RequestScope(JsonApiDocument jsonApiDocument, RequestScope outerReques
...
         this.objectEntityCache = outerRequestScope.objectEntityCache;
         this.securityMode = outerRequestScope.securityMode;
-        this.newResources = outerRequestScope.newResources;
+        this.newPersistentResources = outerRequestScope.newPersistentResources;
         this.commitTriggers = outerRequestScope.commitTriggers;
         this.permissionExecutor = new PermissionExecutor(this);
         this.failedAuthorizations = outerRequestScope.failedAuthorizations;
     }
 
+    public Set<com.yahoo.elide.security.PersistentResource> getNewResources() {
```
aklish: Override?

aklish:
:+1:

PR #154, opened 2016-02-25
Removed required read checks for updates.

DeathByTape:
This PR allows for retrieving values **without** checking read permissions *as long as* that value is never returned to the user.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -123,7 +123,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
         // Initialize null ToMany collections
         requestScope.getDictionary().getRelationships(entityClass).stream()
                 .filter(relationName -> newResource.getRelationshipType(relationName).isToMany()
-                && newResource.getValue(newResource.getObject(), relationName, newResource.dictionary) == null)
+                && newResource.getValueNoCheck(relationName) == null)
                 .forEach(relationName -> newResource.setValue(relationName, new LinkedHashSet<>()));
 
         // Keep track of new resources for non shareable resources
@@ -290,8 +290,8 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
      * @return true if object updated, false otherwise
      */
     public boolean updateAttribute(String fieldName, Object newVal) {
-        checkFieldAwarePermissions(UpdatePermission.class, fieldName, newVal, getValue(fieldName));
-        Object val = getAttribute(fieldName);
+        Object val = getValueNoCheck(fieldName);
+        checkFieldAwarePermissions(UpdatePermission.class, fieldName, newVal, val);
         if (val != newVal && (val == null || !val.equals(newVal))) {
             this.setValueChecked(fieldName, newVal);
             transaction.save(obj);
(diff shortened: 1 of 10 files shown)


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -123,7 +123,7 @@
         // Initialize null ToMany collections
         requestScope.getDictionary().getRelationships(entityClass).stream()
                 .filter(relationName -> newResource.getRelationshipType(relationName).isToMany()
-                && newResource.getValue(newResource.getObject(), relationName, newResource.dictionary) == null)
+                && newResource.getValueNoCheck(relationName) == null)
```
clayreimann: Perhaps convert this to a block lambda for clarity?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -748,8 +762,17 @@ public PersistentResource getRelation(String relation, String id) {
...
 
+        return (Set) filter(ReadPermission.class, (Set) getRelationUnchecked(relationName, filters));
+    }
+
+    /**
+     * Retrieve an uncheck/unfiltered set of relations.
+     *
+     * @return
+     */
+    private Set<PersistentResource> getRelationUnchecked(String relationName, Set<Predicate> filters) {
```
clayreimann: Should be `getRelationFiltered`

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1016,6 +1039,16 @@ protected Object getValue(String fieldName) {
     }
 
     /**
+     * Retrieve an object without checking read permissions (i.e. value is used internally and not sent to others)
+     *
+     * @param fieldName the field name
+     * @return Value
+     */
+    protected Object getValueNoCheck(String fieldName) {
```
clayreimann: `getValueUnchecked` for consistency

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -708,13 +708,27 @@ public PersistentResource getRelation(String relation, String id) {
      * @return collection relation
      */
     public Set<PersistentResource> getRelation(String relationName) {
```
clayreimann: Should be `getRelationChecked`

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
DeathByTape: Why `Filtered`? That does not accurately describe this method. I will rename this for consistency, however, using the `Checked` and `Unchecked` form.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -748,8 +762,17 @@ public PersistentResource getRelation(String relation, String id) {
...
+
+    /**
+     * Retrieve an uncheck/unfiltered set of relations.
+     *
+     * @return
+     */
+    private Set<PersistentResource> getRelationUnchecked(String relationName, Set<Predicate> filters) {
         RelationshipType type = getRelationshipType(relationName);
-        Object val = this.getValue(relationName);
+        Object val = getValueNoCheck(relationName);
```
aklish: Move shouldSkipCollection to the checked version.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
clayreimann: Looks good now, the method overloading hurt my head.

aklish:
:+1:

yahoocla:
CLA is valid!

PR #158, opened 2016-02-29
Move prefab checks from core to annotations/checks project

wcekan:
Prefab checks should be available without dependency on elide-core


DeathByTape:
:+1:

yahoocla:
CLA is valid!

PR #164, opened 2016-03-04
Run DataStore specific integration tests

wcekan:
The integration test refactoring skipped all DataStore specific tests.

### DIFF
--- a/elide-integration-tests/src/test/java/com/yahoo/elide/tests/DataStoreIT.java
+++ b/elide-integration-tests/src/test/java/com/yahoo/elide/tests/DataStoreIT.java
@@ -1,0 +1,12 @@
+/*
+ * Copyright 2015, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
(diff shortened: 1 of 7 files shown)


clayreimann:
For some reason it looks like a lot more `Filtered` objects are being created now...

yahoocla:
CLA is valid!

wcekan:
@clayreimann Seems Hibernate 5 uses an odd mechanism to generate IDs.   Each object of subclass BaseID gets a unique ID regardless of Entity table.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/DataStoreIT.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright 2015, Yahoo Inc.
```
DeathByTape: Should be 2015

DeathByTape:
:+1:

PR #166, opened 2016-03-04
Update check deferment for Patch Extension

DeathByTape:
Update to properly defer checks and modify elements within a patch extension request.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -899,14 +899,10 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
 
     @Override
     public int hashCode() {
-        return 1;
-//        final int prime = 31;
-//        int result = 1;
-//        String id = dictionary.getId(getObject());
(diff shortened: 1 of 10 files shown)


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -894,22 +899,19 @@ public String getType() {
...
-        int result = 1;
-        String id = dictionary.getId(getObject());
-        result = prime * result + (uuid.isPresent() ? uuid.hashCode() : 0);
-        result = prime * result + (id == null ? 0 : id.hashCode());
-        result = prime * result + (type == null ? 0 : type.hashCode());
-        return result;
+        if (getObject() == null) {
+            return -1;
+        }
+        return getObject().getClass().hashCode();
```
wcekan: Perhaps this should be `getObject().getType().hashCode()` since object _may_ be a proxy

DeathByTape: Just `getType().hashCode()`? Otherwise we're assuming `getObject()` has a `getType()` method.

wcekan:
:+1:

yahoocla:
CLA is valid!

PR #167, opened 2016-03-04
More minor cleanup

wcekan:
See #163

### DIFF
--- a/elide-core/src/test/java/com/yahoo/elide/jsonapi/JsonApiTest.java
+++ b/elide-core/src/test/java/com/yahoo/elide/jsonapi/JsonApiTest.java
@@ -208,8 +208,8 @@ public class JsonApiTest {
         assertEquals(data.getType(), "parent");
         assertEquals(data.getId(), "123");
         assertEquals(attributes.get("firstName"), "bob");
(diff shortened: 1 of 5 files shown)


elide-core/src/test/java/com/yahoo/elide/jsonapi/JsonApiTest.java
```diff
@@ -208,8 +208,8 @@ public void readSingle() throws IOException {
         assertEquals(data.getType(), "parent");
         assertEquals(data.getId(), "123");
         assertEquals(attributes.get("firstName"), "bob");
-        assertEquals(relations.get("children").getData().get().iterator().next().getType(), "child");
-        assertEquals(relations.get("children").getData().get().iterator().next().getId(), "2");
+        assertEquals(relations.get("children").getData().getSingleValue().getType(), "child");
```
clayreimann: This works but doesn't make sense. Not sure why our doc indicates that `children` is a `toOne` field

DeathByTape:
:+1: 

NOTE: We figured out that `children` is modeled as `ToOne`. While this doesn't make sense grammatically (i.e. should probably be `child` instead), it does make sense as modeled.

yahoocla:
CLA is valid!

PR #173, opened 2016-03-09
Defer update on create

DeathByTape:
For any sort of creation request, update checks _must_ be deferred.


wcekan:
:+1: Passes my tests

yahoocla:
CLA is valid!

Issue #170, opened 2016-03-07
Avoid Excessive save() Calls


DeathByTape:
We should avoid our excessive calls to `save()`. In particular, non-modified objects should be avoided whenever possible but we should also investigate other situations where this may occur.

PR #174, opened 2016-03-11
Removed redundant and unnecessary calls to save().

DeathByTape:
PR in response to: https://github.com/yahoo/elide/issues/170

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -126,6 +126,7 @@ public class Elide {
             ElideResponse response = buildResponse(responder.get());
             auditLogger.commit();
             transaction.commit();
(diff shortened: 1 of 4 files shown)


wcekan:
:+1:

yahoocla:
CLA is valid!

wcekan:
@DeathByTape Found a problem here.  Will need to revert.

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -126,6 +126,7 @@ public ElideResponse get(
             ElideResponse response = buildResponse(responder.get());
             auditLogger.commit();
             transaction.commit();
+            requestScope.saveObjects();
```
wcekan: @DeathByTape Missed this.  saveObjects occurs **after** commit.

PR #175, opened 2016-03-12
restoring persistent resource hash code

sciolizer:
For a query involving one level of include, server was getting pegged at 100% cpu, with the hotspot being inside `java.util.HashMap`. Changing `hashCode` function to [its original implementation](https://github.com/yahoo/elide/commit/37881cfddb62ae934bd3939a6532e3bf80e9f965#diff-01adfd7c58c229be01f3e188e23974e3R902) reduced query time from 14 seconds to 3 seconds.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -1027,10 +1027,13 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
 
     @Override
     public int hashCode() {
-        if (getType() == null) {
-            return -1;
-        }
-        return getType().hashCode();
+        final int prime = 31;
+        int result = 1;
+        String id = dictionary.getId(getObject());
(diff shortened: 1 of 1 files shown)


DeathByTape:
Thanks for submitting this. I'll think about this a little more and @wcekan you should too. 

This was initially changed due to a broken contract between hashCode and equals that caused problems elsewhere. Seeing as how it's passing the added tests, however, it may or may not be an issue still. Going to investigate. 

Thanks again!

yahoocla:
CLA is valid!

wcekan:
Unfortunately this breaks the contract for [hashCode](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode())/Equals.


> Whenever it is invoked on the same object more than once during an execution of a Java application, the hashCode method must consistently return the same integer, provided no information used in equals comparisons on the object is modified.

DeathByTape:
@sciolizer could you review this change here:
https://github.com/yahoo/elide/pull/176

Does this help your performance issue? I believe this should allow us to maintain the contract and achieve better distribution results.

sciolizer:
That does the trick. Thank you!

PR #176, opened 2016-03-14
Added supporting tests for hashcode and equals and improved hashcode …

DeathByTape:
…performance.


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1027,10 +1028,29 @@ public String getType() {
 
     @Override
     public int hashCode() {
-        if (getType() == null) {
-            return -1;
+        if (hashCode == 0) {
+            final int prime = 31;
+            hashCode = 1;
```
wcekan: This is not thread safe.  Make this local and set this.hashCode once.

wcekan:
:+1:

PR #177, opened 2016-03-15
Fix patch extension for rootable collections and error handling

DeathByTape:


swathi9488:
Looks good.

yahoocla:

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -255,7 +255,10 @@ public boolean matchesId(String checkId) {
         }
 
         PersistentResource<T> resource = new PersistentResource<>(obj, requestScope);
-        resource.checkFieldAwarePermissions(ReadPermission.class);
+        // No need to have read access for a newly created object
+        if (!requestScope.getNewResources().contains(resource)) {
+            resource.checkFieldAwarePermissions(ReadPermission.class);
```
aklish: Why would loadRecord be called for a newly created resource?

aklish:
:+1:

PR #178, opened 2016-03-16
Checking for size to prevent success in case of no change

duseja2:


wcekan:
:+1: Please add unit tests

yahoocla:
CLA is valid!

PR #179, opened 2016-03-17
Relationship filtering

DeathByTape:
This allows filtering. To filter collections, traverse a path relative to the parent. For instance,

`/parent?include=[children]&filter[parent.children.id]=4`

will give all parents that have child id 4.

`/parent?include[children]&filter[child.id]=4`

will provide all parents and include a child with id 4.

This allows us to continue to honor existing behavior while providing greater flexibility in filtering at the collection level.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
@@ -45,27 +45,18 @@ public class Predicate {
             if (matcher.find()) {
                 final String[] keyParts = matcher.group(1).split("\\.");
 
-                if (keyParts.length != 2) {
+                if (keyParts.length < 2) {
                     throw new InvalidPredicateException("Invalid filter format: " + queryParameter.getKey());
                 }
 
-                final String type = keyParts[0];
-                final String field = keyParts[1];
 
                 final Operator operator = (matcher.group(3) == null) ? Operator.IN
                         : Operator.fromString(matcher.group(3));
 
-                final Class<?> entityClass = dictionary.getBinding(type);
-                if (entityClass == null) {
-                    throw new InvalidPredicateException("Unknown entity in filter: " + type);
-                }
-
-                final Class<?> fieldType = ("id".equals(field.toLowerCase(Locale.ENGLISH)))
-                                           ? dictionary.getIdType(entityClass)
-                                           : dictionary.getParameterizedType(entityClass, field);
-                if (fieldType == null) {
-                    throw new InvalidPredicateException("Unknown field in filter: " + field);
-                }
+                final Class<?>[] types = getTypes(keyParts, dictionary);
+                final String type = getType(types);
+                final String field = getField(keyParts);
+                final Class<?> fieldType = getFieldType(types);
 
                 final List<Object> values = new ArrayList<>();
                 for (String valueParams : queryParameter.getValue()) {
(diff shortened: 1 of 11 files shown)


elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -83,4 +74,87 @@
...
+     * NOTE: Types are now fully qualified by (ROOT_TYPE.FIELD1.FIELD2..FIELD(N-1)).ACCESSED_FIELD
+     *
+     * @param types Array of types
+     * @return Type of object to be filtered. Empty string if nothing found.
+     */
+    private static String getType(final String[] types) {
+        if (types == null || types.length <= 1) {
+            return "";
+        }
+        StringBuilder sb = new StringBuilder();
```
shadanan: Looks like you could just do: `String.join(".", types)`

DeathByTape: Great idea :) Going to change/rebase (I was actually thinking join was only in python for some reason lol)

shadanan: Actually, I just noticed that you're dropping the last element in types. I'm not sure how to use String.join in this case without making a copy of the array.

DeathByTape: So close, but it includes the field too :( Check the `for` loop bounds.

elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
```diff
@@ -867,6 +889,41 @@ class FilterIT extends AbstractIntegrationTestInitializer {
         }
     }
 
+    @Test
+    public void testGetBadRelationshipNameWithNestedFieldFilter() {
+        def result = mapper.readTree(RestAssured.get("book?filter[book.author12.name]=Null%20Ned").asString());
```
shadanan: Perhaps we should keep it consistent: use a space instead of %20

clayreimann: Or `%20` it everywhere.

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
shadanan: You could do: `Arrays.asList(types).subList(0, types.length - 1);`
Looks like these operations do not involve any copying:
https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#asList(T...)
https://docs.oracle.com/javase/7/docs/api/java/util/List.html#subList(int,%20int)

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -83,4 +74,87 @@
...
+        if (keyParts == null || keyParts.length <= 0) {
+            return new Class[0];
+        }
+        Class<?>[] types = new Class[keyParts.length];
+        String type = keyParts[0];
+        types[0] = dictionary.getBinding(type);
+        for (int i = 1 ; i < keyParts.length ; ++i) {
+            final String field = keyParts[i];
+            final Class<?> entityClass = types[i - 1];
+            if (entityClass == null) {
```
shadanan: As far as I can tell, this should only ever happen on types[0], because you check for all other possible ones in line 140. So maybe you should move this check after `types[0] = dictionary.getBinding(type);`

elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
DeathByTape: I'm conflicted lol. Both *should* work. Should we have a case to verify it? Like a separate test case that uses `%20` for verification? Or do we not care that much? :)

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
DeathByTape: I think you're right. Good catch

elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
```diff
@@ -867,6 +889,41 @@ class FilterIT extends AbstractIntegrationTestInitializer {
         }
     }
 
+    @Test
+    public void testGetBadRelationshipNameWithNestedFieldFilter() {
```
shadanan: We should have tests for:
1. an invalid root with the new filter method
2. an invalid leaf

shadanan:
:+1: after adding the two extra tests

yahoocla:
CLA is valid!

shadanan:
:+1:

PR #183, opened 2016-03-21
Update shareable

DeathByTape:


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -629,9 +629,10 @@ protected void checkSharePermission(Set<PersistentResource> resourceIdentifiers)
 
         for (PersistentResource persistentResource : resourceIdentifiers) {
             if (!newResources.contains(persistentResource)) {
-                if (persistentResource.isShareable()) {
+                boolean isInLineage = lineage.getRecord(persistentResource.getType()).contains(persistentResource);
+                if (persistentResource.isShareable() && !isInLineage) {
```
wcekan: Reverse the check since we already resolved the local boolean:
```java
if (!isInLineage && persistentResource.isShareable())

wcekan:
:+1:

yahoocla:
CLA is valid!

PR #186, opened 2016-03-28
Added builder interface for Elide.

DeathByTape:
Added support for configurable PermissionExecutor.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -66,6 +66,8 @@ public class Elide {
     /**
      * Instantiates a new Elide.
      *
+     * <i>deprecated since 2.0.11</i>
+     *
      * @param auditLogger the audit logger
      * @param dataStore the dataStore
      * @param dictionary the dictionary
@@ -78,6 +80,8 @@ public class Elide {
     /**
      * Instantiates a new Elide.
      *
+     * <i>deprecated since 2.0.11</i>
+     *
      * @param auditLogger the audit logger
      * @param dataStore the dataStore
      */
@@ -88,6 +92,8 @@ public class Elide {
 
     /**
      * Instantiates a new Elide.
+     *
+     * <i>deprecated since 2.0.11</i>
      *
      * @param auditLogger the audit logger
      * @param dataStore the dataStore
@@ -199,6 +205,7 @@ public class Elide {
      * @param securityMode only for test mode
      * @return Elide response object
      */
+    @Deprecated
     public ElideResponse get(
             String path,
             MultivaluedMap<String, String> queryParams,
@@ -206,6 +213,7 @@ public class Elide {
             SecurityMode securityMode) {
 
         RequestScope requestScope = null;
+        boolean isVerbose = false;
         try (DataStoreTransaction transaction = dataStore.beginReadTransaction()) {
             final User user = transaction.accessUser(opaqueUser);
             requestScope = new RequestScope(
@@ -218,6 +226,7 @@ public class Elide {
                     queryParams,
                     securityMode,
                     permissionExecutor);
+            isVerbose = requestScope.getPermissionExecutor().isVerbose();
             GetVisitor visitor = new GetVisitor(requestScope);
             Supplier<Pair<Integer, JsonNode>> responder = visitor.visit(parse(path));
             requestScope.getPermissionExecutor().executeCommitChecks();
@@ -230,13 +239,13 @@ public class Elide {
             return response;
         } catch (ForbiddenAccessException e) {
             debugLogSecurityExceptions(requestScope);
-            return buildErrorResponse(e, securityMode);
+            return buildErrorResponse(e, isVerbose);
         } catch (HttpStatusException e) {
-            return buildErrorResponse(e, securityMode);
+            return buildErrorResponse(e, isVerbose);
         } catch (IOException e) {
-            return buildErrorResponse(new TransactionException(e), securityMode);
+            return buildErrorResponse(new TransactionException(e), isVerbose);
         } catch (ParseCancellationException e) {
-            return buildErrorResponse(new InvalidURLException(e), securityMode);
+            return buildErrorResponse(new InvalidURLException(e), isVerbose);
         }
     }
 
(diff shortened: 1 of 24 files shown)


wcekan:
Can we deprecate the BYPASS SECURITY now and show how to do that with a special PermissionExcectutor?

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -85,12 +93,60 @@ public Elide(AuditLogger auditLogger, DataStore dataStore) {
...
+        @Getter @Setter private JsonApiMapper jsonApiMapper;
+        @Getter @Setter private Class<? extends PermissionExecutor> permissionExecutor;
+
+        /**
+         * Constructor.
+         *
+         * @param auditLogger
+         * @param dataStore
+         */
+        public Builder(AuditLogger auditLogger, DataStore dataStore) {
```
aklish: Doesn't the builder normally have 'with' methods for each optional thing you are adding to it.  (withAuditLogger, withDataStore, withPermissionExecutor), etc.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -88,7 +92,25 @@ public RequestScope(JsonApiDocument jsonApiDocument,
...
         commitTriggers = new LinkedHashSet<>();
-        permissionExecutor = new PermissionExecutor(this);
+        if (permissionExecutorClass == null) {
+            this.permissionExecutor = new ActivePermissionExecutor(this);
+        } else {
+            try {
+                PermissionExecutor permissionExecutor;
+                try {
+                    // Try to find a constructor using a request scope first
+                    Constructor<? extends PermissionExecutor> ctor =
```
aklish: I dont' think you need to make one of these per request.  The security test framework can just instantiate different instances of Elide with different PermissionExecutors.  I would remove this construction and just attach the executor in the builder.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -0,0 +1,219 @@
...
+
+    private final RequestScope requestScope;
+    private final ExpressionBuilder expressionBuilder;
+
+    /**
+     * Constructor.
+     *
+     * @param requestScope Request scope.
+     */
+    public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestScope) {
```
aklish: Remove dependency on Request scope.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -0,0 +1,219 @@
...
+        HashMap<Class<? extends Check>, Map<PersistentResource, ExpressionResult>> cache = new HashMap<>();
+
+        this.requestScope = requestScope;
```
aklish: Maybe this should be passed in during checks instead of during construction.  It already is in almost all cases (as part of the resource).

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -62,7 +65,8 @@ public RequestScope(JsonApiDocument jsonApiDocument,
                         JsonApiMapper mapper,
                         AuditLogger auditLogger,
                         MultivaluedMap<String, String> queryParams,
-                        SecurityMode securityMode) {
+                        SecurityMode securityMode,
+                        Class<? extends PermissionExecutor> permissionExecutorClass) {
```
aklish: I don't think a permissionExecutor should be 'request scoped'.  It should be global.

elide-core/src/main/java/com/yahoo/elide/Elide.java
DeathByTape: We have the setters for the other options (i.e. `Builder.setAuditLogger.setDataStore.build()`). We can use the word `with` or any qualifier.

As far as these two particular arguments in the constructor are concerned: they're required. Bill and I spoke and decided to require them in the constructor of the builder to enforce this.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
DeathByTape: The problem is you cannot instantiate a permission executor without a requestscope (at least, not our executor). We need one of these per request, in general.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
DeathByTape: That would certainly make construction cleaner but the interface a little dirtier. I actually think it is cleaner the way that you're proposing, though.

@wcekan any opinions?

DeathByTape: Talked to @wcekan offline. He suggests, instead, we leave this as-is and use `Supplier<PermissionExecutor>` elsewhere.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
DeathByTape: It holds caching and state that is relative to your request; it cannot be easily made global.

elide-core/src/main/java/com/yahoo/elide/Elide.java
wcekan: Generally the builder pattern looks like

Builder().dictionary(new Dictionary()).jsonApiMapper(new Mapper()).permissionExecuter(() -> new MyPermissionExecuter()).build();

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -119,7 +216,8 @@ public ElideResponse get(
                     mapper,
                     auditLogger,
                     queryParams,
-                    securityMode);
+                    securityMode,
```
wcekan: securityMode is still used here?

DeathByTape: Yes for back-compat. Until we fully remove `securityMode`, this unfortunately needs to remain in place :-1:

wcekan:
:+1: We can remove the deprecated mode next major release.

PR #188, opened 2016-03-29
Hackday

clayreimann:
Work complete, ready for full review and merge.

This should be released as `2.2.0`.


elide-core/src/main/antlr4/com/yahoo/elide/generated/parsers/Expression.g4
```diff
@@ -0,0 +1,48 @@
...
+    | permissionClass                               #PERMISSION
+    ;
+
+/*
+No only numeric because class names can't be only numeric.
+Do not name this 'class' as that is a keyword.
+*/
+permissionClass: LEGAL_NAME;
+
+NOT         : 'NOT' | 'Not' | 'not';
```
DeathByTape: Maybe for these you should allow all casing? Something like:

```antlr
NOT : ('N'|'n')('O'|'o')('T'|'t');
```

should do, I think.

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -51,77 +50,63 @@
...
         relationships = dequeToList(relationshipsDeque);
+        entityPermissions = new EntityPermissions(cls, fieldOrMethodList);
     }
 
     /**
      * Bind fields of an entity including the Id field, attributes, and relationships.
      *
-     * @param cls Class type to bind fields
-     * @param type JSON API type identifier
+     * @param cls               Class type to bind fields
```
DeathByTape: Just a comment about whitespace like this: it actually is hard to maintain (due to variable name changes, etc.) and doesn't buy very much. I suggest preferring to simply use a single space rather than aligning.

elide-annotations/src/main/java/com/yahoo/elide/annotation/ReadPermission.java
```diff
@@ -37,5 +38,12 @@
      *
      * @return the class [ ]
      */
+    @Deprecated
     Class<? extends InlineCheck>[] all() default {};
+
+    /**
+     * An expression of checks that will be parsed via ANTLR.
```
shadanan: A little example here would be very useful to someone writing their first check expression.

elide-core/src/main/antlr4/com/yahoo/elide/generated/parsers/Expression.g4
```diff
@@ -0,0 +1,48 @@
...
+expression
+    : NOT WS expression                             #NOT
+    | LPAREN WS? expression WS? RPAREN              #PAREN
+    | left=expression WS AND WS right=expression    #AND
+    | left=expression WS  OR WS right=expression    #OR
+    | permissionClass                               #PERMISSION
+    ;
+
+/*
+No only numeric because class names can't be only numeric.
```
shadanan: Should this be `*Not* only numeric`?

elide-core/src/main/antlr4/com/yahoo/elide/generated/parsers/Expression.g4
clayreimann: preprocess perhaps, but I don't want to allow weird things like `nOt`,  `nOT, or `noT`

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -76,8 +108,30 @@ protected EntityBinding entityBinding(Class<?> entityClass) {
      * @param entityClass the entity class
      * @return binding class
      */
-    public String getBinding(Class<?> entityClass) {
-        return entityBinding(entityClass).jsonApi;
+    public String getJsonAliasFor(Class<?> entityClass) {
```
DeathByTape: Can we add comments for these new methods?

checkstyle-style.xml
```diff
@@ -212,7 +212,11 @@
     <property name="severity" value="error"/>
     <property name="lineSeparator" value="lf"/>
   </module>
-  <module name="SuppressionCommentFilter"/>
+  <module name="SuppressionCommentFilter">
```
DeathByTape: For consistency sake, I'm going to suggest that we just get rid of this rule. An IDE should quickly be able to fix the formatting on the other whitespace in comments.

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
```diff
@@ -0,0 +1,184 @@
...
+
+    /**
+     * Encapsulates the rules for a permission annotation.
+     */
+    public static class PermissionDefinition {
+        public final PermissionType permissionType;
+        public final Class<? extends Check>[] checks;
+        public final ParseTree expressionTree;
+
+        public PermissionDefinition(Class<? extends Check>[] checks, PermissionType type) {
```
shadanan: This should be marked `@Deprecated`

elide-core/src/main/antlr4/com/yahoo/elide/generated/parsers/Expression.g4
DeathByTape: fair enough. I'm just thinking from a usability perspective: if I could get by with `not` and `Not` I would assume case-insensitivity. Either way, my opinion about this isn't too strong.

shadanan: I think it's important to maintain consistency. nOt may not be conventional, but it is allowed in SQL and other language whose operators are case insensitive. Either we make it case sensitive and only allow one form, or we make it case insensitive and allow all forms.

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
```diff
@@ -0,0 +1,184 @@
+/*
+ * Copyright 2015, Yahoo Inc.
```
DeathByTape: Year is 2016.

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
```diff
@@ -0,0 +1,184 @@
...
+    public enum PermissionType {
+        EXPRESSION, ANY, ALL
+    }
+
+    /**
+     * Encapsulates the rules for a permission annotation.
+     */
+    public static class PermissionDefinition {
+        public final PermissionType permissionType;
+        public final Class<? extends Check>[] checks;
```
DeathByTape: I'm not sure I like this representation at all. We really should only have a single, definitive representation for checks. What this does is force a second code path below to construct an expression.

I propose we convert these checks to an expression and use the `ParseTree` as the definitive format since that is the long-term goal. I believe this should be straightforward:

```java
boolean anyChecks = ...; // Computed as usual
String conjunction = (anyChecks) ? " or " : " and ";
String expression = Arrays.asList(checks)
                               .stream()
                               .map(Class::getCanonicalName)
                               .reduce("", (left, right) -> (left.isEmpty()) ? right : left + conjunction + right);
```

You could then parse this string. 

This way we can guarantee everything is being constructed as intended for when we're ready to remove this functionality altogether.

I notice that this may seem like extra overhead, but since we're transitioning it will be logically more consistent and this should be short-lived pain. Since we plan on encouraging expression use anyway, we should not experience much pain here in the short-term either.

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionBuilder.java
```diff
@@ -200,106 +191,127 @@ public ExpressionBuilder(HashMap<Class<? extends Check>, Map<PersistentResource,
...
+            return null;
+        }
+
+        final BiFunction<Expression, Expression, Expression> expressionJoiner = getJoiner(permissions.permissionType);
+        switch (permissions.permissionType) {
+            case EXPRESSION:
+                ExpressionVisitor expressionBuilder = new ExpressionVisitor(entityDictionary, checkFn);
+                return expressionBuilder.visit(permissions.expressionTree);
+
+            case ANY:
```
DeathByTape: See comment above about `PermissionDefinition`. It would be nice to eliminate this code if possible.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -76,8 +108,30 @@ protected EntityBinding entityBinding(Class<?> entityClass) {
...
+
+    public PermissionDefinition getPermissionsForField(Class<?> resourceClass,
+                                                       String field,
+                                                       Class<? extends Annotation> annotationClass) {
+        EntityBinding binding = getEntityBinding(resourceClass);
+        return binding.entityPermissions.getFieldChecksForPermission(field, annotationClass);
+    }
+
+    public Class<? extends Check> getCheck(String checkIdentifier) {
+        return checkNames.get(checkIdentifier);
```
aklish: If checkNames don't contain the identifier, we should throw a runtime exception here with meaningful description.

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AnyFieldExpression.java
```diff
@@ -19,9 +19,9 @@
  */
 public class AnyFieldExpression implements Expression {
     private final Expression entityExpression;
-    private final OrExpression fieldExpression;
+    private final Expression fieldExpression;
 
-    public AnyFieldExpression(final Expression entityExpression, final OrExpression fieldExpression) {
+    public AnyFieldExpression(final Expression entityExpression, final Expression fieldExpression) {
```
aklish: Why did you make this more general?

elide-core/src/main/antlr4/com/yahoo/elide/generated/parsers/Expression.g4
clayreimann: I think consistency trumps in this instance, reverting to `[Nn][Oo][Tt]` et. al.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
clayreimann: Not a new method, just renamed

elide-annotations/src/main/java/com/yahoo/elide/security/checks/prefab/Collections.java
```diff
@@ -0,0 +1,70 @@
...
+        @Override
+        public boolean ok(T record, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
+            if (!changeSpecIsCollection(changeSpec)) {
+                return false;
+            }
+
+            Collection originalCollection = (Collection) changeSpec.get().getOriginal();
+            Collection modifiedCollection = (Collection) changeSpec.get().getModified();
+
+            return collectionIsStrictSuperset(originalCollection, modifiedCollection);
```
aklish: Should this be strict?  What if they modified the collection to be exactly what it was before?  Seems like that should pass.

elide-annotations/src/main/java/com/yahoo/elide/security/checks/prefab/Collections.java
```diff
@@ -0,0 +1,70 @@
...
+        @Override
+        public boolean ok(T record, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
+            if (!changeSpecIsCollection(changeSpec)) {
+                return false;
+            }
+
+            Collection originalCollection = (Collection) changeSpec.get().getOriginal();
+            Collection modifiedCollection = (Collection) changeSpec.get().getModified();
+
+            return collectionIsStrictSuperset(modifiedCollection, originalCollection);
```
aklish: Similar comment as above.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/CriterionExpressionVisitor.java
```diff
@@ -0,0 +1,105 @@
...
+    public T visitAND(ExpressionParser.ANDContext ctx) {
+        return joinSubexpressions(ctx.left, ctx.right, andCriterionJoiner);
+    }
+
+    @Override
+    public T visitPermissionClass(ExpressionParser.PermissionClassContext ctx) {
+        CriterionCheck<T, ?> criterionCheck = getCriterionCheck(ctx.getText());
+
+        if (criterionCheck == null) {
+            return null;
```
aklish: Is this safe?

elide-annotations/src/main/java/com/yahoo/elide/security/checks/prefab/Collections.java
clayreimann: Just mirroring the logic from the existing check

elide-annotations/src/main/java/com/yahoo/elide/security/checks/prefab/Collections.java
clayreimann: Same response.

PR #189, opened 2016-03-30
Performance enhancement using JOIN

wcekan:
Reduces queries by eager loading all **sparse** and **include** beans.
Allows subclasses to determine if/when to enable the join logic.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/RequestScopedTransaction.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/RequestScopedTransaction.java
@@ -1,0 +1,14 @@
+/*
+ * Copyright 2016, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core;
+
+/**
+ * Scope aware transaction
+ */
+public interface RequestScopedTransaction extends DataStoreTransaction {
+
+    void setRequestScope(RequestScope requestScope);
+}
--- a/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
+++ b/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
@@ -5,9 +5,13 @@
  */
 package com.yahoo.elide.datastores.hibernate3;
 
-import com.yahoo.elide.core.DataStoreTransaction;
+import com.google.common.base.Objects;
+import com.yahoo.elide.annotation.ReadPermission;
 import com.yahoo.elide.core.EntityDictionary;
 import com.yahoo.elide.core.FilterScope;
+import com.yahoo.elide.core.RequestScope;
+import com.yahoo.elide.core.RequestScopedTransaction;
+import com.yahoo.elide.core.exceptions.ForbiddenAccessException;
 import com.yahoo.elide.core.exceptions.TransactionException;
 import com.yahoo.elide.core.filter.HQLFilterOperation;
 import com.yahoo.elide.core.filter.Predicate;
@@ -15,8 +19,12 @@ import com.yahoo.elide.core.pagination.Pagination;
 import com.yahoo.elide.core.sort.Sorting;
 import com.yahoo.elide.datastores.hibernate3.filter.CriteriaExplorer;
 import com.yahoo.elide.datastores.hibernate3.filter.CriterionFilterOperation;
+import com.yahoo.elide.security.PersistentResource;
 import com.yahoo.elide.security.User;
+import lombok.AccessLevel;
+import lombok.Getter;
 import org.hibernate.Criteria;
+import org.hibernate.FetchMode;
 import org.hibernate.Hibernate;
 import org.hibernate.HibernateException;
 import org.hibernate.ObjectNotFoundException;
@@ -30,8 +38,11 @@ import org.hibernate.criterion.Restrictions;
 
 import java.io.IOException;
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
@@ -42,7 +53,7 @@ import java.util.stream.Collectors;
 /**
  * Hibernate Transaction implementation.
  */
-public class HibernateTransaction implements DataStoreTransaction {
+public class HibernateTransaction implements RequestScopedTransaction {
     private static final Function<Criterion, Criterion> NOT = Restrictions::not;
     private static final BiFunction<Criterion, Criterion, Criterion> AND = Restrictions::and;
     private static final BiFunction<Criterion, Criterion, Criterion> OR = Restrictions::or;
@@ -50,6 +61,8 @@ public class HibernateTransaction implements DataStoreTransaction {
     private final Session session;
     private final LinkedHashSet<Runnable> deferredTasks = new LinkedHashSet<>();
     private final CriterionFilterOperation criterionFilterOperation = new CriterionFilterOperation();
+    @Getter(value = AccessLevel.PROTECTED)
+    private RequestScope requestScope = null;
 
     /**
      * Instantiates a new Hibernate transaction.
@@ -105,6 +118,15 @@ public class HibernateTransaction implements DataStoreTransaction {
     @Override
     public <T> T loadObject(Class<T> loadClass, Serializable id) {
         try {
+            if (isJoinQuery()) {
+                Criteria criteria = session.createCriteria(loadClass).add(Restrictions.idEq(id));
+                if (requestScope != null) {
+                    joinCriteria(criteria, loadClass);
+                }
+                @SuppressWarnings("unchecked")
+                T record = (T) criteria.uniqueResult();
+                return record;
+            }
             @SuppressWarnings("unchecked")
             T record = (T) session.load(loadClass, id);
             Hibernate.initialize(record);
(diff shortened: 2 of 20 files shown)


elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -188,15 +201,134 @@ public void commit() {
...
+                // continue
+            }
+        }
+
+        for (String include : getIncludeList()) {
+            criteria.setFetchMode(include, FetchMode.JOIN);
+        }
+    }
+
+    private List<String> getIncludeList() {
```
shadanan: It would be nice if this function could get a unit test.

shadanan:
:+1:

yahoocla:
CLA is valid!

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
wcekan: Done

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -188,15 +201,134 @@ public void commit() {
...
+
+    private <T> void joinCriteria(Criteria criteria, final Class<T> loadClass) {
+        EntityDictionary dictionary = requestScope.getDictionary();
+        String type = dictionary.getBinding(loadClass);
+        Set<String> fields = Objects.firstNonNull(
+                requestScope.getSparseFields().get(type), Collections.<String>emptySet());
+        List<String> relationships = dictionary.getRelationships(loadClass);
+        for (String field : fields) {
+            try {
+                checkFieldReadPermission(loadClass, field);
```
aklish: Why are we doing security checks inside the data store?  These should really be isolated to persistent resource.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -188,15 +201,134 @@ public void commit() {
...
+                return true;
+            }
+        }
+        return false;
+    }
+
+    private <T> void joinCriteria(Criteria criteria, final Class<T> loadClass) {
+        EntityDictionary dictionary = requestScope.getDictionary();
+        String type = dictionary.getBinding(loadClass);
+        Set<String> fields = Objects.firstNonNull(
```
aklish: Objects.firstNonNull is deprecated.  There is another interface in guava that does the same now.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -188,15 +201,134 @@ public void commit() {
...
+                requestScope.getSparseFields().get(type), Collections.<String>emptySet());
+        List<String> relationships = dictionary.getRelationships(loadClass);
```
aklish: This is unused.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -188,15 +201,134 @@ public void commit() {
...
+            return Collections.emptyList();
+        }
+
+        ArrayList<String> list = new ArrayList<>();
+        for (String includeList : includeParam) {
+            for (String includeItem : includeList.split(",")) {
+                for (int idx = 0; idx != -1;) {
+                    idx = includeItem.indexOf('.', idx + 1);
+                    String field = (idx == -1) ? includeItem : includeItem.substring(0, idx);
+                    list.add(field);
```
aklish: How will this work with nested attributes?  comments.author.  I thought setFetchMode is supposed to be a dot separated path.  Do we need to do any processing here at all?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -282,4 +414,14 @@ public void close() throws IOException {
     public User accessUser(Object opaqueUser) {
         return new User(opaqueUser);
     }
+
+    @Override
+    public void setRequestScope(RequestScope requestScope) {
```
aklish: Could this be done in the constructor?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -117,18 +135,13 @@ public void commit() {
 
     @Override
     public <T> Iterable<T> loadObjects(Class<T> loadClass) {
-        @SuppressWarnings("unchecked")
-        Iterable<T> list = new ScrollableIterator(session.createCriteria(loadClass)
-                .scroll(ScrollMode.FORWARD_ONLY));
-        return list;
+        throw new IllegalStateException("" + loadClass);
```
aklish: Should we just remove this function?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -5,10 +5,14 @@
  */
 package com.yahoo.elide.datastores.hibernate3;
 
+import com.google.common.base.Objects;
```
aklish: Can all this be duplicated for hibernate 5?

elide-core/src/main/java/com/yahoo/elide/core/RequestScopedTransaction.java
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2016, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core;
```
aklish: I spent some time thinking about this.  So we have a situation today where the RequestScope takes the transaction as an argument to the constructor and the Transaction now needs access to the RequestScope.  This circular dependency doesn't seem right to me.  If you look at the RequestScope, the only reason it needs a transaction is to save dirty objects.  This save could easily (and more appropriately) been called on the transaction itself.  I think a bit of minor refactoring can fix this.  I would remove transaction from RequestScope, move 'save' from RequestScope to the transaction, and now transaction can take a RequestScope as an argument to the constructor.  Throughts?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -106,10 +116,18 @@ public void commit() {
     @Override
     public <T> T loadObject(Class<T> loadClass, Serializable id) {
         try {
-            @SuppressWarnings("unchecked")
-            T record = (T) session.load(loadClass, id);
-            Hibernate.initialize(record);
-            return record;
+            if (isJoinQuery()) {
```
aklish: I thought the plan was to have some header or request param to enable/disable the join?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
wcekan: We need a way to short circuit **here** if these fields are not readable by _this_ user.  When these remained lazy collections, there is no impact when the user has no read access.  They simply did not look at this collection.  However, doing a join on a column that has denied access is incredibly inefficient.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
wcekan: But only in Guava 18.  Are we ready to force all users to upgrade?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
wcekan: Hibernate also uses the same dot separated path.  However, we need to each element of the entire path.
`include=spouse.children.friend` requires 3 joins:  `spouse`, `spouse.children` and `spouse.children.friend`

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
wcekan: No.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
wcekan: This interface could be deprecated, but would not be removed until 3.x.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
wcekan: In Elide itself or in the implementation of the web service?

wcekan:
Merged and rebased

elide-core/src/main/java/com/yahoo/elide/core/RequestScopedTransaction.java
DeathByTape: I agree with this comment. It does seem unnecessary to place the dirty objects in with `requestScope`. That said, would this cause a breaking change to the interface? I believe it would since we have no good way of defaulting that behavior afaict.

wcekan: We also need to figure out what to do with User in RequestScope.  User cannot be hydrated until a transaction is available.

aklish:
:+1:

PR #191, opened 2016-04-05
Updated expression logic to validate Inline-Only checks.

DeathByTape:
This data unfortunately gets lost along with type-safety from the new and improved expression logic. Expressions do make overall understanding far clearer, however, so it's a worthwhile sacrifice.


clayreimann:
:+1:

yahoocla:
CLA is valid!

PR #195, opened 2016-04-06
Corrected annotation lookup in ExpressionBuilder

wcekan:
The ExpressionBuilder was asking only the class for annotations instead of using EntityDictionary lookup.


aklish:
:+1:

yahoocla:
CLA is valid!

DeathByTape:
:+1:

PR #196, opened 2016-04-07
feature: simplify Builder constructor

clayreimann:


DeathByTape:
:+1:

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -129,31 +130,39 @@ private Elide(AuditLogger auditLogger,
...
+
+        /**
+         * A new builder used to generate Elide instances. Instantiates an {@link EntityDictionary} without
+         * providing a mapping of security checks and uses the provided {@link Slf4jLogger} for audit.
+         *
+         * @param dataStore the datastore used to communicate with the persistence layer
+         */
+        public Builder(DataStore dataStore) {
+            this.dataStore = dataStore;
+            this.auditLogger = new Slf4jLogger();
```
xiaoyao1991: We should probably add an `Builder auditLogger(AuditLogger auditLogger);` method to the builder, otherwise auditLogger cannot be overwritten given the previous builder is deprecated

clayreimann: Good catch. I meant to do that but forgot.

PR #199, opened 2016-04-08
Fix building of criteria checks to occur through PermissionExecutor.

DeathByTape:


wcekan:
Looks fine here for a minimal change to implement a `BypassPermissionExecutor`.
However, would we want more flexibility to customize ParseTree building?

yahoocla:
CLA is valid!

wcekan:
:+1:

wcekan:
:+1: Thank you.  That looks better.

PR #209, opened 2016-04-11
Use Criteria.DISTINCT_ROOT_ENTITY and disable join by default

wcekan:
Disabled any JOIN by default due to cross-product expansion of many queries.

### DIFF
--- a/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
+++ b/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
@@ -238,6 +238,7 @@ public class HibernateTransaction implements RequestScopedTransaction {
             joinCriteria(sessionCriteria, loadClass);
         }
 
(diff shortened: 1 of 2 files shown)


DeathByTape:
:+1:

yahoocla:
CLA is valid!

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -238,14 +238,15 @@ public void commit() {
...
+        sessionCriteria.setResultTransformer(Criteria.DISTINCT_ROOT_ENTITY);
         if (!isScrollEnabled || isJoinQuery()) {
             return sessionCriteria.list();
         }
         return new ScrollableIterator(sessionCriteria.scroll(scrollMode));
     }
 
     /**
-     * Should this transaction use JOINs. By default will be enabled if Sparse fields are available.
+     * Should this transaction use JOINs. By default will be enabled only if parameter <code>join=true</code>.
```
DeathByTape: I wonder if this should be a server parameter since it can be dangerous if not properly configured.

DeathByTape:
:+1:  after it builds

PR #214, opened 2016-04-13
Dropwizard elide more getters

xiaoyao1991:


shadanan:
:+1:

yahoocla:
CLA is valid!

elide-contrib/dropwizard-elide/src/main/java/com/yahoo/elide/contrib/dropwizard/elide/ElideBundle.java
```diff
@@ -130,6 +137,14 @@ public DataStore getDataStore(T configuration, Environment environment) {
         return new HibernateStore(sessionFactory);
     }
 
+    public ImmutableList<Class<?>> getEntities() {
```
clayreimann: Why this?

xiaoyao1991: so that classes that implement `ElideBundle` can get access to the entities. They can build different dataStore out of it

DeathByTape:
:+1:

Issue #207, opened 2016-04-11
Filtering on java.sql.Timestamp throwing InvalidValueException


DeathByTape:
We're throwing an `InvalidValueException` when receiving a `Long` on a `java.sql.Timestamp` object. I suspect this is due to poor coercion. What we likely should do first for a given type is try to find a constructor that accepts this type (i.e. if direct coercion is not possible) and use this. Will require some more investigation first to confirm this issue/solution.

DeathByTape:
After further investigation, this issue appears to be artificial. Conversion requires a timestamp; we were eating the actual exception from our conversion utility:

```
String must be in JDBC format [yyyy-MM-dd HH:mm:ss.fffffffff] to create a java.sql.Timestamp
```

clayreimann:
Would it be useful to pass this warning back to the caller? Perhaps behind `verboseMessage`?

cbou:
I had the same issue. I had to use the unix timestamp in ms.

`{"data":{"id":"3","attributes":{"name":"My project","myDate":"1479249799770"},"type":"projects"}}`

aklish:
What version of Elide are you using?

Did you configure Elide to use ISO8601 dates?

https://github.com/yahoo/elide/blob/master/elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java#L66-L69

Thanks.

Aaron

cbou:
Thanks Aaron!

I have to admit I totally miss this settings. No idea how to configure them. However I'm not using the Standalone.  I use `elide-spring-boot` which is based on `elide-datastore-hibernate5`. Will look at this.

PR #215, opened 2016-04-13
Added verbose message for invalid value conversions.

DeathByTape:
Addresses @clayreimann's comment on https://github.com/yahoo/elide/issues/207


xiaoyao1991:
👍

PR #218, opened 2016-04-15
Improve CheckExpression performance

wcekan:
Now caches UserCheck results once for user instead of per resource.
Reduced EntityPermissions maps.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
@@ -27,8 +27,8 @@ import java.lang.reflect.AccessibleObject;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Extract permissions related annotation data for a model.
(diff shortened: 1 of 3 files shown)


elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/ImmediateCheckExpression.java
```diff
@@ -69,11 +72,12 @@ public ExpressionResult evaluate() {
         Class<? extends Check> checkClass = check.getClass();
 
         final ExpressionResult result;
-        if (!cache.hasStoredResultFor(checkClass, resource)) {
+        if (cache.hasStoredResultFor(checkClass, resource)) {
```
clayreimann: So here we cache `UserCheck`s against `null`?

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
```diff
@@ -45,49 +45,46 @@
...
-    //CHECKSTYLE.OFF: LineLength
-    private final ConcurrentHashMap<Class<? extends Annotation>, ParseTree> classPermissions = new ConcurrentHashMap<>();
-    private final ConcurrentHashMap<String, Map<Class<? extends Annotation>, ParseTree>> fieldPermissions = new ConcurrentHashMap<>();
-    //CHECKSTYLE.ON: LineLength
+    private static final AnnotationBinding EMPTY_BINDING = new AnnotationBinding(null, Collections.emptyMap());
+    private final HashMap<Class<? extends Annotation>, AnnotationBinding> bindings = new HashMap<>();
+
+    private static class AnnotationBinding {
+        final ParseTree classPermission;
+        Map<String, ParseTree> fieldPermissions;
```
shadanan: final?

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/ImmediateCheckExpression.java
wcekan: Yes

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
wcekan: sure

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
```diff
@@ -27,71 +27,74 @@
...
+            this.fieldPermissions = fieldPermissions.isEmpty() ? Collections.emptyMap() : fieldPermissions;
+        }
     }
 
-    public EntityPermissions(Class<?> cls, Collection<AccessibleObject> fieldOrMethodList) {
-        initFieldPermissionsMap(fieldOrMethodList);
 
-        for (Class annotationClass : PERMISSION_ANNOTATIONS) {
-            bindClassPermissions(cls, annotationClass);
+    private EntityPermissions() {
```
aklish: What's this for?

aklish:
👍

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
DeathByTape: Looks like it's the single constructor for the `EMPTY_PERMISSIONS` (L#47). I suggest we add a comment here.

wcekan: See [line 53 above](https://github.com/yahoo/elide/pull/218/files#diff-556a747049d9e86339fc6176b6bd183dL53)

PR #219, opened 2016-04-18
Adding full className upon instantiation failure.

DeathByTape:


clayreimann:
seems reasonable to me 👍

PR #221, opened 2016-04-19
Added sane defaults for Audit.

DeathByTape:


elide-annotations/src/main/java/com/yahoo/elide/annotation/Audit.java
```diff
@@ -46,33 +46,33 @@
...
     String path() default "";
 
     /**
      * Operation code.
      *
-     * @return the int
+     * @return the int (default = 1)
      */
-    int operation();
+    int operation() default 1;
```
aklish: -1?

elide-annotations/src/main/java/com/yahoo/elide/annotation/Audit.java
```diff
@@ -46,33 +46,33 @@
...
 
     /**
      * Logging string template.
-     * @return the string
+     * @return the string (default = "Generic log")
      */
-    String logStatement();
+    String logStatement() default "Generic log";
```
aklish: empty string

aklish:
:+1:

yahoocla:
CLA is valid!

elide-annotations/src/main/java/com/yahoo/elide/annotation/Audit.java
```diff
@@ -46,33 +46,33 @@
...
-    String logStatement();
+    String logStatement() default "";
 
     /**
      * Unified expression language expressions that will be evaluated and substituted into the logging template.
-     * @return the string [ ]
+     * @return the string [ ] (default = "")
      */
-    String [] logExpressions();
+    String [] logExpressions() default "";
```
clayreimann: Unspace this to match the definition above.

Issue #225, opened 2016-04-22
Provide a way to disable inheriting permissions from a super class.


aklish:
@Include should have an attribute to disable permission inheritance from its super class.  While permissions can be overridden on a field by field basis, there is no way to start with clean slate.

aklish:
Hibernate isn't happy with this idea (can't override getters for persistent fields - even if they are from a mapped superclass) - so we are going to accomplish this another way.

PR #226, opened 2016-04-25
Provide a way to override permissions when inheriting

aklish:
#225

### DIFF
--- a/elide-annotations/src/main/java/com/yahoo/elide/annotation/Include.java
+++ b/elide-annotations/src/main/java/com/yahoo/elide/annotation/Include.java
@@ -32,4 +32,10 @@ public @interface Include {
      * @return the string
      */
     String type() default "";
+
+    /**
+     * (Optional) Whether or not this class inherits permissions from a super class.
+     * @return
+     */
+    boolean inheritPermissions() default true;
 }
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
@@ -550,7 +550,7 @@ public class EntityDictionary {
             throw new DuplicateMappingException(type + " " + cls.getName() + ":" + duplicate.getName());
         }
 
-        entityBindings.putIfAbsent(lookupEntityClass(cls), new EntityBinding(cls, type));
+        entityBindings.putIfAbsent(lookupEntityClass(cls), new EntityBinding(cls, type, include.inheritPermissions()));
         if (include.rootLevel()) {
             bindEntityRoots.add(cls);
         }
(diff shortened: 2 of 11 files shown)


elide-annotations/src/main/java/com/yahoo/elide/annotation/Include.java
```diff
@@ -32,4 +32,10 @@
      * @return the string
      */
     String type() default "";
+
+    /**
+     * (Optional) Whether or not this class inherits permissions from a super class.
+     * @return
+     */
+    boolean inheritPermissions() default true;
```
clayreimann: Do we want to invert this? If we expect micro-services to be the norm this probably wants to be `false`, because the default should be what we expect people to use in the normal case. We shouldn't default to `true` just because defaulting to `false` would be a breaking change.

aklish: There is another norm that might be the common case though.  That is having mapped super classes and sticking permissions in them that users want to be inherited.  I'm not sure how common our microservice plan is.  It seems like something that a mature organization might encounter - but most folks I would expect to have a single service.  Anyway, in the absence of knowing what is truly comment, I would vote to not break the contract.

clayreimann: That seems reasonable.

DeathByTape:
👍

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -629,12 +629,17 @@ public void bindEntity(Class<?> cls) {
...
+     * @param skipClassHierarchy Whether or not to include the class hierarchy in the search
      * @return annotation found
      */
     public static Annotation getFirstAnnotation(Class<?> entityClass,
-                                                List<Class<? extends Annotation>> annotationClassList) {
+                                                List<Class<? extends Annotation>> annotationClassList,
+                                                boolean skipClassHierarchy) {
         Annotation annotation = null;
         for (Class<?> cls = entityClass; annotation == null && cls != null; cls = cls.getSuperclass()) {
+            if (!cls.equals(entityClass) && skipClassHierarchy) {
```
wcekan: It is unnecessary to loop through remaining classes.  This could simply move down as
```java
if (skipClassHierarchy) {
   break;
}
```

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -655,6 +660,18 @@ public static Annotation getFirstAnnotation(Class<?> entityClass,
     }
```
wcekan:
```java
if (skipClassHierarchy) {
   break;
}
```

elide-core/src/test/java/example/FosterParent.java
```diff
@@ -0,0 +1,25 @@
...
+import com.yahoo.elide.security.checks.prefab.Role;
+import lombok.ToString;
+
+import javax.persistence.Entity;
+import java.util.Set;
+
+@Include(rootLevel = true, type = "foster", inheritPermissions = false)
+@Entity
+@ToString
+public class FosterParent extends Parent {
```
wcekan: I seem to have a Hibernate error if I include this FosterParent class in the integration tests.

aklish:
Won't fix - see issue description.

PR #230, opened 2016-05-02
Avoid read on create

wcekan:
@DeathByTape Added a test to reproduce the post-processing read collection on create problem.

### DIFF
--- a/elide-integration-tests/src/test/java/example/CreateButNoRead.java
+++ b/elide-integration-tests/src/test/java/example/CreateButNoRead.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015, Yahoo Inc.
+ * Copyright 2016, Yahoo Inc.
  * Licensed under the Apache License, Version 2.0
  * See LICENSE file in project root for terms.
  */
--- a/elide-integration-tests/src/test/java/example/CreateButNoRead.java
+++ b/elide-integration-tests/src/test/java/example/CreateButNoRead.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2015, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package example;
+
+import com.yahoo.elide.annotation.CreatePermission;
+import com.yahoo.elide.annotation.Include;
+import com.yahoo.elide.annotation.ReadPermission;
+import com.yahoo.elide.security.ChangeSpec;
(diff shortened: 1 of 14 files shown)


yahoocla:
CLA is valid!

elide-integration-tests/src/test/java/example/CreateButNoRead.java
```diff
@@ -0,0 +1,48 @@
...
+ * A model intended to be ONLY created and read, but never updated
+ */
+@Include(rootLevel = true)
+@Entity
+@CreatePermission(any = {Role.ALL.class})
+public class CreateButNoRead extends BaseId {
+    private Set<CreateButNoReadChild> otherObjects;
+
+    @OneToMany(fetch = FetchType.LAZY)
+    @ReadPermission(any = {NOREAD.class})
```
wcekan: @DeathByTape This read check is deferred on patchext add for the newly created child object during post-processing.

elide-integration-tests/src/test/java/example/CreateButNoRead.java
```diff
@@ -0,0 +1,48 @@
+/*
+ * Copyright 2015, Yahoo Inc.
```
DeathByTape: 2016

elide-integration-tests/src/test/java/example/CreateButNoReadChild.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2015, Yahoo Inc.
```
DeathByTape: 2016

DeathByTape:
Tests failing

wcekan:
@DeathByTape Yes, tests are failing.  I recreated the issue, but not the solution.

DeathByTape:
I should have some cycles to look into this soon.

DeathByTape:
After some offline discussion, this appears to be working as intended (as per hierarchical security assumptions). A quick recap below:

We have two objects, `A` and `B`. Each of which can be created but not read. The case we're hitting here is trying to create a rootable object `A` and then another object `B` **through** `A`. Essentially, what this would look like outside of a transaction is the following:

```
POST /A
POST /A/1/B
```

Clearly such a request should not work. If we relied on hierarchical security (i.e. allow all access to `B` since we believe its only exposition to be _through_ `A`), then we effectively open up any action to `B` and this breaks our model.

This test should remain, but the fix should be to expect the failure rather than a successful operation.

aklish:
👍

PR #231, opened 2016-05-02
Audit change details

DeathByTape:


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -282,7 +282,7 @@ public boolean matchesId(String checkId) {
         Iterable<T> list;
         FilterScope filterScope = new FilterScope(requestScope, loadClass);
         list = tx.loadObjects(loadClass, filterScope);
-        Set<PersistentResource<T>> resources = new PersistentResourceSet(list, requestScope);
+        Set<PersistentResource<T>> resources = new PersistentResourceSet(null, list, requestScope);
```
aklish: Can we make another constructor that doesn't take a parent instead of passing null.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1217,8 +1217,10 @@ public Resource toResource(final Supplier<Map<String, Relationship>> relationshi
      * @param newValue the new value
      */
     protected void setValueChecked(String fieldName, Object newValue) {
-        checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, newValue, getValueUnchecked(fieldName));
+        final Object original = getValueUnchecked(fieldName);
+        checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, newValue, original);
```
aklish: Let's have a version of checkPermission that takes a changeSpec - and use that here.

aklish:
👍

PR #233, opened 2016-05-05
Add string to date support

willmsn98:


yahoocla:
CLA is valid!

DeathByTape:
My only question is does this now require a formatted string? Can we accept either long or formatted string?

willmsn98:
I was thinking that users could annotate their bean with either long or java.util.Date.  Date would require a formatted string, but they could also use long if they don't want to use a string.

DeathByTape:
I think that's fair. If we find there are a lot of people who want longs to just work out of the box in the future, we can always add additional support.

👍

PR #236, opened 2016-05-10
KunderaStore

QubitPi:


yahoocla:
CLA is valid!

clayreimann:
Would you please add an example of how this is used in the examples pom.

QubitPi:
@clayreimann Do you want me to add the example now or after Kundera has been integrated into Elide?

clayreimann:
You can add the examples in this PR, should work just fine

QubitPi:
@clayreimann OK

elide-datastore/elide-datastore-kundera/src/main/java/com/yahoo/elide/datastores/kundera/KunderaStore.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Begin Kundera transaction.
+     *
+     * @return transaction
+     */
+    @Override
+    public DataStoreTransaction beginTransaction() {
+        EntityManager entityManager = entityManagerFactory.createEntityManager();
+        Preconditions.checkNotNull(entityManager);
+        return new KunderaTransaction(entityManager);
```
aklish: Where is the KunderaTransaction?  How about tests similar to what we have for other data stores?

QubitPi: @aklish I was intending to work progressively so that I could get helpful comments on the way. This is actually an incompleted work. I am modifying this to move to Redis.

Yeah, I will implement DataStoreTransaction, all necessary tests, and example projects in separate commits.

PR #237, opened 2016-05-11
Allow any HTTP status code to be thrown from within custom business l…

aklish:
…ogic

This PR allows custom logic inside a DataStoreTransaction or a trigger to control the error returned by Elide by simply using HttpStatusException (rather than having to explicitly override it).


yahoocla:
CLA is valid!

DeathByTape:
So, I believe this does what you say it does. What I don't believe is that we won't accidentally break it in the future :) could we get a test for this?

DeathByTape:
👍

PR #238, opened 2016-05-11
Enhance Logging for security failures (minus 'share' which will be separate PR)

aklish:
Share permission will require more work so getting this tiny change out first.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
@@ -47,6 +47,7 @@ import java.util.concurrent.ConcurrentLinkedDeque;
  * @see com.yahoo.elide.annotation.Include#type
  */
 class EntityBinding {
+
     private static final List<Method> OBJ_METHODS = Arrays.asList(Object.class.getMethods());
 
     public final Class<?> entityClass;
@@ -55,7 +56,9 @@ class EntityBinding {
     @Getter private String idFieldName;
     @Getter private Class<?> idType;
     @Getter @Setter private Initializer initializer;
+
     public final EntityPermissions entityPermissions;
+    private final EntityDictionary dictionary;
 
     public final List<String> attributes;
     public final List<String> relationships;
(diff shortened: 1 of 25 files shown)


yahoocla:
CLA is valid!

aklish:
Don't merge yet.  I want to improve the expression logging as well.

aklish:
Ok - ready to look at.  Another PR will come to address logging of share.

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -55,7 +56,9 @@
     @Getter private String idFieldName;
     @Getter private Class<?> idType;
     @Getter @Setter private Initializer initializer;
+
     public final EntityPermissions entityPermissions;
+    private final EntityDictionary dictionary;
```
DeathByTape: Unused? Is it necessary to store this?

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionResult.java
```diff
@@ -5,54 +5,23 @@
...
 
     @Override
     public String toString() {
-        StringBuilder sb = new StringBuilder();
-        sb.append(status);
-        if (failureMessage != null) {
-            sb.append(": ").append(failureMessage);
-        }
-        return sb.toString();
+        return name;
```
DeathByTape: I recommend using [Jansi](https://github.com/fusesource/jansi) here. It would be great to get some colors based on the message. Perhaps in the constructor for the enum, a color code could be included and this statement would be come:

```java
return ansi().fg(colorCode).a(name).reset();
```

elide-core/src/test/java/com/yahoo/elide/security/PermissionExecutorTest.java
```diff
@@ -117,6 +117,7 @@ public void testReadFieldAwareSuccessAll() {
         PersistentResource resource = newResource(SampleBean.class);
         RequestScope requestScope = resource.getRequestScope();
         requestScope.getPermissionExecutor().checkSpecificFieldPermissions(resource, new ChangeSpec(null, null, null, null), ReadPermission.class, "allVisible");
+        //requestScope.getPermissionExecutor().checkSpecificFieldPermissions(resource, null, ReadPermission.class, "allVisible");
```
DeathByTape: Stale

elide-core/src/main/java/com/yahoo/elide/security/permissions/ExpressionResult.java
aklish: Maybe another PR?

wcekan:
👍  Found it very useful for debugging permission errors.

DeathByTape:
👍

PR #243, opened 2016-05-18
Added support for calling lucene queries

willmsn98:
DONT MERGE - added a pull request to best talk about design.


yahoocla:
CLA is valid!

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/filter/LuceneFilter.java
```diff
@@ -0,0 +1,75 @@
+/*
+ * Copyright 2015, Yahoo Inc.
```
DeathByTape: 2016

DeathByTape:
This PR appears to be stale. Closing. Can reopen in the future if we want to continue working on this.

PR #244, opened 2016-05-18
[DON'T MERGE] HBase design proposal

xiaoyao1991:
For design review purpose. Still WIP for TBD items.

### DIFF
--- a/elide-hbase.md
+++ b/elide-hbase.md
@@ -1,0 +1,159 @@
+# Proposal: HBase with Elide
+
+### Goal
+- To support basic Java persistence functionalities(CRUD) for HBase entities through Elide.
+- To bridge with other existed data store types in Elide.
+
+### HBase JPA adaptor
+1. Define `HBasePersistence` that implements `javax.persistence.spi.PersistenceProvider` that reads a `persistence.xml` with custom configs(including hbase nodes, ports, namespace)
+1. Define `HBaseEntityManagerFactory` that implements `javax.persistence.EntityManagerFactory`
+1. (Optional) Define `HBaseEntityManager` that implements `javax.persistence.EntityManager`. It should use the HBase Java Client to perform CRUD operations with HBase.
+1. Define `HBaseStore` that implements `com.yahoo.elide.core.DataStore`
+1. Define `HBaseTransaction` that implements `com.yahoo.elide.core.DataStoreTransaction`
+
+### Entity beans
+In order to support having byte arrays(or any arbitrary primitive types) as column qualifiers and column family names in HBase, some indirection hacks will need to be done with the `@Column` annotation.
+
+The entity bean should contain a map where the key is a String, and value is an Object that represents the actual column family/column qualifier names.
+
+The `name` attribute in `@Column` should look like `"cfPointer1:cqPointer1"`, where `cfPointer1` and `cqPointer1` are keys in the map described above. If keys cannot be found, we consider the name as is.
+
+For example:
+If there's a Person HTable, and looks like:  
+```
+ROW    COLUMN+CELL
+\x01   \x02:\x03, timestamp=1431924360459, value=\x04
+```
+
+Then the HBase entity bean should look like:
+```
+@Entity
+@Table(name = "person")
+@DataStore(cls = HBaseStore.class)
+public class Person {
+  // This map can be injected, doesn't have to be static
+  private final Map<String, Object> columnMap = ImmutableMap.of(
+      "cf1", new byte[]{2}, // Column Family
+      "cq1", new byte[]{3}  // Column Qualifier
+  );
+
+  @Id
+  private byte[] rowKey;
+
+  @Column(name = "timestamp")
+  private Long timestamp;
+
+  @Column(name = "cf1:cq1")
+  private byte[] value;
+
+  @OneToOne/@OneToMany/@ManyToOne/@ManyToMany
+  @DataStore(cls = HibernateStore.class)
+  public Set<HobbyGroup> getHobbyGroups() { return Collections.emptySet(); }
+}
+```
+
+A Relational Database entity bean should look like:
+```
+@Entity
+@Table(name = "hobbyGroup")
+public class HobbyGroup {
+  @Id
+  private long id;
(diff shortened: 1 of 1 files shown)


yahoocla:
CLA is valid!

elide-hbase.md
```diff
@@ -0,0 +1,159 @@
...
+      return rowKey;
+    }
+  }
+}
+```
+
+### Bridging DataStores
+1. Must use a `MultiplexDataStore`
+1. `PersistenceResource` will still be responsible for holding security check logics. `DataStoreTransaction` will have more to do with getting/setting attr, relations, etc. Methods in `DataStoreTransaction` aren't aware of securities at all
+1. Have a custom annotation `@DataStore(cls)`. This will be used to annotate relations of a different data store type in an entity bean.
```
aklish: We'll need to know the relationship cardinality oneToOne, oneToMany, ManToOne, ManyToMany so the entity dictionary can reflect this.  I suggest we either make an elide set of these annotations OR have a single annotation with an attribute that describes this cardinality.

elide-hbase.md
```diff
@@ -0,0 +1,159 @@
...
+1. Add a field `parentTransaction` to `DataStoreTransaction`, so that when using a `MultiplexTransaction`, a detail data store transaction object can get access to its parent(the MultiplexTransaction) and get to its peer(a transaction of a different data store type):
+```
+class DataStoreTransaction {
+  default DataStoreTransaction getParentTransaction() { return null; }
+}
+```
+1. Define `getRelation(Object entity, String relationName)` in `DataStoreTransaction`:
+```
+class DataStoreTransaction {
+  <T> Collection<T> getRelation(Object entity, String relationName) {
```
aklish: We should also push filters, pagination, and sorting to the data store - so there should be a similar function with those args.

elide-hbase.md
```diff
@@ -0,0 +1,159 @@
...
+    // Default implementation should call the getter method(just like the getValue() in PersistenceResource)
+    // if the getter method is annotated with @DataStore, retrieve the transaction for the specified datastore from the parent multiplex transaction,
+    // and then use it to loadObjects, with the Criteria/FilterScope computed by @RowKeyFn(or maybe rename it) function.
+  }
+}
+```
+1. Define `updateRelation(Object entity, String fieldName, Set<PersistentResource> resourceIdentifiers)` in `DataStoreTransaction`:
```
aklish: Do we need to distinguish between added and deleted here?

elide-hbase.md
```diff
@@ -0,0 +1,159 @@
...
+  boolean setRelation(Object entity, String fieldName, Set<PersistentResource> resourceIdentifiers) {
+    // Default implementation should call the setter method(just like the setValue() in PersistenceResource)
+    // if the setter method is annotated with @DataStore, retrieve the transaction for the specified datastore from the parent multiplex transaction,
+    // and then use it to save(serialize), with the Criteria/FilterScope computed by @RowKeyFn(or maybe rename it) function.
+  }
+}
+```
+
+#### Getting & Setting attributes
+Attributes of an entity should be mostly within the same data store. It's not usual that a MySQL entity has an *attribute* that lives in HBase(in those cases, a better data model design would be making that attribute a relation.)
```
aklish: While I agree this is probably unusual, I'm not sure we can make the case that this won't be needed.

elide-hbase.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Bridging DataStores
+1. Must use a `MultiplexDataStore`
+1. `PersistenceResource` will still be responsible for holding security check logics. `DataStoreTransaction` will have more to do with getting/setting attr, relations, etc. Methods in `DataStoreTransaction` aren't aware of securities at all
+1. Have a custom annotation `@DataStore(cls)`. This will be used to annotate relations of a different data store type in an entity bean.
+```
+@DataStore(cls=HBaseStore.class)
+public Collection<HBaseEntity> getHBaseEntities() { return Collections.emptySet(); }
+```
+1. Add a field `parentTransaction` to `DataStoreTransaction`, so that when using a `MultiplexTransaction`, a detail data store transaction object can get access to its parent(the MultiplexTransaction) and get to its peer(a transaction of a different data store type):
```
aklish: Why is this needed?

elide-hbase.md
xiaoyao1991: The `@DataStore` annotation is not for registering classes. It's used to annotate a relation(field or method) that any CRUD on that relation should use that dataStore specified in the annotation.

xiaoyao1991: I think you're right that `@DataStore` is unnecessary.

elide-hbase.md
xiaoyao1991: My original thought is to have a parent transaction reference, so when seeing `@DataStore(cls)` annotation, we can get the transaction of the specific data store type by traversing up to the parent and down to the leaf. But after reading the comment above, `@DataStore` is not necessary, and transaction can be retrieved by looking up in `EntityDictionary` I guess.

elide-hbase.md
xiaoyao1991: Q: Will Hibernate (polyglot persistenc) respect the relationship annotations (OneToOne, etc) or will it blow up and think it needs to 'manage' the entity? 
A: Just experimented, Hibernate blows up when it sees a relationship annotated field which it is not aware of. Hibernate has its own validation logic on the validity of annotated fields. To workaround that may involve overriding and replacing a lot of Hibernate code. So it looks like an easy way around is to leverage the `@Transient` and `@ComputedAttributes` to annotate the relation fields.

xiaoyao1991: One thing that may be semantically incorrect if we do that way is that, with `@ComputedAttribute`, the results of that field will appear in the `attributes` section of the JSON output. Ideally, I would prefer to have a `@ComputedRelation` where the results of that field appear in the `relationships` section of the JSON

elide-hbase.md
xiaoyao1991: I feel like attributes are mostly just primitives or simple object types(like Date). I think it'll be hard to know which dataStore to source from for a given primitive attribute.

PR #245, opened 2016-05-24
ComputedRelationship Annotation

xiaoyao1991:
Have a `@ComputedRelationship` annotation that resembles the `@ComputedAttribute` except that the results of annotated field/method will show up in `relationships` section instead of `attributes` in JSON response.


yahoocla:
CLA is valid!

yahoocla:
CLA is valid!

DeathByTape:
Change LGTM, but what is the use-case for this?

DeathByTape:
👍

PR #248, opened 2016-05-26
Fix inverse relationship checks to have changeSpec on delete.

DeathByTape:


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1297,14 +1297,32 @@ protected boolean addToCollection(Collection collection, String collectionName,
...
                 collectionName,
                 CollectionUtils.disjunction(collection, Collections.singleton(toDelete.getObject())),
-                original);
+                original
+        );
 
         String inverseField = getInverseRelationField(collectionName);
         if (!inverseField.isEmpty()) {
-            checkPermission(UpdatePermission.class, toDelete);
+            final Object originalValue = toDelete.getValueUnchecked(inverseField);
```
wcekan: Briefly describe the problem here.

asifmansoora:
Tested locally.. its all good 👍

PR #249, opened 2016-05-27
Redis

QubitPi:


yahoocla:
CLA is valid!

elide-example/elide-hibernate5-redis-example/src/main/java/com/yahoo/elide/example/persistence/ElideResourceConfig.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Example application for resource config
+ */
+public class ElideResourceConfig extends ResourceConfig {
+    public ElideResourceConfig() {
+        register(new AbstractBinder() {
+            @Override
+            protected void configure() {
+                JsonApiEndpoint.DefaultOpaqueUserFunction noUserFn = v -> null;
```
DeathByTape: Not really a comment for change here, but we really should provide an example for how to properly interleave users and user auth.

DeathByTape:
The code looks good to me, but I would just suggest we add a redis-specific integration test to the hibernate5 datastore.

DeathByTape:
Is this branch now stale @QubitPi? (i.e. should we close it?)

QubitPi:
@aklish would like to add Redis after Elide 3.0 is released. I'm OK with closing this one, but I will re-do this PR anyway if we still plan to add Redis

PR #250, opened 2016-05-27
Delegate getRelation to DataStoreTransaction

xiaoyao1991:
As the first step of changes for bridging multiple data stores. Following proposal at https://github.com/yahoo/elide/blob/dd8751e258f9170eac1ddb1b1385c687b392268c/elide-hbase.md

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
@@ -118,7 +118,9 @@ public interface DataStoreTransaction extends Closeable {
      * @param entityClass the class of the entities in the collection
      * @param predicates  the set of Predicate's to filter by
      * @return the filtered collection
+     * @deprecated Since 2.4, instead implement the filtering logic in detail methods in implementations
      */
+    @Deprecated
     default <T> Collection filterCollection(Collection collection, Class<T> entityClass, Set<Predicate> predicates) {
         return collection;
     }
@@ -133,7 +135,9 @@ public interface DataStoreTransaction extends Closeable {
      * @param pagination The optional Pagination object
      * @param <T> The type parameter
      * @return The optionally filtered, sorted and paginated collection
+     * @deprecated Since 2.4, instead implement the filtering logic in detail methods in implementations
      */
+    @Deprecated
     default <T> Collection filterCollectionWithSortingAndPagination(Collection collection, Class<T> entityClass,
                                                           EntityDictionary dictionary, Optional<Set<Predicate>> filters,
                                                           Optional<Sorting> sorting, Optional<Pagination> pagination) {
@@ -148,17 +152,8 @@ public interface DataStoreTransaction extends Closeable {
             EntityDictionary dictionary,
             Set<Predicate> filters
     ) {
-        Object val = PersistentResource.getValue(entity, relationName, dictionary);
-        if (val instanceof Collection) {
-            Collection filteredVal = (Collection) val;
-
-            if (!filters.isEmpty()) {
-                filteredVal = filterCollection(filteredVal, relationClass, filters);
-            }
-            return filteredVal;
-        }
-
-        return val;
+        // Default implementation ignores filters
+        return PersistentResource.getValue(entity, relationName, dictionary);
     }
 
     default <T> Object getRelationWithSortingAndPagination(
@@ -171,22 +166,7 @@ public interface DataStoreTransaction extends Closeable {
             Sorting sorting,
             Pagination pagination
     ) {
-        Object val = PersistentResource.getValue(entity, relationName, dictionary);
-        if (val instanceof Collection) {
-            Collection filteredVal = (Collection) val;
-            // sorting/pagination supported on last entity only eg /v1/author/1/books? books would be valid
-            final boolean hasSortRules = sorting.isDefaultInstance();
-            final boolean isPaginated = pagination.isDefaultInstance();
-
-            if (!filters.isEmpty() || hasSortRules || isPaginated) {
-                final Optional<Sorting> sortingRules = hasSortRules ? Optional.of(sorting) : Optional.empty();
-                final Optional<Pagination> paginationRules = isPaginated ? Optional.of(pagination) : Optional.empty();
-                filteredVal = filterCollectionWithSortingAndPagination(filteredVal,
-                        relationClass, dictionary, Optional.of(filters), sortingRules, paginationRules);
-            }
-            return filteredVal;
-        }
-
-        return val;
+        // Default implementation ignores filters, sorting and pagination
+        return PersistentResource.getValue(entity, relationName, dictionary);
     }
 }
--- a/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
@@ -139,4 +139,28 @@ public interface DataStoreTransaction extends Closeable {
                                                           Optional<Sorting> sorting, Optional<Pagination> pagination) {
         return collection;
     }
+
+    default <T> Object getRelation(
+            Object entity,
+            RelationshipType relationshipType,
+            String relationName,
+            Class<T> relationClass,
+            EntityDictionary dictionary,
+            Set<Predicate> filters
+    ) {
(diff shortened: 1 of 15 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -888,7 +892,10 @@ protected boolean checkRelation(String relationName) {
      */
     private Set<PersistentResource> getRelationUnchecked(String relationName, Set<Predicate> filters) {
         RelationshipType type = getRelationshipType(relationName);
-        Object val = getValueUnchecked(relationName);
+        final Class<?> entityClass = dictionary.getParameterizedType(obj, relationName);
```
clayreimann: Is there a good reason for `Class<?>` as opposed to `Class`?

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
```diff
@@ -97,8 +101,10 @@ public PersistentResourceTest() {
                 null,
                 new RequestScope(null, null, null, new EntityDictionary(new HashMap<>()), null, MOCK_AUDIT_LOGGER)
         );
-        goodUserScope = new RequestScope(null, null, new User(1), dictionary, null, MOCK_AUDIT_LOGGER);
-        badUserScope = new RequestScope(null, null, new User(-1), dictionary, null, MOCK_AUDIT_LOGGER);
+        goodUserScope = new RequestScope(null, mock(DataStoreTransaction.class, Answers.CALLS_REAL_METHODS),
```
clayreimann: does `CALLS_REAL_METHODS` cause the mock to fall back on the actual method if you don't provide a mock value?

clayreimann:
Seems reasonable to me.

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
xiaoyao1991: I believe so. The reason I use it is because normal mock doesn't respect the `default` implementations of an interface, which in this PR is a point of change.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
clayreimann: From @xiaoyao1991: http://stackoverflow.com/questions/9921676/what-does-class-mean-in-java

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -139,4 +139,28 @@ default void flush() {
...
+
+    default <T> Object getRelation(
+            Object entity,
+            RelationshipType relationshipType,
+            String relationName,
+            Class<T> relationClass,
+            EntityDictionary dictionary,
+            Set<Predicate> filters
+    ) {
+        return PersistentResource.getValue(entity, relationName, dictionary);
```
DeathByTape: Logically, something feels wrong calling in from `DataStoreTransaction` to `PersistentResource`. Maybe we should find a common place to put the reflection code that runs `getValue()`?

xiaoyao1991: I agree. Calling methods in PersistentResource from transaction feels weird. The two classes should be parallel. 

I'm not against still keeping the Dictionary tho. What's the concern of sharing dictionary around?

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -219,19 +219,19 @@ private void bindAttrOrRelation(Class<?> cls, AccessibleObject fieldOrMethod) {
             RelationshipType type;
             String mappedBy;
             if (oneToMany) {
-                type = RelationshipType.ONE_TO_MANY;
+                type = computedRelationship ? RelationshipType.COMPUTED_ONE_TO_MANY : RelationshipType.ONE_TO_MANY;
```
DeathByTape: Just to verify: using the existing `@ManyToMany`, `@OneToOne`, etc. annotations will require any compound models to be _distinct_ from their persistent representations. Otherwise, even though we claim this is a computed attribute, hibernate schema generation and similar will pick these up as concrete fields.

DeathByTape:
Just to verify: using the existing `@ManyToMany`, `@OneToOne`, etc. annotations will require any compound models to be _distinct_ from their persistent representations. Otherwise, even though we claim this is a computed attribute, hibernate schema generation and similar will pick these up as concrete fields.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
DeathByTape: I don't think it is necessary to share a dictionary; these are static methods. That being said, I don't think they belong in the dictionary because it's actually operating on the entity rather than scanning its metadata. Maybe that is alright. @aklish, @clayreimann , or @wcekan-- any opinions on this?

DeathByTape: The more I think about it, perhaps the dictionary is the appropriate place. I recommend moving `getValue` into the dictionary as a static method and using it from there rather than the `PersistentResource`. Any objections, speak now! :)

clayreimann: Acutally, I have a question. Why are we passing `Object entity` rather than `PersistentResource resource` and then just calling `resource.getRelation` directly. We're trying to express our current assumption of everything being a proxy, right, so let's just call the method on the proxy.

xiaoyao1991: I don't think we're gonna make that assumption now. We haven't ready yet to make proxy-ize everything. Designing proxies for other data stores might be a lot of work. 

I'm trying to avoid involving PersistentResource into transactions, they would be pointing to each other and logic may entangle.

xiaoyao1991:
Eh oh, my new commit just folded our discussion @DeathByTape @clayreimann 
The discussion on where `PersistentResource.getValue` should be placed is still valid.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -139,4 +139,54 @@ default void flush() {
...
+
+    default <T> Object getRelation(
+            Object entity,
+            RelationshipType relationshipType,
+            String relationName,
+            Class<T> relationClass,
+            EntityDictionary dictionary,
+            Set<Predicate> filters
+    ) {
+        Object val = PersistentResource.getValue(entity, relationName, dictionary);
```
clayreimann: To try and resurrect the previous discussion: why are we passing `entity` in as an `Object`. Why not just pass it in as a `PersistentResource` and call `getValue` on it directly?

xiaoyao1991: I'm trying to avoid involving PersistentResource into transactions, they would be pointing to each other and logic may entangle.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -139,4 +139,54 @@ default void flush() {
...
+        if (val instanceof Collection) {
+            Collection filteredVal = (Collection) val;
+            // sorting/pagination supported on last entity only eg /v1/author/1/books? books would be valid
+            final boolean hasSortRules = sorting.isDefaultInstance();
+            final boolean isPaginated = pagination.isDefaultInstance();
+
+            if (!filters.isEmpty() || hasSortRules || isPaginated) {
+                final Optional<Sorting> sortingRules = hasSortRules ? Optional.of(sorting) : Optional.empty();
+                final Optional<Pagination> paginationRules = isPaginated ? Optional.of(pagination) : Optional.empty();
+                filteredVal = filterCollectionWithSortingAndPagination(filteredVal,
```
aklish: Let's deprecate filterCollection and filterCollectionWithSortingAndPagination.  That functionality should move into the specific implementation of the data store transactions.

aklish:
👍

DeathByTape:
👍

PR #253, opened 2016-06-03
Add a preCommit method to DataStoreTransaction

aklish:
This method will be invoked prior to commit checks, save, flush, and commit.


yahoocla:
CLA is valid!

DeathByTape:
👍

wcekan:
👍  Will need some documentation update as well.

clayreimann:
👍

PR #256, opened 2016-06-07
Revert total deprecating of filterCollection

xiaoyao1991:
Reverting https://github.com/yahoo/elide/pull/250/commits/d2fb23ee4f4db3d2b3aa78723e9d0a353d9fa7a5

We don't want to totally deprecate `filterCollection` in elide-2.0 per discussion with @aklish


yahoocla:
CLA is valid!

aklish:
👍

PR #257, opened 2016-06-07
Load fields then method bindings

wcekan:
Remove duplicate `accessibleObject` instead using `fieldsToValues`.


yahoocla:
CLA is valid!

DeathByTape:
👍

PR #262, opened 2016-06-15
Generalize filter strategies

aklish:
This is the first of several restructurings for more general filtering


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -207,4 +221,17 @@ default void preCommit() {
 
         return val;
     }
+
+    default <T> Object getRelationWithSortingAndPagination(
```
DeathByTape: I think it would be nice not to require a separate method for handling sorting and pagination. I think this entered the code as more of a band-aid. If we're going to go through a major refactoring of this in the first place, then we should consider consolidating this functionality.

One effective way may be to always give sane defaults in the event sorting and/or pagination is not specified.

aklish: Sounds good - will fix

elide-core/src/main/java/com/yahoo/elide/core/filter/strategy/DefaultFilterStrategy.java
```diff
@@ -0,0 +1,195 @@
...
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core.filter.strategy;
+
+import com.yahoo.elide.core.EntityDictionary;
+import com.yahoo.elide.core.filter.Operator;
+import com.yahoo.elide.core.filter.Predicate;
+import com.yahoo.elide.core.filter.expression.AndExpression;
+import com.yahoo.elide.core.filter.expression.Expression;
```
DeathByTape: `Expression` feels to generic. As I read through the code, I keep finding myself having to recall what namespace this `Expression` is under. Should we rename this to something like `FilterExpression`? Similarly, for Elide 3.0 we should consider this update for Permission `Expression`'s as well.

elide-core/src/main/java/com/yahoo/elide/core/filter/strategy/JoinFilterStrategy.java
```diff
@@ -0,0 +1,31 @@
...
+ * both predicates.
+ *
+ * This filter strategy is invoked whenever the first entity is loaded from the DataStoreTransaction.
+ * For example, the above filter on '/books' would be invoked when 'book' is loaded from the DataStoreTransaction.
+ */
+public interface JoinFilterStrategy {
+    /**
+     * @param path the URL path
+     * @param queryParams the subset of query parameters that start with 'filter'
+     * @return
```
DeathByTape: Missing description of what should be returned. Since this is an interface, we should make sure this is well articulated.

elide-core/src/main/java/com/yahoo/elide/core/filter/strategy/SubqueryFilterStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+ * This strategy is invoked whenever a collection of elements is returned via Elide.  This includes:
+ *  - GET on a collection whose type is referenced in the filter
+ *  - GET on a relationship whose type is referenced in the filter
+ *  - A compound document includes a relationship whose type is referenced in the filter
+ */
+public interface SubqueryFilterStrategy {
+    /**
+     * @param path The URL path
+     * @param queryParams The subset of queryParams that start with 'filter'
+     * @return
```
DeathByTape: Same comment as above. This should be well-defined for interface.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -346,6 +365,80 @@ public T getObject() {
...
     @Override
+    public <T> Object getRelation(
+            Object entity,
+            RelationshipType relationshipType,
+            String relationName,
+            Class<T> relationClass,
+            EntityDictionary dictionary,
+            Optional<Expression> filterExpression
+    ) {
+        Object val = com.yahoo.elide.core.PersistentResource.getValue(entity, relationName, dictionary);
```
DeathByTape: Doesn't this method get called from `PersistentResource`? Would `getValue` be more appropriately placed in the `EntityDictionary` (assuming it's not "checked")?

Realistically, `getValue` is just reflection code that extracts the value from a POJO. We should probably isolate the notion of looking up values in beans and the security around being able to access that value anyway.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -346,6 +365,80 @@ public T getObject() {
...
+            Object entity,
+            RelationshipType relationshipType,
+            String relationName,
+            Class<T> relationClass,
+            EntityDictionary dictionary,
+            Optional<Expression> filterExpression,
+            Sorting sorting,
+            Pagination pagination
+    ) {
+        Object val = com.yahoo.elide.core.PersistentResource.getValue(entity, relationName, dictionary);
```
DeathByTape: Same comment as above.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
aklish: I agree but I think this is out of scope for this PR.  I don't want to boil the ocean (PR is already pretty massive).

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
aklish: Agree but maybe in a different PR.

DeathByTape:
Since those other changes are going to be postponed to a later request, LGTM.

👍

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -94,13 +114,36 @@ public RequestScope(JsonApiDocument jsonApiDocument,
...
+            /* Extract any query param that starts with 'filter' */
+            MultivaluedMap filterParams = getFilterParams(queryParams);
+
+            if (! filterParams.isEmpty()) {
+
+                /* First check to see if there is a global, cross-type filter */
+                try {
+                    globalFilterExpression = filterStrategy.parseGlobalExpression(path, filterParams);
+                } catch (ParseException e) {
+                    //TODO - perhaps aggregate this error message with any subsequent filter parsing error */
```
DeathByTape: We should probably honor this TODO right here. It could make debugging difficult if we only ever log the failure for a single strategy.

DeathByTape:
👍  LGTM but it's pretty large. 

Would be worth getting a second set of reviewer eyes on this from either @clayreimann or @wcekan

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -119,13 +138,19 @@ protected Elide(AuditLogger auditLogger,
                   DataStore dataStore,
                   EntityDictionary dictionary,
                   JsonApiMapper mapper,
-                  Function<RequestScope, PermissionExecutor> permissionExecutor) {
+                  Function<RequestScope, PermissionExecutor> permissionExecutor,
+                  List<JoinFilterStrategy> joinFilterStrategies,
+                  List<SubqueryFilterStrategy> subqueryFilterStrategies,
+                  boolean useFilterExpressions) {
```
clayreimann: This is a breaking change because sub-classes could be depending on this constructor.

elide-core/src/main/java/com/yahoo/elide/core/filter/strategy/MultipleFilterStrategy.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+@AllArgsConstructor
+@Slf4j
+public class MultipleFilterStrategy implements JoinFilterStrategy, SubqueryFilterStrategy {
+    private List<JoinFilterStrategy> joinStrategies;
+    private List<SubqueryFilterStrategy> subqueryStrategies;
+
+    public MultipleFilterStrategy(EntityDictionary dictionary) {
+        DefaultFilterStrategy defaultStrategy = new DefaultFilterStrategy(dictionary);
+        joinStrategies = new ArrayList<>();
```
clayreimann: Is there a reason not to inline these?

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/filter/HQLFilterOperation.java
```diff
@@ -60,4 +65,47 @@ public String applyAll(Set<Predicate> predicates) {
...
+    public String apply(FilterExpression filterExpression) {
+        HQLQueryVisitor visitor = new HQLQueryVisitor();
+        return "WHERE " + filterExpression.accept(visitor);
+
+    }
+
+    /**
+     * Filter expression visitor which builds an HQL query.
+     */
+    public class HQLQueryVisitor implements Visitor<String> {
```
clayreimann: `public static class`?

aklish: THis visitor calls apply on the HQLFilterOperation - so it cannot be static as it stands.

elide-core/src/main/java/com/yahoo/elide/Elide.java
aklish: Fixed

elide-core/src/main/java/com/yahoo/elide/core/filter/strategy/MultipleFilterStrategy.java
aklish: I usually only inline if I have not defined a constructor.

wcekan:
@aklish Looks like a good direction to take.  Will we add some examples and more IT tests?

aklish:
Existing it tests cover the code.  When I add FIQL, I'll add more.

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/filter/HQLFilterOperation.java
clayreimann: Ahh, missed that

DeathByTape:
👍

wcekan:
👍  Ok

Issue #255, opened 2016-06-07
Support for different filtering strategies


aklish:
# Multiple Filter Support

Elide should have the ability to support multiple filtering strategies - including those that allow complete boolean expressions supporting
_and_, _or_, _not_ operators, parenthesis, and various comparison operators.

## FIQL

One standard in this space is [FIQL](https://tools.ietf.org/html/draft-nottingham-atompub-fiql-00).

FIQL is lacking operators for:
* in/any
* all
* is null
* is not null
* not in/none

I propose we extend FIQL to include these operators:
* in would be denotedy by `=in=`
* any would be denotedy by `=any=`
* all would be denotedy by `=all=`
* none would be denotedy by `=none=`
* notin would be denotedy by `=!in=`
* is null would be denotedy by `=null=`
* not null would be denotedy by `=!null=`

## Supporting Multiple Strategies

Each strategy should have a formal grammar that compiles to an expression of filter predicates.
The expression is an abstract syntax tree that supports the visitor pattern.  The syntax tree will be
common for all Elide filtering strategies.  Each grammar will have its own syntax tree that must
be transformed into the common syntax tree that will be passed to `DataStoreTransaction` methods.

`Elide` will extend its builder pattern to select one or more filtering strategies that are supported:

 - `withFilqFilters` to add support for FILQ filtering strategy
 - `withJsonApiFilters` to add support for JSON-API recommended strategy

If more than one strategy is selected, the strategies will be applied in order to parse any query parameter
starting with 'filter'.

If parsing fails with one strategy, the next strategy in the list of selected strategies will be attempted until
all strategies have been exhausted.

## Not Breaking Elide 2.0

Methods which currently take a `Set<Predicate>` will still be supported:
```
    @Deprecated
    default <T> Collection filterCollection(Collection collection, Class<T> entityClass, Set<Predicate> predicates) {
        ...
    }

    @Deprecated
    default <T> Collection filterCollectionWithSortingAndPagination(Collection collection, Class<T> entityClass,
                                                          EntityDictionary dictionary, Optional<Set<Predicate>> filters,
                                                          Optional<Sorting> sorting, Optional<Pagination> pagination) {
        ...
    }

    default <T> Object getRelation(
            Object entity,
            RelationshipType relationshipType,
            String relationName,
            Class<T> relationClass,
            EntityDictionary dictionary,
            Set<Predicate> filters
    ) {
        ...
    }

    default <T> Object getRelationWithSortingAndPagination(
            Object entity,
[...]

PR #265, opened 2016-06-19
Add rsql filter dialect

aklish:
This PR adds RSQL and FIQL filter support to Elide.

The 2nd commit renames FilterStrategy to FilterDialect

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/MultipleFilterDialect.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/MultipleFilterDialect.java
@@ -3,7 +3,7 @@
  * Licensed under the Apache License, Version 2.0
  * See LICENSE file in project root for terms.
  */
-package com.yahoo.elide.core.filter.strategy;
+package com.yahoo.elide.core.filter.dialect;
 
 import com.yahoo.elide.core.EntityDictionary;
 import com.yahoo.elide.core.filter.expression.FilterExpression;
@@ -16,35 +16,35 @@ import java.util.List;
 import java.util.Map;
 
 /**
- * A filter strategy that supports an ordered list of different strategies.  Strategies
- * are attempted in sequence.  The first strategy that successfully parses a filter expression
- * is used.  If no strategy succeeds, the error from the last strategy is returned.
+ * A filter dialect that supports an ordered list of different dialects.  Dialects
+ * are attempted in sequence.  The first dialect that successfully parses a filter expression
+ * is used.  If no dialect succeeds, the error from the last dialect is returned.
  */
 @AllArgsConstructor
 @Slf4j
-public class MultipleFilterStrategy implements JoinFilterStrategy, SubqueryFilterStrategy {
-    private List<JoinFilterStrategy> joinStrategies;
-    private List<SubqueryFilterStrategy> subqueryStrategies;
+public class MultipleFilterDialect implements JoinFilterDialect, SubqueryFilterDialect {
+    private List<JoinFilterDialect> joinDialects;
+    private List<SubqueryFilterDialect> subqueryDialects;
 
-    public MultipleFilterStrategy(EntityDictionary dictionary) {
-        DefaultFilterStrategy defaultStrategy = new DefaultFilterStrategy(dictionary);
-        joinStrategies = new ArrayList<>();
-        joinStrategies.add(defaultStrategy);
-        subqueryStrategies = new ArrayList<>();
-        subqueryStrategies.add(defaultStrategy);
+    public MultipleFilterDialect(EntityDictionary dictionary) {
+        DefaultFilterDialect defaultDialect = new DefaultFilterDialect(dictionary);
+        joinDialects = new ArrayList<>();
+        joinDialects.add(defaultDialect);
+        subqueryDialects = new ArrayList<>();
+        subqueryDialects.add(defaultDialect);
     }
 
     @Override
     public FilterExpression parseGlobalExpression(String path,
                                             MultivaluedMap<String, String> queryParams) throws ParseException {
-        if (joinStrategies.isEmpty()) {
+        if (joinDialects.isEmpty()) {
             throw new ParseException("Heterogeneous type filtering not supported");
         }
 
         ParseException lastFailure = null;
-        for (JoinFilterStrategy strategy : joinStrategies) {
+        for (JoinFilterDialect dialect : joinDialects) {
             try {
-                return strategy.parseGlobalExpression(path, queryParams);
+                return dialect.parseGlobalExpression(path, queryParams);
             } catch (ParseException e) {
                 log.trace("Parse Failure: {}", e.getMessage());
                 if (lastFailure != null) {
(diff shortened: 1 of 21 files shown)


yahoocla:
CLA is valid!

clayreimann:
FYI to reviewers, the first 4 commits are actually #262 so this will be much easier to review when that goes in.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -0,0 +1,315 @@
...
+    public RSQLFilterDialect(EntityDictionary dictionary) {
+        this(dictionary, RSQLOperators.defaultOperators());
+    }
+
+
+    @Override
+    public FilterExpression parseGlobalExpression(String path, MultivaluedMap<String, String> queryParams)
+            throws ParseException {
+        if (queryParams.size() != 1) {
+            throw new ParseException("There can only be a single filter query parameter");
```
clayreimann: is `queryParams` just the subset of params that start with `filter` or do we mean that you can't use RSql filters with sparse fieldsets, includes, and other features?

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -0,0 +1,315 @@
...
+            path = path.substring(1);
+        }
+
+        String[] pathComponents = path.split("/");
+        String lastPathComponent = "";
+        if (pathComponents.length > 0) {
+            lastPathComponent = pathComponents[pathComponents.length - 1];
+        }
+
+        Class entityType = dictionary.getEntityClass(lastPathComponent);
```
clayreimann: does this really map relationship names to entity classes? I thought we just mapped json-api types to entity classes.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
aklish: It is just the subset of params that start with filter (this is true for all filter dialects).

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
aklish: It works by virtue that global filters only work for root collections.  This could be fixed with a custom Visitor on the JSON-API path - that returns the 'type' of the last URL path segment.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -0,0 +1,315 @@
...
+            Matcher matcher = typedFilterPattern.matcher(paramName);
+            if (matcher.find()) {
+                String typeName = matcher.group(1);
+                if (paramValues.size() != 1) {
+                    throw new ParseException("Exactly one RSQL expression must be defined for type : " + typeName);
+                }
+
+                String expressionText = paramValues.get(0);
+
+                Class entityType = dictionary.getEntityClass(typeName);
```
clayreimann: How would this work for a filter like: `filter[book.author.id]`. I'm guessing that in `dictionary.getEntityClass('book.author.id')` is going to break stuff

clayreimann:
👍  here

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
xiaoyao1991: we should then rename to `filterParams` maybe?

DeathByTape: I 👍 @xiaoyao1991's recommendation. That would make this clearer overall to the reader.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -0,0 +1,315 @@
...
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * FilterDialect which implements support for RSQL filter dialect.
+ */
+public class RSQLFilterDialect implements SubqueryFilterDialect, JoinFilterDialect {
+
+    private final Pattern typedFilterPattern;
```
DeathByTape: This is the same across all instances right? We can probably safely make this `static`.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -0,0 +1,315 @@
...
+    private final RSQLParser parser;
+    private final EntityDictionary dictionary;
+    private final Map<ComparisonOperator, Operator> operatorMap;
```
DeathByTape: It looks like this is the same for all instances, can we make this `static`? Also, can we make this an `ImmutableMap`?

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -0,0 +1,315 @@
...
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+/**
+ * FilterDialect which implements support for RSQL filter dialect.
+ */
+public class RSQLFilterDialect implements SubqueryFilterDialect, JoinFilterDialect {
+
+    private final Pattern typedFilterPattern;
+    private final RSQLParser parser;
```
DeathByTape: I only see this used once, does it need to be a member variable?

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -0,0 +1,315 @@
...
+                throw new ParseException("Invalid query parameter: " + paramName);
+            }
+        }
+        return expressionByType;
+    }
+
+    /**
+     * Visitor which converts RSQL abstract syntax tree into an Elide filter expression.
+     */
+    public class RSQL2ExpressionFilterVisitor implements RSQLVisitor<FilterExpression, Class> {
```
DeathByTape: Can we move this out to its own file? The constructor would also have to take a dictionary (and operatorMap if we don't make it static), but it may be easier to digest.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/MultipleFilterDialect.java
```diff
@@ -0,0 +1,84 @@
...
+                } else {
+                    lastFailure = e;
+                }
+            }
+        }
+        throw lastFailure;
+    }
+
+    @Override
+    public Map<String, FilterExpression> parseTypedExpression(String path,
```
DeathByTape: This method and `parseGlobalExpression` seem some close... I wonder if we can reduce the code duplication like this:

```java
...
    @Override
    public FilterExpression parseGlobalExpression(String path,
                                            MultivaluedMap<String, String> queryParams) throws ParseException {
        if (joinDialects.isEmpty()) {
            throw new ParseException("Heterogeneous type filtering not supported");
        }

        return parseExpression(joinDialects, (dialect) -> dialect.parseGlobalExpression(path, queryParams));
    }

    @Override
    public Map<String, FilterExpression> parseTypedExpression(String path,
                                                        MultivaluedMap<String, String> queryParams)
            throws ParseException {

        if (subqueryDialects.isEmpty()) {
            throw new ParseException("Type filtering not supported");
        }

        return parseExpression(subqueryDialects, (dialect) -> dialect.parseTypedExpression(path, queryParams));
    }

    private static <T, R> R parseExpression(List<T> dialects, ParseFunction<T, R> parseFunction) throws ParseException {
        ParseException lastFailure = null;
        for (T dialect : dialects) {
            try {
                return parseFunction.apply(dialect);
            } catch (ParseException e) {
                log.trace("Parse Failure: {}", e.getMessage());
                if (lastFailure != null) {
                    lastFailure = new ParseException(e.getMessage() + "\n" + lastFailure.getMessage());
                } else {
                    lastFailure = e;
                }
            }
        }
        throw lastFailure;
    }

    @FunctionalInterface
    public interface ParseFunction<T, R> {
        R apply(T dialect) throws ParseException;
    }
...
```

DeathByTape:
👍

aklish:
Fixes #255

Issue #264, opened 2016-06-16
Pagination needs total page and record count


johnoscott:
Most pagination user experiences need to know the total number of pages and records e.g.

`Page 6 of 12 | 1999 Records`

JSON API suggests adding this as `meta` data so for a request like:

`api/product?filter[person.surname][infix]=smith&page[number]=6&page[limit]=20`

something like this :

``` JSON
{
"meta": {
   "page": {
     "number": 6,
     "limit": 20,
     "totalPages": 12,
     "totalRecords":1999
   },
data : { ...
```

Can Elide return something like this ?

DeathByTape:
This is a good point and sounds pretty reasonable to me.

@aklish @wcekan  @clayreimann -- do we have a way to do this now (if we do, it's escaping me right now)? If we have no such mechanism already existing, are you guys opposed to this proposal?

clayreimann:
@DeathByTape I'm not aware of any way we have to do this currently. Sounds like an extension to the `DatastoreTransaction` interface to me. I think it's a good idea though it might add some overhead. It doesn't look like hibernate's results include this information, so we'll probably have to run a separate query to compute this information–unless @wcekan knows about something I didn't see.

DeathByTape:
@clayreimann you're probably right. Perhaps it would be useful to guard this with an additional query param to determine whether or not we want to send the count query to return the number of available records.

I suspect a simple `SELECT count(*) FROM ... WHERE ...` _shouldn't_ take an exorbitant amount of time to execute on most tables, but we could minimize the general latency if we have users explicitly enable this behavior.

DeathByTape:
@aklish we should probably make an annotation rather than updating the builder, right? It seems to me like you would want to control this on a per-object basis.

johnoscott:
agree this should be optional since it will likely require another query to get the count. maybe

mjvanleeuwen:
I have made an attempt at this at https://github.com/ActioPtyLtd/elide/commit/1c9f7b6c051cbddd33a28b72a63e685ae32d43d5
and would appreciate any comments. 

At this stage the logic to query for the totals is only implemented for Hibernate5. The code could be made more efficient by using a single Criteria object to first fetch the total first, then apply the pagination limit and offset to load all entities afterwards. However, that would have involved a much bigger change.

A couple of things I am not sure about:
- Ensuring that a commit is triggered when the total is fetched. As the total is being fetched in the same session as the data, it seemed safe to me assume that a commit will be executed.
[...]

PR #266, opened 2016-06-20
Add support for page totals and JSON-API page meta data

mjvanleeuwen:
For enhancement proposal https://github.com/yahoo/elide/issues/264.

The page meta data will be returned in the JSON-API response if the page size/limit/number/offset query parameters have been specified.

If the query parameter "page[totals]=true" is also specified, the page totals will be determined and included in the returned page meta data.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
@@ -9,6 +9,7 @@ import com.yahoo.elide.core.exceptions.InvalidValueException;
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.ToString;
+import org.apache.commons.lang3.mutable.MutableBoolean;
 
 import javax.ws.rs.core.MultivaluedMap;
 import java.util.HashMap;
@@ -30,7 +31,7 @@ public class Pagination {
     public static final int DEFAULT_PAGE_SIZE = 500;
     public static final int MAX_PAGE_SIZE = 10000;
 
-    private static final Pagination DEFAULT_PAGINATION = new Pagination(0, 0);
+    private static final Pagination DEFAULT_PAGINATION = new Pagination(0, 0, false);
 
     public static final Map<String, PaginationKey> PAGE_KEYS = new HashMap<>();
     static {
@@ -40,11 +41,17 @@ public class Pagination {
         PAGE_KEYS.put("page[offset]", PaginationKey.offset);
     }
 
+    // For requesting total pages/records be included in the response
+    public static String PAGE_TOTAL_KEY = "page[totals]";
+
     @Getter
     private int offset;
 
     @Getter
     private int limit;
+
+    @Getter
+    private boolean pageTotalsRequested = false;
 
     /**
      * Know if this is the default instance.
@@ -70,6 +77,8 @@ public class Pagination {
     public static Pagination parseQueryParams(final MultivaluedMap<String, String> queryParams)
             throws InvalidValueException {
         final Map<PaginationKey, Integer> pageData = new HashMap<>();
+        final MutableBoolean pageTotalsRequested = new MutableBoolean(false);
+
         queryParams.entrySet()
                 .forEach(paramEntry -> {
                     final String queryParamKey = paramEntry.getKey();
@@ -80,9 +89,18 @@ public class Pagination {
                         } catch (ClassCastException e) {
                             throw new InvalidValueException("page values must be integers");
                         }
+                    } else if (queryParamKey.equals(PAGE_TOTAL_KEY)) {
+                        final String value = paramEntry.getValue().get(0);
+                        if (value.equals("true")) {
+                            pageTotalsRequested.setTrue();
+                        } else if (value.equals("false")) {
+                            pageTotalsRequested.setFalse();
+                        } else {
+                            throw new InvalidValueException(PAGE_TOTAL_KEY + " must be either true or false");
+                        }
                     } else if (queryParamKey.startsWith("page[")) {
                         throw new InvalidValueException("Invalid Pagination Parameter. Accepted values are page[number]"
-                                + ",page[size],page[offset],page[limit]");
+                                + ",page[size],page[offset],page[limit],page[totals]");
                     }
                 });
 
@@ -103,8 +121,8 @@ public class Pagination {
         if (!pageData.containsKey(PaginationKey.offset)) {
             pageData.put(PaginationKey.offset, 0);
         }
-        // offset, page, limit
-        return new Pagination(pageData.get(PaginationKey.offset), limit);
+        // offset, page, limit, pageTotalsRequested
+        return new Pagination(pageData.get(PaginationKey.offset), limit, pageTotalsRequested.getValue());
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
@@ -14,6 +14,7 @@ import org.apache.commons.lang3.mutable.MutableBoolean;
 import javax.ws.rs.core.MultivaluedMap;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 /**
  * Encapsulates the pagination strategy.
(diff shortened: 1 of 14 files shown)


yahoocla:

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -80,9 +89,18 @@ public static Pagination parseQueryParams(final MultivaluedMap<String, String> q
                         } catch (ClassCastException e) {
                             throw new InvalidValueException("page values must be integers");
                         }
+                    } else if (queryParamKey.equals(PAGE_TOTAL_KEY)) {
+                        final String value = paramEntry.getValue().get(0);
+                        if (value.equals("true")) {
```
DeathByTape: Is there a reason we care about the values here? I think it would probably be ok to check for the existence of this parameter (regardless of its value) to enable this feature-- by default it should probably stay disabled anyway.

DeathByTape:
This looks pretty good to me overall; just a couple of additional comments.

First, we should probably have some tests here. Whether it's unit tested or integration tested doesn't matter as much as long as we have looked at some edge cases and so forth. This is less about verifying it right now, but more about not breaking it in the future with additional changes (and having an automated way to test it). Integration tests are usually done through `ResourceIT.java` which is a bit of a mess right now and needs to be cleaned up, but for features that jump across many objects it typically makes more sense.

Next, we need 2 access control points here; one of which you have already implemented.

  1. **Caller access**. This is the access you have implemented here. The caller should be able to selectively request this data from the server only when it needs it.
  1. **Server access**. The server should also be able to control on a per-object basis whether or not it chooses to expose this information to the caller. In some cases, you could imagine that we would _never_ want to leak this cardinality info for some reason. I suggest we implement this through an additional annotation (say `@CanSelectCount` or something?).

This does bring another question @aklish, @wcekan, @clayreimann: if we're going to have a server-side annotation scheme, should we do it for the whole pagination feature? We could start the implementation here (something like `@Paginate(count = true)`) and then expand for other pagination-like things later in additional PR's (to selectively enable/disable pagination for an object, for instance). I am not sure how useful this would actually be in practice, but an `@Paginate` annotation may provide us with a couple of useful things. The most useful feature I could think of for an `@Paginate` (other than enabling/disabling `count`) is for providing **default** limits and other access controls over how pagination is exposed. For example, `@Paginate(defaultLimit = 100, maxLimit = 200)` could be an annotation describing that if no pagination is specified by the caller, the server will assume that is equivalent to paginating to `100` rows and the `maxLimit` would denote that the greatest number of rows for this object we're willing to return at any one time is `200`.

Thoughts?

Thanks for working on this @mjvanleeuwen!

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -248,6 +249,20 @@ public void commit() {
...
+    @Override
+    public <T> Long getTotalRecords(Class<T> entityClass, FilterScope filterScope) {
+        final Criterion criterion = filterScope.getCriterion(NOT, AND, OR);
+        final CriteriaExplorer criteriaExplorer = new CriteriaExplorer(
+                entityClass, filterScope.getRequestScope(), criterion);
+        final Criteria sessionCriteria = session.createCriteria(entityClass);
+
+        criteriaExplorer.buildCriteria(sessionCriteria, session);
+
+        sessionCriteria.setProjection(Projections.rowCount());
```
DeathByTape: I think this is equivalent to a `select count(*) ...`, right? So if there was a filter applied, you could select the right set of objects, however, if (by permission) any of those records were filtered out, we would still count them here. I believe the correct value would be something like:

```
this_record_count - total_num_records_filtered
```

This way you're not getting a count for the records which are validly returned by the SQL query, but disallowed by your permissions in Elide.

clayreimann:
I think we _do_ need the `@Paginate` annotation. For the purposes of this PR it should probably just be `@Paginate(countable = true)` but I agree that we should expand it to include at least `@Paginate(defaultLimit = 100, maxLimit = 200)`

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
mjvanleeuwen: It would seem a bit odd if "page[totals]" value of "false" or "no" was passed and the page totals were still generated.

Also, I am thinking that the client will specify "true" only for the first page request, then with subsequent page requests the totals will normally not need to be recalculated. There is some convenience with being able to just change the value to "false" to stop the totals recalculation.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
clayreimann: This is a thorny problem, because `total_num_records_filtered` depends on evaluating permissions on EVERY record in SQL. This sounds like something that should be done OOB after the first query. 

The first query would say that this is page 1 of ?? pages and we would kick off a background job that fetches all of the objects from the database. After we evaluate permissions on all of the returned objects and stores the count into a map keyed on user and object requested so that on subsequent requests we can say this is page 2 of 8.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -19,30 +19,33 @@
...
+    // For requesting total pages/records be included in the response
+    public static String PAGE_TOTALS_KEY = "page[totals]";
 
     public static final Map<String, PaginationKey> PAGE_KEYS = new HashMap<>();
     static {
         PAGE_KEYS.put("page[size]", PaginationKey.limit);
         PAGE_KEYS.put("page[limit]", PaginationKey.limit);
         PAGE_KEYS.put("page[number]", PaginationKey.page);
         PAGE_KEYS.put("page[offset]", PaginationKey.offset);
+        PAGE_KEYS.put(PAGE_TOTALS_KEY, PaginationKey.totals);
```
clayreimann: While you're here (and since it's a good idea) would you mind replacing the other pagination keys with constants?

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -83,46 +91,57 @@ public static Pagination parseQueryParams(final MultivaluedMap<String, String> q
...
+    public Pagination evaluate(final Class entityClass) {
+        Paginate paginate = entityClass != null ? (Paginate) entityClass.getAnnotation(Paginate.class) : null;
 
-        boolean isPageBased = pageData.containsKey(PaginationKey.page);
-        int limit = !pageData.containsKey(PaginationKey.limit) ? DEFAULT_PAGE_SIZE : pageData.get(PaginationKey.limit);
-        if (limit < 0) {
-            throw new InvalidValueException("page[size] and page[limit] must contain positive values.");
-        }
+        int defaultLimit = paginate != null ? paginate.defaultLimit() : this.DEFAULT_PAGE_SIZE;
+        int maxLimit = paginate != null ? paginate.maxLimit() : this.MAX_PAGE_SIZE;
```
clayreimann: I'd drop `this` and use the naked variable name, or prefix with the class name if you prefer, since this is a class static.

DeathByTape:
These changes are looking good to me! :) Would you mind squashing at least the `merge upstream changes` commit?

DeathByTape:
The squash is almost perfect, but you merged back into this branch. I would recommend replacing this branch with just those three commits. One easy way to do this is:

```
$ git checkout YOUR_SQUASHED_BRANCH_WITH_ONLY_THOSE_THREE_COMMITS
$ git branch -D add_page_totals_and_meta_page # This will delete your local branch
$ git checkout -b add_page_totals_and_meta_page # This will create a new branch from the one you're currently on and call it 'add_page_totals_and_meta_page'
$ git push <YOUR_REMOTE> add_page_totals_and_meta_page --force
```

This should replace this branch and this PR will only have those 3 commits! :)

Once that is done, I don't see any blocks. @clayreimann does this looks good to you too?

DeathByTape:
This looks good, it just needs to be rebased to resolve conflicts for master. Then I think we can merge this in! 👍

DeathByTape:
:+1: thanks for working on this!

PR #268, opened 2016-06-21
Fixed the NPE problem when POST resource with not 'include' relations…

tarrantzhang:
…hip.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidRelationException.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidRelationException.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright 2016, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core.exceptions;
+
+import com.yahoo.elide.core.HttpStatus;
+
+/**
+ * Invalid Relationship exception.
+ */
+public class InvalidRelationException extends HttpStatusException {
+    public InvalidRelationException(String objectOrFieldName) {
+        super(HttpStatus.SC_NOT_FOUND, "Unknown resource type in relationship '" + objectOrFieldName + "'.");
+    }
+}
--- a/elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidRelationException.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidRelationException.java
@@ -11,7 +11,7 @@ import com.yahoo.elide.core.HttpStatus;
  * Invalid Relationship exception.
  */
 public class InvalidRelationException extends HttpStatusException {
-    public InvalidRelationException(String objectOrFieldName) {
-        super(HttpStatus.SC_NOT_FOUND, "Unknown resource type in relationship '" + objectOrFieldName + "'.");
+    public InvalidRelationException(String entityType) {
+        super(HttpStatus.SC_BAD_REQUEST, "Unknown entity in relationship '" + entityType + "'");
     }
 }
--- a/elide-core/src/main/java/com/yahoo/elide/core/exceptions/UnknownEntityException.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/exceptions/UnknownEntityException.java
@@ -12,6 +12,6 @@ import com.yahoo.elide.core.HttpStatus;
  */
 public class UnknownEntityException extends HttpStatusException {
     public UnknownEntityException(String entityType) {
-        super(HttpStatus.SC_BAD_REQUEST, "Unknown entity in relationship '" + entityType + "'");
+        super(HttpStatus.SC_BAD_REQUEST, "Unknown entity type: '" + entityType + "'");
     }
 }
--- a/elide-core/src/main/java/com/yahoo/elide/core/exceptions/UnknownEntityException.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/exceptions/UnknownEntityException.java
@@ -8,10 +8,10 @@ package com.yahoo.elide.core.exceptions;
 import com.yahoo.elide.core.HttpStatus;
 
 /**
- * Invalid Relationship exception.
+ * Unknown entity exception.
  */
-public class InvalidRelationException extends HttpStatusException {
-    public InvalidRelationException(String entityType) {
+public class UnknownEntityException extends HttpStatusException {
+    public UnknownEntityException(String entityType) {
         super(HttpStatus.SC_BAD_REQUEST, "Unknown entity in relationship '" + entityType + "'");
     }
 }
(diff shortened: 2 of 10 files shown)


yahoocla:

yahoocla:
CLA is valid!

elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
```diff
@@ -1447,6 +1447,21 @@ public void testInverseDeleteFromCollection() {
...
+    public void testPostInvalidRelationship() {
+        // Note: This tests the correct response when POSTing a resource with a not "include" relationship. The server
+        // should returns InvalidRelationException rather than NPE.
+        String createRoot = jsonParser.getJson("/ResourceIT/testPostWithInvalidRelationship.json");
+
+        given()
+                .contentType(JSONAPI_CONTENT_TYPE)
+                .accept(JSONAPI_CONTENT_TYPE)
+                .body(createRoot)
+                .post("/ResoureWithInvalidRelationship")
```
wcekan: Did you mean `resourceWithInvalidRelationship `

elide-core/src/main/java/com/yahoo/elide/jsonapi/models/Relationship.java
```diff
@@ -80,6 +81,8 @@ public Relationship(@JsonProperty("links") Map<String, String> links,
                     res.add(resource.toPersistentResource(requestScope));
                 } catch (ForbiddenAccessException e) {
                     //skip resource
+                } catch (java.lang.NullPointerException e) {
+                    throw new InvalidRelationException(resource.getType());
```
clayreimann: What causes this NPE? Wouldn't it be better to throw this exceptions instead of an NPE?

clayreimann: I'm guessing that the exception is ultimately being thrown [from here](https://github.com/yahoo/elide/blob/65929a6b13e1798afd7ba06b12775e660803881c/elide-core/src/main/java/com/yahoo/elide/jsonapi/models/Resource.java#L140) and so we should throw this `InvalidRelationshipException` there.

elide-core/src/main/java/com/yahoo/elide/jsonapi/models/Relationship.java
```diff
@@ -80,6 +81,8 @@ public Relationship(@JsonProperty("links") Map<String, String> links,
                     res.add(resource.toPersistentResource(requestScope));
                 } catch (ForbiddenAccessException e) {
                     //skip resource
+                } catch (java.lang.NullPointerException e) {
```
xiaoyao1991: Why throwing with the full pkg path? Is there another NullPointerException class used in this class?

xiaoyao1991:
I'm not convinced that catching the NPE is a good idea. NPE is very general, and it could be caused by many reasons at many places. Instead of catching it, could we do a sanity check against all the Elide entities -- to see if their relationships are also known by Elide, and fail early.

tarrantzhang:
I follow @xiaoyao1991 suggestion to locate where the elide first "scans" models, and that would be the best place to exclude elide from "scanning" resources without "@include" annotation. But I'm not able to trace the place where it happens. The best guess I have now is [here](https://github.com/yahoo/elide/blob/master/elide-example/elide-blog-example/src/main/java/com/yahoo/elide/example/persistence/ElideResourceConfig.java#L32), I need confirmation though.

clayreimann:
@tarrantzhang You should look at [EntityBinding](https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java#L204-L235) for how we bind things. 

The problem with trying to validate that relationships are `@Include`d at boot time is that (I don't think) we can be sure what classes need to have `@Include` and which Hibernate (or HBase, or some other datastore) can handle natively.

I think the better approach is to try and figure out where that NPE is coming from and throw your `InvalidRelationshipType` exception there.

elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidRelationException.java
```diff
@@ -0,0 +1,17 @@
...
+package com.yahoo.elide.core.exceptions;
+
+import com.yahoo.elide.core.HttpStatus;
+
+/**
+ * Invalid Relationship exception.
+ */
+public class InvalidRelationException extends HttpStatusException {
+    public InvalidRelationException(String objectOrFieldName) {
+        super(HttpStatus.SC_NOT_FOUND, "Unknown resource type in relationship '" + objectOrFieldName + "'.");
```
clayreimann: This should just read `"Unknown resource type '" + type  + "'."`

elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidRelationException.java
```diff
@@ -0,0 +1,17 @@
...
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core.exceptions;
+
+import com.yahoo.elide.core.HttpStatus;
+
+/**
+ * Invalid Relationship exception.
+ */
+public class InvalidRelationException extends HttpStatusException {
```
xiaoyao1991: Shall we rename this class to something like `UnknownEntityException` ?

clayreimann: 👍

elide-integration-tests/src/test/java/example/NotIncludedResource.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2015, Yahoo Inc.
```
xiaoyao1991: 2016

elide-integration-tests/src/test/java/example/ResourceWithInvalidRelationship.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2015, Yahoo Inc.
```
xiaoyao1991: 2016

elide-core/src/main/java/com/yahoo/elide/core/exceptions/UnknownEntityException.java
```diff
@@ -8,10 +8,10 @@
...
 
 /**
- * Invalid Relationship exception.
+ * Unknown entity exception.
  */
-public class InvalidRelationException extends HttpStatusException {
-    public InvalidRelationException(String entityType) {
+public class UnknownEntityException extends HttpStatusException {
+    public UnknownEntityException(String entityType) {
         super(HttpStatus.SC_BAD_REQUEST, "Unknown entity in relationship '" + entityType + "'");
```
clayreimann: Message should read `Unknown entity type: '<type>'.`

aklish:
👍

xiaoyao1991:
👍

Issue #260, opened 2016-06-10
CoerceUtil bug


xiaoyao1991:
I found a problem when you try to coerce a non-number String, like "a", to a number, like Long.class, it will convert it to 0, which is bad

xiaoyao1991:
Found the root cause. The ConvertUtils from commons-beanutils pkg will by default ignore conversion exceptions, and return a defaultValue instead. In this case, the defaultValue for any Number is 0. All we need to do is to enable throwing exception.

PR #269, opened 2016-06-23
Fix CoerceUtil Bug

xiaoyao1991:
#260

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
+++ b/elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
@@ -44,7 +44,6 @@ public class CoerceUtil {
 
         try {
             return ConvertUtils.convert(value, cls);
-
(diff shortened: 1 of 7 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
```diff
@@ -55,6 +54,11 @@ public static Object coerce(Object value, Class<?> cls) {
      */
     private static void setup() {
         BeanUtilsBean.setInstance(new BeanUtilsBean(new ConvertUtilsBean() {
+            {
+                // https://github.com/yahoo/elide/issues/260
+                // disable returning defaultValues when conversions fail
```
clayreimann: This note should be changed from a negation to an assertion like: `enable throwing when conversion fails`

DeathByTape:
👍  after addressing @clayreimann  comment

clayreimann:
👍 here

PR #270, opened 2016-06-23
In memory filters and visitors

xiaoyao1991:
Depend on #269

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/InMemoryFilterOperation.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/InMemoryFilterOperation.java
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2016, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core.filter;
+
+import com.yahoo.elide.core.EntityDictionary;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+/**
+ * InMemoryFilterOperation
+ */
+public class InMemoryFilterOperation implements FilterOperation<Set<java.util.function.Predicate>> {
+    private final EntityDictionary dictionary;
+
+    public InMemoryFilterOperation(EntityDictionary dictionary) {
+        this.dictionary = dictionary;
+    }
+
+    @Override
+    public Set<java.util.function.Predicate> apply(Predicate predicate) {
+        return Collections.singleton(predicate.getOperator().apply(predicate, dictionary));
+    }
+
+    @Override
+    public Set<java.util.function.Predicate> applyAll(Set<Predicate> predicates) {
+        return predicates.stream()
+                .map(this::apply)
+                .flatMap(Collection::stream)
+                .collect(Collectors.toSet());
+    }
+
+    public EntityDictionary getDictionary() {
+        return dictionary;
+    }
+}
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
@@ -24,29 +24,28 @@ import java.util.stream.Collectors;
 public enum Operator implements BiFunction<Predicate, EntityDictionary, java.util.function.Predicate> {
     IN("in", true) {
         @Override
-        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
             return (entity) -> {
                 Object val = PersistentResource.getValue(entity, field, dictionary);
                 List<Object> coercedValues = values.stream()
-                        .map(v -> CoerceUtil.coerce(v, val.getClass())).collect(Collectors.toList());
+                        .map(v -> CoerceUtil.coerce(v, val.getClass()))
+                        .collect(Collectors.toList());
                 return coercedValues.contains(val);
             };
         }
     },
     NOT("not", true) {
         @Override
-        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
-            return (entity) -> {
-                Object val = PersistentResource.getValue(entity, field, dictionary);
-                List<Object> coercedValues = values.stream()
-                        .map(v -> CoerceUtil.coerce(v, val.getClass())).collect(Collectors.toList());
-                return !coercedValues.contains(val);
-            };
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> !IN.getFilterFunction(field, values, dictionary).test(entity);
         }
     },
     PREFIX("prefix", true) {
         @Override
-        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
             return (entity) -> {
                 if (values.size() != 1) {
                     throw new InvalidPredicateException("PREFIX can only take one argument");
@@ -55,13 +54,14 @@ public enum Operator implements BiFunction<Predicate, EntityDictionary, java.uti
                 Object val = PersistentResource.getValue(entity, field, dictionary);
                 String valStr = (String) CoerceUtil.coerce(val, String.class);
                 String filterStr = (String) CoerceUtil.coerce(values.get(0), String.class);
-                return valStr.startsWith(filterStr);
+                return valStr != null && filterStr != null && valStr.startsWith(filterStr);
             };
         }
     },
     POSTFIX("postfix", true) {
         @Override
-        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
             return (entity) -> {
                 if (values.size() != 1) {
                     throw new InvalidPredicateException("POSTFIX can only take one argument");
@@ -70,13 +70,14 @@ public enum Operator implements BiFunction<Predicate, EntityDictionary, java.uti
                 Object val = PersistentResource.getValue(entity, field, dictionary);
                 String valStr = (String) CoerceUtil.coerce(val, String.class);
                 String filterStr = (String) CoerceUtil.coerce(values.get(0), String.class);
-                return valStr.endsWith(filterStr);
+                return valStr != null && filterStr != null && valStr.endsWith(filterStr);
             };
         }
     },
     INFIX("infix", true) {
         @Override
-        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
             return (entity) -> {
                 if (values.size() != 1) {
                     throw new InvalidPredicateException("INFIX can only take one argument");
@@ -85,13 +86,14 @@ public enum Operator implements BiFunction<Predicate, EntityDictionary, java.uti
                 Object val = PersistentResource.getValue(entity, field, dictionary);
                 String valStr = (String) CoerceUtil.coerce(val, String.class);
                 String filterStr = (String) CoerceUtil.coerce(values.get(0), String.class);
-                return valStr.contains(filterStr);
+                return valStr != null && filterStr != null && valStr.contains(filterStr);
             };
         }
     },
     ISNULL("isnull", false) {
         @Override
-        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
             return (entity) -> {
                 Object val = PersistentResource.getValue(entity, field, dictionary);
                 return val == null;
@@ -100,16 +102,15 @@ public enum Operator implements BiFunction<Predicate, EntityDictionary, java.uti
     },
     NOTNULL("notnull", false) {
         @Override
-        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
-            return (entity) -> {
-                Object val = PersistentResource.getValue(entity, field, dictionary);
-                return val != null;
-            };
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> !ISNULL.getFilterFunction(field, values, dictionary).test(entity);
         }
     },
     LT("lt", true) {
         @Override
-        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
             return (entity) -> {
                 if (values.size() != 1) {
                     throw new InvalidPredicateException("LT can only take one argument");
@@ -125,7 +126,8 @@ public enum Operator implements BiFunction<Predicate, EntityDictionary, java.uti
     },
     LE("le", true) {
         @Override
-        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
             return (entity) -> {
                 if (values.size() != 1) {
                     throw new InvalidPredicateException("LE can only take one argument");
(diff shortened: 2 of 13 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
-    PREFIX("prefix", true),
-    POSTFIX("postfix", true),
-    INFIX("infix", true),
-    ISNULL("isnull", false),
-    NOTNULL("notnull", false),
-    LT("lt", true),
-    LE("le", true),
-    GT("gt", true),
-    GE("ge", true);
+public enum Operator implements BiFunction<Predicate, EntityDictionary, java.util.function.Predicate> {
```
clayreimann: Why is one of these `Predicate`s fully qualified? I can't see an import for `*.Predicate` that would conflict.

elide-core/src/main/java/com/yahoo/elide/core/filter/InMemoryFilterOperation.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public class InMemoryFilterOperation implements FilterOperation<Set<java.util.function.Predicate>> {
+    private final EntityDictionary dictionary;
+
+    public InMemoryFilterOperation(EntityDictionary dictionary) {
+        this.dictionary = dictionary;
+    }
+
+    @Override
+    public Set<java.util.function.Predicate> apply(Predicate predicate) {
```
clayreimann: Why is this fully qualified?

elide-core/src/main/java/com/yahoo/elide/core/filter/InMemoryFilterOperation.java
```diff
@@ -0,0 +1,41 @@
...
+        return Collections.singleton(predicate.getOperator().apply(predicate, dictionary));
+    }
+
+    @Override
+    public Set<java.util.function.Predicate> applyAll(Set<Predicate> predicates) {
```
clayreimann: Same question.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
+    IN("in", true) {
+        @Override
+        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                List<Object> coercedValues = values.stream()
+                        .map(v -> CoerceUtil.coerce(v, val.getClass())).collect(Collectors.toList());
```
clayreimann: `.collect` on new line

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
+            };
+        }
+    },
+    NOT("not", true) {
+        @Override
+        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                List<Object> coercedValues = values.stream()
+                        .map(v -> CoerceUtil.coerce(v, val.getClass())).collect(Collectors.toList());
```
clayreimann: `.collect` on new line

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
+    PREFIX("prefix", true) {
+        @Override
+        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                if (values.size() != 1) {
+                    throw new InvalidPredicateException("PREFIX can only take one argument");
+                }
+
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                String valStr = (String) CoerceUtil.coerce(val, String.class);
```
clayreimann: if `val == null` here will valStr be `"null"` or `""` or `null`?

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
+    POSTFIX("postfix", true) {
+        @Override
+        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                if (values.size() != 1) {
+                    throw new InvalidPredicateException("POSTFIX can only take one argument");
+                }
+
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                String valStr = (String) CoerceUtil.coerce(val, String.class);
```
clayreimann: Same as with `prefix`

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
+    INFIX("infix", true) {
+        @Override
+        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                if (values.size() != 1) {
+                    throw new InvalidPredicateException("INFIX can only take one argument");
+                }
+
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                String valStr = (String) CoerceUtil.coerce(val, String.class);
```
clayreimann: Same as with `prefix` and `infix`

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
+        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                if (values.size() != 1) {
+                    throw new InvalidPredicateException("LT can only take one argument");
+                }
+
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                Comparable filterComp = (Comparable) CoerceUtil
+                        .coerce(CoerceUtil.coerce(values.get(0), val.getClass()), Comparable.class);
+                Comparable valComp = (Comparable) CoerceUtil.coerce(val, Comparable.class);
```
clayreimann:
```java
Object val = PersistentResource.getValue(entity, field, dictionary);
Comparable valComp = (Comparable) CoerceUtil.coerce(val, Comparable.class);
Object filterVal = CoerceUtil.coerce(values.get(0), val.getClass());
Comparable filterComp = (Comparable) CoerceUtil.coerce(filterVal, Comparable.class);
```

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
+        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                if (values.size() != 1) {
+                    throw new InvalidPredicateException("LE can only take one argument");
+                }
+
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                Comparable filterComp = (Comparable) CoerceUtil
+                        .coerce(CoerceUtil.coerce(values.get(0), val.getClass()), Comparable.class);
+                Comparable valComp = (Comparable) CoerceUtil.coerce(val, Comparable.class);
```
clayreimann: See `lt`

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
+        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                if (values.size() != 1) {
+                    throw new InvalidPredicateException("GT can only take one argument");
+                }
+
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                Comparable filterComp = (Comparable) CoerceUtil
+                        .coerce(CoerceUtil.coerce(values.get(0), val.getClass()), Comparable.class);
+                Comparable valComp = (Comparable) CoerceUtil.coerce(val, Comparable.class);
```
clayreimann: See `lt`

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,172 @@
...
+        java.util.function.Predicate getFilterFunction(String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                if (values.size() != 1) {
+                    throw new InvalidPredicateException("GE can only take one argument");
+                }
+
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                Comparable filterComp = (Comparable) CoerceUtil
+                        .coerce(CoerceUtil.coerce(values.get(0), val.getClass()), Comparable.class);
+                Comparable valComp = (Comparable) CoerceUtil.coerce(val, Comparable.class);
```
clayreimann: See `lt`

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -44,4 +190,12 @@ public static Operator fromString(final String string) {
 
         throw new InvalidPredicateException("Unknown operator in filter: " + string);
     }
+
+    abstract java.util.function.Predicate getFilterFunction(
+            String field, List<Object> values, EntityDictionary dictionary);
```
clayreimann: Why 2 lines?

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -38,6 +40,17 @@
     @Getter @NonNull private Operator operator;
     @Getter @NonNull private List<Object> values;
 
+    public Predicate(PathElement pathElement, Operator op, List<Object> values) {
```
clayreimann: We should rename this. See above comment.

elide-core/src/test/java/com/yahoo/elide/core/filter/OperatorTest.java
```diff
@@ -0,0 +1,285 @@
...
+        // When name is null
+        author.setName(null);
+        fn = Operator.ISNULL.getFilterFunction("name", null, dictionary);
+        Assert.assertTrue(fn.test(author));
+        fn = Operator.NOTNULL.getFilterFunction("name", null, dictionary);
+        Assert.assertFalse(fn.test(author));
+    }
+
+    @Test
+    public void prefixAndPostfixAndInfixTest() throws Exception {
```
clayreimann: Should add cases where value is `null`

elide-core/src/main/java/com/yahoo/elide/core/filter/InMemoryFilterOperation.java
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright 2016, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core.filter;
```
clayreimann: Why is this in elide-core when the transaction the uses it is in the memory-store child project?

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
xiaoyao1991: I totally agree with you. I like the idea of changing elide's Predicate to FilterPredicate. Having conflicting names with java builtin packages is such a pain. Do you think it would be a breaking change to rename? If it's not I'm glad to change

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
xiaoyao1991: Just tried, it will be null. I'm gonna add a test case for that too.

elide-core/src/main/java/com/yahoo/elide/core/filter/InMemoryFilterOperation.java
xiaoyao1991: It's not restricted to be used in InMemoryDataStore only. Every datastore could have use cases when they need to do some in-memory filtering in addition to pushing the filters to the backend data store

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
xiaoyao1991: 1 line exceeds 120 characters

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
xiaoyao1991: Good catch. It's a possible point of NPE

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
clayreimann: Also, did you intend to leave out an access modifier?

elide-core/src/main/java/com/yahoo/elide/core/filter/InMemoryFilterOperation.java
clayreimann: Fair enough

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
xiaoyao1991: Yes. It's abstract

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
xiaoyao1991: Discussed with @aklish, we can do the renaming later.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -55,13 +54,14 @@
                 Object val = PersistentResource.getValue(entity, field, dictionary);
                 String valStr = (String) CoerceUtil.coerce(val, String.class);
                 String filterStr = (String) CoerceUtil.coerce(values.get(0), String.class);
-                return valStr.startsWith(filterStr);
+                return valStr != null && filterStr != null && valStr.startsWith(filterStr);
```
clayreimann: break into multiple lines for readability

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
clayreimann: Addressed

clayreimann:
Just a few lingering code-style things that you can disagree with. 👍  otherwise

elide-core/src/main/java/com/yahoo/elide/core/filter/InMemoryFilterOperation.java
```diff
@@ -0,0 +1,41 @@
...
+    @Override
+    public Set<java.util.function.Predicate> apply(Predicate predicate) {
+        return Collections.singleton(predicate.getOperator().apply(predicate, dictionary));
+    }
+
+    @Override
+    public Set<java.util.function.Predicate> applyAll(Set<Predicate> predicates) {
+        return predicates.stream()
+                .map(this::apply)
+                .flatMap(Collection::stream)
```
DeathByTape: We can remove this `flatMap()` by either using `reduce()` or the other version of `collect` (that doesn't take a `Collector` but, instead, can take 3 lambdas or method references).

Either way, I suppose I don't feel too strongly about it.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,191 @@
...
+    IN("in", true) {
+        @Override
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                if (val == null) {
+                    return false;
+                }
+                List<Object> coercedValues = values.stream()
```
DeathByTape: No need to hold this `List` object just to scan it again. We can either use a `Set` (faster set membership lookup) or, probably a better solution, instead of running `collect()` just scan for an `anyMatch`. Something like:

```java
values.stream()
  .map(v -> CoerceUtil.coerce(v, val.getClass()))
  .anyMatch(val::equals)
```

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -5,26 +5,191 @@
...
+        public java.util.function.Predicate getFilterFunction(
+                String field, List<Object> values, EntityDictionary dictionary) {
+            return (entity) -> {
+                if (values.size() != 1) {
+                    throw new InvalidPredicateException("PREFIX can only take one argument");
+                }
+
+                Object val = PersistentResource.getValue(entity, field, dictionary);
+                String valStr = (String) CoerceUtil.coerce(val, String.class);
+                String filterStr = (String) CoerceUtil.coerce(values.get(0), String.class);
```
DeathByTape: I wish we didn't have to perform these casts 👎 

Is there a reason we cannot change the signature of `coerce()` in `CoerceUtil` to:

```java
public static <T> T coerce(Object value, Class<T> cls);
```

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -44,14 +44,6 @@ public Predicate(PathElement pathElement, Operator op, List<Object> values) {
...
     }
 
-    public Predicate(PathElement pathElement, Operator op, Object value) {
-        this(Collections.singletonList(pathElement), op, Collections.singletonList(value));
-    }
-
-    public Predicate(List<PathElement> path, Operator op, Object value) {
-        this(path, op, Collections.singletonList(value));
-    }
-
```
clayreimann: What's the rationale for removing this?

xiaoyao1991: They are just redundant.

clayreimann:
👍  again

elide-core/src/main/java/com/yahoo/elide/core/filter/InMemoryFilterOperation.java
clayreimann: Fixed in #276

PR #271, opened 2016-06-23
When user tries to create an object without ID return 400 not 403

### LINKED ISSUE #101: Wonky exception type
When Elide [fails to assign an ID](https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java#L185) it responds with `403 Forbidden`, this feels misleading. I think elide should respond with `400 Bad Request` instead.

clayreimann:
Closes #101

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/exceptions/HttpStatusException.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/exceptions/HttpStatusException.java
@@ -19,7 +19,7 @@ import java.util.Map;
  * Creates fast exceptions without stack traces since filter checks can throw many of these.
  */
 @Slf4j
-public class HttpStatusException extends RuntimeException {
+public abstract class HttpStatusException extends RuntimeException {
     private static final long serialVersionUID = 1L;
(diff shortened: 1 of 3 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/exceptions/HttpStatusException.java
```diff
@@ -19,7 +19,7 @@
  * Creates fast exceptions without stack traces since filter checks can throw many of these.
  */
 @Slf4j
-public class HttpStatusException extends RuntimeException {
+abstract public class HttpStatusException extends RuntimeException {
```
DeathByTape: *sigh* this is _very_ pedantic (sorry about that) but it's more common to see `public abstract class` than `abstract public class` (i.e. flip `abstract` and `public`).

Would you mind updating this so we can remain consistent?

DeathByTape:
Just my one nit-pick comment :)

Then 👍

PR #273, opened 2016-06-23
Don't leak data from relationships

duseja2:


yahoocla:
CLA is valid!

DeathByTape:
Mostly looks good. 2 things that remain:

  1. Need to check the result of `UpdatePermission` when there is no change to a relationship and return that result code (currently it is always `201` which leaves this vulnerable to a brute-force leak).
  1. Please add a unit or integration test for this! :)

clayreimann:
👍  after @DeathByTape's comments are addressed

DeathByTape:
👍

PR #276, opened 2016-06-24
Clean up elide predicate

clayreimann:
Depends on #270

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
@@ -13,172 +13,113 @@ import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 
 import java.util.List;
-import java.util.function.BiFunction;
+import java.util.function.Predicate;
 
 /**
  * Operator enum for predicates.
  */
 @RequiredArgsConstructor
-public enum Operator implements BiFunction<Predicate, EntityDictionary, java.util.function.Predicate> {
+public enum Operator {
     IN("in", true) {
         @Override
-        public java.util.function.Predicate getFilterFunction(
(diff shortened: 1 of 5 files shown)


yahoocla:
CLA is valid!

xiaoyao1991:
👍

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -13,191 +13,265 @@
...
-                    throw new InvalidPredicateException("LE can only take one argument");
-                }
-                Object val = PersistentResource.getValue(entity, field, dictionary);
-                return val != null && getComparisonResult(val, values.get(0)) >= 0;
-            };
+            return Operator.ge(field, values, dictionary);
         }
     };
 
-    @Getter private final String string;
```
xiaoyao1991: wow, I didn't notice we have such a bad naming exist

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -13,191 +13,265 @@
...
+
+            return val != null && values.stream()
+                    .map(v -> CoerceUtil.coerce(v, val.getClass()))
+                    .anyMatch(val::equals);
+        };
+    }
+
+    //
+    // Predicate generation
+    //
```
xiaoyao1991: Maybe we should use `/**` here?

clayreimann: Not really javadoc, I just wanted some visual seperation

xiaoyao1991:
👍  LGTM

DeathByTape:
👍

PR #278, opened 2016-06-24
Added filter RSQL '=isnull=' support.

tarrantzhang:
Implemented option 2 in [this ticket](https://github.com/jirutka/rsql-parser/issues/21).

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
@@ -45,6 +45,14 @@ public class Predicate implements FilterExpression, Function<EntityDictionary, j
         this(Collections.singletonList(pathElement), op, values);
     }
 
+    public Predicate(PathElement pathElement, Operator op) {
+        this(Collections.singletonList(pathElement), op, Collections.emptyList());
+    }
+
+    public Predicate(List<PathElement> path, Operator op) {
+        this(path, op, Collections.emptyList());
+    }
+
     public String getField() {
         PathElement last = path.get(path.size() - 1);
         return last.getFieldName();
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
@@ -7,6 +7,7 @@ package com.yahoo.elide.core.filter.dialect;
 
 import com.google.common.collect.ImmutableMap;
 import com.yahoo.elide.core.EntityDictionary;
+import com.yahoo.elide.core.exceptions.InvalidValueException;
 import com.yahoo.elide.core.filter.Operator;
 import com.yahoo.elide.core.filter.Predicate;
 import com.yahoo.elide.core.filter.expression.AndFilterExpression;
@@ -45,29 +46,32 @@ public class RSQLFilterDialect implements SubqueryFilterDialect, JoinFilterDiale
 
     private static final Pattern TYPED_FILTER_PATTERN = Pattern.compile("filter\\[([^\\]]+)\\]");
 
+    private static final ComparisonOperator ISNULL_OP = new ComparisonOperator("=isnull=", false);
+
     /* Subset of operators that map directly to Elide operators */
     private static final Map<ComparisonOperator, Operator> OPERATOR_MAP =
             ImmutableMap.<ComparisonOperator, Operator>builder()
-            .put(RSQLOperators.IN, Operator.IN)
-            .put(RSQLOperators.NOT_IN, Operator.NOT)
-            .put(RSQLOperators.LESS_THAN, Operator.LT)
-            .put(RSQLOperators.GREATER_THAN, Operator.GT)
-            .put(RSQLOperators.GREATER_THAN_OR_EQUAL, Operator.GE)
-            .put(RSQLOperators.LESS_THAN_OR_EQUAL, Operator.LE)
-            .build();
+                    .put(RSQLOperators.IN, Operator.IN)
+                    .put(RSQLOperators.NOT_IN, Operator.NOT)
+                    .put(RSQLOperators.LESS_THAN, Operator.LT)
+                    .put(RSQLOperators.GREATER_THAN, Operator.GT)
+                    .put(RSQLOperators.GREATER_THAN_OR_EQUAL, Operator.GE)
+                    .put(RSQLOperators.LESS_THAN_OR_EQUAL, Operator.LE)
+                    .build();
 
     private final RSQLParser parser;
     private final EntityDictionary dictionary;
 
-    public RSQLFilterDialect(EntityDictionary dictionary, Set<ComparisonOperator> rsqlOperators) {
-        // Match "filter[<type>]"
-
-        parser = new RSQLParser(rsqlOperators);
+    public RSQLFilterDialect(EntityDictionary dictionary) {
+        parser = new RSQLParser(getDefaultOperatorsWithIsnull());
         this. dictionary = dictionary;
     }
 
-    public RSQLFilterDialect(EntityDictionary dictionary) {
-        this(dictionary, RSQLOperators.defaultOperators());
+    //add rsql isnull op to the default ops
+    private static Set<ComparisonOperator> getDefaultOperatorsWithIsnull() {
+        Set<ComparisonOperator> operators = RSQLOperators.defaultOperators();
+        operators.add(ISNULL_OP);
+        return operators;
     }
 
     @Override
@@ -274,6 +278,11 @@ public class RSQLFilterDialect implements SubqueryFilterDialect, JoinFilterDiale
                 throw new RSQLParseException(String.format("No such association %s", relationship));
             }
 
+            //handles '=isnull=' op before coerce arguments
+            if (op.equals(ISNULL_OP)) {
+                return buildIsNullOperator(path, arguments);
+            }
+
             Class<?> relationshipType = path.get(0).getFieldType();
 
             //Coerce arguments to their correct types
@@ -318,6 +327,28 @@ public class RSQLFilterDialect implements SubqueryFilterDialect, JoinFilterDiale
             }
 
             throw new RSQLParseException(String.format("Invalid Operator %s", op.getSymbol()));
+        }
+
+        /**
+         * Returns Predicate for '=isnull=' case depending on its arguments.
+         *
+         * NOTE: Filter Expression builder specially for '=isnull=' case.
+         *
+         * @return Returns Predicate for '=isnull=' case depending on its arguments.
+         */
+        private FilterExpression buildIsNullOperator(List<Predicate.PathElement> path, List<String> arguments) {
+            Operator elideOP;
+            try {
+                boolean argBool = (boolean) CoerceUtil.coerce(arguments.get(0), boolean.class);
+                if (argBool) {
+                    elideOP = Operator.ISNULL;
+                } else {
+                    elideOP = Operator.NOTNULL;
+                }
+                return new Predicate(path, elideOP);
+            } catch (InvalidValueException e) {
+                throw new RSQLParseException(String.format("Invalid value for operator =isnull="));
+            }
         }
     }
 }
--- a/elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
+++ b/elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
@@ -799,6 +799,52 @@ class FilterIT extends AbstractIntegrationTestInitializer {
             Assert.assertTrue(book.get("attributes").get("genre").isNull())
             Assert.assertTrue(bookIdsWithNullGenre.contains(book.get("id")))
         }
+
+        /* Test RSQL Typed */
+        /* param = true */
+        result = mapper.readTree(
+                RestAssured.get("book?filter[book]=genre=isnull=true").asString())
+
+        Assert.assertEquals(result.get("data").size(), bookIdsWithNullGenre.size())
+
+        for (JsonNode book : result.get("data")) {
+            Assert.assertTrue(book.get("attributes").get("genre").isNull())
+            Assert.assertTrue(bookIdsWithNullGenre.contains(book.get("id")))
+        }
+
+        /* param = 1 */
+        result = mapper.readTree(
+                RestAssured.get("book?filter[book]=genre=isnull=1").asString())
+
+        Assert.assertEquals(result.get("data").size(), bookIdsWithNullGenre.size())
+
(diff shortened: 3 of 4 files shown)


yahoocla:
CLA is valid!

elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
```diff
@@ -1328,6 +1328,88 @@ class FilterIT extends AbstractIntegrationTestInitializer {
                         "Invalid query parameter: filter");
     }
 
+    @Test
+    public void testNull() {
```
aklish: There were already tests here for the default filter dialect to test isNull and isNotNull.  You just need to add to those existing tests.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -314,7 +318,28 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
                     return new NotFilterExpression(new Predicate(path, Operator.IN, values));
                 }
             } else if (OPERATOR_MAP.containsKey(op)) {
-                return new Predicate(path, OPERATOR_MAP.get(op), values);
+                if (op == isNullOp) {
```
aklish: We should move this to a separate function I believe to reduce cyclomatic complexity.

aklish: The type coercion above for values should NOT be applied for isnull operator.  We need special handling for this operator that should be treated separately.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -54,14 +56,16 @@
             .put(RSQLOperators.GREATER_THAN, Operator.GT)
             .put(RSQLOperators.GREATER_THAN_OR_EQUAL, Operator.GE)
             .put(RSQLOperators.LESS_THAN_OR_EQUAL, Operator.LE)
+            .put(isNullOp, Operator.ISNULL)
```
aklish: I don't think this is needed if we make the refactor I mention below.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -45,6 +45,8 @@
 
     private static final Pattern TYPED_FILTER_PATTERN = Pattern.compile("filter\\[([^\\]]+)\\]");
 
+    private static ComparisonOperator isNullOp = new ComparisonOperator("=isnull=", false);
```
aklish: public static final.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -54,14 +56,16 @@
...
             .build();
 
     private final RSQLParser parser;
     private final EntityDictionary dictionary;
 
     public RSQLFilterDialect(EntityDictionary dictionary, Set<ComparisonOperator> rsqlOperators) {
         // Match "filter[<type>]"
-
+        //Adding "=isnull=" operator
+        rsqlOperators.add(isNullOp);
```
aklish: I don't think this constructor works as intended.  We should probably remove it.

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -43,6 +44,14 @@ public Predicate(PathElement pathElement, Operator op, List<Object> values) {
...
     }
 
+    public Predicate(PathElement pathElement, Operator op){
+        this(Collections.singletonList(pathElement), op, new ArrayList<Object>());
+    }
+
+    public Predicate(List<PathElement> path, Operator op){
+        this(path, op, new ArrayList<Object>());
+    }
+
```
clayreimann: These were just removed.

tarrantzhang: There is no constructor in Predicate class allow instantiate without `values`. And "ISNULL" "NOTNULL" operators are Nullary filters, so I think it's good to make them.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -61,13 +63,19 @@
 
     public RSQLFilterDialect(EntityDictionary dictionary, Set<ComparisonOperator> rsqlOperators) {
         // Match "filter[<type>]"
-
```
aklish: Can we remove this constructor?  I think it was an error when I added it.  I

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -313,11 +321,29 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
                 } else {
                     return new NotFilterExpression(new Predicate(path, Operator.IN, values));
                 }
+            } else if (op == isnullOp) {
```
aklish: It is too late to put this here as we already have attempted to coerce the values above.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -313,11 +321,29 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
...
+                Operator elideOP = rsqlIsnull2Operator(values);
+                return new Predicate(path, elideOP);
             } else if (OPERATOR_MAP.containsKey(op)) {
                 return new Predicate(path, OPERATOR_MAP.get(op), values);
             }
 
             throw new RSQLParseException(String.format("Invalid Operator %s", op.getSymbol()));
         }
+
+        private Operator rsqlIsnull2Operator(List<Object> values) {
```
aklish: I would have a function which just returns the correct FilterExpression predicate.  This should be called near the beginning of the visit function - before we attempt to coerce values.

elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
```diff
@@ -799,6 +799,29 @@ class FilterIT extends AbstractIntegrationTestInitializer {
             Assert.assertTrue(book.get("attributes").get("genre").isNull())
             Assert.assertTrue(bookIdsWithNullGenre.contains(book.get("id")))
         }
+
+        /* Test RSQL Typed */
+        /* param = true */
+        result = mapper.readTree(
```
aklish: we should check global filters too.

elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
```diff
@@ -823,6 +846,29 @@ class FilterIT extends AbstractIntegrationTestInitializer {
             Assert.assertTrue(!book.get("attributes").get("genre").isNull())
             Assert.assertTrue(bookIdsWithNonNullGenre.contains(book.get("id")))
         }
+
+        /* Test RSQL Typed */
+        /* param = false */
+        result = mapper.readTree(
+                RestAssured.get("book?filter[book]=genre=isnull=false").asString())
```
aklish: Same comment.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -45,6 +45,8 @@
 
     private static final Pattern TYPED_FILTER_PATTERN = Pattern.compile("filter\\[([^\\]]+)\\]");
 
+    private static final ComparisonOperator isnullOp = new ComparisonOperator("=isnull=", false);
```
xiaoyao1991: For static final variable, you may wanna use ALL_UPPER_CASE_NAMING

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
xiaoyao1991: let's do `rsqlIsNullToOperator`, `2O` here looks really like twenty

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -43,6 +44,14 @@ public Predicate(PathElement pathElement, Operator op, List<Object> values) {
         this(Collections.singletonList(pathElement), op, values);
     }
 
+    public Predicate(PathElement pathElement, Operator op){
+        this(Collections.singletonList(pathElement), op, new ArrayList<Object>());
```
xiaoyao1991: Could use `Collections.emptyList()` so that it's *sorta* an immutable list

xiaoyao1991: I believe we do have a constructor for Predicate where it takes a pathElement instead of a list of it

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -43,6 +44,14 @@ public Predicate(PathElement pathElement, Operator op, List<Object> values) {
...
+    }
+
+    public Predicate(List<PathElement> path, Operator op){
+        this(path, op, new ArrayList<Object>());
```
xiaoyao1991: Could use Collections.emptyList() so that it's sorta an immutable list

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
xiaoyao1991: Should it be `.equals`?

tarrantzhang: @aklish could you explain a bit on why it's too late here?  I think these if..else statements are just matching the `op` and then assign correct handler for it.

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
tarrantzhang: @xiaoyao1991 Yes we do. We have both the `@AllArgsConstructor` for list of `pathElement` and also the non-list version for 3-params constructor. So I follow it to make both versions for 2-params constructor.

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -43,6 +44,14 @@ public Predicate(PathElement pathElement, Operator op, List<Object> values) {
...
     }
 
+    public Predicate(PathElement pathElement, Operator op){
+        this(Collections.singletonList(pathElement), op, Collections.emptyList());
+    }
+
+    public Predicate(List<PathElement> path, Operator op){
+        this(path, op, Collections.emptyList());
+    }
+
```
clayreimann: These were recently removed by @xiaoyao1991's PR. Are they necessary to your PR or do you need to rebase?

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -319,5 +327,33 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
...
+                    break;
+                case "1":
+                    elideOP = Operator.ISNULL;
+                    break;
+                case "0":
+                    elideOP = Operator.NOTNULL;
+                    break;
+                default:
+                    throw new RSQLParseException(String.format("Invalid value for operator =isnull="));
+            }
```
clayreimann: You should make use of case fall through to condense this `switch`.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -319,5 +327,29 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
...
+        /**
+         * Returns Predicate for '=isnull=' case depending on its arguments.
+         *
+         * NOTE: Filter Expression builder specially for '=isnull=' case.
+         *
+         * @return Returns Predicate for '=isnull=' case depending on its arguments.
+         */
+        private FilterExpression buildIsNullOperator(List<Predicate.PathElement> path, List<String> arguments) {
+            Operator elideOP;
+            switch (arguments.get(0)) {
```
aklish: Do coercion to boolean instead of checking equality against true, false, 0, and 1.  You don't handle case for example in this switch statement.

aklish:
👍

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -15,13 +14,12 @@
 
 import java.util.Collections;
 import java.util.List;
-import java.util.function.Function;
 
 /**
  * Predicate class.
  */
 @AllArgsConstructor
-public class Predicate implements FilterExpression, Function<EntityDictionary, java.util.function.Predicate> {
```
clayreimann: I think this is unintended.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -274,12 +278,17 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
...
+            if (op.equals(ISNULL_OP)) {
+                return buildIsNullOperator(path, arguments);
+            }
+
+            Class relationshipType = path.get(0).getFieldType();
 
             //Coerce arguments to their correct types
             List<Object> values = arguments
                     .stream()
-                    .map((argument) -> (Object) CoerceUtil.coerce(argument, relationshipType))
```
clayreimann: This might be required.

elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
```diff
@@ -1328,6 +1466,7 @@ class FilterIT extends AbstractIntegrationTestInitializer {
                         "Invalid query parameter: filter");
     }
 
+
```
clayreimann: Boo, hiss. (also this might break the linter)

clayreimann:
lgtm, but you should squash again.

PR #279, opened 2016-06-24
[WIP] delete cascade parent

asifmansoora:


yahoocla:

yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -668,6 +668,15 @@ public void deleteResource() throws ForbiddenAccessException {
                         inverseResource.markDirty();
                     }
                 }
+                Object obj = getValueUnchecked(relationName);
+                if (obj instanceof Collection) {
+                    Collection coll = (Collection) obj;
+                    try {
+                        coll.clear();
+                    } catch (UnsupportedOperationException e) {
+//                        log.debug("Caught unsupported exception");
```
DeathByTape: We should actually log here

elide-integration-tests/src/test/resources/logback.xml
```diff
@@ -11,7 +11,7 @@
     </encoder>
   </appender>
 
-  <root level="error">
+  <root level="trace">
```
DeathByTape: Can we change this back to error?

elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
```diff
@@ -130,11 +135,25 @@ public void testRootCollection() throws Exception {
         String actual = given().when().get("/parent").then().statusCode(HttpStatus.SC_OK)
                 .extract().body().asString();
 
+        System.out.println("test root" + actual);
```
DeathByTape: Remove

DeathByTape:
I hope to be able to look at this soon

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -99,7 +99,10 @@ public void delete(Object object) {
 
     @Override
     public void save(Object object) {
-        deferredTasks.add(() -> session.saveOrUpdate(object));
+        deferredTasks.add(() -> {
+            session.saveOrUpdate(object);
+            session.evict(object);
```
clayreimann: Can you provide a comment explaining why we're saving and then evicting the object. I'm assuming this is some magical Hibernate incantation, but it would be nice to provide the explanation so that if it stops working at some point we can have an idea of why this was needed to begin with.

elide-integration-tests/src/test/java/example/Child.java
```diff
@@ -18,15 +18,7 @@
...
-import javax.persistence.CascadeType;
-import javax.persistence.Column;
-import javax.persistence.Entity;
-import javax.persistence.FetchType;
-import javax.persistence.GeneratedValue;
-import javax.persistence.GenerationType;
-import javax.persistence.Id;
-import javax.persistence.ManyToMany;
-import javax.persistence.OneToOne;
+import javax.persistence.*;
```
clayreimann: Un-star this import please.

elide-integration-tests/src/test/java/example/Parent.java
```diff
@@ -18,12 +18,7 @@
 import com.yahoo.elide.security.checks.prefab.Role;
 import lombok.ToString;
 
-import javax.persistence.CascadeType;
-import javax.persistence.Column;
-import javax.persistence.Entity;
-import javax.persistence.JoinColumn;
-import javax.persistence.JoinTable;
-import javax.persistence.ManyToMany;
+import javax.persistence.*;
```
clayreimann: Un-star please.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
DeathByTape: I realized after I started commenting, this is some scratch work we were doing in-person on Asif's laptop last Friday. He submitted the test case for me so we could better debug and fix offline :)

elide-integration-tests/src/test/java/example/Parent.java
xiaoyao1991: You can actually refer to use [our IntelliJ code style template](https://github.com/yahoo/elide/blob/master/CONTRIBUTING.md)

asifmansoora:
This is a test brach to reproduce a bug that Elide causes when a parent entity is hard removed. So please do not review for now, once the bug is fixed - will send out a clean reviewable PR .
Thanks

clayreimann:
Ok, just remove the `[wip]` from the PR title when you're ready for review.

asifmansoora:
Sure

Koka:
Doesn't this PR relates to https://github.com/yahoo/elide/issues/300 ?

asifmansoora:
@Koka Yes . But this PR was to show the devs the test cases that was failing because of broken delete cascades. We have still not worked on the actual implementation

Koka:
Actually I've managed to fix cascade deletion by just commenting out whole code block (in PersistentResource class) which nullifies inverse relationships, but I suspect it was needed for something.

asifmansoora:
Yup - I did that before and tested the app which used Elide and it broke other test cases around inverse relations

DeathByTape:
@aklish is going to end up providing a complete solution to this problem in another PR here soon. Closing.

Issue #282, opened 2016-06-29
How to include "links" attribute in response?


bduisenov:
I was not able to find any example.

clayreimann:
I don't think we support `links` at the moment. @DeathByTape might know differently

bduisenov:
Hi @clayreimann @DeathByTape, I can create PR with this. I was able to supplement Elide responses with "self" & "related" links. But I'm not sure whether initial url path should set with Elide.Builder or somehow differently.

DeathByTape:
I believe the links can be relative. While I can't find any explicit mention of this in the jsonapi spec, [they do have an example](http://jsonapi.org/format/#document-resource-objects) using relative URIs. In this case, a prefix uri would be unnecessary unless you were redirecting to another service. 

What do you guys think? I feel like using relative links probably avoids the problem of having to specify your service's root prefix. It's also possible we can support both, but perhaps unnecessary.

jimirocks:
Hello @DennisMcWherter , what is the status of links in Elide's JSON:API now? Seems like this effort didn't reach the finalization? Would you consider to finish it?

aklish:
I think the existing approach from @bduisenov looks pretty solid.  I think some things that would be nice to add:
1. The way the lineage is stored in the PersistentResource in a map is not ideal.  Ideally, the lineage is just an ordered list of resources.  I think we could update the lineage to include this structure as well and use it instead of the map keys.
2. As referenced, it would be nice to tie this into elide-spring-boot and elide-standalone with configuration that pulls in the URL.
3. I think we should have an option to turn this off/on in settings as links inflates the payload response.  By default, it should be off.

For merging this to master, the only absolute must have on my list is 3.  The other two could be done in another PR (or together).

rehab-reda0:
That will be great. we need links but we need also to customize it 
I mean the links object instead of containing {self; " " ,related:" "}
w may need it to contain {self: " ", name1: " ", name2:" "} is that will be applicable by this pull request ?

PR #284, opened 2016-07-01
added "links" member (#282)

bduisenov:


yahoocla:
CLA is valid!

clayreimann:
@bduisenov First, thanks for the PR. Second, I'm not against having `links` in our data (for all those HATEOAS fans out there), but this implementation is broken. Many of the `links` you are generating are invalid (for example `"self": "/child/8"` is not a viable url). 

I see two reasonable solutions here:

* If you want to create links relative to the current resource then they should start with a `.`. 
* If you want to create links that are relative to the service (which I'm inclined to say is the best course) then you should use the `PersistentEntity`'s `ResourceLineage` to reconstitute the path to the resource being accessed.

bduisenov:
Hi @clayreimann,
Thanks for your response.
In regards to your first point, the syntax of relative url comes from the example given in [Resource Object](http://jsonapi.org/format/#document-resource-objects) section. 
@DeathByTape any thoughts on that?

clayreimann:
@bduisenov I understand but what I think you're missing is that the urls that are shown in the spec are relative to the root of the API. There's nothing wrong with this, but if that's what you're trying to achieve then you have missed your target. As I mentioned above the url created from the link `/child/8/relationships/friends` (which would resolve to `http://your_elide_API/child/8/relationships/friends`) _is invalid_. 

`/child/8/relationships/friends` could potentially be made to be valid if you consider all URLs to be relative to the current resource (i.e. not idempotent urls). If you wish to go this route, which I highly discourage, then the link you generate should be `./children/8/relationships/friends`–the link starts with `.` to symbolize that it is relative to the position of the resource being accessed, and uses `children` (the field name) rather than `child` (the model name).

DeathByTape:
@bduisenov any progress or questions? :)

bduisenov:
@DeathByTape no progress so far. Trying to catch up on my backlog that grew to an interesting size during summer. Definitely would not be able to continue working on this until November.

DeathByTape:
sounds good, no rush.

If (maybe a big if?) we can chew through our backlog before then we may also pick up where you've left off so far.

Thanks! 😄

nicolasrueegg:
Hi guys, I just wanted to point to the recommendations jsonapi has considering including links in a resource response: http://jsonapi.org/recommendations/#including-links
This information might be useful when considering the details of the implementation.

bduisenov:
Hi @DeathByTape @clayreimann, I'm back with code fixes. Could you please make a review.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1220,6 +1221,7 @@ public Resource toResource(final Supplier<Map<String, Relationship>> relationshi
                 : dictionary.getId(obj));
         resource.setRelationships(relationshipSupplier.get());
         resource.setAttributes(attributeSupplier.get());
+        resource.setLinks(getResourceLevelLink());
```
clayreimann: (I feel) This should be configurable on a bunch of levels. You should be able to enable/disable links

* when creating the Elide instance
* when annotating a particular resource or relationship
* as a query param

I think a tri-state would make sense too. (`ForbidLinks`, `PermitLinks`, `ForceLinks`)

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1267,14 +1269,52 @@ public Resource toResource(final Supplier<Map<String, Relationship>> relationshi
...
+     * Get base resource url of current resource.
+     * @return resource url
+     */
+    protected String getResourceUrl() {
+        StringBuilder result = new StringBuilder();
+        for (String name : lineage.getKeys()) {
+            if (name.equals(getType())) {
+                break;
+            }
+            PersistentResource value = lineage.getRecord(name).get(0);
```
DeathByTape: You said you were concerned about this line. I believe this is a fine approach. In fact, this is exactly what the lineage was created for: to be able to tell where you came from.

Perhaps we could clean up the lineage interface, however, to make this a bit cleaner to use.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
DeathByTape: @clayreimann @bduisenov, from the perspective of preserving current behavior I think we should probably make this configurable before finally merging this PR in. Playing off of the tri-state idea:

We could default the behavior to `PermitLinks` for the whole elide instance and unspecified beans. This way the behavior is the same for anyone pulling this change in.

I am assuming the following:

`ForbidLinks` -- Links is never displayed on a resource/elide instance
`PermitLinks` -- Links will be displayed if and only if requested by the caller (via a query param) for a resource/elide instance
`ForceLinks` -- Links are always displayed for a resource/elide instance

Provided those assumptions hold, I believe there shouldn't be much work in the configuration bit for this.

btw, thanks a lot for your work on this @bduisenov! :)

bduisenov: @clayreimann @DeathByTape looks good to me. I'll try to spend some time on this weekends.

DeathByTape: Thanks!

celdridge91190:
Is this still in progress? I'm interested in the links functionality and was curious if this would be merged in in the near future :)

clayreimann:
Unknown, as you can see we haven't heard from @bduisenov recently. If you wanted to pick it up we'd be happy to review it, otherwise I might have time to pick it up (but it's not high on my priority list)

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
bduisenov: Hi @clayreimann @DeathByTape,
I'm back and want to finish this PR. 
I have a question about `PermitLinks` mode, think that I'm not quite sure how `url` should look like. I assume that the `url` might be something like `/people?links=true`. Do you have any thoughts on that?

aklish:
Could you rebase instead of merge?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
aklish: links as a query param seems reasonable to me.  I don't see anything in jsonapi that would preclude it.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
aklish: I actually don't think this right.  We're using the persistent resource's _type_ rather than the _relationship name_ to build the URL.  These links won't be correct.  

I'm wondering if a better approach is pass the URL through the request scope?  Either that or the lineage needs to be augmented to include the relationship names used to traverse to a given persistent resource.

elide-integration-tests/src/test/resources/ResourceIT/createChildRelateExisting.json
```diff
@@ -8,16 +8,27 @@
...
+                    "links": {
+                        "self": "/parent/7/child/8/relationships/friends",
+                        "related": "/parent/7/child/8/friends"
+                    },
                     "data":[]
                 },
                 "parents":{
+                    "links": {
+                        "self": "/parent/7/child/8/relationships/parents",
+                        "related": "/parent/7/child/8/parents"
```
aklish: Here is an example of the above comment.  This should be /parent/7/children/8/parents

belovaf:
Hi all! What links should be returned for resource in included section which is
1. Not root resource
2. Referenced by multiple include paths

Example:
Resource `magazine` has 2 collections `products` and `bestProducts` (subset of products).
Request url is `{endpoint}/magazine/1?include=products,bestProducts`.
What link should be in included `bestProducts` resources?
Variants:
1. `/magazine/1/products/{id}`
2. `/magazine/1/bestProducts/{id}`
3. both

clayreimann:
I would think that you would include both types of links. There would be some overlap, but that's ok since a member of `products` can be in `bestProducts` now, but might be removed at some later time.

kelseybrennan:
I'm very interested in having this functionality available, is there a reason this PR wasn't accepted? I might pick it up.

clayreimann:
@kelseybrennan the original author disappeared. We’d be more than happy to review a PR from you that adds links

PR #286, opened 2016-07-01
Adding FilterExpressionCheck support. 

tarrantzhang:
This PR allows merging permission restriction(especially ReadPermission)  to `FilterExpression` which transfer the workload of checking permission from elide layer to datastore layer. 

Finished: implemented the corresponding check interface and AST visitor. 
                integrating the check in datastore query building process.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
@@ -90,6 +90,10 @@ public interface DataStoreTransaction extends Closeable {
      */
     <T> T loadObject(Class<T> entityClass, Serializable id);
 
+    default <T> T loadObject(Class<T> entityClass, Serializable id, Optional<FilterExpression> filterExpression) {
+        return loadObject(entityClass, id);
+    }
+
     /**
      * Read entity records from database table.
      *
@@ -112,6 +116,7 @@ public interface DataStoreTransaction extends Closeable {
         return loadObjects(entityClass);
     }
 
+
     /**
      * Read entity records from database table with applied criteria.
      *
@@ -124,6 +129,7 @@ public interface DataStoreTransaction extends Closeable {
         // default to ignoring criteria
         return loadObjects(entityClass);
     }
+
 
     /**
      * Filter a collection by the Predicates in filterScope.
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -23,6 +23,7 @@ import com.yahoo.elide.core.exceptions.InvalidEntityBodyException;
 import com.yahoo.elide.core.exceptions.InvalidObjectIdentifierException;
 import com.yahoo.elide.core.filter.Operator;
 import com.yahoo.elide.core.filter.Predicate;
+import com.yahoo.elide.core.filter.expression.AndFilterExpression;
 import com.yahoo.elide.core.filter.expression.FilterExpression;
 import com.yahoo.elide.core.filter.expression.PredicateExtractionVisitor;
 import com.yahoo.elide.extensions.PatchRequestScope;
@@ -256,8 +257,10 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
         T obj = (T) cache.get(dictionary.getJsonAliasFor(loadClass), id);
         if (obj == null) {
             // try to load object
+            Optional<FilterExpression> permissionFilter = getPermissionFilterExpression(loadClass,
+                    requestScope);
             Class<?> idType = dictionary.getIdType(loadClass);
-            obj = tx.loadObject(loadClass, (Serializable) CoerceUtil.coerce(id, idType));
+            obj = tx.loadObject(loadClass, (Serializable) CoerceUtil.coerce(id, idType), permissionFilter);
             if (obj == null) {
                 throw new InvalidObjectIdentifierException(id, loadClass.getSimpleName());
             }
@@ -297,6 +300,23 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
             requestScope.queueCommitTrigger(resource);
         }
         return resources;
+    }
+
+    /**
+     * Get a FilterExpression parsed from FilterExpressionCheck.
+     *
+     * @param <T> the type parameter
+     * @param loadClass the load class
+     * @param requestScope the request scope
+     * @return a FilterExpression defined by FilterExpressionCheck.
+     */
+    private static <T> Optional<FilterExpression> getPermissionFilterExpression(Class<T> loadClass,
+                                                                      RequestScope requestScope) {
+        try {
+            return requestScope.getPermissionExecutor().getReadPermissionFilter(loadClass);
+        } catch (ForbiddenAccessException e) {
+            return Optional.empty();
+        }
     }
 
     /**
@@ -909,6 +929,18 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
                                                          Optional<FilterExpression> filterExpression) {
         RelationshipType type = getRelationshipType(relationName);
         final Class<?> relationClass = dictionary.getParameterizedType(obj, relationName);
+
+        //Invoke filterExpressionCheck and then merge with filterExpression.
+        Optional<FilterExpression> permissionFilter = getPermissionFilterExpression(relationClass, requestScope);
+        if (permissionFilter.isPresent()) {
+            if (filterExpression.isPresent()) {
+                FilterExpression mergedExpression =
+                        new AndFilterExpression(filterExpression.get(), permissionFilter.get());
+                filterExpression = Optional.of(mergedExpression);
+            } else {
+                filterExpression = permissionFilter;
+            }
+        }
 
         Object val;
 
--- a/elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
@@ -9,6 +9,7 @@ import com.yahoo.elide.annotation.OnCommit;
 import com.yahoo.elide.audit.AuditLogger;
 import com.yahoo.elide.core.exceptions.InvalidPredicateException;
 import com.yahoo.elide.core.filter.dialect.MultipleFilterDialect;
+import com.yahoo.elide.core.filter.expression.AndFilterExpression;
 import com.yahoo.elide.core.filter.expression.FilterExpression;
 import com.yahoo.elide.core.filter.dialect.ParseException;
 import com.yahoo.elide.core.pagination.Pagination;
@@ -327,11 +328,28 @@ public class RequestScope implements com.yahoo.elide.security.RequestScope {
      * @return The global filter expression evaluated at the first load
      */
     public Optional<FilterExpression> getLoadFilterExpression(Class<?> loadClass) {
+        Optional<FilterExpression> permissionFilter;
+        permissionFilter = getPermissionExecutor().getReadPermissionFilter(loadClass);
+        Optional<FilterExpression> globalFilterExpressionOptional = null;
         if (globalFilterExpression == null) {
             String typeName = dictionary.getJsonAliasFor(loadClass);
-            return getFilterExpressionByType(typeName);
+            globalFilterExpressionOptional =  getFilterExpressionByType(typeName);
+        } else {
+            globalFilterExpressionOptional = Optional.of(globalFilterExpression);
         }
-        return Optional.of(globalFilterExpression);
+
+        if (globalFilterExpressionOptional.isPresent() && permissionFilter.isPresent()) {
+            return Optional.of(new AndFilterExpression(globalFilterExpressionOptional.get(),
+                    permissionFilter.get()));
+        }
+        else if (globalFilterExpressionOptional.isPresent()) {
+            return globalFilterExpressionOptional;
+        }
+        else if (permissionFilter.isPresent()) {
+            return permissionFilter;
+        } else {
+            return Optional.empty();
+        }
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
@@ -9,9 +9,11 @@ import com.yahoo.elide.core.EntityDictionary;
 import com.yahoo.elide.core.filter.expression.FilterExpression;
 import com.yahoo.elide.core.filter.expression.Visitor;
 import lombok.AllArgsConstructor;
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
-import lombok.NonNull;
 import lombok.ToString;
+import lombok.NonNull;
+
 
 import java.util.Collections;
 import java.util.List;
@@ -21,6 +23,7 @@ import java.util.function.Function;
  * Predicate class.
  */
 @AllArgsConstructor
+@EqualsAndHashCode
 public class Predicate implements FilterExpression, Function<EntityDictionary, java.util.function.Predicate> {
 
     /**
@@ -30,6 +33,7 @@ public class Predicate implements FilterExpression, Function<EntityDictionary, j
      */
     @AllArgsConstructor
     @ToString
+    @EqualsAndHashCode
     public static class PathElement {
         @Getter Class type;
         @Getter String typeName;
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
@@ -5,11 +5,13 @@
  */
 package com.yahoo.elide.core.filter.expression;
 
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 
 /**
  * An 'And' Filter FilterExpression.
  */
+@EqualsAndHashCode
 public class AndFilterExpression implements FilterExpression {
 
     @Getter private FilterExpression left;
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/NotFilterExpression.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/NotFilterExpression.java
@@ -5,11 +5,13 @@
  */
 package com.yahoo.elide.core.filter.expression;
 
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 
 /**
  * A 'Not' Filter FilterExpression.
  */
+@EqualsAndHashCode
 public class NotFilterExpression implements FilterExpression {
 
     @Getter private FilterExpression negated;
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/OrFilterExpression.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/OrFilterExpression.java
@@ -6,11 +6,13 @@
 
 package com.yahoo.elide.core.filter.expression;
 
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 
 /**
  * An 'Or' Filter FilterExpression.
  */
+@EqualsAndHashCode
 public class OrFilterExpression implements FilterExpression {
 
     @Getter private FilterExpression left;
--- a/elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
+++ b/elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2016, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+
+package com.yahoo.elide.security;
+
+import com.yahoo.elide.core.EntityDictionary;
+import com.yahoo.elide.core.filter.Predicate;
+import com.yahoo.elide.core.filter.expression.FilterExpression;
+import com.yahoo.elide.parsers.expression.FilterExpressionCheckEvaluationVisitor;
+import com.yahoo.elide.security.checks.InlineCheck;
+
+import java.util.Optional;
+
+/**
+ * Check for FilterExpression. This is a super class for user defined FilterExpression check. The subclass should
+ * override getFilterExpression function and return a FilterExpression which will be passed down to datastore.
+ * @param <T> Type of class
+ */
+public abstract class FilterExpressionCheck<T> extends InlineCheck<T> {
+
+    /**
+     * Returns a FilterExpression from FilterExpressionCheck.
+     * @param requestScope Request scope object
+     * @return FilterExpression for FilterExpressionCheck.
+     */
+    public abstract FilterExpression getFilterExpression(RequestScope requestScope);
+
+
+    /**
+     * The filter expression is evaluated in memory if it cannot be pushed to the data store by elide for any reason.
+     * @param object object returned from datastore
+     * @param requestScope Request scope object
+     * @param changeSpec Summary of modifications
+     * @return true if the object pass evaluation against FilterExpression.
+     */
+    @Override
+    public final boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
+        FilterExpression filterExpression = getFilterExpression(requestScope);
+        return filterExpression.accept(new FilterExpressionCheckEvaluationVisitor(object, this, requestScope));
+    }
+
+    /**
+     *
+     * @param object object returned from datastore
+     * @param predicate A predicate from filterExpressionCheck
+     * @param requestScope Request scope object
+     * @return true if the object pass evaluation against Predicate.
+     */
+    public boolean applyPredicateToObject(T object, Predicate predicate, RequestScope requestScope) {
+        Predicate.PathElement path = predicate.getPath().get(predicate.getPath().size() - 1);
+        Class type = path.getType();
+        String fieldName = path.getFieldName();
+        if (fieldName == null) {
+            return false;
+        }
+        EntityDictionary dictionary = ((com.yahoo.elide.core.RequestScope) requestScope).getDictionary();
+        java.util.function.Predicate fn = predicate.getOperator()
+                .contextualize(fieldName, predicate.getValues(), dictionary);
+        return fn.test(object);
+    }
+}
--- a/elide-core/src/main/java/com/yahoo/elide/security/PermissionExecutor.java
+++ b/elide-core/src/main/java/com/yahoo/elide/security/PermissionExecutor.java
@@ -5,9 +5,11 @@
  */
 package com.yahoo.elide.security;
 
+import com.yahoo.elide.core.filter.expression.FilterExpression;
 import org.antlr.v4.runtime.tree.ParseTree;
 
 import java.lang.annotation.Annotation;
+import java.util.Optional;
 import java.util.function.BiFunction;
 import java.util.function.Function;
 
@@ -94,6 +96,8 @@ public interface PermissionExecutor {
      * @param <A> type parameter
      */
     <A extends Annotation> void checkUserPermissions(Class<?> resourceClass, Class<A> annotationClass);
+
+    Optional<FilterExpression> getReadPermissionFilter(Class<?> resourceClass);
 
     /**
      * Execute commmit checks.
--- a/elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
+++ b/elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
@@ -5,11 +5,15 @@
  */
 package com.yahoo.elide.security.executors;
 
+import static com.yahoo.elide.security.permissions.ExpressionResult.DEFERRED;
+import static com.yahoo.elide.security.permissions.ExpressionResult.FAIL;
+
 import com.yahoo.elide.annotation.DeletePermission;
 import com.yahoo.elide.annotation.ReadPermission;
 import com.yahoo.elide.annotation.SharePermission;
 import com.yahoo.elide.core.RequestScope;
 import com.yahoo.elide.core.exceptions.ForbiddenAccessException;
+import com.yahoo.elide.core.filter.expression.FilterExpression;
 import com.yahoo.elide.parsers.expression.CriterionExpressionVisitor;
 import com.yahoo.elide.security.ChangeSpec;
 import com.yahoo.elide.security.PermissionExecutor;
@@ -20,19 +24,19 @@ import com.yahoo.elide.security.permissions.ExpressionResultCache;
 import com.yahoo.elide.security.permissions.PermissionExpressionBuilder;
 import com.yahoo.elide.security.permissions.PermissionExpressionBuilder.Expressions;
 import com.yahoo.elide.security.permissions.expressions.Expression;
+
+import org.antlr.v4.runtime.tree.ParseTree;
+
 import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
-import org.antlr.v4.runtime.tree.ParseTree;
 
 import java.lang.annotation.Annotation;
+import java.util.Optional;
 import java.util.Queue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.function.BiFunction;
 import java.util.function.Function;
-
-import static com.yahoo.elide.security.permissions.ExpressionResult.DEFERRED;
-import static com.yahoo.elide.security.permissions.ExpressionResult.FAIL;
 
 /**
  * Default permission executor.
@@ -194,6 +198,21 @@ public class ActivePermissionExecutor implements PermissionExecutor {
                 requestScope
         );
         executeExpressions(expressions, annotationClass);
+    }
+
+    /**
+     * Get permission filter on an entity.
+     *
+     * @param resourceClass Resource class
+     */
+    public Optional<FilterExpression> getReadPermissionFilter(Class<?> resourceClass) {
+        if (requestScope.getSecurityMode() == SecurityMode.SECURITY_INACTIVE) {
+            return Optional.empty(); // Bypass
+        }
+        FilterExpression filterExpression =
+                expressionBuilder.buildAnyFieldFilterExpression(resourceClass, requestScope);
+
+        return Optional.ofNullable(filterExpression);
     }
 
     /**
--- a/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
+++ b/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
@@ -162,6 +162,32 @@ public class HibernateTransaction implements RequestScopedTransaction {
         }
     }
 
+    /**
+     * load a single record with id and filter.
+     *
+     * @param loadClass class of query object
+     * @param id id of the query object
+     * @param filterExpression FilterExpression contains the predicates
+     */
+    @Override
+    public <T> T loadObject(Class<T> loadClass, Serializable id, Optional<FilterExpression> filterExpression) {
+
+        try {
+            Criteria criteria = session.createCriteria(loadClass).add(Restrictions.idEq(id));
+            if (requestScope != null && isJoinQuery()) {
+                joinCriteria(criteria, loadClass);
+            }
+            if (filterExpression.isPresent()) {
+                CriterionFilterOperation filterOpn = new CriterionFilterOperation(criteria);
+                criteria = filterOpn.apply(filterExpression.get());
+            }
+            T record = (T) criteria.uniqueResult();
+            return record;
+        } catch (ObjectNotFoundException e) {
+            return null;
+        }
+    }
+
     @Deprecated
     @Override
     public <T> Iterable<T> loadObjects(Class<T> loadClass) {
@@ -172,7 +198,8 @@ public class HibernateTransaction implements RequestScopedTransaction {
     public <T> Iterable<T> loadObjects(Class<T> loadClass, FilterScope filterScope) {
         Criterion securityCriterion = filterScope.getCriterion(NOT, AND, OR);
 
-        Optional<FilterExpression> filterExpression = filterScope.getRequestScope().getLoadFilterExpression(loadClass);
+        Optional<FilterExpression> filterExpression =
+                filterScope.getRequestScope().getLoadFilterExpression(loadClass);
 
         Criteria criteria = session.createCriteria(loadClass);
         if (securityCriterion != null) {
@@ -188,7 +215,8 @@ public class HibernateTransaction implements RequestScopedTransaction {
     }
 
     @Override
-    public <T> Iterable<T> loadObjectsWithSortingAndPagination(Class<T> entityClass, FilterScope filterScope) {
+    public <T> Iterable<T> loadObjectsWithSortingAndPagination(Class<T> entityClass,
+                                                               FilterScope filterScope) {
         Criterion securityCriterion = filterScope.getCriterion(NOT, AND, OR);
 
         Optional<FilterExpression> filterExpression =
--- a/elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
+++ b/elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
@@ -122,6 +122,29 @@ public class HibernateTransaction implements DataStoreTransaction {
         }
     }
 
+    /**
+     * load a single record with id and filter.
+     *
+     * @param loadClass class of query object
+     * @param id id of the query object
+     * @param filterExpression FilterExpression contains the predicates
+     */
+    @Override
+    public <T> T loadObject(Class<T> loadClass, Serializable id,  Optional<FilterExpression> filterExpression) {
+
+        try {
+            Criteria criteria = session.createCriteria(loadClass).add(Restrictions.idEq(id));
+            if (filterExpression.isPresent()) {
+                CriterionFilterOperation filterOpn = new CriterionFilterOperation(criteria);
+                criteria = filterOpn.apply(filterExpression.get());
+            }
+            T record = (T) criteria.uniqueResult();
+            return record;
+        } catch (ObjectNotFoundException e) {
+            return null;
+        }
+    }
+
     @Override
     public <T> T loadObject(Class<T> loadClass, Serializable id) {
         try {
--- a/elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionCheckEvaluationVisitor.java
+++ b/elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionCheckEvaluationVisitor.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2016, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+
+package com.yahoo.elide.parsers.expression;
+
+
+import com.yahoo.elide.core.filter.Predicate;
+import com.yahoo.elide.core.filter.expression.AndFilterExpression;
+import com.yahoo.elide.core.filter.expression.FilterExpression;
+import com.yahoo.elide.core.filter.expression.NotFilterExpression;
+import com.yahoo.elide.core.filter.expression.OrFilterExpression;
+import com.yahoo.elide.core.filter.expression.Visitor;
+import com.yahoo.elide.security.FilterExpressionCheck;
+import com.yahoo.elide.security.RequestScope;
+
+/**
+ * FilterExpressionCheckEvaluationVisitor evaluate a check against fields of a returning object from datastore.
+ */
+
+public class FilterExpressionCheckEvaluationVisitor implements Visitor<Boolean> {
+    private final Object object;
+    private final FilterExpressionCheck filterExpressionCheck;
+    private final RequestScope requestScope;
+    public FilterExpressionCheckEvaluationVisitor(Object object,
+                                                  FilterExpressionCheck filterExpressionCheck,
+                                                  RequestScope requestScope) {
+        this.object = object;
+        this.filterExpressionCheck = filterExpressionCheck;
+        this.requestScope = requestScope;
+    }
+
+    @Override
+    public Boolean visitPredicate(Predicate predicate) {
+        return filterExpressionCheck.applyPredicateToObject(object, predicate, requestScope);
+    }
+
+    @Override
+    public Boolean visitAndExpression(AndFilterExpression expression) {
+        FilterExpression left = expression.getLeft();
+        FilterExpression right = expression.getRight();
+        return left.accept(this) && right.accept(this);
+    }
+
+    @Override
+    public Boolean visitOrExpression(OrFilterExpression expression) {
+        FilterExpression left = expression.getLeft();
+        FilterExpression right = expression.getRight();
+        return left.accept(this) || right.accept(this);
+    }
+
+    @Override
+    public Boolean visitNotExpression(NotFilterExpression expression) {
+        FilterExpression negation = expression.getNegated();
+        return !negation.accept(this);
+    }
+}
--- a/elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
+++ b/elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2016, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+
+package com.yahoo.elide.parsers.expression;
+
+import com.yahoo.elide.core.CheckInstantiator;
+import com.yahoo.elide.core.EntityDictionary;
+import com.yahoo.elide.core.filter.expression.AndFilterExpression;
+import com.yahoo.elide.core.filter.expression.FilterExpression;
+import com.yahoo.elide.core.filter.expression.OrFilterExpression;
+import com.yahoo.elide.core.filter.expression.Visitor;
+import com.yahoo.elide.generated.parsers.ExpressionBaseVisitor;
+import com.yahoo.elide.generated.parsers.ExpressionParser;
+import com.yahoo.elide.security.FilterExpressionCheck;
+import com.yahoo.elide.security.RequestScope;
+import com.yahoo.elide.security.checks.Check;
+import com.yahoo.elide.security.checks.UserCheck;
+
+
+/**
+ * PermissionToFilterExpressionVisitor parses a permission parseTree and returns the corresponding FilterExpression
+ * representation of it. This allows passing a security permission predicate down to datastore level to reduce
+ * in-memory permission verification workload.
+ * A few cases is not allow and will throw exception:
+ *      1. User define FilterExpressionCheck which returns null in getFilterExpression function.
+ *      2. User put a FilterExpressionCheck with a non-userCheck type check in OR relation.
+ */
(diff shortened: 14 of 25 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -35,6 +35,21 @@
...
+            if(other == null) {
+                return false;
+            }
+            if (!PathElement.class.isAssignableFrom(other.getClass())) {
+                return false;
+            }
+            final PathElement pe = (PathElement) other;
+            if(pe.type.equals(type) && pe.typeName.equals(typeName) && pe.fieldName.equals(fieldName) && pe.fieldType.equals(fieldType)) {
+                return true;
+            } else
```
clayreimann: Need space before `if`

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -35,6 +35,21 @@
...
+                return false;
+        }
     }
```
clayreimann: Need braces around return

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -75,4 +90,19 @@ public String toString() {
...
+        }
+        if (!Predicate.class.isAssignableFrom(other.getClass())) {
+            return false;
+        }
+        final Predicate predicate = (Predicate) other;
+        if(predicate.path.equals(path) && predicate.operator.equals(operator) && predicate.values.equals(values)) {
+            return true;
+        } else
+            return false;
+    }
```
clayreimann: Same style issues as above

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
```diff
@@ -29,4 +29,24 @@ public AndFilterExpression(FilterExpression left, FilterExpression right) {
...
+    @Override
+    public boolean equals(Object other){
+        if(other == null) {
+            return false;
+        }
+        if (!AndFilterExpression.class.isAssignableFrom(other.getClass())) {
+            return false;
+        }
+        final AndFilterExpression andExp = (AndFilterExpression) other;
+        if ((this.left == null) ? (andExp.left != null) : !this.left.equals(andExp.left)) {
```
clayreimann: Ternary expression in the if clause is too clever let's KISS this

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
```diff
@@ -29,4 +29,24 @@ public AndFilterExpression(FilterExpression left, FilterExpression right) {
...
+            return false;
+        }
+        if ((this.right == null) ? (andExp.right != null) : !this.right.equals(andExp.right)) {
```
clayreimann: Same

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/NotFilterExpression.java
```diff
@@ -27,4 +27,19 @@ public NotFilterExpression(FilterExpression negated) {
...
+    public boolean equals(Object other){
+        if(other == null) {
+            return false;
+        }
+        if (!NotFilterExpression.class.isAssignableFrom(other.getClass())) {
+            return false;
+        }
+        final NotFilterExpression negaExp = (NotFilterExpression) other;
+        if ((this.negated == null) ? (negaExp.negated != null) : !this.negated.equals(negaExp.negated)) {
+            return false;
```
clayreimann: Same as AndExpression

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/OrFilterExpression.java
```diff
@@ -30,4 +30,22 @@ public OrFilterExpression(FilterExpression left, FilterExpression right) {
...
+    @Override
+    public boolean equals(Object other){
+        if(other == null) {
+            return false;
+        }
+        if (!OrFilterExpression.class.isAssignableFrom(other.getClass())) {
+            return false;
+        }
+        final OrFilterExpression OrExp = (OrFilterExpression) other;
+        if ((this.left == null) ? (OrExp.left != null) : !this.left.equals(OrExp.left)) {
```
clayreimann: Same as AndExpression

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -35,6 +35,21 @@
...
+        @Override
+        public boolean equals(Object other){
+            if(other == null) {
+                return false;
+            }
+            if (!PathElement.class.isAssignableFrom(other.getClass())) {
+                return false;
+            }
+            final PathElement pe = (PathElement) other;
+            if(pe.type.equals(type) && pe.typeName.equals(typeName) && pe.fieldName.equals(fieldName) && pe.fieldType.equals(fieldType)) {
```
clayreimann: Need space before `if`

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+import com.yahoo.elide.generated.parsers.ExpressionParser;
+import com.yahoo.elide.security.FilterExpressionCheck;
+import com.yahoo.elide.security.RequestScope;
+import com.yahoo.elide.security.checks.Check;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/*
+FilterExpressionVisitor parses a permission parseTree and returns the corresponding Predicate representation of it.
```
clayreimann: This should be javadoc'ed.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+ */
+public class FilterExpressionVisitor extends ExpressionBaseVisitor<FilterExpression> implements CheckInstantiator {
+    private final EntityDictionary dictionary;
+    private final RequestScope requestScope;
+
+    public FilterExpressionVisitor(EntityDictionary dictionary,RequestScope requestScope) {
```
clayreimann: needs space after comma. It would be cleaner to do `@AllArgsConstructor` though.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+
+    @Override
+    public FilterExpression visitOR(ExpressionParser.ORContext ctx) {
+        return new OrFilterExpression(visit(ctx.left), visit(ctx.right));
+    }
+
+    @Override
+    public FilterExpression visitAND(ExpressionParser.ANDContext ctx) {
+        FilterExpression left = visit(ctx.left);
+        FilterExpression right = visit(ctx.right);
```
clayreimann: You should either make local variables like this or do things in line, but most importantly you should be consistent.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /*
+    Return FilterExpression representative of permission
+     */
+    @Override
+    public FilterExpression visitPermissionClass(ExpressionParser.PermissionClassContext ctx) {
+        //should call filterExpressionCheck here
+        Check check = getCheck(dictionary, ctx.getText());
+        if(FilterExpressionCheck.class.isAssignableFrom(check.getClass())) {
```
clayreimann: Space after `if`

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+            return ((FilterExpressionCheck) check).getFilterExpression(requestScope);
+        } else
```
clayreimann: braces around `else` block

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,74 @@
...
+            return null;
+    }
+
+}
```
clayreimann: should have trailing newline

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.yahoo.elide.core.filter.Predicate;
+import com.yahoo.elide.security.checks.InlineCheck;
+import java.util.Optional;
+
+public abstract class FilterExpressionCheck extends InlineCheck {
+
+    @Override
+    public boolean ok(Object object, RequestScope requestScope, Optional optional) {
+        return true;
```
clayreimann: Is this really the case, if so we should explain why this is.

clayreimann:
It would probably be beneficial to you to run the checkstyle locally before you open a PR.

elide-core/src/main/java/com/yahoo/elide/core/filter/Predicate.java
```diff
@@ -35,6 +35,34 @@
...
+            final PathElement pe = (PathElement) other;
+            if (pe.type.equals(type)
+                    && pe.typeName.equals(typeName)
+                    && pe.fieldName.equals(fieldName)
+                    && pe.fieldType.equals(fieldType)) {
+                return true;
+            } else {
+                return false;
+            }
+        }
```
xiaoyao1991: You can probably use `@EqualsAndHashCode` from lombok.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    @Override
+    public FilterExpression visitOR(ExpressionParser.ORContext ctx) {
+        return new OrFilterExpression(visit(ctx.left), visit(ctx.right));
+    }
+
+    @Override
+    public FilterExpression visitAND(ExpressionParser.ANDContext ctx) {
+        FilterExpression left = visit(ctx.left);
```
aklish: These cases need to handle nulls.  If one side is null and not the other, we shouldn't make this an AND or NOT.  Also need to handle the case where both sides are null.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,24 @@
...
+import com.yahoo.elide.security.checks.InlineCheck;
+import java.util.Optional;
+
+/**
+ * A check for filterExpression.
+ */
+public abstract class FilterExpressionCheck extends InlineCheck {
+
+    @Override
+    public boolean ok(Object object, RequestScope requestScope, Optional optional) {
```
aklish: You might want to add the type here for readability.

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
DeathByTape: Agree. Cute but maybe a little too cute :)

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,83 @@
...
+    public FilterExpression visitNOT(ExpressionParser.NOTContext ctx) {
+        return new NotFilterExpression(visit(ctx.expression()));
+    }
+
+    @Override
+    public FilterExpression visitOR(ExpressionParser.ORContext ctx) {
+        FilterExpression left = visit(ctx.left);
+        FilterExpression right = visit(ctx.right);
+        if (left == null) {
+            return right;
```
aklish: These should throw an exception.  We can't push the predicate to the datastore because it will filter out rows that otherwise might have evaluated to true.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        if (right == null) {
+            return left;
```
aklish: Same comment as above.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -197,6 +200,29 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
+     * @param <A> type parameter
+     */
+    public <A extends Annotation> FilterExpression getPermissionFilter(Class<?> resourceClass,
+                                                                       Class<A> annotationClass) {
+        if (requestScope.getSecurityMode() == SecurityMode.SECURITY_INACTIVE) {
+            return null; // Bypass
+        }
+        EntityDictionary dictionary = requestScope.getDictionary();
+        FilterExpressionVisitor fev = new FilterExpressionVisitor(dictionary, requestScope);
+        ParseTree classPermissions = dictionary.getPermissionsForClass(resourceClass, annotationClass);
```
aklish: I don't think this is correct.  Look at how read permissions works with 

expressions = expressionBuilder.buildAnyFieldExpressions(resource, annotationClass, changeSpec);

We have to look at more than just the class, but rather the union of the class and field permissions to evaluate.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -197,6 +200,29 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
 
     /**
+     * Get permission filter on an entity.
+     *
+     * @param resourceClass Resource class
+     * @param annotationClass Annotation class
+     * @param <A> type parameter
+     */
+    public <A extends Annotation> FilterExpression getPermissionFilter(Class<?> resourceClass,
+                                                                       Class<A> annotationClass) {
```
aklish: This will only apply for ReadPermission.  I don't think we need to pass annotationClass here.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,26 @@
...
+ * Check for FilterExpression
+ *
+ * @param <T> Type of class
+ */
+public abstract class FilterExpressionCheck<T> extends InlineCheck<T> {
+
+    public abstract FilterExpression getFilterExpression(RequestScope requestScope);
+
+    @Override
+    public boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
```
aklish: I wonder if this should evaluate the FilterExpression in memory returned from getFilterExpression.  Also, maybe make this final so the behavior can't be changed to something that would break assumptions.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
aklish: We also have to be careful here because ActivePermissionExecutor.getFilterExpression can return null if security is disabled.  We should probably throw an exception only if both one side of the OR returns null AND security is enabled.

elide-core/src/main/java/com/yahoo/elide/security/PermissionExecutor.java
```diff
@@ -95,6 +96,8 @@
      */
     <A extends Annotation> void checkUserPermissions(Class<?> resourceClass, Class<A> annotationClass);
 
+    <A extends Annotation> FilterExpression getPermissionFilter(Class<?> resourceClass, Class<A> annotationClass);
```
aklish: Maybe getReadPermissionFilter

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
aklish: We also have to be careful here because ActivePermissionExecutor.getFilterExpression can return null if security is disabled.  We should probably throw an exception only if both one side of the OR returns null AND security is enabled.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -90,6 +90,11 @@ default void preCommit() {
      */
     <T> T loadObject(Class<T> entityClass, Serializable id);
 
+    default <T> T loadObject(Class<T> entityClass, Serializable id, FilterScope filterScope) {
```
aklish: Instead of having FilterScope here, why not make this more forward looking and directly pass the FilterExpression.

elide-core/src/main/java/com/yahoo/elide/core/FilterScope.java
```diff
@@ -27,6 +29,23 @@ public FilterScope(RequestScope requestScope, Class<?> resourceClass) {
         this.permissions = requestScope.getDictionary().getPermissionsForClass(resourceClass, ReadPermission.class);
     }
 
+    /*
+     *This constructor with FilterExpression allows merging the filterExpression parsed from URL and the one from
+     *FilterExpressionCheck.
+     */
+    public FilterScope(RequestScope requestScope, Class<?> resourceClass, FilterExpression permissionExpression) {
```
aklish: You don't need a new constructor here.  The filterExpression can be derived from the PermissionExecutor which can be retrieved from the RequestScope.

elide-core/src/main/java/com/yahoo/elide/core/FilterScope.java
```diff
@@ -27,6 +29,23 @@ public FilterScope(RequestScope requestScope, Class<?> resourceClass) {
...
+        this.requestScope = requestScope;
+        this.permissions = requestScope.getDictionary().getPermissionsForClass(resourceClass, ReadPermission.class);
+        if (permissionExpression != null) {
+            if (this.requestScope.getGlobalFilterExpression() != null) {
+                this.requestScope.setGlobalFilterExpression(
+                        new AndFilterExpression(this.requestScope.getGlobalFilterExpression(), permissionExpression));
```
aklish: This logic might make more sense in the getGlobalFilterExpression function in RequestScope.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -300,6 +310,25 @@ public boolean matchesId(String checkId) {
...
     /**
+     * Get a FilterExpression parsed from FilterExpressionCheck.
+     *
+     * @param <T> the type parameter
+     * @param loadClass the load class
+     * @param PermissionClass the annotation class
+     * @param requestScope the request scope
+     * @return a FilterExpression defined by FilterExpressionCheck.
+     */
+    private static <T> FilterExpression getPermissionFilterExpression(Class<T> loadClass,
```
aklish: getReadPermissionFilterExpression.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -300,6 +310,25 @@ public boolean matchesId(String checkId) {
...
+                                                                      Class<? extends Annotation> PermissionClass,
```
aklish: I don't think we need the annotation param.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -256,8 +257,13 @@ public boolean matchesId(String checkId) {
...
         if (obj == null) {
             // try to load object
+            //filterExpressionCheck
+            FilterExpression permissionFilter = getPermissionFilterExpression(loadClass,
+                    ReadPermission.class,
+                    requestScope);
+            FilterScope filterScope = new FilterScope(requestScope, loadClass, permissionFilter);
             Class<?> idType = dictionary.getIdType(loadClass);
-            obj = tx.loadObject(loadClass, (Serializable) CoerceUtil.coerce(id, idType));
+            obj = tx.loadObject(loadClass, (Serializable) CoerceUtil.coerce(id, idType), filterScope);
```
aklish: This goes away if we pass the filterExpression directly.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -288,8 +294,12 @@ public boolean matchesId(String checkId) {
             return Collections.emptySet();
         }
 
+        //filterExpressionCheck
+        FilterExpression permissionFilter = getPermissionFilterExpression(loadClass,
+                ReadPermission.class, requestScope);
+
         Iterable<T> list;
-        FilterScope filterScope = new FilterScope(requestScope, loadClass);
+        FilterScope filterScope = new FilterScope(requestScope, loadClass, permissionFilter);
```
aklish: I believe this can be removed if we modify the RequestScope.getGlobalFilterExpression function.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -58,7 +59,7 @@
     private final Map<String, FilterExpression> expressionsByType;
 
     /* Used to filter across heterogeneous types during the first load */
-    private FilterExpression globalFilterExpression;
+    @Setter @Getter private FilterExpression globalFilterExpression;
```
aklish: I don't think this should have a setter.  There is already a method to get the globalFilterExpression.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -162,6 +162,38 @@ public void commit() {
...
+
+        try {
+            Optional<FilterExpression> filterExpression;
+            if (filterScope != null) {
+                filterExpression = filterScope.getRequestScope().getLoadFilterExpression(loadClass);
+            } else {
+                filterExpression = Optional.empty();
+            }
+            Criteria criteria = session.createCriteria(loadClass).add(Restrictions.idEq(id));
+            if (requestScope != null) {
```
aklish: joinCriteria shoudl only be called if isJoinQuery() is true.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,161 @@
...
+    public final boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
+        FilterExpression filterExpression = getFilterExpression(requestScope);
+        return applyFilterExpressionToObject(object, filterExpression);
+    }
+
+    private boolean applyFilterExpressionToObject(T object, FilterExpression filterExpression) {
+        if (filterExpression == null) {
+            return true;
+        }
+        if (AndFilterExpression.class.isAssignableFrom(filterExpression.getClass())) {
```
aklish: Generally not advised to use isAssignableFrom and instanceof when it can be avoided.   A visitor would be better here.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,161 @@
...
+            Object expectedValue = predicate.getValues().get(predicate.getValues().size() - 1);
+            Operator op = predicate.getOperator();
+            return handleByOperatorCases(op, fieldValue, expectedValue);
+        } catch (NoSuchFieldException e) {
+            return true;
+        }
+
+    }
+
+    private boolean handleByOperatorCases(Operator op, Object fieldValue, Object predicateConstrain) {
```
aklish: This should already be implemented inside the Operator.  You shouldn't have to reimplement the ability to evaluate predicates.

aklish: Go and look at InMemoryTransaction.filterCollection.  Almost everything here is already implemented in Elide - that is the starting place to look for how it works.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,134 @@
...
+import com.yahoo.elide.generated.parsers.ExpressionBaseVisitor;
+import com.yahoo.elide.generated.parsers.ExpressionParser;
+import com.yahoo.elide.security.FilterExpressionCheck;
+import com.yahoo.elide.security.RequestScope;
+import com.yahoo.elide.security.checks.Check;
+import com.yahoo.elide.security.checks.UserCheck;
+
+
+/**
+ * FilterExpressionVisitor parses a permission parseTree and returns the corresponding Predicate representation of it.
```
aklish: Provide a better explanation of the purpose of this class.

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/FilterExpression.java
```diff
@@ -10,4 +10,8 @@
  */
 public interface FilterExpression {
     public <T> T accept(Visitor<T> visitor);
+
+    default public boolean isDirty() {
```
aklish: I'm not sure what is meant here by the concept of 'dirty'?

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionCheckOkVisitor.java
```diff
@@ -0,0 +1,54 @@
...
+    public Boolean visitOrExpression(OrFilterExpression expression) {
+        FilterExpression left = expression.getLeft();
+        FilterExpression right = expression.getRight();
+        return left.accept(this) || right.accept(this);
+    }
+
+    @Override
+    public Boolean visitNotExpression(NotFilterExpression expression) {
+        FilterExpression negation = expression.getNegated();
+        return negation.accept(this);
```
aklish: Should this be return !negation.accept(this)?

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    public FilterExpression visitOR(ExpressionParser.ORContext ctx) {
+        FilterExpression left = visit(ctx.left);
+        FilterExpression right = visit(ctx.right);
+        if (left.isDirty() && right.isDirty()) {
+            //case where both side don't contain FilterExpressionCheck
+            return NO_EVALUATION_EXPRESSION;
+        }
+        if (left == NO_EVALUATION_EXPRESSION || right == NO_EVALUATION_EXPRESSION) {
+            throw new IllegalPermissionExpressionException(ctx.left.getText(), ctx.right.getText());
```
aklish: Just return NO_EVALUATION_EXPRESSION here - and let the caller of the visitor throw an exception if needed.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,130 @@
...
+        return new OrFilterExpression(left, right);
+    }
+
+    @Override
+    public FilterExpression visitPermissionClass(ExpressionParser.PermissionClassContext ctx) {
+        Check check = getCheck(dictionary, ctx.getText());
+        if (FilterExpressionCheck.class.isAssignableFrom(check.getClass())) {
+            FilterExpression filterExpression = ((FilterExpressionCheck) check).getFilterExpression(requestScope);
+            if (filterExpression == null) {
+                throw new IllegalPermissionExpressionException();
```
aklish: return NO_EVALUATION_EXPRESSION

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,130 @@
...
+    public FilterExpressionVisitor(EntityDictionary dictionary, RequestScope requestScope) {
+        this.dictionary = dictionary;
+        this.requestScope = requestScope;
+    }
+
+    @Override
+    public FilterExpression visitOR(ExpressionParser.ORContext ctx) {
+        FilterExpression left = visit(ctx.left);
+        FilterExpression right = visit(ctx.right);
+        if (left.isDirty() && right.isDirty()) {
```
aklish: I don't think we need this.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,130 @@
...
+            return NO_EVALUATION_EXPRESSION;
+        }
+    }
+
+    @Override
+    public FilterExpression visitAND(ExpressionParser.ANDContext ctx) {
+        FilterExpression left = visit(ctx.left);
+        FilterExpression right = visit(ctx.right);
+
+        if (left == NO_EVALUATION_EXPRESSION && right == USER_CHECK_EXPRESSION) {
```
aklish: I don't think we need this check.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,130 @@
...
+            return left;
+        }
+
+        if (right == NO_EVALUATION_EXPRESSION && left == USER_CHECK_EXPRESSION) {
```
aklish: I don't think we need this check.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,50 @@
...
+    public final boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
+        FilterExpression filterExpression = getFilterExpression(requestScope);
+        return filterExpression.accept(new FilterExpressionCheckOkVisitor(object, this));
+    }
+
+    public boolean applyPredicateToObject(T object, Predicate predicate) {
+        Predicate.PathElement path = predicate.getPath().get(predicate.getPath().size() - 1);
+        Class type = path.getType();
+        //If predicate is not related to object, vacuously true.
+        if (!type.isAssignableFrom(object.getClass())) {
```
aklish: I don't think you need to do this check.  We can safely assume predicates were built correctly given this is our own code.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,50 @@
...
+            return true;
+        }
+        String fieldName = path.getFieldName();
+        //same reason as above.
+        if (fieldName == null) {
+            return false;
+        }
+        EntityDictionary dictionary = new EntityDictionary();
+        dictionary.bindEntity(object.getClass());
```
aklish: We only bind entities at startup.  This should be removed.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,50 @@
...
+import java.util.Optional;
+
+/**
+ * Check for FilterExpression
+ *
+ * @param <T> Type of class
+ */
+public abstract class FilterExpressionCheck<T> extends InlineCheck<T> {
+
+    public abstract FilterExpression getFilterExpression(RequestScope requestScope);
```
aklish: This needs javadoc comments.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public final boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
+        FilterExpression filterExpression = getFilterExpression(requestScope);
+        return filterExpression.accept(new FilterExpressionCheckOkVisitor(object, this));
```
aklish: You can get the dictionary that this visitor needs by casting RequestScope to the core version of it.  Not ideal, but better than creating an empty dictionary below.

elide-core/src/main/java/com/yahoo/elide/core/exceptions/IllegalPermissionExpressionException.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ *
+ * Exception showing invalid combination of FilterExpressionCheck and other checks.
+ *
+ */
+public class IllegalPermissionExpressionException extends HttpStatusException {
+
+    public IllegalPermissionExpressionException(String left, String right) {
+        super(HttpStatus.SC_BAD_REQUEST, "Invalid Permission Expression: '" + left + "' with " + "'" + right + "'");
```
aklish: An invalid permission expression doesn't have anything to do with a bad client request, but rather bad server config.  We should probably just remove this exception and throw an IllegalStateException instead.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -197,6 +202,59 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
+        EntityDictionary entityDictionary = requestScope.getDictionary();
+        List<String> fields = entityDictionary.getAllFields(resourceClass);
+        for (String field : fields) {
+            ParseTree fieldPermissions =
+                    entityDictionary.getPermissionsForField(resourceClass, field, ReadPermission.class);
+            if (fieldPermissions == null) {
+                continue;
+            }
+            FilterExpression permissionFilter = fev.visit(fieldPermissions);
+            if (permissionFilter == null || permissionFilter.isDirty()) {
```
aklish: change to if (permissionFilter == NO_EXECUTION_EXPRESSION || permissionFilter == USER_EXPRESSION_FILTER)

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -197,6 +202,59 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
+        FilterExpression allFieldsExpression = null;
+        EntityDictionary entityDictionary = requestScope.getDictionary();
+        List<String> fields = entityDictionary.getAllFields(resourceClass);
+        for (String field : fields) {
+            ParseTree fieldPermissions =
+                    entityDictionary.getPermissionsForField(resourceClass, field, ReadPermission.class);
+            if (fieldPermissions == null) {
+                continue;
+            }
+            FilterExpression permissionFilter = fev.visit(fieldPermissions);
```
aklish: This whole logic might be cleaner if we added visit methods to Expression, and instead of creating individual visitors on the permissions per field, we create a single composite Expression for the entire ReadPermission - and then visit that.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
aklish: In general, we shouldn't be checking isDirty and null, we should be checking if the expression equals NO_EVALUATION_EXPRESSION or USER_CHECK_EXPRESSION.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -335,6 +336,22 @@ protected RequestScope(String path, JsonApiDocument jsonApiDocument, RequestScop
     }
 
     /**
+     * Merge the global filterExpression with the one provided.
+     * @param filterExpression the filterExpression to merge
+     * @return The global filter expression evaluated at the first load
+     */
+    public Optional<FilterExpression> addGlobalFilterExpression(FilterExpression filterExpression) {
```
aklish: Methods that begin with the word 'add' imply that they mutate the object.  This method doesn't do that.   In retrospect, this probably doesn't belong in RequestScope.

aklish: Check out getLoadFilterExpression in RequestScope.  This is the method that probably should be augmented to AND the globalFilterExpression with the filter expression from the active permission executor.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -288,6 +292,12 @@ public boolean matchesId(String checkId) {
             return Collections.emptySet();
         }
 
+        //filterExpressionCheck
+        FilterExpression permissionFilter = getPermissionFilterExpression(loadClass, requestScope);
```
aklish: I dont' think this code actually changes the load.  I think this change can be removed.  Look at how loadObjects is implemented in HibernateTransaction.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -315,6 +342,11 @@ public boolean matchesId(String checkId) {
...
```
aklish: I don't think this does anything.  We can remove.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
aklish: Can we add an example of why this check is needed.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
tarrantzhang: @aklish I looked for an existing function for getting an expression for all ReadPermission, and found `buildAnyFieldExpressions(final PersistentResource resource,
                                                                       final Class<A> annotationClass,
                                                                       final ChangeSpec changeSpec)` this one. The problem with this is that it needs a PersistentResource as param,  but it is difficult for me to access one in this context. 

Also the function calling `getFieldPermissionFilter(Class<?> resourceClass, FilterExpressionVisitor fev)` is `loadRecords(Class<T> loadClass, RequestScope requestScope)` which is static. So there is no way I can pass a PersistenceResource instance from the parent.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -118,6 +122,21 @@ default void preCommit() {
      * @param <T>         the type parameter
      * @param entityClass the entity class
      * @param filterScope scope for filter processing
+     * @param filterExpression expression for permission filters
+     * @return records iterable
+     */
+    default <T> Iterable<T> loadObjects(Class<T> entityClass,
+                                        FilterScope filterScope,
+                                        Optional<FilterExpression> filterExpression) {
```
aklish: Let's not push this to the DataStoreTransaction.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -126,6 +145,21 @@ default void preCommit() {
...
     /**
+     * Read entity records from database table with applied criteria.
+     *
+     * @param <T>         the type parameter
+     * @param entityClass the entity class
+     * @param filterScope scope for filter processing
+     * @param filterExpression expression for permission filters
+     * @return records iterable
+     */
+    default <T> Iterable<T> loadObjectsWithSortingAndPagination(Class<T> entityClass,
```
aklish: Same comment.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -300,6 +306,23 @@ public boolean matchesId(String checkId) {
...
+     * @param loadClass the load class
+     * @param requestScope the request scope
+     * @return a FilterExpression defined by FilterExpressionCheck.
+     */
+    private static <T> Optional<FilterExpression> getPermissionFilterExpression(Class<T> loadClass,
+                                                                      RequestScope requestScope) {
+        try {
+            return requestScope.getPermissionExecutor().getReadPermissionFilter(loadClass);
+        } catch (ForbiddenAccessException e) {
+            return null;
```
aklish: Instead of returning null, this should return Optional.empty()

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -288,6 +292,8 @@ public boolean matchesId(String checkId) {
             return Collections.emptySet();
         }
 
+        //filterExpressionCheck
+        //Optional<FilterExpression> permissionFilter = getPermissionFilterExpression(loadClass, requestScope);
```
aklish: Let's remove commented out code.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -315,6 +338,8 @@ public boolean matchesId(String checkId) {
             return Collections.emptySet();
         }
 
+        //filterExpressionCheck
```
aklish: Remove

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -327,11 +329,30 @@ protected RequestScope(String path, JsonApiDocument jsonApiDocument, RequestScop
      * @return The global filter expression evaluated at the first load
      */
     public Optional<FilterExpression> getLoadFilterExpression(Class<?> loadClass) {
-        if (globalFilterExpression == null) {
-            String typeName = dictionary.getJsonAliasFor(loadClass);
-            return getFilterExpressionByType(typeName);
+        Optional<FilterExpression> permissionFilter;
+        try {
+            permissionFilter = getPermissionExecutor().getReadPermissionFilter(loadClass);
+        } catch (ForbiddenAccessException e) {
```
aklish: This shouldn't throw ForbiddenAccessException because no permissions are being executed here.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -327,11 +329,30 @@ protected RequestScope(String path, JsonApiDocument jsonApiDocument, RequestScop
...
+            permissionFilter = Optional.empty();
+        }
+        if (!permissionFilter.isPresent()) {
+            if (globalFilterExpression == null) {
```
aklish: I would pull getting the global filter expression out of this if statement - so we don't duplicate this logic.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionCheckOkVisitor.java
```diff
@@ -0,0 +1,58 @@
...
+import com.yahoo.elide.core.filter.expression.OrFilterExpression;
+import com.yahoo.elide.core.filter.expression.Visitor;
+import com.yahoo.elide.security.FilterExpressionCheck;
+import com.yahoo.elide.security.RequestScope;
+
+/**
+ * FilterExpressionVisitor parses a permission parseTree and returns the corresponding Predicate representation of it.
+ */
+
+public class FilterExpressionCheckOkVisitor implements Visitor<Boolean> {
```
aklish: Let's rename this - maybe FilterExpressionCheckEvaluationVisitor

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * FilterExpressionVisitor parses a permission parseTree and returns the corresponding FilterExpression representation
+ * of it. This allows passing a security permission predicate down to datastore level to reduce in-memory permission
+ * verification workload.
+ * A few cases is not allow and will throw exception:
+ *      1. User define FilterExpressionCheck which returns null in getFilterExpression function.
+ *      2. User put a FilterExpressionCheck with a non-userCheck type check in OR relation.
+ */
+public class FilterExpressionVisitor extends ExpressionBaseVisitor<FilterExpression> implements CheckInstantiator {
```
aklish: Let's rename this to maybe PermissionToFilterExpressionVisitor

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public FilterExpression visitOR(ExpressionParser.ORContext ctx) {
+        FilterExpression left = visit(ctx.left);
+        FilterExpression right = visit(ctx.right);
+
+        if (left == NO_EVALUATION_EXPRESSION || right == NO_EVALUATION_EXPRESSION) {
+            return NO_EVALUATION_EXPRESSION;
+        }
+
+        if (left == USER_CHECK_EXPRESSION) {
```
aklish: I'm wondering if this is actually wrong.  I'm wondering if we should treat this the same as NO_EVALUATION_EXPRESSION.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,44 @@
...
+import com.yahoo.elide.core.EntityDictionary;
+import com.yahoo.elide.core.filter.Predicate;
+import com.yahoo.elide.core.filter.expression.FilterExpression;
+import com.yahoo.elide.parsers.expression.FilterExpressionCheckOkVisitor;
+import com.yahoo.elide.security.checks.InlineCheck;
+
+import java.util.Optional;
+
+/**
+ * Check for FilterExpression. This provide an super class for user defined FilterExpression check. The subclass should
```
aklish: Change 'This provide an super class' to 'This is a super class ...'

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -0,0 +1,44 @@
...
+ * override getFilterExpression function and return a FilterExpression which will be passed down to datastore.
+ * @param <T> Type of class
+ */
+public abstract class FilterExpressionCheck<T> extends InlineCheck<T> {
+
+    public abstract FilterExpression getFilterExpression(RequestScope requestScope);
+
+    @Override
+    public final boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
```
aklish: Add a comment that the filter expression is evaluated in memory if it cannot be pushed to the data store by elide for any reason.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
aklish: The function needs javadoc comments too.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -170,9 +196,16 @@ public void commit() {
 
     @Override
     public <T> Iterable<T> loadObjects(Class<T> loadClass, FilterScope filterScope) {
+        return loadObjects(loadClass, filterScope, Optional.empty());
+    }
+
+    @Override
+    public <T> Iterable<T> loadObjects(Class<T> loadClass, FilterScope filterScope,
```
aklish: We shoudl remove this change.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -189,10 +222,17 @@ public void commit() {
 
     @Override
     public <T> Iterable<T> loadObjectsWithSortingAndPagination(Class<T> entityClass, FilterScope filterScope) {
+        return loadObjectsWithSortingAndPagination(entityClass, filterScope, Optional.empty());
+    }
+
```
aklish: We should remove this change.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -189,10 +222,17 @@ public void commit() {
...
+    @Override
+    public <T> Iterable<T> loadObjectsWithSortingAndPagination(Class<T> entityClass,
+                                                               FilterScope filterScope,
+                                                               Optional<FilterExpression> permissionExpression) {
         Criterion securityCriterion = filterScope.getCriterion(NOT, AND, OR);
 
         Optional<FilterExpression> filterExpression =
-                filterScope.getRequestScope().getLoadFilterExpression(entityClass);
+                filterScope.getRequestScope().getLoadFilterExpression(entityClass, permissionExpression);
```
aklish: Same comment.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -144,9 +167,16 @@ public void commit() {
 
     @Override
     public <T> Iterable<T> loadObjects(Class<T> loadClass, FilterScope filterScope) {
+        return loadObjects(loadClass, filterScope, Optional.empty());
+    }
```
aklish: We should remove this change.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -144,9 +167,16 @@ public void commit() {
...
+
+    @Override
+    public <T> Iterable<T> loadObjects(Class<T> loadClass, FilterScope filterScope,
+                                       Optional<FilterExpression> permissionExpression) {
         Criterion securityCriterion = filterScope.getCriterion(NOT, AND, OR);
 
-        Optional<FilterExpression> filterExpression = filterScope.getRequestScope().getLoadFilterExpression(loadClass);
+        Optional<FilterExpression> filterExpression = filterScope.getRequestScope().getLoadFilterExpression(loadClass,
```
aklish: Same

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -163,10 +193,16 @@ public void commit() {
 
     @Override
     public <T> Iterable<T> loadObjectsWithSortingAndPagination(Class<T> entityClass, FilterScope filterScope) {
+        return loadObjectsWithSortingAndPagination(entityClass, filterScope, Optional.empty());
+    }
```
aklish: Same

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -163,10 +193,16 @@ public void commit() {
...
+
+    @Override
+    public <T> Iterable<T> loadObjectsWithSortingAndPagination(Class<T> entityClass, FilterScope filterScope,
+                                                               Optional<FilterExpression> permissionExpression) {
         Criterion securityCriterion = filterScope.getCriterion(NOT, AND, OR);
 
         Optional<FilterExpression> filterExpression =
-                filterScope.getRequestScope().getLoadFilterExpression(entityClass);
+                filterScope.getRequestScope().getLoadFilterExpression(entityClass, permissionExpression);
```
aklish: Same

aklish:
👍

PR #288, opened 2016-07-08
Fixing pagination, sorting and Integration tests

duseja2:


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -189,7 +189,8 @@ default void preCommit() {
             EntityDictionary dictionary,
             Optional<FilterExpression> filterExpression,
             Sorting sorting,
-            Pagination pagination
+            Pagination pagination,
+            String relationNameForSortAndPage
```
aklish: Instead of pushing this responsibility to the data store to differentiate when to paginate or when to not, there should be versions of getRelation - one with pagination and one without (or Pagination could be Optional).  The persistent resource should know when to call which interface.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -203,24 +204,28 @@ default void preCommit() {
             EntityDictionary dictionary,
             Set<Predicate> filters,
             Sorting sorting,
-            Pagination pagination
+            Pagination pagination,
+            String relationNameForSortAndPage
```
aklish: Same comment.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -380,4 +380,11 @@ public void saveObjects() {
     public boolean useFilterExpressions() {
         return useFilterExpressions;
     }
+
+    public String getQueriedCollectionRelationName() {
```
aklish: There are a couple of issues here:
1) A collection isn't always a 'relation'.  It could be a root collection.  The name of this function is misleading.
2) This function doesn't do the right thing if the path is not a collection but rather an element (/authors/1).

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
aklish: These are breaking changes.  We don't want to break the contract until Elide 3.0

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -18,48 +17,45 @@
...
+    public static final int DEFAULT_OFFSET = 0;
     public static final int DEFAULT_PAGE_SIZE = 500;
     public static final int MAX_PAGE_SIZE = 10000;
 
-    private static final Pagination DEFAULT_PAGINATION = new Pagination(0, 0);
+    private static final Pagination DEFAULT_PAGINATION = new Pagination(DEFAULT_OFFSET, DEFAULT_PAGE_SIZE);
 
     public static final Map<String, PaginationKey> PAGE_KEYS = new HashMap<>();
     static {
-        PAGE_KEYS.put("page[size]", PaginationKey.limit);
```
aklish: This breaks the API for anyone using pagination.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HQLTransaction.java
```diff
@@ -153,7 +153,7 @@ else if (filters != null) {
                     query = session.createFilter(collection, "");
                 }
                 query.setFirstResult(pagination.getOffset());
-                query.setMaxResults(pagination.getLimit());
+                query.setMaxResults(pagination.getSize());
```
aklish: Ideally, we shouldn't have to change the data stores.  I have some ideas for how this might work without requiring changes.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -914,9 +914,18 @@ protected boolean checkRelation(String relationName) {
 
         /* If elide was configured for Elide 3.0 data store interface */
         if (requestScope.useFilterExpressions()) {
-            val = requestScope.getTransaction()
-                    .getRelation(obj, type, relationName, relationClass, dictionary,
-                            filterExpression, requestScope.getSorting(), requestScope.getPagination());
+            String path = requestScope.getPath();
+            String relationNameForSortAndPage = path.substring(path.lastIndexOf("/") + 1);
+            if (relationName.equals(relationNameForSortAndPage)) {
```
aklish: I'm confused why we need these checks to extract the relationship from the path.  The state machine knows when pagination is needed and when it isn't.  I would assume the persistent resource would always call the non paginated version of getRelation.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HQLTransaction.java
```diff
@@ -129,7 +129,7 @@ else if (filters != null) {
             }
 
             // add sorting into HQL string query generation
-            if (!sortingRules.isEmpty()) {
+            if (sortingRules != null && !sortingRules.isEmpty()) {
```
aklish: Passing null is still a breaking change in some sense.  I don't understand why we need to change the DataStores at all (unless there was a bug).

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -18,48 +17,46 @@
...
-        return this.offset == 0 && this.limit == 0;
-    }
-
-    /**
-     * Alias for isDefault.
-     * @return true if there are no pagination rules
-     */
-    public boolean isEmpty() {
-        return isDefaultInstance();
+        return this.offset == DEFAULT_OFFSET && this.limit == DEFAULT_PAGE_LIMIT;
```
aklish: Is this method still used?

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -90,21 +87,32 @@ public static Pagination parseQueryParams(final MultivaluedMap<String, String> q
             return DEFAULT_PAGINATION;
         }
 
-        boolean isPageBased = pageData.containsKey(PaginationKey.page);
```
aklish: Should we verify pageData.size() <=2?

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -90,21 +87,32 @@ public static Pagination parseQueryParams(final MultivaluedMap<String, String> q
...
-            throw new InvalidValueException("page[size] and page[limit] must contain positive values.");
-        }
 
-        if (isPageBased && !pageData.containsKey(PaginationKey.offset)) {
-            int page = pageData.get(PaginationKey.page);
-            pageData.put(PaginationKey.offset, (page > 0 ? page - 1 : 0) * limit);
-        }
-        if (!pageData.containsKey(PaginationKey.offset)) {
-            pageData.put(PaginationKey.offset, 0);
+        if (pageData.containsKey(PaginationKey.limit) || pageData.containsKey(PaginationKey.offset)) {
```
aklish: Are combinations like limit and size legal?  Should we throw an error when specified?

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -90,21 +87,32 @@ public static Pagination parseQueryParams(final MultivaluedMap<String, String> q
...
+            int limit = pageData.containsKey(PaginationKey.limit)
+                    ? pageData.get(PaginationKey.limit) : DEFAULT_PAGE_LIMIT;
+            int offset = pageData.containsKey(PaginationKey.offset) ? pageData.get(PaginationKey.offset) : 0;
+            if (limit < 0 || offset < 0) {
+                throw new InvalidValueException("page[offset] and page[limit] must contain positive values.");
+            }
+            return new Pagination(offset, limit);
+        } else if (pageData.containsKey(PaginationKey.size)) {
+            // Page-based pagination strategy
+            int limit = Integer.min(pageData.get(PaginationKey.size), MAX_PAGE_LIMIT);
```
aklish: If the provide a size greater than MAX_PAGE_LIMIT, we should probably return a 400 error.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -206,10 +206,10 @@ public void commit() {
 
         final Pagination pagination = filterScope.hasPagination()
                 ? filterScope.getRequestScope().getPagination()
-                : null;
+                : Pagination.getDefaultPagination();
 
         // if we have sorting and sorting isn't empty, then we should pull dictionary to validate the sorting rules
-        Set<Order> validatedSortingRules = null;
+        List<Order> validatedSortingRules = null;
```
aklish: Why did we change to List?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -378,14 +378,11 @@ public T getObject() {
...
             Collection filteredVal = (Collection) val;
-            // sorting/pagination supported on last entity only eg /v1/author/1/books? books would be valid
-            final boolean hasSortRules = sorting.isDefaultInstance();
-            final boolean isPaginated = pagination.isDefaultInstance();
 
-            if ((filterExpression.isPresent() || hasSortRules || isPaginated)
-                    && (filteredVal instanceof AbstractPersistentCollection)) {
-                final Optional<Sorting> sortingRules = hasSortRules ? Optional.of(sorting) : Optional.empty();
-                final Optional<Pagination> paginationRules = isPaginated ? Optional.of(pagination) : Optional.empty();
+            // sorting/pagination supported on last entity only eg /v1/author/1/books? books would be valid
```
aklish: I'm not sure what this comment is about in the old code.  It looks wrong to me.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
duseja2: Changed it to preserve the ordering of the sorting rules. I'll change it to ordered set instead

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
duseja2: That would mean enforcing usage of only one of the pagination strategies. We'll do that

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -88,23 +76,65 @@ public static Pagination parseQueryParams(final MultivaluedMap<String, String> q
...
+            }
+
+            int computedOffset = pageNumber > 0 ? (pageNumber - 1) * limit : 0;
+            return new Pagination(computedOffset, limit);
+
+        } else if (isOffsetBased(pageData)) {
+
+            // Offset-based pagination strategy
+            int limit = pageData.containsKey(PaginationKey.limit)
+                    ? pageData.get(PaginationKey.limit) : DEFAULT_PAGE_LIMIT;
```
aklish: Above you check and make sure the limit <= MAX_PAGE_LIMIT (or else you throw an error).  In this clause you don't.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -88,23 +76,65 @@ public static Pagination parseQueryParams(final MultivaluedMap<String, String> q
...
+            // Contains page[number] but not page[size]
+            throw new InvalidValueException("page[size] parameter is required for page-based pagination.");
         }
-        // offset, page, limit
-        return new Pagination(pageData.get(PaginationKey.offset), limit);
+        return false;
+    }
+
+    private static boolean isOffsetBased(Map<PaginationKey, Integer> pageData) {
+        return  (pageData.containsKey(PaginationKey.limit) || pageData.containsKey(PaginationKey.offset));
```
aklish: What if this includes page[number]?  Shouldn't that return false?

elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java
```diff
@@ -91,8 +91,8 @@ public CollectionTerminalState(Class<?> entityClass, Optional<PersistentResource
 
     private Set<PersistentResource> getResourceCollection(RequestScope requestScope) {
         final Set<PersistentResource> collection;
-        final boolean hasSortingOrPagination = !requestScope.getPagination().isDefaultInstance()
-                || !requestScope.getSorting().isDefaultInstance();
+        final boolean hasSortingOrPagination = requestScope.getPagination() != null
```
aklish: Is this ever equal to null now that we always return DEFAULT instance?

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -180,7 +180,7 @@ public void commit() {
 
         final Pagination pagination = filterScope.hasPagination()
                 ? filterScope.getRequestScope().getPagination()
-                : null;
+                : Pagination.getDefaultPagination();
```
aklish: We need to be careful with loads.  If there was a filter join, pagination won't work and we'd be better off passing null (in the cases where there is no pagination defined at all).

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
duseja2: Good catch

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
duseja2: If that was the case, it would have thrown an exception in the call to isPageBased function

aklish: Hmm.  I know what you mean but I'm not sure if the function should live on its own or rely on the order in which another function is called.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -208,19 +209,12 @@ default void preCommit() {
...
-            final boolean isPaginated = pagination.isDefaultInstance();
-
-            if (!filters.isEmpty() || hasSortRules || isPaginated) {
-                final Optional<Sorting> sortingRules = hasSortRules ? Optional.of(sorting) : Optional.empty();
-                final Optional<Pagination> paginationRules = isPaginated ? Optional.of(pagination) : Optional.empty();
-                filteredVal = filterCollectionWithSortingAndPagination(filteredVal,
-                        relationClass, dictionary, Optional.of(filters), sortingRules, paginationRules);
-            }
+            Optional<Sorting> sortingRules = sorting != null ? Optional.of(sorting) : Optional.empty();
+            Optional<Pagination> paginationRules = pagination != null ? Optional.of(pagination) : Optional.empty();
```
clayreimann: Should use `Optional.ofNullable` here

elide-datastore/testng-audit.xml
```diff
@@ -5,10 +5,10 @@
   ~ See LICENSE file in project root for terms.
   -->
 <!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd">
-<suite name="Suite">
+<suite name="Audit Test Suite">
     <test name="com.yahoo.elide.tests.AuditIT">
         <classes>
```
clayreimann: Is there a reason not to use a package here too?

duseja2: There is only 1 class in this package. But, I'll change this to packages instead to pick up all the classes, just in case somebody adds more classes in this package and forgets to update this

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
duseja2: Good point

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -23,77 +22,119 @@
...
                 .forEach(paramEntry -> {
                     final String queryParamKey = paramEntry.getKey();
                     if (PAGE_KEYS.containsKey(queryParamKey)) {
-                        final String value = paramEntry.getValue().get(0);
-                        try {
-                            pageData.put(PAGE_KEYS.get(queryParamKey), Integer.parseInt(value, 10));
-                        } catch (NumberFormatException e) {
-                            throw new InvalidValueException("page values must be integers");
+                        PaginationKey paginationKey = PAGE_KEYS.get(queryParamKey);
+                        if (queryParamKey.equals(PAGE_TOTALS_KEY)) {
```
aklish: You can compare the enum value here instead of the string.

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -23,77 +22,119 @@
...
-        } else if (pageData.size() > 2) {
-            throw new InvalidValueException("Invalid usage of pagination parameters.");
-        }
+    /**
+     * Evaluates the pagination variables. Uses the Paginate annotation if it has been set for the entity to be
+     * queried.
+     *
+     * @param entityClass
+     */
+    public Pagination evaluate(final Class entityClass) {
```
aklish: Instead of allowing null here, why not have two functions - one which doesn't take a class and one which takes the default and max limits.

aklish: One thing I'm not sure about is that we will evaluate certain rules more than once.  Once parsing the query params, and then again when evaluate is called for a given entity class.  Maybe this could be split into two evaluation function.  One which validates limits (done per class), and one which evaluates the legal set of parameters - done at query param parsing time.

aklish:
👍

PR #289, opened 2016-07-11
Default loadObjectsWithSortingAndPagination calls correct loadObjects

wcekan:
When pagination was called the filter was lost in Multiplex transactions.


yahoocla:
CLA is valid!

xiaoyao1991:
👍

PR #296, opened 2016-07-22
Cache User permission result to avoid duplicate evaluations

wcekan:
Vastly improves processing of user permission checks for filter collection.
Ignores changeSpec for User checks allowing the cached results to be used.


yahoocla:
CLA is valid!

aklish:
Did this not help?

wcekan:
@aklish Turns out the duplication was calling User checks when a commit had a changeSpec

PR #297, opened 2016-07-25
Fix problems when FilterExpressionCheck OR UserCheck

tarrantzhang:
Support FilterExpressionCheck "OR" with UserCheck.  If UserCheck evaluate to true, ignore FilterExpression, otherwise push FilterExpression to datastore.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
+++ b/elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
@@ -39,7 +39,8 @@ public class PermissionToFilterExpressionVisitor extends ExpressionBaseVisitor<F
             return null;
         }
     };
-    public static final FilterExpression USER_CHECK_EXPRESSION = new FilterExpression() {
+
+    public static final FilterExpression FALSE_USER_CHECK_EXPRESSION = new FilterExpression() {
         @Override
         public <T> T accept(Visitor<T> visitor) {
             return null;
@@ -60,9 +61,10 @@ public class PermissionToFilterExpressionVisitor extends ExpressionBaseVisitor<F
             return NO_EVALUATION_EXPRESSION;
         }
 
-        //This special case require future reconsideration.
-        if (left == USER_CHECK_EXPRESSION || right == USER_CHECK_EXPRESSION) {
-            return USER_CHECK_EXPRESSION;
+        if (left == FALSE_USER_CHECK_EXPRESSION) {
+            return right;
+        } else if (right == FALSE_USER_CHECK_EXPRESSION) {
+            return left;
         }
 
         return new OrFilterExpression(left, right);
@@ -78,7 +80,8 @@ public class PermissionToFilterExpressionVisitor extends ExpressionBaseVisitor<F
             }
             return ((FilterExpressionCheck) check).getFilterExpression(requestScope);
         } else if (UserCheck.class.isAssignableFrom(check.getClass())) {
-            return USER_CHECK_EXPRESSION;
+            boolean userCheckResult = check.ok(requestScope.getUser());
+            return userCheckResult ? NO_EVALUATION_EXPRESSION : FALSE_USER_CHECK_EXPRESSION;
         } else {
             return NO_EVALUATION_EXPRESSION;
         }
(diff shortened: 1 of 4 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
```diff
@@ -39,7 +42,17 @@
             return null;
         }
     };
-    public static final FilterExpression USER_CHECK_EXPRESSION = new FilterExpression() {
+
+    /**
+     * A FilterExpression representing userCheck.
+     */
+    public class UserCheckExpression implements FilterExpression {
```
aklish: I'm wondering if we could keep the original USER_CHECK_EXPRESSION and do the following edits (see below):

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
```diff
@@ -78,7 +108,8 @@ public FilterExpression visitPermissionClass(ExpressionParser.PermissionClassCon
             }
             return ((FilterExpressionCheck) check).getFilterExpression(requestScope);
         } else if (UserCheck.class.isAssignableFrom(check.getClass())) {
-            return USER_CHECK_EXPRESSION;
+            boolean userCheckResult = check.ok(requestScope.getUser());
+            return new UserCheckExpression(userCheckResult);
```
aklish: if userCheckResult is true, return NO_EVALUATION_EXPRESSION.  Otherwise return USER_CHECK_EXPRESSION.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
```diff
@@ -60,9 +73,26 @@ public FilterExpression visitOR(ExpressionParser.ORContext ctx) {
             return NO_EVALUATION_EXPRESSION;
         }
 
-        //This special case require future reconsideration.
-        if (left == USER_CHECK_EXPRESSION || right == USER_CHECK_EXPRESSION) {
-            return USER_CHECK_EXPRESSION;
+        if (left instanceof UserCheckExpression || right instanceof UserCheckExpression) {
```
aklish: This would become:
```
if (left == FALSE_USER_CHECK_EXPRESSION) {
  return right
} else if (right == FALSE_USER_CHECK_EXPRESSION) {
  return left;
}
```

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
```diff
@@ -78,7 +108,8 @@ public FilterExpression visitPermissionClass(ExpressionParser.PermissionClassCon
             }
             return ((FilterExpressionCheck) check).getFilterExpression(requestScope);
         } else if (UserCheck.class.isAssignableFrom(check.getClass())) {
-            return USER_CHECK_EXPRESSION;
+            boolean userCheckResult = check.ok(requestScope.getUser());
```
aklish: I think this might be OK - BUT we have to be careful this doesn't break Xiaoyao's work to optimize user check expression evaluation.  That work uses the permission executor and has a new cache.  This would bypass that.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
```diff
@@ -91,19 +122,19 @@ public FilterExpression visitAND(ExpressionParser.ANDContext ctx) {
 
         // (NO_EVA and UserCheck) should return NO_EVA let the upper level know.
         // For example, FilterExpressionCheck Or (NO_EVA and UserCheck) should evaluate to  NO_EVALUATION_EXPRESSION.
-        if (left == NO_EVALUATION_EXPRESSION && right == USER_CHECK_EXPRESSION) {
+        if (left == NO_EVALUATION_EXPRESSION && right instanceof UserCheckExpression) {
```
aklish: This would change as well if we encounter FALSE_USER_CHECK_EXPRESSION.  

Anything AND FALSE_USER_CHECK_EXPRESSION should return FALSE_USER_CHECK_EXPRESSION.

elide-core/src/main/java/com/yahoo/elide/security/permissions/PermissionExpressionBuilder.java
```diff
@@ -351,10 +354,8 @@ private FilterExpression filterExpressionFromParseTree(ParseTree permissions, Re
         FilterExpression permissionFilter = new PermissionToFilterExpressionVisitor(entityDictionary,
                 requestScope).visit(permissions);
         //case where the permissions does not have ANY filterExpressionCheck
-        if (permissionFilter == NO_EVALUATION_EXPRESSION) {
-            return null;
-        }
-        if (permissionFilter == USER_CHECK_EXPRESSION) {
+        if (permissionFilter == NO_EVALUATION_EXPRESSION
+                || permissionFilter instanceof PermissionToFilterExpressionVisitor.UserCheckExpression) {
```
aklish: would change back to `|| permissionFilter == FALSE_USER_CHECK_EXPRESSION`

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
```diff
@@ -91,19 +94,19 @@ public FilterExpression visitAND(ExpressionParser.ANDContext ctx) {
 
         // (NO_EVA and UserCheck) should return NO_EVA let the upper level know.
         // For example, FilterExpressionCheck Or (NO_EVA and UserCheck) should evaluate to  NO_EVALUATION_EXPRESSION.
-        if (left == NO_EVALUATION_EXPRESSION && right == USER_CHECK_EXPRESSION) {
+        if (left == NO_EVALUATION_EXPRESSION && right == FALSE_USER_CHECK_EXPRESSION) {
```
aklish: I believe this simplifies now.  If either side is FALSE_USER_CHECK_EXPRESSION, the entire expression will evaluate to false - so we can just return FALSE_USER_CHECK_EXPRESSION here.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
tarrantzhang: This does bypass calling executor, but since we already have the check itself at this point, is it necessary to call the checkUserPermission with executor?

aklish:
👍

PR #299, opened 2016-08-02
Support sorting on related entity attributes in queries

mjvanleeuwen:
Elide currently only supports sorting on attributes of the main entity which is being queried.

This enhancement adds support for supporting on attributes of entities related to the main entity.

For example, given entities Author and Book, allow a query on "/book?sort=author.name,title" which will sort books firstly by author name, then by book title.

The sort join path can be of arbitrary length as long as the corresponding entity relationships exist in the data model.


yahoocla:

yahoocla:
CLA is valid!

aklish:
I haven't looked at the PR yet, but one thing to consider is that any join performed by the DataStore will almost certainly break pagination if the datastore performs it.  There is another PR about to go out that will determine whether the dataStore can correctly paginate or not (there are 2 cases where it can't - and a join is one such case).  If the datastore can't paginate correctly, Elide still can (in memory).  However, paginating in Elide for very large collections isn't super great.

elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/SortIT.groovy
```diff
@@ -280,49 +280,70 @@ class SortIT extends AbstractIntegrationTestInitializer {
...
         Assert.assertEquals(firstChapterBookId, "4")
-        final String firstChapterTitle = firstChapter.get("attributes").get("title").asText()
+        final String firstChapterTitle = firstChapter["attributes"]["title"].asText()
         Assert.assertEquals(firstChapterTitle, "Roast Mutton")
     }
 
+    @Test
+    public void testSortOnMultipleJoinEntityAttributesWithPagination() {
+        def result = mapper.readTree(RestAssured.get(
+                "/chapter?include=books,books.authors&sort=books.authors.name,books.title,title" +
```
aklish: I don't think this test fully captures the join problem.  If you have a parent/child, one to many relationship and query the children (joined with a single parent), the correct number or rows will be returned by the join - one per child.  In your case, you will have one per chapter.  Try inverting this test.  Instead of querying/paginating the children, query/paginate the parent.

For example, query and paginate /books but sort or filter on chapters.title.  It probably doesn't make much sense to sort in this manner, but the filter does make sense and illustrates the issue with joins and pagination. 

Here is a description of the problem in stack overflow:
[http://stackoverflow.com/questions/31581978/pagination-in-hibernate-with-joined-tables]

I'm torn on joins in general.  I think Elide is simpler only supporting subqueries (that's how relationship filters currently work).  Also, joins are very problematic for some DataStores (like key/values stores).  I think pushing down sorting to the DataStore (Hibernate) in your test makes sense because the join works in this case, but I'm not convinced it works in the general case.  I don't think Elide should just return incorrect results in those cases.  So the question remains what to do in that case.

aklish: One idea to address this problem:  Sorting and joins will work just fine assuming the relationship we are traversing is OneToOne or ManyToOne.  Elide could return an error in the other cases.

DeathByTape:
@mjvanleeuwen any updates? :)

aklish:
Elide already supports this capability.

Issue #294, opened 2016-07-21
Elide 3.0 DataStoreTransaction Changes


aklish:
Major changes:
1. `createObject` now takes the populated object instead of the class.
2. Every method should take a `RequestScope`
3. Removed redundant parameters that can be derived from the RequestScope
4. This solution bakes multiplexing into Elide instead of a separate data store transaction.  Or at least get/setRelation now take a `DataStoreTransaction` so they can use it to bridge between data stores
5. Using `Optional` for filters, sorting, and pagination
6. Remove unnecessary generics
7. Replace `precommit` with new hooks.

```
/**
 * Wraps the Database Transaction type.
 */
public interface DataStoreTransaction extends Closeable {

    /**
     * Wrap the opaque user.
     *
     * @param opaqueUser the opaque user
     * @return wrapped user context
     */
    default User accessUser(Object opaqueUser) {
        return new User(opaqueUser);
    }

    /**
     * Write any outstanding entities before processing response.
     */
    default void flush(RequestScope scope) {
    }

    /**
     * End the current transaction.
     */
    void commit(RequestScope scope);

    /**
     * Elide will create and populate the object with the attributes and relationships before
     * calling this method.  Operation security checks will be evaluated before invocation but commit
     * security checks will be called later immediately prior to `commit` being called.
     * @param entity - the object to create in the data store.
     * @param scope - contains request level metadata.
     */
    void createObject(Object entity, RequestScope scope) {}

    /**
     * Loads an object by ID.
     * @param id - the ID of the object to load.
     * @param filterExpression - security filters that can be evaluated in the data store.   It is optional for the data store to attempt evaluation.
     * @return the loaded object if it exists AND any provided security filters pass.
     */
    Object loadObject(Class<?> entityClass,
                     Serializable id,
                     Optional<FilterExpression> filterExpression,
                     RequestScope scope) {}

    /**
     * Loads a collection of objects.
     * @param filterExpression - filters that can be evaluated in the data store.   It is optional for the data store to attempt evaluation.
     * @param sorting - sorting which can be pushed down to the data store.
     * @param pagination - pagination which can be pushed down to the data store.
     * @param scope - contains request level metadata.
     * @returns the set of filtered objects.
     */
    Iterable<Object> loadObjects(
            Class<?> entityClass,
            Optional<FilterExpression> filterExpression,
            Optional<Sorting> sorting,
            Optional<Pagination> pagination,
            RequestScope scope) {}

    /**
     * Save the updated object.
[...]

PR #301, opened 2016-08-02
Change datastore interface-wip

tarrantzhang:
This PR dedicates to reimplement datastore interface in elide. [Details](https://github.com/yahoo/elide/issues/294).

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
@@ -5,14 +5,21 @@
  */
 package com.yahoo.elide.core;
 
+import com.yahoo.elide.core.exceptions.InvalidAttributeException;
 import com.yahoo.elide.core.filter.Predicate;
 import com.yahoo.elide.core.filter.expression.FilterExpression;
 import com.yahoo.elide.core.pagination.Pagination;
 import com.yahoo.elide.core.sort.Sorting;
 import com.yahoo.elide.security.User;
+import com.yahoo.elide.utils.coerce.CoerceUtil;
+
+import org.apache.commons.lang3.text.WordUtils;
 
 import java.io.Closeable;
 import java.io.Serializable;
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.util.Collection;
 import java.util.Optional;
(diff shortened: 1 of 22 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -71,64 +77,56 @@ default void preCommit() {
...
+     * @param entity - the object to create in the data store.
+     * @param scope - contains request level metadata.
      */
-    <T> T loadObject(Class<T> entityClass, Serializable id);
+    void createObject(Object entity, RequestScope scope);
 
-    default <T> T loadObject(Class<T> entityClass, Serializable id, Optional<FilterExpression> filterExpression) {
-        return loadObject(entityClass, id);
+
+    default <T> T createNewObject(Class<T> entityClass) {
```
aklish: Remove and do this inside the persistent resource.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -199,18 +197,97 @@ default void preCommit() {
...
             String relationName,
-            Class<T> relationClass,
-            EntityDictionary dictionary,
             Optional<FilterExpression> filterExpression,
-            Sorting sorting,
-            Pagination pagination
-    ) {
-        return PersistentResource.getValue(entity, relationName, dictionary);
-    }
+            Optional<Sorting> sorting,
```
aklish: The default method for this should call relationTx.load with a FilterExpression on the parent entity ID.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -199,18 +197,97 @@ default void preCommit() {
...
+     * @param scope - contains request level metadata.
+     */
+    default void setAttribute(
+            Object entity,
+            String attributeName,
+            Object attributeValue,
+            RequestScope scope) {
+        Class<?> entityClass = entity.getClass();
+        EntityDictionary dictionary = scope.getDictionary();
+        String type = dictionary.getJsonAliasFor(entity.getClass());
```
aklish: I would leave default method here empty.

PR #302, opened 2016-08-03
Shortcircuit user permission checks

xiaoyao1991:


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1661,9 +1665,22 @@ protected void addInverseRelation(String relationName, Object relationValue) {
...
+                    = resource.getRequestScope().getExpressionResultShortCircuit();
+            PermissionExecutor permissionExecutor = resource.getRequestScope().getPermissionExecutor();
+
             try {
                 if (!(skipNew && resource.getRequestScope().getNewResources().contains(resource))) {
-                    resource.checkFieldAwarePermissions(permission);
+                    if (!expressionResultShortCircuit
+                            .contains(Triple.of(permission, resource.getResourceClass(), null))) {
+                        permissionExecutor.checkUserPermissions(resource.getResourceClass(), permission);
+                        if (expressionResultShortCircuit
```
aklish: Instead of checking whether the `expressionResultShortCircuit` contains the item (which is relying on a side effect of `checkUserPermissions`, why not just check the return value of checkUserPermissions?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1685,9 +1702,24 @@ protected void addInverseRelation(String relationName, Object relationValue) {
...
                 if (checkIncludeSparseField(requestScope.getSparseFields(), type, field)) {
+                    if (expressionResultShortCircuit
+                            .contains(Triple.of(ReadPermission.class, getResourceClass(), field))) {
+                        filteredSet.add(field);
+                        continue;
+                    }
+
+                    requestScope.getPermissionExecutor().checkUserPermissions(this, ReadPermission.class, field);
+
+                    if (expressionResultShortCircuit
```
aklish: Same comment as above.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -58,6 +66,9 @@
     private final MultipleFilterDialect filterDialect;
     private final Map<String, FilterExpression> expressionsByType;
 
+    @Getter private final Map<String, Long> checkStats;
+    @Getter private final Set<Triple<Class, Class, String>> expressionResultShortCircuit;
```
aklish: You could be more specific about Class here - like Class extends Annotation for one of them.

aklish: I'm wondering if this belongs here since all we have is a getter.  I'm not sure where it ought to go.  Would it make more sense in the ActivePermissionExecutor?

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -398,4 +414,25 @@ public void saveObjects() {
...
+     * Print the permission check statistics
+     * @return the permission check statistics
+     */
+    public String printCheckStats() {
+        StringBuilder sb = new StringBuilder("Permission Check Statistics:\n");
+
+        Map<String, Long> result = new LinkedHashMap<>();
+        checkStats.entrySet().stream()
+                .sorted(Map.Entry.comparingByValue())
+                .forEachOrdered(e -> result.put(e.getKey(), e.getValue()));
```
aklish: Why don't you append to the string builder here instead of putting them in the map - and then iterating over that again?

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -48,6 +52,7 @@
 
     private final RequestScope requestScope;
     private final PermissionExpressionBuilder expressionBuilder;
+    private final Map<Triple<Class, Class, String>, ExpressionResult> userPermissionCheckCache;
```
aklish: One of the classes could be Class extends Annotation

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -90,64 +97,79 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
+                                                                                 ChangeSpec changeSpec,
+                                                                                 Class<A> annotationClass,
+                                                                                 String field) {
         if (requestScope.getSecurityMode() == SecurityMode.SECURITY_INACTIVE) {
-            return; // Bypass
+            return ExpressionResult.PASS; // Bypass
         }
+
+        ExpressionResult expressionResult = this.checkUserPermissions(resource, annotationClass, field);
+        if (requestScope.getExpressionResultShortCircuit()
```
aklish: It would be more natural to check for PASS here directly (by comparing with the expressionResult)

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -90,64 +97,79 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
+                                                                                         Class<A> annotationClass,
+                                                                                         String field) {
         if (requestScope.getSecurityMode() == SecurityMode.SECURITY_INACTIVE) {
-            return; // Bypass
+            return ExpressionResult.PASS; // Bypass
         }
+
+        // If the user check has already been evaluated before, return the result directly and save the building cost
+        ExpressionResult expressionResult = this.checkUserPermissions(resource, annotationClass, field);
+        if (requestScope.getExpressionResultShortCircuit()
```
aklish: Same comment as above.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -59,22 +64,24 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
      * @see com.yahoo.elide.annotation.ReadPermission
      * @see com.yahoo.elide.annotation.UpdatePermission
      * @see com.yahoo.elide.annotation.DeletePermission
      */
     @Override
-    public <A extends Annotation> void checkPermission(Class<A> annotationClass, PersistentResource resource) {
-        checkPermission(annotationClass, resource, null);
+    public <A extends Annotation> ExpressionResult checkPermission(
+            Class<A> annotationClass, PersistentResource resource) {
+        return checkPermission(annotationClass, resource, null);
```
aklish: Should we also short circuit here for user permissions as well?

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -90,64 +97,79 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
+            return ExpressionResult.PASS; // Bypass
         }
         Expressions expressions;
         if (SharePermission.class == annotationClass) {
             expressions = expressionBuilder.buildSharePermissionExpressions(resource);
         } else {
             expressions = expressionBuilder.buildAnyFieldExpressions(resource, annotationClass, changeSpec);
         }
-        executeExpressions(expressions, annotationClass);
+        return executeExpressions(expressions, annotationClass);
```
aklish: Same comment - should we also short circuit here?

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -268,11 +324,16 @@ public void executeCommitChecks() {
...
      */
-    private void executeExpressions(final Expressions expressions,
+    private ExpressionResult executeExpressions(final Expressions expressions,
                                     final Class<? extends Annotation> annotationClass) {
         Expression expression = expressions.getOperationExpression();
         ExpressionResult result = expression.evaluate();
-        if (result == DEFERRED) {
+
+        // Record the check
+        Long checkOccurrences = requestScope.getCheckStats().getOrDefault(expression.toString(), 0L) + 1;
```
aklish: Maybe checkstats should live in this class.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
xiaoyao1991: 👍

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
xiaoyao1991: Moving it(and the checkStats) to `ActivePermissionExecutor` should also require the `PermissionExecutor` to have the getter declaration. I'm not sure if it makes sense for `BypassPermissionExecutor` to have these getters.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
xiaoyao1991: Moving it(and the shortcircuit map) to `ActivePermissionExecutor` should also require the `PermissionExecutor` to have the getter declaration. I'm not sure if it makes sense for `BypassPermissionExecutor` to have these getters.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
xiaoyao1991: done in the subsequent method

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -90,64 +97,81 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
                                                        PersistentResource resource,
                                                        ChangeSpec changeSpec) {
         if (requestScope.getSecurityMode() == SecurityMode.SECURITY_INACTIVE) {
-            return; // Bypass
+            return ExpressionResult.PASS; // Bypass
         }
+
         Expressions expressions;
         if (SharePermission.class == annotationClass) {
             expressions = expressionBuilder.buildSharePermissionExpressions(resource);
```
xiaoyao1991: SharePermission seems to be a special case. If I bypass sharepermission check, [this shareable test](https://github.com/yahoo/elide/blob/master/elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/ShareableIT.groovy#L63-L76) will fail

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -158,46 +182,80 @@ public ActivePermissionExecutor(final com.yahoo.elide.core.RequestScope requestS
...
+                                                                        Class<A> annotationClass,
+                                                                        String field) {
         if (requestScope.getSecurityMode() == SecurityMode.SECURITY_INACTIVE) {
-            return; // Bypass
+            return ExpressionResult.PASS; // Bypass
+        }
+
+        // If the user check has already been evaluated before, return the result directly and save the building cost
+        ExpressionResult expressionResult
+                = userPermissionCheckCache.get(Triple.of(annotationClass, resource.getResourceClass(), field));
```
aklish: Does the building cost save much?  If this cache was removed, do we see an appreciable slowdown?

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
aklish: Instead of adding getters, why not just add a single method to the PermissionExecutor to check whether or not to short circuit.  This also hides the implementation details from PersistentResource.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
aklish: Any thoughts on this?  I'm gong to +1 regardless.

aklish:
👍

DeathByTape:
👍

PR #303, opened 2016-08-04
Implement new API

wcekan:
Add missing API calls to MultiplexTransaction and InMemoryTransaction


yahoocla:
CLA is valid!

clayreimann:
👍  Seems reasonable to me

PR #304, opened 2016-08-08
Proper handling of pagination

wcekan:
"Default" pagination was to return 0 rows per query


yahoocla:
CLA is valid!

xiaoyao1991:
👍

wcekan:
Will revisit and refactor pagination

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -227,8 +227,8 @@ default void preCommit() {
         Object val = PersistentResource.getValue(entity, relationName, dictionary);
         if (val instanceof Collection) {
             Collection filteredVal = (Collection) val;
-            Optional<Sorting> sortingRules = Optional.ofNullable(sorting);
-            Optional<Pagination> paginationRules = Optional.ofNullable(pagination);
+            Optional<Sorting> sortingRules = Optional.ofNullable(sorting.isDefaultInstance() ? null : sorting);
```
DeathByTape: We should talk to @aklish and @duseja2. I think the default instance just isn't behaving as intended.

duseja2:
Yeah, so the idea from what I think we should have:
1. Do not apply pagination if the @Paginate annotation is not put on a bean. There's unit tests for these.
2. For the one's where we have that annotation:
  a. If pagination related parameters are applied, we use those values, capped on max page limit.
  b. If no pagination parameters are specified, we apply default pagination (i.e applying default page limit with offset of 0)

PR #305, opened 2016-08-09
elide 3.0 createObject

tarrantzhang:
Implement new createObject interface in each datastore.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
@@ -72,14 +72,23 @@ public interface DataStoreTransaction extends Closeable {
     }
 
     /**
-     * Create new entity record.
+     * Create entity record of populated object.
      *
-     * @param <T>         the type parameter
-     * @param entityClass the entity class
-     * @return new record
+     * @param entity   the persisted Object
+     * @param scope  the request scope
      */
-    <T> T createObject(Class<T> entityClass);
+    void createObject(Object entity, RequestScope scope);
 
+
+    default <T> T createNewObject(Class<T> entityClass) {
+        T obj = null;
+        try {
+            obj = entityClass.newInstance();
+        } catch (java.lang.InstantiationException | IllegalAccessException e) {
+            //do nothing
+        }
+        return obj;
+    }
     /**
      * Read entity record from database table.
      *
@@ -195,7 +204,6 @@ public interface DataStoreTransaction extends Closeable {
             }
             return filteredVal;
         }
-
         return val;
     }
 
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -124,9 +124,10 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
             Class<T> entityClass,
             RequestScope requestScope,
             String uuid) {
-        DataStoreTransaction tx = requestScope.getTransaction();
 
-        T obj = tx.createObject(entityClass);
+        //instead of calling transcation.createObject, create the new object here.
+        T obj = requestScope.getTransaction().createNewObject(entityClass);
+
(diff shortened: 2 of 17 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -124,9 +124,10 @@ public String toString() {
             Class<T> entityClass,
             RequestScope requestScope,
             String uuid) {
-        DataStoreTransaction tx = requestScope.getTransaction();
 
-        T obj = tx.createObject(entityClass);
+        //instead of calling transcation.createObject, create the new object here.
+        T obj = requestScope.getTransaction().createNewObject(entityClass);
```
aklish: Can't we just use reflection here directly to create the new instance?

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -71,15 +71,33 @@ default void preCommit() {
...
-     * @param entityClass the entity class
-     * @return new record
+     * @param entity   the persisted Object
+     * @param scope  the request scope
      */
-    <T> T createObject(Class<T> entityClass);
+    void createObject(Object entity, RequestScope scope);
 
+
+    default <T> T createNewObject(Class<T> entityClass) {
```
aklish: Can we remove this method?

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -71,15 +71,33 @@ default void preCommit() {
...
     }
 
+//    /**
+//     * Create new entity record.
+//     *
+//     * @param <T>         the type parameter
+//     * @param entityClass the entity class
+//     * @return new record
+//     */
+//    <T> T createObject(Class<T> entityClass);
```
aklish: We can remove this dead code.

elide-datastore/elide-datastore-inmemorydb/src/main/java/com/yahoo/elide/datastores/inmemory/InMemoryTransaction.java
```diff
@@ -98,21 +100,16 @@ public void commit() {
...
-        try {
-            T instance = entityClass.newInstance();
-            setId(instance, id);
-            return instance;
-        } catch (InstantiationException | IllegalAccessException e) {
-            e.printStackTrace();
-        }
-        return null;
+        setId(entity, id);
+        operations.add(new Operation(id, entity, entity.getClass(), false));
```
aklish: What is this doing?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
wcekan: `T obj = (T) entityClass.newInstance();`

elide-datastore/elide-datastore-inmemorydb/src/main/java/com/yahoo/elide/datastores/inmemory/InMemoryTransaction.java
tarrantzhang: operations holds all changes made, and will execute each operation during commit. createObject() used to follow by save() which does this job, but now we don't call save() after it anymore (we create with populated object, and then commit directly). So we need to add this to operations.

elide-datastore/elide-datastore-inmemorydb/src/main/java/com/yahoo/elide/datastores/inmemory/InMemoryTransaction.java
```diff
@@ -57,7 +58,8 @@ public void save(Object object) {
         }
         String id = dictionary.getId(object);
         if (id.equals("0")) {
-            setId(object, dictionary.getId(createObject(object.getClass())));
+            createObject(object, null);
```
aklish: Let's see if this is still needed.  Let's throw an exception if id ever equals 0.

Issue #292, opened 2016-07-18
CoerceUtil cannot convert String/Long to Date field properly


xiaoyao1991:
`filter[author.creationDate][gt]=1454638927411` will receive an error complaining `"InvalidValueException: Invalid value: 1454638927411"`

DeathByTape:
This is typically done by adding custom jackson serializers. Does that approach not work here for some reason?

Koka:
@DeathByTape I'm afraid it's not true - almost everything is converted with `CoerceUtil` class wich isn't related to Jenkins at all and relies to Apache BeanUtils instead to perform conversion. To provide BeanUtils with new converter you have to do some ugly classloading tricks like this:

    private void hackElideDateParser() {
        try {
            Class.forName(CoerceUtil.class.getName());
        } catch (ClassNotFoundException e) {
            throw new RuntimeException(e.getMessage(), e);
        }
        DateConverter dateConverter = new DateConverter();
        dateConverter.setPatterns(new String[]{
                "yyyy-MM-dd",
                "yyyy-MM-dd'T'HH:mm:ss",
                "yyyy-MM-dd'T'HH:mm:ss.SSSX",
        });
        ConvertUtils.register(dateConverter, Date.class);
        ConvertUtils.register(new Base64Converter(), byte[].class);
    }

clayreimann:
@DeathByTape @Koka We should provide an easy way to customize how we serialize/deserialize data in elide

xiaoyao1991:
I'm gonna address this with a custom coerce converter like we did in https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/utils/coerce/converters/ToEnumConverter.java

DeathByTape:
@clayreimann this can be done via the jackson `ObjectMapper`. I have added a new constructor to make this more obvious: https://github.com/yahoo/elide/pull/317

We should also add this to the docs!

DeathByTape:
@xiaoyao1991 your change is merged. Can you or @Koka confirm that this fixes the issue?

I will be cutting a new release very shortly so it should push out to maven central.

PR #308, opened 2016-08-11
Fix coerce to date

xiaoyao1991:


yahoocla:
CLA is valid!

xiaoyao1991:
This should fix #292

DeathByTape:
LGTM. Should we add an integration test to make sure this doesn't get mangled somewhere else?

DeathByTape:
👍

xiaoyao1991:
@DeathByTape I added the integration tests, and more problem reveals. Turns out I can construct a date from long, but when it is serialized through hibernate, the value stored in db will carry an implicit timezone. For example, for timestamp `1454638927412 (2016-02-05 02:22:07 GMT)`, when I construct a `java.util.Date` out of it and store it in db, it will be cast with my local timezone and become `2016-02-04, 20:22:07`. The consequence of it will be: if I query the object by its original timestamp, this row will never return.

elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
```diff
@@ -70,6 +74,9 @@ public Converter lookup(Class<?> sourceType, Class<?> targetType) {
                     return TO_ENUM_CONVERTER;
                 } else if (Map.class.isAssignableFrom(sourceType)) {
                     return FROM_MAP_CONVERTER;
+                } else if ((String.class.isAssignableFrom(sourceType) || Long.class.isAssignableFrom(sourceType))
```
Koka: @xiaoyao1991  @DeathByTape  I've found a bug here. If you try to create entity with date field and set value to 1970-01-01 (timestamp is -14400000) it will be automactically converted into `Integer` object, not `Long` inside elide because the value is too small and it seems like the smallest allowed number type is selected on conversion. Then it will miss this if branch and wrong date converter will be chosen.

PR #307, opened 2016-08-11
Elide 3.0 add requestscope

tarrantzhang:
#294 Added Requestscope as param for datastoreTranscation functions.

PR #310, opened 2016-08-12
Elide 3.0 get/setRelation functions

tarrantzhang:
This PR added functions above #307 . Please ignore duplicated changes since #307 is not merged yet.

### DIFF
--- a/elide-annotations/src/main/java/com/yahoo/elide/annotation/OnCommit.java
+++ b/elide-annotations/src/main/java/com/yahoo/elide/annotation/OnCommit.java
@@ -14,6 +14,7 @@ import java.lang.annotation.Target;
 
 /**
  * On Commit trigger annotation.
+ * The onCommit invoked function takes a Requestscope as parameter.
  */
 @Target({METHOD, FIELD})
 @Retention(RUNTIME)
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -125,7 +125,7 @@ public class Elide {
                 ActivePermissionExecutor::new,
                 Collections.singletonList(new DefaultFilterDialect(dictionary)),
                 Collections.singletonList(new DefaultFilterDialect(dictionary)),
-                false
+                true
         );
     }
 
@@ -151,7 +151,7 @@ public class Elide {
             ActivePermissionExecutor::new,
             Collections.singletonList(new DefaultFilterDialect(dictionary)),
             Collections.singletonList(new DefaultFilterDialect(dictionary)),
-            false
+            true
         );
     }
 
@@ -247,7 +247,7 @@ public class Elide {
                     permissionExecutorFunction,
                     joinFilterDialects,
                     subqueryFilterDialects,
-                    useFilterExpressions);
+                    true);
         }
 
         @Deprecated
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -361,8 +361,7 @@ public class Elide {
                     queryParams,
                     securityMode,
                     permissionExecutor,
-                    new MultipleFilterDialect(joinFilterDialects, subqueryFilterDialects),
-                    useFilterExpressions);
+                    new MultipleFilterDialect(joinFilterDialects, subqueryFilterDialects));
 
             isVerbose = requestScope.getPermissionExecutor().isVerbose();
             GetVisitor visitor = new GetVisitor(requestScope);
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -79,7 +79,6 @@ public class Elide {
     private final Function<RequestScope, PermissionExecutor> permissionExecutor;
     private final List<JoinFilterDialect> joinFilterDialects;
     private final List<SubqueryFilterDialect> subqueryFilterDialects;
-    private final boolean useFilterExpressions;
 
     /**
      * Instantiates a new Elide.
@@ -124,8 +123,7 @@ public class Elide {
                 mapper,
                 ActivePermissionExecutor::new,
                 Collections.singletonList(new DefaultFilterDialect(dictionary)),
-                Collections.singletonList(new DefaultFilterDialect(dictionary)),
-                true
+                Collections.singletonList(new DefaultFilterDialect(dictionary))
         );
     }
 
@@ -150,8 +148,7 @@ public class Elide {
             mapper,
             ActivePermissionExecutor::new,
             Collections.singletonList(new DefaultFilterDialect(dictionary)),
-            Collections.singletonList(new DefaultFilterDialect(dictionary)),
-            true
+            Collections.singletonList(new DefaultFilterDialect(dictionary))
         );
     }
 
@@ -165,7 +162,6 @@ public class Elide {
      * @param permissionExecutor Custom permission executor implementation
      * @param joinFilterDialects A list of filter parsers to use for filtering across types
      * @param subqueryFilterDialects A list of filter parsers to use for filtering by type
-     * @param useFilterExpressions Whether or not to use Elide 3.0 filter expressions for DataStore interactions
      */
     protected Elide(AuditLogger auditLogger,
                   DataStore dataStore,
@@ -173,8 +169,7 @@ public class Elide {
                   JsonApiMapper mapper,
                   Function<RequestScope, PermissionExecutor> permissionExecutor,
                   List<JoinFilterDialect> joinFilterDialects,
-                  List<SubqueryFilterDialect> subqueryFilterDialects,
-                  boolean useFilterExpressions) {
+                  List<SubqueryFilterDialect> subqueryFilterDialects) {
         this.auditLogger = auditLogger;
         this.dataStore = dataStore;
         this.dictionary = dictionary;
@@ -183,7 +178,6 @@ public class Elide {
         this.permissionExecutor = permissionExecutor;
         this.joinFilterDialects = joinFilterDialects;
         this.subqueryFilterDialects = subqueryFilterDialects;
-        this.useFilterExpressions = useFilterExpressions;
     }
 
     /**
@@ -197,7 +191,6 @@ public class Elide {
         private Function<RequestScope, PermissionExecutor> permissionExecutorFunction = ActivePermissionExecutor::new;
         private List<JoinFilterDialect> joinFilterDialects;
         private List<SubqueryFilterDialect> subqueryFilterDialects;
-        private boolean useFilterExpressions;
 
         /**
          * A new builder used to generate Elide instances. Instantiates an {@link EntityDictionary} without
@@ -246,8 +239,7 @@ public class Elide {
                     jsonApiMapper,
                     permissionExecutorFunction,
                     joinFilterDialects,
-                    subqueryFilterDialects,
-                    true);
+                    subqueryFilterDialects);
         }
 
         @Deprecated
@@ -317,13 +309,11 @@ public class Elide {
         }
 
         public Builder withJoinFilterDialect(JoinFilterDialect dialect) {
-            useFilterExpressions = true;
             joinFilterDialects.add(dialect);
             return this;
         }
 
         public Builder withSubqueryFilterDialect(SubqueryFilterDialect dialect) {
-            useFilterExpressions = true;
             subqueryFilterDialects.add(dialect);
             return this;
         }
--- a/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
@@ -91,7 +91,6 @@ public interface DataStoreTransaction extends Closeable {
         return obj;
     }
 
-
     /**
      * Loads an object by ID.
      * @param id - the ID of the object to load.
--- a/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
@@ -168,6 +168,8 @@ public interface DataStoreTransaction extends Closeable {
 
 
     /**
+     * Elide core will update the in memory representation of the objects to the requested state.
+     * These functions allow a data store to optionally persist the relationship if needed.
      * @param relationTx - The datastore that governs objects of the relationhip's type.
      * @param entity - The object which owns the relationship.
      * @param relationName - name of the relationship.
@@ -183,6 +185,8 @@ public interface DataStoreTransaction extends Closeable {
                                       RequestScope scope) { };
 
     /**
+     * Elide core will update the in memory representation of the objects to the requested state.
+     * These functions allow a data store to optionally persist the relationship if needed.
      * @param relationTx - The datastore that governs objects of the relationhip's type.
      * @param entity - The object which owns the relationship.
      * @param relationName - name of the relationship.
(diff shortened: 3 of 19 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -125,7 +125,7 @@ public Elide(AuditLogger auditLogger, DataStore dataStore, EntityDictionary dict
                 ActivePermissionExecutor::new,
                 Collections.singletonList(new DefaultFilterDialect(dictionary)),
                 Collections.singletonList(new DefaultFilterDialect(dictionary)),
-                false
+                true
```
aklish: We can just remove this option.  It was temporary for Elide 2.0 to allow for FilterExpressions.

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -149,53 +149,33 @@ protected DataStoreTransaction getTransaction(Class<?> cls) {
...
-        return transaction.getRelation(entity, relationshipType, relationName,
-                relationClass, dictionary, filterExpression, sorting, pagination);
+    public Object getRelation(DataStoreTransaction relationTx,
+                              Object entity,
+                              String relationName,
+                              Optional<FilterExpression> filterExpression,
+                              Optional<Sorting> sorting,
+                              Optional<Pagination> pagination,
+                              RequestScope scope) {
+        return relationTx.getRelation(relationTx, entity, relationName, filterExpression, sorting, pagination, scope);
```
aklish: In this case, the Multiplex manager actually has to lookup the relationTx (because the PersistentResource doesn't do that).  It needs to locate the TX by the relationship class.

elide-annotations/src/main/java/com/yahoo/elide/annotation/OnCommit.java
```diff
@@ -14,6 +14,7 @@
 
 /**
  * On Commit trigger annotation.
+ * The onCommit invoked function takes a Requestscope as parameter.
```
wcekan: @aklish Should RequestScope be a required parameter, or should it be optional?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1355,7 +1318,8 @@ protected void nullValue(String fieldName, PersistentResource oldValue) {
      */
     protected Object getValueChecked(String fieldName) {
         checkFieldAwareDeferPermissions(ReadPermission.class, fieldName, (Object) null, (Object) null);
-        return getValue(getObject(), fieldName, dictionary);
+        Optional<FilterExpression> filterExpression = requestScope.getLoadFilterExpression(getObject().getClass());
```
aklish: I don't think filter expressions should be applied for getting attributes.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1365,7 +1329,8 @@ protected Object getValueChecked(String fieldName) {
      * @return Value
      */
     protected Object getValueUnchecked(String fieldName) {
-        return getValue(getObject(), fieldName, dictionary);
+        Optional<FilterExpression> filterExpression = requestScope.getLoadFilterExpression(getObject().getClass());
```
aklish: Same comment.

aklish:
I didn't see any logic for setAttribute and setRelation.  Is that in another PR?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateStore.java
```diff
@@ -116,7 +116,7 @@ public DataStoreTransaction beginTransaction() {
         Session session = sessionFactory.getCurrentSession();
         Preconditions.checkNotNull(session);
         session.beginTransaction();
-        return new HibernateTransaction(session);
+        return new HibernateTransaction(session, true, ScrollMode.FORWARD_ONLY);
```
aklish: What was this change for?

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -150,40 +141,52 @@ protected DataStoreTransaction getTransaction(Class<?> cls) {
...
+        com.yahoo.elide.core.RequestScope requestScope;
+        try {
+            requestScope  = (com.yahoo.elide.core.RequestScope) scope;
+        } catch (ClassCastException e) {
+            throw new ClassCastException("Fail trying to cast requestscope");
+        }
+        EntityDictionary dictionary = requestScope.getDictionary();
+        Class<?> relationClass = dictionary.getParameterizedType(entity, relationName);
+        DataStoreTransaction subTransaction = getTransaction(relationClass);
+        return relationTx.getRelation(subTransaction, entity,
```
aklish: This isn't quite right.  You need to get two different transactions here - the relation transaction and the one for the entity.  Then you would call: entityTransaction.getRelation(relationTransaction, ....)

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexWriteTransaction.java
```diff
@@ -191,44 +187,20 @@ public Object loadObject(Class<?> entityClass,
...
-        Object relation = transaction
-                .getRelation(entity, relationshipType, relationName, relationClass, dictionary, filters);
+    public Object getRelation(DataStoreTransaction relationTx,
+                              Object entity, String relationName,
+                              Optional<FilterExpression> filterExpression,
+                              Optional<Sorting> sorting,
+                              Optional<Pagination> pagination,
+                              RequestScope scope) {
+        Object relation = relationTx
+                .getRelation(relationTx, entity, relationName, filterExpression, sorting, pagination, scope);
```
aklish: This should work like MultiplexTransaction.getRelation.  I'm confused why this is even overridden.  Is it because of the hold?

elide-annotations/src/main/java/com/yahoo/elide/annotation/OnCommit.java
aklish: I think it should be required.  The implementer can ignore it if needed.

aklish: Hmmm.  I guess it could be optional.  That might be more elegant.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -807,4 +807,12 @@ public AccessibleObject getAccessibleObject(Class<?> targetClass, String fieldNa
...
         return fields;
     }
+
+    public boolean isRelation(Class<?> entityClass, String relationName) {
+        return getEntityBinding(entityClass).relationships.contains(relationName);
+    }
+
+    public boolean isAttribute(Class<?> entityClass, String attributeName) {
+        return getEntityBinding(entityClass).attributes.contains(attributeName);
+    }
```
tarrantzhang: These are functions for getting conditions for calling setAttribute/relation.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1495,10 +1475,17 @@ protected void setValue(String fieldName, Object value) {
...
         for (Method method : methods) {
             try {
-                method.invoke(obj);
+                method.invoke(obj, requestScope);
+            } catch (IllegalArgumentException exception) {
+                try {
+                    method.invoke(obj);
+                } catch (ReflectiveOperationException e) {
+                    throw new IllegalArgumentException(e);
+                }
```
tarrantzhang: I made these onCreate/Delete/Update.. hooks take requestScope Optionally.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -509,6 +508,12 @@ protected boolean updateToManyRelation(String fieldName,
                 });
 
 
+        //Hook for setRelation
+        Class<?> targetClass = obj.getClass();
+        if (dictionary.isRelation(targetClass, fieldName)) {
+            transaction.setRelation(transaction, obj, fieldName, collection, requestScope);
+        }
+
```
tarrantzhang: I feel like I should put the setRelation hook here, because only updating toOne relation calls `setValue()` function. Correct me if I'm wrong.

tarrantzhang: I'm not sure if the setrelation hook should also go into `addInverseRelation()`, since it's also updating the inverse relation.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -207,19 +149,20 @@ default Object getRelation(
...
             throw new ClassCastException("Fail trying to cast requestscope");
         }
+        EntityDictionary dictionary = requestScope.getDictionary();
+        Object val = PersistentResource.getValue(entity, relationName, dictionary);
+        if (val instanceof Collection) {
+            Collection filteredVal = (Collection) val;
+            return filteredVal;
+        }
+
+        return val;
```
aklish: On setRelation below, let's change the signature to be:
```
void updateToManyRelation(DataStoreTransaction relationTx, Object entity, String relationName, Set<Object> newRelationships, Set<Object> deletedRelationships, RequestScope scope)

void updateToOneRelation(DataStoreTransaction relationTx, Object entity, String relationName, Object relationshipValue, RequestScope scope)
```

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -207,55 +150,54 @@ default Object getRelation(
...
-     * @param relationValue - the desired contents of the relationship.
-     * @param scope - contains request level metadata.
-     */
-    default void setRelation(
-            DataStoreTransaction relationTx,
-            Object entity,
-            String relationName,
-            Object relationValue,
-            RequestScope scope) { }
+    default void updateToManyRelation(DataStoreTransaction relationTx,
```
aklish: We need some javadoc here.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1497,10 +1498,17 @@ protected void setValue(String fieldName, Object value) {
         Collection<Method> methods = dictionary.getTriggers(targetClass, annotationClass, fieldName);
         for (Method method : methods) {
             try {
-                method.invoke(obj);
+                method.invoke(obj, requestScope);
+            } catch (IllegalArgumentException exception) {
+                try {
+                    method.invoke(obj);
```
aklish: Instead of guess and check by invocation, you can actually look at the signature of the method and see if it takes a RequestScope.

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -150,40 +141,48 @@ protected DataStoreTransaction getTransaction(Class<?> cls) {
...
-        return transaction.getRelation(entity, relationshipType, relationName, relationClass, dictionary, filters);
+    public Object getRelation(DataStoreTransaction relationTx,
+                              Object entity,
+                              String relationName,
+                              Optional<FilterExpression> filterExpression,
+                              Optional<Sorting> sorting,
+                              Optional<Pagination> pagination,
+                              RequestScope scope) {
+        DataStoreTransaction entityTransaction = getTransaction(entity.getClass());
+        return entityTransaction.getRelation(relationTx, entity,
```
aklish: YOu have to lookup the relationTX as well here.  The passed in version will be the multiplex transaction.

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -150,40 +141,48 @@ protected DataStoreTransaction getTransaction(Class<?> cls) {
...
+    }
+
+    @Override
+    public void updateToManyRelation(DataStoreTransaction relationTx,
+                                     Object entity, String relationName,
+                                     Set<Object> newRelationships,
+                                     Set<Object> deletedRelationships,
+                                     RequestScope scope) {
+        DataStoreTransaction entityTransaction = getTransaction(entity.getClass());
+        entityTransaction.updateToManyRelation(relationTx, entity, relationName,
```
aklish: Same comment.

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -150,40 +141,48 @@ protected DataStoreTransaction getTransaction(Class<?> cls) {
...
+                newRelationships, deletedRelationships, scope);
+    }
+
+    @Override
+    public void updateToOneRelation(DataStoreTransaction relationTx, Object entity,
+                                    String relationName, Object relationshipValue, RequestScope scope) {
+        DataStoreTransaction entityTransaction = getTransaction(entity.getClass());
+        entityTransaction.updateToOneRelation(relationTx, entity, relationName,
```
aklish: Same comment.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1498,17 +1498,14 @@ protected void setValue(String fieldName, Object value) {
         Collection<Method> methods = dictionary.getTriggers(targetClass, annotationClass, fieldName);
         for (Method method : methods) {
             try {
-                method.invoke(obj, requestScope);
-            } catch (IllegalArgumentException exception) {
-                try {
+                if (method.getParameterCount() != 0) {
```
aklish: Should this be (==2)?

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -140,6 +141,18 @@ protected DataStoreTransaction getTransaction(Class<?> cls) {
         return transaction;
     }
 
+    protected DataStoreTransaction getRelationTransaction(Object object, String relationName, RequestScope scope) {
```
aklish: Why not just pass in the dictionary or something here.  THen you don't need to cast, etc.

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -140,6 +141,18 @@ protected DataStoreTransaction getTransaction(Class<?> cls) {
...
+        com.yahoo.elide.core.RequestScope requestScope;
+        try {
```
aklish: The only reason we do this cast is to get the dictionary.  However, it looks like the dictionary can be fetched from multiplexManager.getDictionary().

tarrantzhang:
@aklish .If this looks good, can I merge it to elide-3.0-wip ?

PR #312, opened 2016-08-13
Adding check whether data store can paginate based on execution of in…

aklish:
This is the first of three edits to make sure pagination doesn't return an incorrect result:

1) Disallow paginating collections where the security checks are not user checks or filter expression checks.  That is the subject of this PR.

2) Disallow paginating collections where a filter/sort resulted in a join across a ToMany collection.

3) If either (1) or (2) is true AND the pagination annotation says it is OK, paginate in memory.


yahoocla:
CLA is valid!

DeathByTape:
👍

PR #313, opened 2016-08-17
Allow to use dasherized resource names in filter expressions

Koka:
I'm using elide with hibernate datastore as backend for my ember-cli application and I stuck with one problem related to ember-data. By default ember-data uses dasherized resource names (like `my-very-own-resource`) and those names work well with elide by using `@Include(rootLevel = true, type = "my-very-own-resource")` annotation to resource class.

Unfortunately when I try to filter using basic syntax I have a problem with dasherized resource names. Filters like `/storage-items?filter[storage-items.manufacturer.id]=1` generate invalid SQL queries which contain table aliases like `storage-it1` and so on.

This pull request replaces dash sign in generated table aliases to underscore to make aliases valid database identifiers.


yahoocla:
CLA is valid!

clayreimann:
👍 LGTM, @wcekan: could you double check with your hibernate-nuance-knowing eyes

wcekan:
@clayreimann This should work fine.  Perhaps we would want the "alias"-ator to handle any invalid SQL characters.

aklish:
👍

aklish:
Might be good to add a comment to the line though explaining.  It is a bit cryptic.

PR #317, opened 2016-08-17
Added additional constructor for json api mapper.

DeathByTape:
This should make it more obvious on how to add custom (de)serializers to elide. Rather than requiring `jsonApiMapper.getMapper().registerModule(...)` or similar, you can now pass in your own pre-constructed `ObjectMapper` for Elide to use.


yahoocla:
CLA is valid!

DeathByTape:
NOTE: We need to add documentation about doing this properly.

aklish:
👍

PR #318, opened 2016-08-18
Avoid jcabi Mysql complaining too long socket path

xiaoyao1991:
Not sure if it's just me, but I got this error from jcabi on my laptop when building elide:
`The socket file path is too long (> 103): /Users/xiaoyaoqian/projects/elide/elide-datastore/elide-datastore-hibernate5/target/mysql-data/mysql.sock`

So I made this change to avoid it. Thought it might be helpful for others.


yahoocla:
CLA is valid!

tarrantzhang:
:+1:

PR #319, opened 2016-08-18
Don't set inverse relationships to null on cascading deletes

### LINKED ISSUE #300: Cascade delete is broken when Hibernate datastore is used and there are bidirectional relationships
I've tried to implement simple bidirectional relationship while using hibernate5 datastore implementation:

Parent class:

    .....
    @Entity
    public class Parent {
        ......
        private Set<Child> children;

        @OneToMany(mappedBy = "parent", cascade = CascadeType.ALL, orphanRemoval = true)
        public Set<Child> getChildren() {
            ....
        }
        .....
    }

Child class:

    .....
    @Entity
    public class Child {
        ......
        private Parent parent;

        @ManyToOne
        public Parent getParent() {
            ....
        }
        .....
    }

All works properly except of DELETE requests to /parent resource. I observe the following behavior (through hibernate SQL logger):

1. Hibernate deletes all child records (as cascade in my `@OneToMany` annotation requires)
2. Hibernate deletes parent record (because this is DELETE request)
3. Hibernate reinserts all child records and sets `parent_id` field to null (for unknown reason, and fails due to constraint violation in my database)

As the result I can't delete any parent record wich has at least one child - I get `423 Locked` error instead with cryptic hibernate related description.

I've searched code base and found this commit https://github.com/yahoo/elide/commit/7a9b9db34ef04cd2549e959f6ee4f3dd272bd981 and now I think this is intended behavior - someone forcefully sets inverse part of each relationship to null and marks those resources dirty, so (unfortunately) hibernate is trying to persist this change on commit ignoring the fact that those records were deleted actually.

Defenitely there are some conflict between intentions of this one commit and ability to do cascade deletes using hibernate datastore.

Could anyone please suggest any solution which will allow me to do cascade deletes using my REST API without the need to delete all child records first manually ?

aklish:
Fixes #300 

EntityDictionary Unit Tests

Added IT test

Finished unit tests

Added javadoc

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
@@ -13,15 +13,21 @@ import com.yahoo.elide.annotation.OnCreate;
 import com.yahoo.elide.annotation.OnDelete;
 import com.yahoo.elide.annotation.OnUpdate;
 import com.yahoo.elide.core.exceptions.DuplicateMappingException;
-
+import lombok.Getter;
+import lombok.Setter;
 import org.apache.commons.collections4.MultiValuedMap;
 import org.apache.commons.collections4.multimap.HashSetValuedHashMap;
 import org.apache.commons.lang3.text.WordUtils;
 import org.apache.commons.lang3.tuple.Pair;
 
-import lombok.Getter;
-import lombok.Setter;
-
+import javax.persistence.CascadeType;
+import javax.persistence.Column;
+import javax.persistence.Id;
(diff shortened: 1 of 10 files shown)


yahoocla:
CLA is valid!

asifmansoora:
👍

DeathByTape:
Will this limit _who_ can cascade? For instance, if I wanted hibernate to cascade `ALL`, I couldn't somehow tell Elide not to cascade, is that correct? Consequently, this would keep all these systems in lockstep making cascading difficult (if possible), right?

DeathByTape:
👍

Issue #321, opened 2016-08-19
JsonApiEndpoint problem with DELETEing resources


Koka:
While using elide with JAX-RS integration (using Resteasy as JAX-RS provider) I've found the following: currently when using JsonApiEndpoint you can't delete resource without specifying `Content-Type` header because `delete` method in this class expects request body.

According to [jsonapi spec](http://jsonapi.org/format/#crud-deleting) DELETE requests need body only when you delete relationship but not when you delete resource itself.

HTTP DELETE requests to resource URLs without body and Content-Type header, like the following one, must be valid (but unfortunately currently they doesn't work, will return 415 error):

```
DELETE /photos/1 HTTP/1.1
Accept: application/vnd.api+json
```

DeathByTape:
Updating from the PR: it looks like this is an implementation-specific issue with the RestEasy implementation of JAX-RS. In any case, we should work to resolve this issue for ease-of-use with any JAX-RS impl.

DeathByTape:
I believe this issue is complete with @Koka's PR. feel free to reopen if it's not complete.

PR #322, opened 2016-08-19
JsonApiEndpoint: Allow delete resource requests without body and content-type header

Koka:
Should fix issue described in https://github.com/yahoo/elide/issues/321 and allow HTTP DELETE requests without body and content-type header when using JsonApiEndpoint in JAX-RS environment

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/resources/JsonApiEndpoint.java
+++ b/elide-core/src/main/java/com/yahoo/elide/resources/JsonApiEndpoint.java
@@ -33,7 +33,6 @@ import javax.ws.rs.core.UriInfo;
  */
 @Singleton
 @Produces("application/vnd.api+json")
-@Consumes("application/vnd.api+json")
 @Path("/")
 public class JsonApiEndpoint {
     protected final Elide elide;
@@ -56,6 +55,7 @@ public class JsonApiEndpoint {
      */
     @POST
     @Path("{path:.*}")
+    @Consumes("application/vnd.api+json")
     public Response post(
         @PathParam("path") String path,
         @Context SecurityContext securityContext,
@@ -93,6 +93,7 @@ public class JsonApiEndpoint {
      */
     @PATCH
     @Path("{path:.*}")
+    @Consumes("application/vnd.api+json")
     public Response patch(
         @HeaderParam("Content-Type") String contentType,
         @HeaderParam("accept") String accept,
@@ -103,7 +104,7 @@ public class JsonApiEndpoint {
     }
 
     /**
-     * Delete handler.
+     * Delete relationship handler (expects body with resource ids and types).
      *
      * @param path request path
      * @param securityContext security context
@@ -112,11 +113,27 @@ public class JsonApiEndpoint {
      */
     @DELETE
     @Path("{path:.*}")
+    @Consumes("application/vnd.api+json")
     public Response delete(
         @PathParam("path") String path,
         @Context SecurityContext securityContext,
         String jsonApiDocument) {
         return build(elide.delete(path, jsonApiDocument, getUser.apply(securityContext)));
+    }
+
+    /**
+     * Delete resource handler (expects no body and no content-type header).
+     *
+     * @param path request path
+     * @param securityContext security context
+     * @return response
+     */
+    @DELETE
+    @Path("{path:.*}")
+    public Response deleteWithoutBody(
+            @PathParam("path") String path,
+            @Context SecurityContext securityContext) {
+        return build(elide.delete(path, null, getUser.apply(securityContext)));
     }
 
     private static Response build(ElideResponse response) {
--- a/elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
+++ b/elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
@@ -737,8 +737,8 @@ public class ResourceIT extends AbstractIntegrationTestInitializer {
 
     @Test(priority = 11)
     public void testDeleteParent() {
-        //Should work without content-type and body
         given()
+            .contentType(JSONAPI_CONTENT_TYPE)
             .accept(JSONAPI_CONTENT_TYPE)
             .delete("/parent/1")
             .then()
--- a/elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
+++ b/elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
@@ -737,8 +737,8 @@ public class ResourceIT extends AbstractIntegrationTestInitializer {
 
     @Test(priority = 11)
     public void testDeleteParent() {
+        //Should work without content-type and body
         given()
-            .contentType(JSONAPI_CONTENT_TYPE)
             .accept(JSONAPI_CONTENT_TYPE)
             .delete("/parent/1")
             .then()
--- a/elide-example/elide-blog-example-resteasy/README.md
+++ b/elide-example/elide-blog-example-resteasy/README.md
@@ -1,0 +1,112 @@
+# Elide Example
+
+## Usage
+
+1. Install and start a MySQL server
+
+2. Create ```elide``` database
+
+        mysql> create database elide;
+
+3. Create ```elide``` user with password ```elide123```
+
+        mysql> grant all on elide.* to 'elide'@'localhost' identified by 'elide123';
+
+4. Launch the example webservice
+
+        ~/elide $ mvn install
+        ~/elide $ cd elide-example/elide-blog-example
+        ~/elide/elide-example/elide-blog-example $ mvn exec:java -Dexec.mainClass="com.yahoo.elide.example.persistence.Main"
+
+5. Create an admin user
+
+        $ curl -H'Content-Type: application/vnd.api+json' \
+               -H'Accept: application/vnd.api+json' --data '
+          {
+            "data": {
+              "type": "user",
+              "attributes": {
+                "name": "Michael Admin",
+                "role": "Admin"
+              }
+            }
+          }
+          ' -X POST http://localhost:4080/user
+
+6. Create a registered user
+
+        $ curl -H'Content-Type: application/vnd.api+json' \
+               -H'Accept: application/vnd.api+json' --data '
+          {
+            "data": {
+              "type": "user",
+              "attributes": {
+                "name": "Ken Registered",
+                "role": "Registered"
+              }
+            }
+          }
+          ' -X POST http://localhost:4080/user
+
+7. Create an unregistered user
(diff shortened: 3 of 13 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/resources/JsonApiEndpoint.java
```diff
@@ -112,13 +113,29 @@ public Response patch(
...
+    /**
+     * Delete resource handler (expects no body and no content-type header).
+     *
+     * @param path request path
+     * @param securityContext security context
+     * @return response
+     */
+    @DELETE
+    @Path("{path:.*}")
+    public Response deleteWithoutBody(
```
clayreimann: Perhaps this should be `delete` and the other should be renamed to `deleteRelationships`? If that makes sense this would be better suited to going into the `elide3.0` branch.

Thoughts @DeathByTape?

Koka: @clayreimann Actually old `delete` method allows you delete resources too - not only relationships. The only difference is it requires mandatory content-type header and empty body in this case.

Also, I think it's very minor change - it doesn't  change existing behavior at all and just adds one more way to delete things. So I'm not sure if it deserves 3.0 branch.

clayreimann: @Koka I was thinking of changing the signatures to be:

```
@DELETE
@Path("{path:.*}")
public Response delete(
         @PathParam("path") String path,
          @Context SecurityContext securityContext)
``` and
```
@DELETE
@Path("{path:.*}")
@Consumes("application/vnd.api+json")
public Response deleteRelationship(
         @PathParam("path") String path,
          @Context SecurityContext securityContext,
          String jsonApiDocument)
```

Your fix is a minor change, changing the signatures like this a breaking change but is more explicit in the need for both versions. I'm not sure what our plans are for cutting more releases before 3.0. If you need #321 fixed before 3.0 drops then let's get this PR merged and clean things up in 3.0. Otherwise my vote is to do things right the first time.

Comment (Koka) [inline elide-core/src/main/java/com/yahoo/elide/resources/JsonApiEndpoint.java]: @clayreimann Fixing #321 isn't urgent because it can easily be fixed by subclasing JsonApiEndpoint in application and providing necessary DELETE method (actually it's what I've done in my app to continue work right now). So I don't insist on anything.

One thought is if we rename signatures in the way you propose - it would be good to ensure that deleteRelationship method can be used only on relationships URLs and delete method can be used only on resource URLs. Maybe some runtime checks or something. Without those checks new method names can be quite misleading because it will be still possible to delete resource using deleteRelationship method.

File: elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java [hunk 2]
```diff
@@ -737,8 +737,8 @@ public void testForbiddenDeleteEmptyCollectionRelationship() throws Exception {
 
     @Test(priority = 11)
     public void testDeleteParent() {
+        //Should work without content-type and body
         given()
```
DeathByTape: I think this may still send a body. Does this test case blow up without your new method?

elide-core/src/main/java/com/yahoo/elide/resources/JsonApiEndpoint.java
DeathByTape: @clayreimann I don't see any reason to entirely freeze any and all patches in waiting for elide 3.0 (at least until we work on releasing it full bore). Unless anyone else has different thoughts, I suspect we'll continue to patch elide 2.x for the near future :)

Anyway, to avoid confusion, can we try something like the following (untested, just an idea of some variation like this):

```
    @DELETE
    @Path("{path:.*}")
    @Consumes({"", "application/vnd.api+json"})
    public Response delete(
        @PathParam("path") String path,
        @Context SecurityContext securityContext,
        @DefaultValue("") String jsonApiDocument) {
        String doc = (jsonApiDocument.isEmpty()) ? null : jsonApiDocument;
        return build(elide.delete(path, doc, getUser.apply(securityContext)));
    }
```

elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
Koka: Without new method it will be 415 http error

elide-core/src/main/java/com/yahoo/elide/resources/JsonApiEndpoint.java
Koka: I'm not sure `@Consumes("")` is allowed by JAX-RS specification. Even if something like this will work in jersey it can be not supported by resteasy etc. 

Afaik the only way to tell jax-rs that you want no request body - is to omit consumes annotation and body method parameter.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
DeathByTape: @Koka unfortunately, I cannot reproduce this behavior. This test works fine for me with the existing code even after removing `contentType`. Are you using Jersey? If not, perhaps this is an issue with specific servlet containers. Either way, we should fix it, but I am interested in reproducing it since this test (from my end) doesn't appear to be behaving as we expect.

Similarly, I ran the [blog example](https://github.com/yahoo/elide/tree/master/elide-example/elide-blog-example) and was able to execute these two queries with success:

```curl -H'Content-Type: application/vnd.api+json' \
               -H'Accept: application/vnd.api+json' --data '
          {
            "data": {
              "type": "user",
              "attributes": {
                "name": "Michael Admin",
                "role": "Admin"
              }
            }
          }
          ' -X POST http://localhost:4080/user
```

immediately followed by:

```
curl -X DELETE http://localhost:4080/user/2
```

Can you reproduce that behavior on your end? We should make sure we're reproducing this correctly so we don't break it in the future.

Koka: @DeathByTape I'm using resteasy - not jersey in my project. I've checked this test again and I've found that it's my fault. This test actually passes with and without content-type header on both code versions - original and this one from pull request. Obviously we need new representative test to cover this case.

This bug is probably specific to resteasy JAX-RS implementation. I've created modified blog example using resteasy instead of jersey and `curl -X DELETE http://localhost:4080/user/2` fails with `415 Unsupported Media Type` with original elide-core code and deletes successfuly (`204 No Content`) with my changes form this pull request.

Koka:
@DeathByTape If we decide not to do any method renaming suggested by @clayreimann then it's ready.

Currently in this PR there is minimal non-breaking change required to make things work on Resteasy (and new example using resteasy).

DeathByTape:
:+1:

HassanAzzam:
Hi!
Coming from mid 2024, trying to delete a resource without a request body but always fails.

Elide version: 7.0.4
Response body: 
```json
{
    "type": "about:blank",
    "title": "Bad Request",
    "status": 400,
    "detail": "Failed to read request",
    "instance": "/api/ingredient/3160"
}
```
Server logs:
```txt
WARN  o.s.w.s.m.m.a.ExceptionHandlerExceptionResolver.logException - Resolved [org.springframework.http.converter.HttpMessageNotReadableException: Required request body is missing: public java.util.concurrent.Callable<org.springframework.http.ResponseEntity<java.lang.String>> com.yahoo.elide.spring.controllers.JsonApiController.elideDeleteRelation(org.springframework.http.HttpHeaders,org.springframework.util.MultiValueMap<java.lang.String, java.lang.String>,java.lang.String,jakarta.servlet.http.HttpServletRequest,org.springframework.security.core.Authentication)]```

**Comment by aklish** [comment]
Are you sure the request doesn't have a body?  The presence (or absence of
a body) is the sole arbitrator in the controller of whether we are deleting
the resource or relationship (before path parsing happens).  Can you
demonstrate this with curl?

**Comment by HassanAzzam** [comment]
That's my cURL:
```sh
curl 'http://localhost:8080/api/ingredient/3160' \
-X 'DELETE' \
-H 'Content-Type: application/vnd.api+json' \
-H 'Pragma: no-cache' \
-H 'Accept: application/vnd.api+json' \
-H 'Sec-Fetch-Site: same-site' \
-H 'Accept-Language: ar' \
-H 'Cache-Control: no-cache' \
-H 'Sec-Fetch-Mode: cors' \
-H 'Accept-Encoding: gzip, deflate' \
-H 'Origin: http://localhost:3000' \
-H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15' \
-H 'Referer: http://localhost:3000/' \
-H 'Content-Length: 0' \
-H 'Connection: keep-alive' \
-H 'Host: localhost:8080' \
-H 'Sec-Fetch-Dest: empty'
```

That's the result:
`{"type":"about:blank","title":"Bad Request","status":400,"detail":"Failed to read request","instance":"/api/ingredient/3160"}`

aklish:
Thanks.  Will try to reproduce.  It sounds like an issue in Spring boot as
that controller code hasn’t really changed and the wrong controller method
is being invoked.  What version of Spring boot are you using and what http
server?

HassanAzzam:
Ah, that could be the problem. I'm using Spring boot 3 😬

HassanAzzam:
> That's my cURL:
> 
> ```shell
> curl 'http://localhost:8080/api/ingredient/3160' \
> -X 'DELETE' \
> -H 'Content-Type: application/vnd.api+json' \
> -H 'Pragma: no-cache' \
> -H 'Accept: application/vnd.api+json' \
> -H 'Sec-Fetch-Site: same-site' \
> -H 'Accept-Language: ar' \
> -H 'Cache-Control: no-cache' \
> -H 'Sec-Fetch-Mode: cors' \
> -H 'Accept-Encoding: gzip, deflate' \
> -H 'Origin: http://localhost:3000' \
> -H 'User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15' \
> -H 'Referer: http://localhost:3000/' \
> -H 'Content-Length: 0' \
> -H 'Connection: keep-alive' \
> -H 'Host: localhost:8080' \
> -H 'Sec-Fetch-Dest: empty'
> ```
> 
> That's the result: `{"type":"about:blank","title":"Bad Request","status":400,"detail":"Failed to read request","instance":"/api/ingredient/3160"}`

Removing `Content-Type` header did the trick.

PR #324, opened 2016-08-23
Changes the rules about whether Elide can paginate to be less restric…

aklish:
…tive for user checks

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/parsers/expression/CanPaginateVisitor.java
+++ b/elide-core/src/main/java/com/yahoo/elide/parsers/expression/CanPaginateVisitor.java
@@ -127,67 +127,101 @@ import java.util.Set;
  *  - The filtered (2) result set will be returned
  *
  */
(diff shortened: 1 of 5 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/parsers/expression/CanPaginateVisitor.java
```diff
@@ -127,67 +127,101 @@
...
+    public CanPaginateVisitor(EntityDictionary dictionary, RequestScope scope) {
         this.dictionary = dictionary;
+        this.scope = scope;
     }
 
-
     @Override
-    public Boolean visitNOT(ExpressionParser.NOTContext ctx) {
-        return visit(ctx.expression());
+    public PaginationStatus visitNOT(ExpressionParser.NOTContext ctx) {
```
wcekan: Does this need to negate CAN_PAGINATE/CAN_PAGINATE as well?

DeathByTape:
👍

wcekan:
👍

PR #326, opened 2016-08-29
Operator field path evaluation

wcekan:
Required for FilterExpressionChecks that process predicates with relationships like
`this.author.book`

### DIFF
--- a/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/filter/CriterionFilterOperation.java
+++ b/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/filter/CriterionFilterOperation.java
@@ -96,8 +96,14 @@ public class CriterionFilterOperation implements FilterOperation<Criterion> {
 
         switch (predicate.getOperator()) {
             case IN:
+                if (predicate.getValues().isEmpty()) {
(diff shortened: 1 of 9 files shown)


yahoocla:
CLA is valid!

DeathByTape:
👍

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/filter/CriterionFilterOperation.java
```diff
@@ -117,6 +117,10 @@ public Criterion apply(Predicate predicate) {
                 return Restrictions.gt(alias, predicate.getValues().get(0));
             case GE:
                 return Restrictions.ge(alias, predicate.getValues().get(0));
+            case TRUE:
```
aklish: Should also cover the IN and NOT cases like for HQL.

wcekan: Done.

DeathByTape:
👍

aklish:
👍 But we still need to investigate why we need '.' separated paths instead of using the Predicate path.  Something isn't right.

aklish:
👍

DeathByTape:
👍

aklish:
👍

Issue #328, opened 2016-09-13
Fix Date Coercion


DeathByTape:
@Koka has found a bug in our date coercion. See his comment in https://github.com/yahoo/elide/pull/308 for details.

PR #329, opened 2016-09-14
Fix date coercion for non-long numeric types

Koka:
Fixes bug #328 by the simpliest available way

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
+++ b/elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
@@ -74,7 +74,7 @@ public class CoerceUtil {
                     return TO_ENUM_CONVERTER;
                 } else if (Map.class.isAssignableFrom(sourceType)) {
                     return FROM_MAP_CONVERTER;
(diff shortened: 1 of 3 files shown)


yahoocla:
CLA is valid!

clayreimann:
LGTM. Just one question, are we concerned that this also affects floating point numbers?

Koka:
I think case when someone requests to convert floating point number to date is pretty exotic.

DeathByTape:
@clayreimann I agree with @Koka. Maybe that's a bridge we should cross when (if) we get to it

👍

xiaoyao1991:
👍  Thanks @Koka. I'm still having trouble with the last comment I have on #308, which is causing https://github.com/yahoo/elide/pull/320 to not pass. Any idea on that or is it just me

Koka:
@xiaoyao1991 I've looked at latest travis log for your integration test #320  and I can see that there is no exceptions related to date parsing itself, but some logical errors in other tests:

```
  FilterIT.testPublishDateGreaterThanFilterSubRecord:1248 expected [true] but found [false]
  PaginateIT.testPageAndSortOnSubRecords:502 expected [1464638927412] but found [1464638927000]
```

Both failures look like someone is truncating milliseconds on Date fields somewhere.

PR #331, opened 2016-09-15
Check if Field Permission has in memory filter

wcekan:
Fix bug where field permissions were ignored when entity contained FilterExpression


yahoocla:
CLA is valid!

DeathByTape:
👍

aklish:
👍

PR #332, opened 2016-09-16
Short circuit OR and AND filter expressions

wcekan:
Will process OR and AND shortcuts when building the filterexpression so 
`where id = 1 and (false)` evaluates to `false` which immediately fails


yahoocla:
CLA is valid!

DeathByTape:
👍

DeathByTape:
👍

PR #333, opened 2016-09-19
Don't swallow cause for InvalidAttributeExceptions.

DeathByTape:


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidAttributeException.java
```diff
@@ -12,8 +12,15 @@
...
 public class InvalidAttributeException extends HttpStatusException {
+    public InvalidAttributeException(String attributeName, String type, Throwable cause) {
+        super(HttpStatus.SC_NOT_FOUND,
+                "Unknown attribute '" + attributeName + "' in " + "'" + type + "'",
+                null,
+                cause);
+    }
+
     public InvalidAttributeException(String attributeName, String type) {
-        super(HttpStatus.SC_NOT_FOUND, "Unknown attribute '" + attributeName + "' in " + "'" + type + "'");
```
wcekan: try
```java
this(attributeName, type, null);
```

PR #340, opened 2016-10-26
fix NPE in InMemoryDataStore

pettermahlen:
This change makes it possible to first add an entry without a generated id,
and then later to generate one. Before this, an NPE would be thrown
at the second step. It also somewhat decreases the risk of ID collision
when combining generated and user-specified IDs.


yahoocla:

yahoocla:
CLA is valid!

DeathByTape:
👍  LGTM. Thanks for this!

PR #341, opened 2016-10-26
Remove deprecated

DeathByTape:
Clean up dead code and remove deprecated methods/interfaces for Elide 3.0. Also upgrade all tests and examples to work properly with changes.


yahoocla:
CLA is valid!

elide-datastore/elide-datastore-inmemorydb/src/main/java/com/yahoo/elide/datastores/inmemory/InMemoryTransaction.java
```diff
@@ -162,32 +163,20 @@ public Object loadObject(Class<?> entityClass,
         if (objs == null) {
             return Collections.emptyList();
         }
+        // Support for filtering
+        if (filterExpression.isPresent()) {
+            Set<Predicate> predicates = filterExpression.get().accept(new PredicateExtractionVisitor());
```
aklish: I don't think this is quite what you want.  A predicate extraction visitor would collapse an expression like
`(A or B) and not C)` to a set of predicates `[A,B,C]` which in turn the code below treats like `(A and B and C)`.

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexWriteTransaction.java
```diff
@@ -194,46 +190,4 @@ public Object loadObject(Class<?> entityClass,
         DataStoreTransaction transaction = getTransaction(entityClass);
         return hold(transaction, transaction.loadObjects(entityClass, filterExpression, sorting, pagination, scope));
     }
-
-    @Override
-    public <T> Object getRelation(
```
aklish: Hmm.  We removed all the defaults.  I get that since these are deprecated, but I feel like there should be new versions of these functions that implement the getRelation functions that take FilterExpressions - for correctness.

elide-datastore/elide-datastore-hibernate5/src/test/java/com/yahoo/elide/datastores/hibernate5/HibernateDataStoreSupplier.java
```diff
@@ -24,6 +27,11 @@
 public class HibernateDataStoreSupplier implements Supplier<DataStore> {
     @Override
     public DataStore get() {
+        // Add additional checks to our static check mappings map.
+        // NOTE: This is a bit hacky. We need to do a major overhaul on our test architecture
+        TestCheckMappings.MAPPINGS.put("filterCheck", Filtered.FilterCheck.class);
```
aklish: Isn't CriteriaCheck also deprecated?  So any CriteriaCheck in our examples should probably be replaced with a FilterExpressionCheck (which is not tightly coupled to Hibernate in any way).

elide-datastore/elide-datastore-hibernate3/src/test/java/com/yahoo/elide/datastores/hibernate3/HibernateDataStoreSupplier.java
```diff
@@ -23,6 +26,11 @@
 public class HibernateDataStoreSupplier implements Supplier<DataStore> {
     @Override
     public DataStore get() {
+        // Add additional checks to our static check mappings map.
+        // NOTE: This is a bit hacky. We need to do a major overhaul on our test architecture
```
aklish: See my comment on the same code in Hibernate 5

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -377,6 +353,7 @@ public String printCheckStats() {
 
     @Override
     public boolean isVerbose() {
-        return requestScope.getSecurityMode() == SecurityMode.SECURITY_ACTIVE_VERBOSE;
+        //TODO: Should make this configurable?
+        return false;
```
aklish: Might be nice to preserve this somehow.

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
```diff
@@ -101,30 +101,15 @@ private void bindMemberPermissions(Map<String, ParseTree> fieldPermissions,
...
             boolean hasExpression = !expression.isEmpty();
 
-            boolean hasConfiguredChecks = hasAnyChecks || hasAllChecks || hasExpression;
-            boolean hasConfiguredOneChecks = hasAnyChecks ^ hasAllChecks ^ hasExpression;
-
-            if (!hasConfiguredChecks || !hasConfiguredOneChecks) {
+            if (!hasExpression) {
                 log.warn("Poorly configured permission: {} {}", annotationClass.getName(),
-                         hasConfiguredChecks ? "more than one set of checks specified" : "no checks specified.");
+                        hasExpression ? "more than one set of checks specified" : "no checks specified.");
```
clayreimann: This ternary is unnecessary now.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
clayreimann: This could be a constructor param/instance variable.

elide-core/src/test/java/com/yahoo/elide/security/PermissionExecutorTest.java
```diff
@@ -45,7 +44,7 @@ public void testSuccessfulOperationCheck() throws Exception {
     public void testFailOperationCheckAll() throws Exception {
         @Entity
         @Include
-        @UpdatePermission(all = {SampleOperationCheck.class, Role.NONE.class})
+        @UpdatePermission(expression = "sampleOperation OR deny all")
```
clayreimann: `sampleOperation AND deny all`

elide-core/src/test/java/com/yahoo/elide/security/PermissionExecutorTest.java
```diff
@@ -474,23 +465,23 @@ public boolean ok(Object object, com.yahoo.elide.security.RequestScope requestSc
...
-        @ReadPermission(all = {Role.NONE.class, SampleOperationCheck.class})
-        @UpdatePermission(all = {SampleCommitCheck.class, SampleOperationCheck.class})
+        @ReadPermission(expression = "deny all AND sampleOperation")
+        @UpdatePermission(expression = "sampleCommit AND sampleOperation")
         public String openAll = "all";
 
-        @ReadPermission(any = {SampleOperationCheck.class, SampleOperationCheck.class})
-        @UpdatePermission(any = {SampleCommitCheck.class, SampleOperationCheck.class})
+        @ReadPermission(expression = "sampleOperation OR sampleOperation")
+        @UpdatePermission(expression = "sampleCommit or sampleOperation")
```
clayreimann: `OR` (nit)

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -150,43 +139,6 @@ protected DataStoreTransaction getTransaction(Class<?> cls) {
     }
 
     @Override
-    public <T> Object getRelation(
```
clayreimann: I'm with Aaron, I think these need to stay. (unless I'm missing something obvious this change totally borks the MultiplexTransaction)

DeathByTape: Yeah, good catch. I must have been getting tired around here. This interface must never have been updated with the new code (this `getRelation` is the Elide 2.0 interface).

elide-datastore/elide-datastore-hibernate3/src/test/java/com/yahoo/elide/datastores/hibernate3/HibernateDataStoreSupplier.java
DeathByTape: See above

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -154,12 +158,22 @@ default Object getRelation(
         } catch (ClassCastException e) {
             throw new ClassCastException("Fail trying to cast requestscope");
         }
+
+        EntityDictionary dictionary = requestScope.getDictionary();
+        Object val = PersistentResource.getValue(entity, relationName, dictionary);
+        if (val instanceof Collection) {
```
aklish: just return val.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -187,35 +202,43 @@ default void updateToOneRelation(DataStoreTransaction relationTx,
...
-    default void setAttribute(
-            Object entity,
-            String attributeName,
-            Object attributeValue,
-            RequestScope scope) {
-    }
+    default void setAttribute(Object entity,
+                      String attributeName,
+                      Object attributeValue,
+                      RequestScope scope) { }
```
aklish: Hmm - should this have a default like getAttribute?

DeathByTape: Interesting question and I think the answer depends on cleanliness of code and separation of concern.

The way this is done right now, `PersistentResource` actually sets the in-memory value of the object. Doing it again here would be redundant. In the case of `InMemory` or `Hibernate` proxies, a noop here is effective.

elide-datastore/elide-datastore-inmemorydb/src/main/java/com/yahoo/elide/datastores/inmemory/InMemoryTransaction.java
```diff
@@ -186,4 +187,36 @@ public void close() throws IOException {
         ConcurrentHashMap<String, Object> objs = dataStore.get(entityClass);
         return objs == null ? 0L : objs.size();
     }
+
+    /**
+     * Visitor to retrieve a properly combined in-memory filter.
+     */
+    public class InMemoryFilter implements Visitor<java.util.function.Predicate> {
```
aklish: There are already two visitors which do something very similar.  Look at InMemoryFilterVisitor and to a lesser extent FilterExpressionCheckEvaluationVisitor

aklish:
👍

Issue #291, opened 2016-07-15
More usable extension points


clayreimann:
Elide currently provides several annotations and a DatastoreTransaction method that allow users to customize its runtime behavior. The current behavior looks like this:

```
--Start Request--
<@OnCreate> / <Load data>
*loop*
    <Update field>
    <@OnUpdate>
*end*
<@OnDelete>
<Request.preCommit>
<Check security>
<Commit transaction>
<@OnCommit>
--End Request--
```

This is not a useful model when:
- the `@OnCreate` relies on the object's values being present from the request
- the `@OnUpdate` hooks capture relationships between the object's attributes
- you want to fix-up an object before security is evaluated without rolling your own `Datastore` + `DatastoreTransaction`
- you want a chance to execute code after you know that the security checks have been successfully evaluated

A more useful set of lifecycle hooks (as it seems to me) would look like:

```
--Start Request--
<<Custom object instantiation>>
<Load data>
<Execute request payload>
<<Fix-up>>
<Check security>
<<Actions conditional on request success>>
<Commit transaction>
<<Final actions>>
```

Ideally all of these actions would be available directly from the beans, without having to create a custom `Datastore` or `DatastoreTransaction`; and these methods would have access to the request scope so they could determine what happened and/or load any additional resources they might require.

aklish:
Custom object instantiation is done today via the Initializer.  I'm not sure what you have in mind for the first step and how it will allow injection of objects to the lifecycle hooks.

DeathByTape:
After further discussion, I'm going to add an addendum to this proposal. I think this captures much of what we need, but as @aklish pointed out, it loses the spirit of its intuitive ties to CRUD. As a result, I propose we merge the two concepts as follows:


```
@PreCreate -> Executed on a create _before_ commit security checks are run, but after object has been fully populated with data (i.e. after all @OnUpdate hooks)
@PostCreate -> Executed on create _after_ commit security checks are run
@OnUpdate -> Executed on update _before_ operation security checks are run
@PreUpdate -> Executed on update _before_ commit security checks but after operation checks are run
@PostUpdate -> Executed on updated _after_ commit security checks are run
@OnDelete -> Executed on delete _before_ operation security checks
@PreDelete -> Executed on delete _before_ commit security checks
@PostDelete -> Executed on delete _after_ commit security checks
```

aklish:
👍

clayreimann:
👍

DeathByTape:
Done in #345

PR #345, opened 2016-11-01
Updated lifecycle triggers as proposed.

DeathByTape:
#First attempt at resolving: https://github.com/yahoo/elide/issues/291

Critical for Elide 3.0 usability.

Still up for discussion:

Should we have an `@OnDelete` hook? I currently removed it, but we could similarly pass the current HttpRequest method into the `RequestScope` and then `@OnDelete` could be covered by `@PostCommit` or similar. Thoughts?


yahoocla:
CLA is valid!

DeathByTape:
NOTE: I am leaning toward the second approach for handling `@OnDelete` since `OnDelete` doesn't really mesh well with our mental model of how these lifestyle hooks are laid out in this new form. Namely, we don't care what the action is (i.e. `GET`, `DELETE`, etc.) but instead we care about when we're executing security checks and updating the datastore. I think allowing the triggered code to perform based on method is more appropriate.

aklish:
There are things I like about the new annotations, but the lifecycle seems devoid of knowing what CRUD action was performed on the entity.  I'm not sure how well this new proposal will map to existing uses of the old annotations.

DeathByTape:
@aklish: see my comment. If we push this information down into the `RequestScope`, we can pass that information down to the developer.

elide-annotations/src/main/java/com/yahoo/elide/annotation/OnCreate.java
```diff
@@ -1,23 +1,23 @@
...
+ * The invoked function takes a RequestScope as parameter.
+ * @see com.yahoo.elide.security.RequestScope
  */
-@Target({METHOD, FIELD})
-@Retention(RUNTIME)
+@Target({ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
 public @interface OnCreate {
-
+    String value() default "";
```
aklish: Not sure we need a value.

elide-annotations/src/main/java/com/yahoo/elide/annotation/PostCreate.java
```diff
@@ -0,0 +1,25 @@
...
+ * This hook will be triggered <em>after</em> all security checks have been run and <em>after</em> the datastore
+ * has been committed.
+ *
+ * The invoked function takes a RequestScope as parameter.
+ * @see com.yahoo.elide.security.RequestScope
+ */
+@Target({ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface PostCreate {
+    String value() default "";
```
aklish: Do we need this?

elide-annotations/src/main/java/com/yahoo/elide/annotation/PreCreate.java
```diff
@@ -0,0 +1,25 @@
...
+ * This hook will be triggered <em>after</em> all security checks have been run, but <em>before</em> the datastore
+ * has been committed.
+ *
+ * The invoked function takes a RequestScope as parameter.
+ * @see com.yahoo.elide.security.RequestScope
+ */
+@Target({ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface PreCreate {
+    String value() default "";
```
aklish: Do we need this?

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -261,12 +261,14 @@ public ElideResponse get(
             GetVisitor visitor = new GetVisitor(requestScope);
             Supplier<Pair<Integer, JsonNode>> responder = visitor.visit(parse(path));
             transaction.preCommit();
+            requestScope.runQueuedOnTriggers();
```
aklish: Do we have the order of 'on' and 'pre' swapped here?

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -261,12 +261,14 @@ public ElideResponse get(
...
             requestScope.getPermissionExecutor().executeCommitChecks();
             transaction.flush(requestScope);
             ElideResponse response = buildResponse(responder.get());
+            requestScope.runQueuedPreTriggers();
             auditLogger.commit();
             transaction.commit(requestScope);
-            requestScope.runCommitTriggers();
+            requestScope.runQueuedPostTriggers();
```
aklish: Should post run immediately before or after commit?

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -353,21 +374,87 @@ else if (permissionFilter.isPresent()) {
...
-        commitTriggers.add(() -> resource.runTriggers(OnCommit.class, fieldName));
+    /**
+     * Queue triggers for a particular resource and its field.
+     *
+     * @param resource Resource on which to execute trigger
+     * @param fieldName Field name for which to specify trigger
+     * @param crudAction CRUD Action
+     */
+    protected void queueTriggers(PersistentResource resource, String fieldName, CRUDAction crudAction) {
+        Consumer<Class> queue = (cls) -> queuedTriggers.get(cls).add(() -> resource.runTriggers(cls, fieldName));
```
clayreimann: `querer` perhaps? `triggerQueuer`? `triggerFinder`?

clayreimann:
👍

PR #346, opened 2016-11-02
Isolate trace handling for performance

wcekan:
Found overhead in trace handling that should be skipped for normal processing.


yahoocla:
CLA is valid!

DeathByTape:
👍

aklish:
👍

PR #347, opened 2016-11-07
Noop store

DeathByTape:
Adds a noop store which aids in performing actions that occur out-of-bounds or do not require persisting to any sort of datastore (in-memory, on disk, or otherwise).


yahoocla:
CLA is valid!

aklish:
👍

PR #348, opened 2016-11-08
Enable filter expressions in Elide 2 RequestScope

wcekan:
Deprecated a number of old RequestScope constructors and fixed the Elide 2.0 instantiations.  This had been effectively disabling FilterExpression use in Elide 2.

There may still be a problem in Elide 2 where `any` will instead processes as `all` when FilterExpressions are used but not enabled.  So be sure to enable filter expressions.  This problem will go away entirely in Elide 3.


yahoocla:
CLA is valid!

aklish:
👍

DeathByTape:
👍

PR #350, opened 2016-11-21
Restrict ANY permission to sparse fields

wcekan:
PermissionExpressionBuilder included all fields in ANY even when sparse field set exists.


yahoocla:
CLA is valid!

DeathByTape:
👍

aklish:
:+1:

PR #354, opened 2016-12-02
Handle RequestScopedTransaction transactions

wcekan:
Process mutiplexed RequestScopedTransactions

### DIFF
--- a/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
+++ b/elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
@@ -15,15 +15,18 @@ import com.yahoo.elide.core.RequestScopedTransaction;
 import com.yahoo.elide.core.exceptions.ForbiddenAccessException;
 import com.yahoo.elide.core.exceptions.TransactionException;
 import com.yahoo.elide.core.filter.HQLFilterOperation;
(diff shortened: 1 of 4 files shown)


yahoocla:
CLA is valid!

clayreimann:
LGTM 👍

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -64,6 +67,7 @@
     private final boolean isScrollEnabled;
     private final ScrollMode scrollMode;
     @Getter(value = AccessLevel.PROTECTED)
+    @Setter
```
clayreimann: Does this also want `AccessLevel` (or is it inherited from the getter)?

aklish:
Let's add a test for it though.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
wcekan: Setter must be public, which is the default AccessLevel.

wcekan:
test added.  If you reverse the change the tests fails with
```
Failed tests: 
  ResourceIT.createChildRelateExisting:962 1 expectation failed.
Expected status code <200> doesn't match actual status code <400>.
```

aklish:
👍

PR #356, opened 2016-12-05
Enable filter expression handling in 3.0 patch extension

wcekan:
Expands on https://github.com/yahoo/elide/pull/354

Handles RequestScopedTransaction in 3.0 and with filter expressions in 2.0


yahoocla:
CLA is valid!

clayreimann:
👍

PR #358, opened 2016-12-06
Unique HQL parameter names

wcekan:
Found a problem where multiple Predicates for the same field attempted to use the same parameter name, causing a name collision.

Added Predicate.getParameterName to generate a unique name based on the predicate properties.


yahoocla:
CLA is valid!

DeathByTape:
👍

PR #359, opened 2016-12-09
Multiplex close

wcekan:
Multiplex Transaction was leaving orphaned child transactions on close.  Moved the clear to the correct location after `close`.


yahoocla:
CLA is valid!

DeathByTape:
👍

PR #360, opened 2016-12-14
Start on implementing Swagger documentation generator

ddsnowboard:
I got all the boilerplate stuff out of the way by writing the classes that map to the JSON objects in the OpenAPI spec, and I also wrote a class that would theoretically build the object tree, but we never managed to succeed in interfacing with the main Elide code, so the classes at the end of the day don't do anything. They do have unit tests though.


yahoocla:
CLA is valid!

Issue #363, opened 2016-12-15
OnRead Hooks


DeathByTape:
Need to have hooks that can be invoked on reads.

PR #364, opened 2016-12-15
On read hook

DeathByTape:
Addresses: https://github.com/yahoo/elide/issues/363


yahoocla:
CLA is valid!

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -357,7 +357,7 @@ public Object getRelation(
             throw new ClassCastException("Fail trying to cast requestscope");
         }
         EntityDictionary dictionary = requestScope.getDictionary();
-        Object val = com.yahoo.elide.core.PersistentResource.getValue(entity, relationName, dictionary);
+        Object val = com.yahoo.elide.core.PersistentResource.getValue(entity, relationName, requestScope);
```
DeathByTape: Doh! This is a remnant. Needs to be removed.

PR #369, opened 2017-01-06
Added JaxRS endpoint for swagger docs

aklish:
Added basic IT test for swagger doc

Fixed checkstyles

Modified pom to run integration tests

Refactored pom to only run mysql stuff for dropwizard


yahoocla:
CLA is valid!

elide-contrib/elide-swagger/testng.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
clayreimann: Do we need this file? Won't the tests be picked up automatically?

DeathByTape: I agree. Best to avoid these `testng.xml` files unless there is a good reason for needing them. In the past these have led to problems of not getting updated and new tests are then not consistently run :(

DeathByTape:
👍

PR #371, opened 2017-01-14
Update swagger code to support large and complex entity graphs.

aklish:
A number of fixes:
 - The model resolver now correctly handles Elide entities that were attributes.
 - Aggressively prune redundant paths.
 - Allow options to generate docs for subgraphs of the entity model.
 - Allow options to limit the filtering operators for each GET 
 - The JAXRS resource now supports serving multiple swagger documents.

### DIFF
--- a/elide-contrib/elide-swagger/src/main/java/com/yahoo/elide/contrib/swagger/JsonApiModelResolver.java
+++ b/elide-contrib/elide-swagger/src/main/java/com/yahoo/elide/contrib/swagger/JsonApiModelResolver.java
@@ -5,6 +5,7 @@
  */
 package com.yahoo.elide.contrib.swagger;
 
+import com.fasterxml.jackson.databind.type.SimpleType;
 import com.yahoo.elide.annotation.CreatePermission;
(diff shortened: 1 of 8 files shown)


yahoocla:
CLA is valid!

elide-contrib/elide-swagger/src/main/java/com/yahoo/elide/contrib/swagger/JsonApiModelResolver.java
```diff
@@ -68,7 +79,14 @@ public Model resolve(Type type, ModelConverterContext context, Iterator<ModelCon
         for (String relationshipName : relationshipNames) {
 
             Class<?> relationshipType = dictionary.getParameterizedType(clazz, relationshipName);
-            Relationship relationship = new Relationship(dictionary.getJsonAliasFor(relationshipType));
+            Relationship relationship;
+            try {
+                relationship = new Relationship(dictionary.getJsonAliasFor(relationshipType));
```
clayreimann: same as above

Issue #373, opened 2017-01-16
Delete Requests throw no ForbiddenAccessException


IDragonfire:
We use elide in JAVA Spring with some Security Annotations (`@ReadPermission`, `@DeletePermission`) .

On GET Requests all work fine, a `ForbiddenAccessException` is thrown internally and `buildErrorResponse` return a valid error response with an non empty code and body.

On DELETE Request this does not happen, the status code is `403` with a `null` body.

Is this behaviour correct?
Did we make something wrong?

![image](https://cloud.githubusercontent.com/assets/1217999/21986713/cda05950-dc00-11e6-8587-6e043fe40406.png)

![image](https://cloud.githubusercontent.com/assets/1217999/21986751/f63bbce2-dc00-11e6-9162-1d42ad5a5fe0.png)

Controller
```java
  @RequestMapping(
      method = RequestMethod.DELETE,
      produces = JSON_API_MEDIA_TYPE,
      value = {"/{entity}/{id}", "/{entity}/{id}/relationships/{entity2}"})
  @Transactional
  public String jsonApiDelete(final HttpServletRequest request,
                           final Authentication authentication) throws JsonApiException {
    ElideResponse response = elide.delete(
        getJsonApiPath(request),
        null,
        getPrincipal(authentication)
    ).getBody();
}
```

Config
```java
@Configuration
public class ElideConfig {

  @Bean
  public Elide elide(EntityManagerFactory entityManagerFactory) {
    ConcurrentHashMap<String, Class<? extends Check>> checks = new ConcurrentHashMap<>();
    checks.put(IsAuthenticated.EXPRESSION, IsAuthenticated.Inline.class);
    checks.put(IsOwner.EXPRESSION, IsOwner.Inline.class);
    checks.put(IsClanLeader.EXPRESSION, IsClanLeader.Inline.class);
    checks.put(IsClanMembershipLeader.EXPRESSION, IsClanMembershipLeader.Inline.class);

    EntityDictionary entityDictionary = new EntityDictionary(checks);

    RSQLFilterDialect rsqlFilterDialect = new RSQLFilterDialect(entityDictionary);

    return new Elide.Builder(new HibernateStore(entityManagerFactory.unwrap(SessionFactory.class)))
        .withAuditLogger(new Slf4jLogger())
        .withEntityDictionary(entityDictionary)
        .withJoinFilterDialect(rsqlFilterDialect)
        .withSubqueryFilterDialect(rsqlFilterDialect)
        .build();
  }

[...]

PR #374, opened 2017-01-16
Surface ForbiddenAccessException for DELETE.

DeathByTape:
Addresses: https://github.com/yahoo/elide/issues/373

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
+++ b/elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
@@ -10,7 +10,6 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.yahoo.elide.core.HttpStatus;
 import com.yahoo.elide.core.PersistentResource;
 import com.yahoo.elide.core.RequestScope;
-import com.yahoo.elide.core.exceptions.ForbiddenAccessException;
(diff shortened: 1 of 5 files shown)


yahoocla:
CLA is valid!

elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
```diff
@@ -85,12 +84,8 @@ public RecordTerminalState(PersistentResource record, CollectionTerminalState co
 
     @Override
     public Supplier<Pair<Integer, JsonNode>> handleDelete(StateContext state) {
-        try {
-            record.deleteResource();
-            return () -> Pair.of(HttpStatus.SC_NO_CONTENT, null);
-        } catch (ForbiddenAccessException e) {
-            return () -> Pair.of(e.getStatus(), null);
-        }
+        record.deleteResource();
```
clayreimann: I'm guessing that we'll have to be cleverer in `deleteResource` to throw `403` instead of `400`

DeathByTape: The only test that isn't passing is a JSON Patch extension, unfortunately :( i think it may be best to clean up that solution.

I'm relatively certain the reason it was this way is because it was easier for us to do at the time

PR #376, opened 2017-01-19
Surface web exceptions thrown by called code.

DeathByTape:


premc:
👍

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1914,4 +1922,21 @@ private Collection copyCollection(final Collection collection) {
...
+     */
+    private static RuntimeException handleInvocationTargetException(InvocationTargetException e) {
+        Throwable exception = e.getTargetException();
+        if (exception instanceof HttpStatusException) {
+            return (HttpStatusException) exception;
+        } else if (exception instanceof WebApplicationException) {
+            return (WebApplicationException) exception;
+        }
+        log.debug("Caught an unexpected exception", e);
+        return new ServerErrorException("Unexpected exception caught", Response.Status.INTERNAL_SERVER_ERROR);
```
wcekan: I would prefer to add `, e);` to propagate the exception.

Issue #93, opened 2015-12-18
Return 200 instead of 204 on update


DeathByTape:
There has been expressed interest in returning a proper 200 OK code (with proper response body) upon updating instead of a 204 No Content. As far as I can tell, there are no technical limitations to us doing this at the moment. This is likely an artifact of an artificial limitation we encountered during our initial development.

That said, this should comply with the JSON API spec: http://jsonapi.org/format/#crud-updating-responses-200

wcekan:
The spec is unclear on this point.  It says to respond with either 200 or 204 when the changes are applied.  However, always choosing 200 would simplify client processing.

pokaru:
Hey William,

I think this issue came from one of my request.  Basically, what I'd like is for the updated object to be returned on updates.  We would like to sent our clients the updated object without having to make another rs call.

QubitPi:
Can I pick up this one? Just make sure this feature is still needed

aklish:
I think this should be OK to do, but we should have an option in Elide builder to enable this behavior.  Default should be what it does today so it won't break anyone's API.

QubitPi:
OK, I will add the option, leave all tests unchanged, and add some new tests to test this option

DeathByTape:
Done in #377, thanks @QubitPi

PR #377, opened 2017-01-21
Allow user to configure update status code(200 or 204)

QubitPi:
https://github.com/yahoo/elide/issues/93

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -327,6 +327,11 @@ public class Elide {
             subqueryFilterDialects.add(dialect);
             return this;
         }
+
+        public Builder withUpdateStatusCode(int updateStatusCode) {
+            HttpStatus.SC_UPDATED = updateStatusCode == HttpStatus.SC_OK ? updateStatusCode : HttpStatus.SC_UPDATED;
+            return this;
+        }
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -71,7 +71,6 @@ import javax.ws.rs.core.MultivaluedMap;
 @Slf4j
 public class Elide {
 
-
     private final AuditLogger auditLogger;
     private final DataStore dataStore;
     private final EntityDictionary dictionary;
@@ -80,6 +79,7 @@ public class Elide {
     private final List<JoinFilterDialect> joinFilterDialects;
     private final List<SubqueryFilterDialect> subqueryFilterDialects;
     private final boolean useFilterExpressions;
+    private final int updateStatusCode;
 
     /**
      * Instantiates a new Elide.
@@ -125,7 +125,8 @@ public class Elide {
                 ActivePermissionExecutor::new,
                 Collections.singletonList(new DefaultFilterDialect(dictionary)),
                 Collections.singletonList(new DefaultFilterDialect(dictionary)),
-                false
+                false,
+                HttpStatus.SC_NO_CONTENT
         );
     }
 
@@ -151,7 +152,8 @@ public class Elide {
             ActivePermissionExecutor::new,
             Collections.singletonList(new DefaultFilterDialect(dictionary)),
             Collections.singletonList(new DefaultFilterDialect(dictionary)),
-            false
+            false,
+            HttpStatus.SC_NO_CONTENT
         );
     }
 
@@ -174,7 +176,8 @@ public class Elide {
                   Function<RequestScope, PermissionExecutor> permissionExecutor,
                   List<JoinFilterDialect> joinFilterDialects,
                   List<SubqueryFilterDialect> subqueryFilterDialects,
-                  boolean useFilterExpressions) {
+                  boolean useFilterExpressions,
+                  int updateStatusCode) {
         this.auditLogger = auditLogger;
         this.dataStore = dataStore;
         this.dictionary = dictionary;
@@ -184,6 +187,7 @@ public class Elide {
         this.joinFilterDialects = joinFilterDialects;
         this.subqueryFilterDialects = subqueryFilterDialects;
         this.useFilterExpressions = useFilterExpressions;
+        this.updateStatusCode = updateStatusCode;
     }
 
     /**
@@ -198,6 +202,7 @@ public class Elide {
         private List<JoinFilterDialect> joinFilterDialects;
         private List<SubqueryFilterDialect> subqueryFilterDialects;
         private boolean useFilterExpressions;
+        private int updateStatusCode;
 
         /**
          * A new builder used to generate Elide instances. Instantiates an {@link EntityDictionary} without
@@ -213,7 +218,7 @@ public class Elide {
             this.jsonApiMapper = new JsonApiMapper(entityDictionary);
             this.joinFilterDialects = new ArrayList<>();
             this.subqueryFilterDialects = new ArrayList<>();
-
+            updateStatusCode = HttpStatus.SC_NO_CONTENT;
         }
 
         /**
@@ -228,6 +233,7 @@ public class Elide {
             this.jsonApiMapper = new JsonApiMapper(entityDictionary);
             this.joinFilterDialects = new ArrayList<>();
             this.subqueryFilterDialects = new ArrayList<>();
+            updateStatusCode = HttpStatus.SC_NO_CONTENT;
         }
 
         public Elide build() {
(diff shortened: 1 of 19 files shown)


elide-core/src/main/java/com/yahoo/elide/core/HttpStatus.java
```diff
@@ -18,4 +18,5 @@
     public static final int SC_NOT_FOUND = 404;
     public static final int SC_LOCKED = 423;
     public static final int SC_INTERNAL_SERVER_ERROR = 500;
+    public static int SC_UPDATED = SC_NO_CONTENT;
```
clayreimann: I disagree with the choice to make this `public static` rather than an instance variable of `Elide`

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -327,6 +327,11 @@ public Builder withSubqueryFilterDialect(SubqueryFilterDialect dialect) {
             subqueryFilterDialects.add(dialect);
             return this;
         }
+
+        public Builder withUpdateStatusCode(int updateStatusCode) {
```
DeathByTape: I don't think we should allow the user to choose an arbitrary status code here and silently "correct" them. Instead, we should be very explicit. `withUpdate200Status` or `withUpdate204Status` or some variety of that flavor.

By default we should retain the existing `204` behavior, of course.

elide-core/src/main/java/com/yahoo/elide/core/HttpStatus.java
DeathByTape: I agree with the previous comment I read: not sure this is how we want to do this. This configuration shouldn't be globally modifiable at-will. We should encapsulate this nicely inside of Elide.

elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
```diff
@@ -79,7 +79,16 @@ public RecordTerminalState(PersistentResource record, CollectionTerminalState co
         }
 
         patch(resource, state.getRequestScope());
-        return () -> Pair.of(HttpStatus.SC_NO_CONTENT, null);
+        return () -> Pair.of(
+                HttpStatus.SC_UPDATED,
+                HttpStatus.SC_UPDATED == HttpStatus.SC_NO_CONTENT
```
DeathByTape: When we refactor, we'll need to change the`SC_UPDATED` bit, but other than that this logic looks good to me.

elide-core/src/main/java/com/yahoo/elide/parsers/state/RelationshipTerminalState.java
```diff
@@ -84,7 +84,38 @@ public RelationshipTerminalState(PersistentResource record, String relationshipN
 
     @Override
     public Supplier<Pair<Integer, JsonNode>> handlePatch(StateContext state) {
-        return handleRequest(state, this::patch);
+        return handlePatchRequest(state, this::patch);
+    }
+
+    private Supplier<Pair<Integer, JsonNode>> handlePatchRequest(StateContext state,
```
DeathByTape: Rather than duplicating the logic, it may make most sense here to modify the existing `handleRequest` method. Namely, either add a parameter that determines the type of response or just check if the `handler == this::patch` for a short-term solution since that seems to be the only use case right now.

elide-core/src/main/java/com/yahoo/elide/parsers/state/RelationshipTerminalState.java
```diff
@@ -84,7 +84,38 @@ public RelationshipTerminalState(PersistentResource record, String relationshipN
...
+                                                            BiFunction<Data<Resource>, RequestScope, Boolean> handler) {
+        Data<Resource> data = state.getJsonApiDocument().getData();
+        handler.apply(data, state.getRequestScope());
+        return () -> Pair.of(
```
DeathByTape: This return statement is duplicated. Perhaps consider a small static method to encapsulate it. Since this logic appears to be identical, it's easier to maintain if we have it only in one place.

elide-integration-tests/src/test/java/com/yahoo/elide/auditTests/AuditIT.java
```diff
@@ -127,4 +129,25 @@ public void testAuditUpdateOnInverseCollection() {
         Assert.assertTrue(logger.logMessages.contains("Entity with id 1 now has inverse list []"));
         Assert.assertTrue(logger.logMessages.contains("Inverse entities: []"));
     }
+
+    @Test(priority = 999)
+    public void testAuditWithDuplicateLineageEntry2() {
```
clayreimann: This test will need to go in a new class that extends `AbstractIntegrationTestInitializer` and creates a `StandardTestBinder` that returns the appropriate status code.

elide-datastore/testng-assignedIdLong.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2016, Yahoo Inc.
```
DeathByTape: 2017 :)

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -166,6 +167,52 @@ public RequestScope(String path,
...
+     * @param user the user making this request
+     * @param dictionary the entity dictionary
+     * @param mapper converts JsonApiDocuments to raw JSON
+     * @param auditLogger logger for this request
+     * @param queryParams the query parameters
+     * @param securityMode the current security mode
+     * @param permissionExecutorGenerator the user-provided function that will generate a permissionExecutor
+     * @param updateStatusCode the response status code on successful path request
+     */
+    public RequestScope(String path,
```
clayreimann: This should be the primary constructor and the other constructor should be `deprecated`.

elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordTerminalState.java
```diff
@@ -79,22 +79,13 @@ public RecordTerminalState(PersistentResource record, CollectionTerminalState co
...
-                        )
-        );
+        return constructResponse(record, state, "patch");
     }
 
     @Override
     public Supplier<Pair<Integer, JsonNode>> handleDelete(StateContext state) {
         record.deleteResource();
-        return () -> Pair.of(HttpStatus.SC_NO_CONTENT, null);
+        return constructResponse(record, state, "delete");
```
clayreimann: Let's not change this since it will never vary

elide-core/src/main/java/com/yahoo/elide/parsers/state/RelationshipTerminalState.java
```diff
@@ -84,24 +84,26 @@ public RelationshipTerminalState(PersistentResource record, String relationshipN
...
-    private Supplier<Pair<Integer, JsonNode>> handleRequest(StateContext state,
-                                                           BiFunction<Data<Resource>, RequestScope, Boolean> handler) {
+    private Supplier<Pair<Integer, JsonNode>> handleRequest(
+            StateContext state,
+            BiFunction<Data<Resource>, RequestScope, Boolean> handler,
+            String requestType) {
         Data<Resource> data = state.getJsonApiDocument().getData();
         handler.apply(data, state.getRequestScope());
-        return () -> Pair.of(HttpStatus.SC_NO_CONTENT, null);
+        return constructResponse(record, state, requestType);
```
clayreimann: This is [invalid](http://jsonapi.org/format/#crud-updating-relationship-responses-204)

elide-core/src/main/java/com/yahoo/elide/parsers/state/BaseState.java
```diff
@@ -150,4 +161,41 @@ public void handle(StateContext state, SubCollectionRelationshipContext ctx) {
     public Supplier<Pair<Integer, JsonNode>> handleDelete(StateContext state) throws HttpStatusException {
         throw new UnsupportedOperationException(this.getClass().toString());
     }
+
+    protected static Supplier<Pair<Integer, JsonNode>> constructResponse(
```
clayreimann: This can be simplified based on my comments below.

clayreimann:
👍  After my changes from #395 are merged here

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
clayreimann: This is addressed.

PR #379, opened 2017-01-25
Bridgeable store

DeathByTape:
This PR cleans up a few things such as logging uncaught exceptions and using the core `RequestScope` instead of secure's (since all datastores require `elide-core` to get the `DataStore` interface anyway).

More importantly, however, this is the first pass at a bridgeable store. Currently only _reads_ are supported. My test cases are using Hibernate5 and a toy/incomplete Redis store I created solely for the purpose of this test.


elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -282,6 +282,9 @@ public ElideResponse get(
             return buildErrorResponse(new TransactionException(e), isVerbose);
         } catch (ParseCancellationException e) {
             return buildErrorResponse(new InvalidURLException(e), isVerbose);
+        } catch (Exception e) {
```
wcekan: Should this be `RuntimeException | Error e`

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/BridgeableTransaction.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public interface BridgeableTransaction {
+
+    Object bridgeableLoadObject(MultiplexTransaction muxTx,
+                                Object parent,
+                                String relationName,
+                                Optional<FilterExpression> filterExpression,
+                                RequestScope scope);
+
+    Object bridgeableLoadObjects(MultiplexTransaction muxTx,
```
aklish: Should this be Iterable\<Object\>?

PR #380, opened 2017-01-26
Support object-level reads through Noop Store and RequestScope params for Computed Fields

DeathByTape:
This PR offers support for performing object-level reads in the noop store. Moreover, it enables `@ComputedAttribute`'s and `@ComputedRelationship`'s to accept a `RequestScope` as a parameter.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
@@ -21,6 +21,7 @@ import com.yahoo.elide.annotation.OnDeletePostCommit;
 import com.yahoo.elide.annotation.OnUpdatePostCommit;
 import com.yahoo.elide.annotation.OnDeletePreCommit;
 import com.yahoo.elide.core.exceptions.DuplicateMappingException;
+import com.yahoo.elide.security.*;
 import lombok.Getter;
 import lombok.Setter;
 import org.apache.commons.collections4.MultiValuedMap;
@@ -49,6 +50,7 @@ import java.util.Deque;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.stream.Stream;
 
 /**
  * Entity Dictionary maps JSON API Entity beans to/from Entity type names.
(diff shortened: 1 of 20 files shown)


elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -21,6 +21,7 @@
 import com.yahoo.elide.annotation.OnUpdatePostCommit;
 import com.yahoo.elide.annotation.OnDeletePreCommit;
 import com.yahoo.elide.core.exceptions.DuplicateMappingException;
+import com.yahoo.elide.security.*;
```
clayreimann: Did you mean to `import *` here?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1594,13 +1597,22 @@ private boolean isValidParameterizedMap(Map<?, ?> values, Class<?> keyType, Clas
...
-    public static Object getValue(Object target, String fieldName, EntityDictionary dictionary) {
+    public static Object getValue(Object target, String fieldName, RequestScope requestScope) {
+        EntityDictionary dictionary = requestScope.getDictionary();
         AccessibleObject accessor = dictionary.getAccessibleObject(target, fieldName);
         try {
             if (accessor instanceof Method) {
+                // Pass RequestScope into @Computed fields if requested
+                if (Stream.of(accessor.getAnnotations())
+                        .map(Annotation::annotationType)
+                        .anyMatch(c -> c == ComputedAttribute.class || c == ComputedRelationship.class)
```
clayreimann: We do this in `EntityBinding` too, do we want to save the value there so we can retrieve it more efficiently? Perhaps also remembering if the computed thing takes the `RequestScope` too?

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
DeathByTape: I did not :(

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
DeathByTape: Fair point, I'll make that change!

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/InMemoryFilterVisitor.java
```diff
@@ -5,23 +5,23 @@
...
 /**
  * Visitor for in memory filterExpressions
  */
 public class InMemoryFilterVisitor implements Visitor<Predicate> {
-    private final EntityDictionary dictionary;
+    private final RequestScope requestScope;
 
-    public InMemoryFilterVisitor(EntityDictionary dictionary) {
-        this.dictionary = dictionary;
+    public InMemoryFilterVisitor(RequestScope requestScope) {
```
wcekan: Should this constructor take the Interface `RequestScope` instead of implementation one?

DeathByTape: I don't think so. This is already located in `elide-core`, so I see little reason to take the more generic interface. That interface seems to be more limiting and everywhere we use it, it gets casted to `elide-core`'s version anyway.

Generally speaking, if someone was to overload RequestScope, they would really have to extend `elide-core`'s `RequestScope`. If they didn't we would likely get cast exceptions in many places.

clayreimann:
👍

PR #388, opened 2017-02-06
Fix RSQL value coercion.

DeathByTape:
This PR fix the value coercion problem for nested types in RSQL.


clayreimann:
Whelp. Ok then 👍

PR #395, opened 2017-02-10
Clean up code and conform to JSON API

clayreimann:


elide-core/src/main/java/com/yahoo/elide/parsers/state/RelationshipTerminalState.java
```diff
@@ -79,33 +79,34 @@ public RelationshipTerminalState(PersistentResource record, String relationshipN
...
+     */
+    private Supplier<Pair<Integer, JsonNode>> handleRequest(StateContext state,
+                                                            BiFunction<Data<Resource>, RequestScope, Boolean> handler) {
         Data<Resource> data = state.getJsonApiDocument().getData();
         handler.apply(data, state.getRequestScope());
-        return requestType.equals("patch")
-                ? constructPatchResponse(record, state)
-                : () -> Pair.of(HttpStatus.SC_NO_CONTENT, null);
+        // TODO: figure out if we've made modifications that differ from those requested by client
+        return () -> Pair.of(HttpStatus.SC_NO_CONTENT, null);
```
QubitPi: Isn't gonna make `SC_OK` not working if we need 200 on update?

QubitPi:
👍

PR #400, opened 2017-02-17
Pin jersey container version for example projects

danchen:
Jersey container is now pulling version 2.26-b02 as RELEASE, and it is causing "java.lang.ClassNotFoundException: javax.ws.rs.client.RxInvokerProvider" with that version. We should pin to last good version.

- Pin jersey container to version 2.25.1 for example projects.


DeathByTape:
👍

DeathByTape:
We should probably look at upgrading all of our dependencies before we push out elide 3.0.

PR #404, opened 2017-02-22
Call session.flush only if intended

wcekan:


elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -83,7 +85,9 @@ public void flush(RequestScope requestScope) {
         try {
             deferredTasks.forEach(Runnable::run);
             deferredTasks.clear();
-            session.flush();
+            if (session.getFlushMode() != FlushMode.COMMIT && session.getFlushMode() != FlushMode.NEVER) {
```
DeathByTape: Should we do this in hibernate5 too?

wcekan: Done

DeathByTape:
👍

Issue #397, opened 2017-02-14
Relationship update returns correct response code


QubitPi:
Base on the [JSON API docs](http://jsonapi.org/format/#crud-updating-relationship-responses-204), relationship updates MUST return 204 unless the server has made additional modification to the relationship. When the relationship is modified, however, Elide should return either 200 or 204 based on users' preference.

DeathByTape:
Can you explain a little deeper what you're intending with this?

QubitPi:
@DeathByTape After https://github.com/yahoo/elide/pull/377, users are free to choose between 200 and 204 on update, but JSON API seems to restrict this choice when user updates relationships according to http://jsonapi.org/format/#crud-updating-relationship-responses-204. Looks like JSON API forces 204 if an PATCH request doesn't change any relationships. @clayreimann brought this up in his comments for https://github.com/yahoo/elide/pull/377

QubitPi:
I guess we keep https://github.com/yahoo/elide/pull/377 for everything except for the case of **updating relationships**

clayreimann:
NB: the behavior of `PATCH` to relationship urls was unchanged in #377.

aklish:
Closing for inactivity.

PR #406, opened 2017-02-24
Relationship update returns correct response code

QubitPi:
https://github.com/yahoo/elide/issues/397

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/parsers/state/RelationshipTerminalState.java
+++ b/elide-core/src/main/java/com/yahoo/elide/parsers/state/RelationshipTerminalState.java
@@ -82,9 +82,31 @@ public class RelationshipTerminalState extends BaseState {
         return () -> Pair.of(HttpStatus.SC_OK, mapper.toJsonObject(doc));
     }
 
+    /**
+     * Handle PATCH requests for relationship updates.
+     * <p>
+     * If request doesn't modify relationship, return 204,
+     * otherwise, return either 200 or 204 based on user's configuration on Elide
+     *
+     * @param state the state containing request information
+     *
+     * @return the proper update status code
+     */
     @Override
     public Supplier<Pair<Integer, JsonNode>> handlePatch(StateContext state) {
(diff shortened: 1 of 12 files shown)


elide-core/src/main/java/com/yahoo/elide/parsers/state/RelationshipTerminalState.java
```diff
@@ -82,9 +82,31 @@ public RelationshipTerminalState(PersistentResource record, String relationshipN
...
-        return handleRequest(state, this::patch);
+        RequestScope requestScope = state.getRequestScope();
+        int updateStatusCode = requestScope.getUpdateStatusCode();
+
+        boolean relationshipChanged = patch(state.getJsonApiDocument().getData(), requestScope);
+
+        if (relationshipChanged) {
+            return () -> Pair.of(
+                    updateStatusCode,
+                    updateStatusCode == HttpStatus.SC_NO_CONTENT ? null : getResponseBody(record, requestScope)
```
clayreimann: This conditional is not needed here, JSON API says that when the server makes modifications in addition to those sent by the user the server **MUST** respond with `200` and the modified documents

elide-core/src/main/java/com/yahoo/elide/parsers/state/RelationshipTerminalState.java
```diff
@@ -82,9 +82,31 @@ public RelationshipTerminalState(PersistentResource record, String relationshipN
...
+     *
+     * @return the proper update status code
+     */
     @Override
     public Supplier<Pair<Integer, JsonNode>> handlePatch(StateContext state) {
-        return handleRequest(state, this::patch);
+        RequestScope requestScope = state.getRequestScope();
+        int updateStatusCode = requestScope.getUpdateStatusCode();
+
+        boolean relationshipChanged = patch(state.getJsonApiDocument().getData(), requestScope);
```
clayreimann: I'm also not sure this is correct. We're not looking for *any* change we're looking for changes that differ from those sent by the client.

Issue #407, opened 2017-02-26
Exception thrown when using forward only result sets


danuid:
I was using SQL Server 2008 and JPA backed by Hibernate5.
The JDBC driver I'm using is `sqljdbc42.jar` from [this file](https://www.microsoft.com/en-us/download/details.aspx?id=11774).

When I try to fetch a collection using the JSON API, `ScrollableIterator.java` line 37 tried to call `this.scroll.first()` on a `ScrollableResults`.
 I'm not sure why the result set is somehow put  to `FORWARD_ONLY` mode, and an exception is thrown.

The stack trace is shown below, if it is of any help:
```
com.microsoft.sqlserver.jdbc.SQLServerException: The requested operation is not supported on forward only result sets.
	at com.microsoft.sqlserver.jdbc.SQLServerException.makeFromDriverError(SQLServerException.java:191)
	at com.microsoft.sqlserver.jdbc.SQLServerResultSet.throwNotScrollable(SQLServerResultSet.java:414)
	at com.microsoft.sqlserver.jdbc.SQLServerResultSet.verifyResultSetIsScrollable(SQLServerResultSet.java:437)
	at com.microsoft.sqlserver.jdbc.SQLServerResultSet.first(SQLServerResultSet.java:1417)
	at com.zaxxer.hikari.pool.HikariProxyResultSet.first(HikariProxyResultSet.java)
	at org.hibernate.internal.ScrollableResultsImpl.first(ScrollableResultsImpl.java:75)
	at com.yahoo.elide.datastores.hibernate5.ScrollableIterator.iterator(ScrollableIterator.java:37)
	at com.yahoo.elide.core.PersistentResourceSet.iterator(PersistentResourceSet.java:35)
	at com.yahoo.elide.core.PersistentResource.filter(PersistentResource.java:1691)
	at com.yahoo.elide.core.PersistentResource.filter(PersistentResource.java:1674)
	at com.yahoo.elide.core.PersistentResource.loadRecordsWithSortingAndPagination(PersistentResource.java:354)
	at com.yahoo.elide.parsers.state.CollectionTerminalState.getResourceCollection(CollectionTerminalState.java:135)
	at com.yahoo.elide.parsers.state.CollectionTerminalState.handleGet(CollectionTerminalState.java:68)
	at com.yahoo.elide.parsers.state.StateContext.handleGet(StateContext.java:110)
	at com.yahoo.elide.parsers.GetVisitor.visitQuery(GetVisitor.java:34)
[...]

PR #408, opened 2017-02-27
No longer call unsafe methods in ScrollableIterator

aklish:
Should fix issue #407 

Would like a review from @wcekan before merging.


DeathByTape:
👍

DeathByTape:
👍

danuid:
Great :+1:

PR #410, opened 2017-02-27
Updated Elide to use an ElideSettings object in lieu of explicit decl…

DeathByTape:
…aration of all settings.


DeathByTape:
**NOTE:** One more thing to do after this PR before we merge elide 3.0 back to master. Namely, we should remove PersistenceStore entirely from the hibernate5 store as it is currently unmaintained and of limited use.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -141,10 +132,12 @@ public RequestScope(String path,
             }
         };
 
+        Function<RequestScope, PermissionExecutor> permissionExecutorGenerator = elideSettings.getPermissionExecutor();
         this.permissionExecutor = (permissionExecutorGenerator == null)
                 ? new ActivePermissionExecutor(this)
                 : permissionExecutorGenerator.apply(this);
 
+
```
clayreimann: Extra blank line here?

PR #413, opened 2017-02-28
Promote OpaqueUserFunction to be a top-level interface

clayreimann:
This is in preparation for GraphQL support (we'll want to share this between our endpoints)


DeathByTape:
👍

PR #414, opened 2017-02-28
The great dependency upgrade of 2017

DeathByTape:


DeathByTape:
According to version eye, there were no major-version upgrades.

That said, it was intentional for the hibernate3 upgrade, of course :)

PR #421, opened 2017-03-03
Allow typed filters to traverse to-one relationships

clayreimann:


elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
```diff
@@ -1407,187 +1428,214 @@ class FilterIT extends AbstractIntegrationTestInitializer {
...
+        }
+    }
+
+    @Test
+    void testFailFilterAuthorBookByChapter() {
+        def result = mapper.readTree(RestAssured.get("/author/$hemingwayId/books?filter[book.chapters.title]=doesn't matter").asString())
+        Assert.assertNotNull(result.get("errors"))
+
+        // This causes hibernate errors
+//        result = mapper.readTree(RestAssured.get("/author/$hemingwayId/books?filter[book]=chapters.title=='Borked'").asString())
```
aklish: Let's cleanup the commented out code.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -157,7 +157,8 @@ public FilterExpression parseGlobalExpression(String path, MultivaluedMap<String
 
                 try {
                     Node ast = parser.parse(expressionText);
-                    RSQL2ExpressionFilterVisitor visitor = new RSQL2ExpressionFilterVisitor(false);
+                    // this is not guarded against to-many joins
+                    RSQL2ExpressionFilterVisitor visitor = new RSQL2ExpressionFilterVisitor(true);
```
aklish: This behavior will allow nested filters on 'to many' relationships as well.  This isn't the same behavior as the default filter dialect.

elide-integration-tests/src/test/groovy/com/yahoo/elide/tests/FilterIT.groovy
```diff
@@ -1407,187 +1428,214 @@ class FilterIT extends AbstractIntegrationTestInitializer {
...
-            String name = author.get("attributes").get("name").asText();
-            Assert.assertTrue(("Isaac Asimov".equals(name) || "Null Ned".equals(name)) && !name.equals(last));
+            String name = author.get("attributes").get("name").asText()
+            Assert.assertTrue(name == "Isaac Asimov" || name == "Null Ned")
         }
     }
 
     @Test
-    public void testGetBadRelationshipRoot() {
+    void testFilterAuthorBookByPublisher() {
```
aklish: We should have tests both for default filter and rsql filter dialects

clayreimann: There are tests for both. I have the default filter test on top followed by the rsql test, like the other tests here.

PR #424, opened 2017-03-07
Fixing page totals for nested/joined filters

aklish:


elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -510,4 +514,8 @@ public Integer getQueryLimit() {
         // no limit
         return null;
     }
+
+    private static String getUniqueAlias(Class<?> entityType) {
```
DeathByTape: Should we rename this to something clearer? It's not exactly unique but merely distinct from `relationType`, right?

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -380,4 +384,8 @@ public void close() throws IOException {
     public User accessUser(Object opaqueUser) {
         return new User(opaqueUser);
     }
+
+    private static String getUniqueAlias(Class<?> entityType) {
```
DeathByTape: Same question as above.

PR #430, opened 2017-03-10
Allow extensible type conversion

clayreimann:


wcekan:
How is does this work with Jackson `.addSerializer` and `ConvertUtils.register`?

DeathByTape:
@clayreimann you're right. We wouldn't want to stomp on two versions of Elide running in the same JVM. Arguably this could be mitigated by simply running 2 JVM's, but for various reasons that could add unnecessary complexity.

Is it possible to add the converter directly to the `ElideSettings` object and pass it with the `RequestScope`? (`ElideSettings` are automatically passed around in the `RequestScope` object).

DeathByTape:
Update on this?

PR #433, opened 2017-03-10
Fix checkstyle

clayreimann:


elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -59,6 +59,8 @@
 class EntityBinding {
 
     private static final List<Method> OBJ_METHODS = Arrays.asList(Object.class.getMethods());
+    private static final List<Class<? extends Annotation>> RELATIONSHIP_TYPES =
```
DeathByTape: Should we use a set here and just perform a `contains`?

Iterating and comparing a constant list of 4 may be faster than the constant time hashing, though. @wcekan thoughts here?

elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -193,6 +163,41 @@ private Pagination evaluate(int defaultLimit, int maxLimit) {
         return this;
     }
 
+    private boolean hasInvalidCombiation(Map<PaginationKey, Integer> pageData) {
```
DeathByTape: Typo. `Combiation` -> `Combination`

DeathByTape:
Can we add these change descriptions to the change log?

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
wcekan: This is fine.  We loop over this list, but do not call `contains`.

PR #442, opened 2017-04-13
Deprecate duplicate methods. Simplify code paths

clayreimann:


clayreimann:
Gah, `javadoc`'ed. Will fix

DeathByTape:
:+1: after build is fixed

PR #452, opened 2017-04-27
Translate top level README to Chinese

QubitPi:
I can spread Elide into Chinese OSS community once Chinese translation is ready


aklish:
Nice!

PR #455, opened 2017-05-01
Restrict read deferrals to new object creation

aklish:


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -812,7 +816,7 @@ public PersistentResource getRelation(String relation, String id) {
             // already).
             skipNew = true;
         } else {
-            Class<?> entityType = dictionary.getParameterizedType(getResourceClass(), relation);
+            //Class<?> entityType = dictionary.getParameterizedType(getResourceClass(), relation);
```
wcekan: Remove

elide-core/src/main/java/com/yahoo/elide/core/exceptions/ForbiddenAccessException.java
```diff
@@ -20,20 +20,22 @@
...
     public ForbiddenAccessException(String message) {
         super(HttpStatus.SC_FORBIDDEN, null, message);
         this.expression = Optional.empty();
+        this.evaluationMode = Optional.empty();
     }
 
-    public ForbiddenAccessException(String message, Expression expression) {
-        super(HttpStatus.SC_FORBIDDEN, null, message);
+    public ForbiddenAccessException(String message, Expression expression, Expression.EvaluationMode mode) {
+        super(HttpStatus.SC_FORBIDDEN, null, message + ": " + expression);
```
wcekan: This does a string manipulation for every ForbiddenAccess which include each filtered element in a large collection.

PR #457, opened 2017-05-01
Improved Filter performance

wcekan:
Avoid `Class.getSimpleName()` in FilterExpression by loading the correct alias when needed. 

Made a few runtime improvements, and some other code cleanup as well.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/exceptions/HttpStatusException.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/exceptions/HttpStatusException.java
@@ -13,6 +13,7 @@ import org.apache.commons.lang3.tuple.Pair;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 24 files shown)


elide-core/src/main/java/com/yahoo/elide/core/exceptions/HttpStatusException.java
```diff
@@ -43,9 +44,13 @@ public HttpStatusException(int status, String message, String verboseMessage) {
     }
 
     public HttpStatusException(int status, String message, String verboseMessage, Throwable cause) {
+        this(status, message, cause, () -> verboseMessage);
```
clayreimann: Should we deprecate this version in favor of the `Supplier` version?

wcekan: We can, but probably cannot remove.

PR #458, opened 2017-05-03
Fix user permission check cache

aklish:


elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -289,7 +281,9 @@ public void executeCommitChecks() {
     @Override
     public boolean shouldShortCircuitPermissionChecks(Class<? extends Annotation> annotationClass,
                                                       Class resourceClass, String field) {
-        return expressionResultShortCircuit.contains(Triple.of(annotationClass, resourceClass, field));
+        ExpressionResult result = userPermissionCheckCache.get(Triple.of(annotationClass, resourceClass, field));
+
+        return (result == PASS);
```
wcekan: Can we short circuit only on `PASS` or can we also short circuit `FAIL`?

clayreimann: I think we can (and should).

DeathByTape:
@clayreimann: fail is actually an exception rather than a return code. @aklish noted offline that he will work to make this change, but cannot do it here. Similarly, it would be simpler to break it up to a separate PR.

PR #460, opened 2017-05-03
[Issue 459] Add String to UUID conversion

### LINKED ISSUE #459: UUID conversion fails
If an entity's `@Id` field is of type `UUID`, you can't access the route `/entity/{id}`. An exception is thrown `java.lang.String cannot be cast to java.util.UUID` because there is currently no conversion from String to UUID available.
Conversion in other direction works of course, so it probably only fails on using it as id or in filters.

Brutus5000:
fixes #459


aklish:
thanks!

PR #463, opened 2017-05-05
Refactor permission cache

aklish:


elide-core/src/main/java/com/yahoo/elide/security/PermissionExecutor.java
```diff
@@ -87,6 +87,7 @@
      * @param field Field
      * @return the results of evaluating the permission
      */
+    @Deprecated
```
clayreimann: We should deprecate these in the javadoc too (perhaps with the reason why)

PR #468, opened 2017-05-11
Account for inherited types in new object detection.

DeathByTape:


elide-core/src/test/java/com/yahoo/elide/core/RequestScopeTest.java
```diff
@@ -30,4 +35,39 @@ public void testFilterQueryParams() throws Exception {
...
+
+        String myId = "myId";
+        // Test that a new inherited class is counted for base type
+        requestScope.setUUIDForObject(dictionary.getJsonAliasFor(MyInheritedClass.class), myId, new MyInheritedClass());
+        Assert.assertNotNull(requestScope.getObjectById(dictionary.getJsonAliasFor(MyBaseClass.class), myId));
+    }
+
+    @Entity
+    @Include
+    public static class MyBaseClass {
```
aklish: Move this inside the test function.

elide-core/src/test/java/com/yahoo/elide/core/RequestScopeTest.java
```diff
@@ -30,4 +35,39 @@ public void testFilterQueryParams() throws Exception {
...
+        @Id
+        public long id;
+    }
+
+    @Entity
+    @Include
+    public static class MyInheritedClass extends MyBaseClass {
```
aklish: Same

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -379,4 +382,16 @@ private void bindTriggerIfPresent(Class<? extends Annotation> annotationClass, A
         }
         return annotation == NO_ANNOTATION ? null : (A) annotation;
     }
+
+    public static List<Class<?>> getInheritedTypes(Class<?> entityClass) {
```
aklish: Is there any reason this is here as opposed to the dictionary?

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -410,10 +411,28 @@ public String getUUIDFor(Object o) {
...
     public void setUUIDForObject(String type, String id, Object object) {
         objectEntityCache.put(type, id, object);
+
+        // Insert for all inherited entities as well
+        dictionary.getSuperClassEntityNames(type).stream()
+                .map(i -> getInheritanceKey(type, i))
+                .forEach((newType) -> objectEntityCache.put(newType, id, object));
+    }
+
+    private String getInheritanceKey(String rootType, String superClass) {
```
aklish: Should rootType be called subClass?

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -379,4 +382,16 @@ private void bindTriggerIfPresent(Class<? extends Annotation> annotationClass, A
...
+        Class<?> cls = entityClass.getSuperclass();
+        ArrayList<Class<?>> results = new ArrayList<>();
+
+        while (cls != null && cls != Object.class) {
```
wcekan: Would this look any cleaner as
```java
for (Class<?> cls = entityClass.getSuperclass(); cls != null && cls != Object.class; cls = cls.getSuperclass()) {
    results.add(cls);
}
```

PR #470, opened 2017-05-11
Allow application to process the uncaught exceptions.

wcekan:
While attempting to handle a custom exception that currently generates 500, I found it was not possible by over-riding `buildErrorResponse`.

Even an "unknown" exception should be converted into a 500 so the application can handle it correctly.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -42,7 +42,9 @@ import org.antlr.v4.runtime.tree.ParseTree;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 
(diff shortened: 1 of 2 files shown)


elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -224,8 +225,11 @@ protected ElideResponse handleRequest(boolean isReadOnly, Object opaqueUser,
         } catch (ParseCancellationException e) {
             return buildErrorResponse(new InvalidURLException(e), isVerbose);
 
-        } catch (RuntimeException | Error e) {
-            log.error("Exception uncaught by Elide", e);
+        } catch (RuntimeException e) {
```
DeathByTape: Probably want all exceptions. What if reflection blows up and a checked exception is thrown? We would probably ignore it?

PR #472, opened 2017-05-16
Remove unused parameter

QubitPi:
I found that `RequestScope` is not used at all in all `commit()` used across project.


clayreimann:
This is a breaking change, and while this parameter is unused that is, I believe, implementation specific.

DeathByTape:
I think we should close this PR. We have custom audit loggers internally that use the `RequestScope`. This is provided for the convenience of the implementor.

QubitPi:
OK

DeathByTape:
Thanks!

PR #473, opened 2017-05-16
Add support for sorting by id.

DeathByTape:


elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
```diff
@@ -1632,6 +1632,23 @@ public void testPostInvalidRelationship() {
                 .statusCode(HttpStatus.SC_BAD_REQUEST);
     }
 
+    @Test(priority = 42)
+    public void testSortByIdTopLevel() {
+        String sortByIdAscendingTopLevel = jsonParser.getJson("/ResourceIT/sortByTopLevelAscending.json");
+        given()
+                .accept(JSONAPI_CONTENT_TYPE)
+                .get("/parent?sort=id")
```
aklish: We should test more than root objects. We should also test relationships as well as non root entities.

aklish: Also, I was hoping we could refactor the IT tests so that we have one for sorting, one for pagination, one for filters, etc.

DeathByTape: Will add a test for non-rootable collections too but I suggest we punt on the total IT refactor (at least in this PR). That way we don't hold up this feature for too long.

PR #474, opened 2017-06-02
Changes to the permission evaluation

suriyasundar92:
The PR makes changes to the format of the debug string related to the permission evaluation that is returned in elide response. This helps in parsing the string in ralisplitter security testing framework easily.


suriyasundar92:
With respect to the SpecificFieldExpression, the parens are required only if there is an OR operand. Its not pertaining to FIELD or ENTITY terms. Since there is no OR operand in SpecificFieldExpression, we can go without the extra parens.

PR #476, opened 2017-06-08
Add data fetcher

clayreimann:
Should be merged by squashing. No useful history here.


elide-core/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
```diff
@@ -0,0 +1,104 @@
...
+    public Response post(
+            @Context SecurityContext securityContext,
+            String graphQLDocument) {
+        String path = "/";
+        boolean isVerbose = false;
+        try (DataStoreTransaction tx = elide.getDataStore().beginTransaction()) {
+            ObjectMapper mapper = elide.getMapper().getObjectMapper();
+            JsonNode jsonDocument = mapper.readTree(graphQLDocument);
+            final User user = tx.accessUser(getUser.apply(securityContext));
+            RequestScope requestScope = new RequestScope(path, null, tx, user, null, elide.getElideSettings());
```
aklish: Let's cast the nulls so we know what is not being passed here.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,244 @@
...
+        Optional<Sorting> sorting = Optional.empty();
+        Optional<Pagination> pagination = Optional.empty();
+
+        RequestScope requestScope = request.requestScope;
+        if (request.outputType instanceof GraphQLList) {
+            if (request.id.isPresent() && request.filters.isPresent()) {
+                throw new WebApplicationException("You may not filter when loading by id");
+            }
+
+            GraphQLObjectType graphQLType = (GraphQLObjectType) ((GraphQLList) request.outputType).getWrappedType();
```
aklish: What if its a list of scalars or a list of enums?

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,244 @@
...
+            String entityType = graphQLType.getName();
+            Class recordType = (Class) requestScope.getDictionary().getEntityClass(entityType);
+
+            if (recordType == null) {
+                throw new UnknownEntityException(entityType);
```
aklish: I believe getEntityClass never returns null.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,244 @@
...
+
+        RequestScope requestScope = request.requestScope;
+        if (request.outputType instanceof GraphQLList) {
+            if (request.id.isPresent() && request.filters.isPresent()) {
+                throw new WebApplicationException("You may not filter when loading by id");
+            }
+
+            GraphQLObjectType graphQLType = (GraphQLObjectType) ((GraphQLList) request.outputType).getWrappedType();
+            String entityType = graphQLType.getName();
+            Class recordType = (Class) requestScope.getDictionary().getEntityClass(entityType);
```
aklish: This (incorrectly) makes the assumption that a List implies the element type is an Elide entity.  We could have an attribute that is a list of integers.

GraphQL.md
```diff
@@ -0,0 +1,139 @@
...
+| Replace   | True         | True       | Overwrite all specified values in body (including id) |
+|           | True         | False      | Overwrite all specified values (no id in body to overwrite) |
+|           | False        | True       | Remove from collection except if id exists, it won't be created |
+|           | False        | False      | Totally new collection values |
+| Fetch     | True         | True       | Boom. |
+|           | True         | False      | Find single id. |
+|           | False        | True       | Boom. |
+|           | False        | False      | Find all. |
+| Delete    | True         | True       | Must match. Remove single. |
+|           | True         | False      | Must match. Remove single. |
```
aklish: What must match in this case?

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,244 @@
...
+        if (!request.id.isPresent() && request.data.size() > 1) {
+            throw new WebApplicationException("Id argument specification with an additional list of id's to delete "
+                    + "is unsupported", HttpStatus.SC_BAD_REQUEST);
+        }
+
+        EntityDictionary dictionary = request.requestScope.getDictionary();
+
+        String loadType = request.field.getName();
+        // TODO: This works at the root-level, but will it blow up in nested deletes?
+        String idFieldName = dictionary.getIdFieldName(dictionary.getEntityClass(loadType));
```
aklish: This will only work for deletes on root level collections.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,244 @@
...
+        String loadId = request.field.getArguments().stream()
```
aklish: I think we need to move some of this logic into the environment.  There should be common methods for getting the set of IDs that need to be loaded, etc.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,244 @@
...
+        return deleted;
+    }
+
+    private Object createObject(Environment request) {
+        EntityDictionary dictionary = request.requestScope.getDictionary();
+
+        GraphQLObjectType objectType;
+        String uuid = UUID.randomUUID().toString();
+        if (request.outputType instanceof GraphQLObjectType) {
+            // No parent
```
aklish: For this branch of the code, we still need to set all the attributes to whatever was provided in the data argument.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,244 @@
...
+        } else if (request.outputType instanceof GraphQLList) {
+            // Has parent
+            objectType = (GraphQLObjectType) ((GraphQLList) request.outputType).getWrappedType();
+            List<PersistentResource> container = new ArrayList<>();
+            for (Map<String, Object> input : request.data) {
+                Class<?> entityClass = dictionary.getEntityClass(objectType.getName());
+                // TODO: See above comment about UUID's.
+                PersistentResource toCreate = PersistentResource.createObject(null, entityClass, request.requestScope,
+                        uuid);
+                input.entrySet().stream()
```
aklish: There is missing support here for relationships.

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -0,0 +1,65 @@
...
+    public final Optional<String> id;
+    public final Object source;
+    public final PersistentResource parentResource;
+    public final GraphQLType parentType;
+    public final GraphQLType outputType;
+    public final Field field;
+    public final List<Map<String, Object>> data;
+    public final Optional<String> filters;
+
+    public Environment(DataFetchingEnvironment environment) {
```
aklish: I'm wondering if we should have multiple classes - one for each operation type (FETCH, REPLACE, ADD, DELETE) that encapsulates the constraints and request state of each operation.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,244 @@
...
+                return deleteObject(context);
+
+            case REPLACE:
+                return new UpdateOperation(context).execute();
+        }
+
+        throw new UnsupportedOperationException("Unknown operation: " + operation);
+    }
+
+    private void logContext(RelationshipOp operation, Environment environment) {
```
aklish: This method should move to the environment.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,244 @@
...
+
+        String loadType = request.field.getName();
+        // TODO: This works at the root-level, but will it blow up in nested deletes?
+        String idFieldName = dictionary.getIdFieldName(dictionary.getEntityClass(loadType));
+        String loadId = request.field.getArguments().stream()
+                .filter(arg -> ARGUMENT_DATA.equals(arg.getName()))
+                .findFirst()
+                .map(Argument::getChildren)
+                // TODO: Iterate over children and determine which contains id.
+                .map(List::toString) // TODO: place holder. remove this line.
```
aklish: There is a big assumption here about the format of the argument data (that it is a list of IDs).  This needs to be reworked.

elide-core/src/main/java/com/yahoo/elide/graphql/operations/UpdateOperation.java
```diff
@@ -0,0 +1,216 @@
...
+        if (isRoot) {
+            return new ArrayList<>(fetchRootableResourceCollection());
+        }
+        return new ArrayList<>(fetchNonRootableResourceCollection());
+    }
+
+    private Collection<PersistentResource> fetchNonRootableResourceCollection() {
+        // NOTE: Currently only handles _single_ object update
+        Optional<String> dataId = (id != null) ? Optional.of(id) : field.getArguments().stream()
+                .filter(arg -> ARGUMENT_DATA.equalsIgnoreCase(arg.getName()))
```
aklish: This code is very similar to some of the code that handles deletes.  I think this kind of logic should move to a Request Specific (Fetch, Delete, Add, Replace) class that provides functions to do common things like get the list of IDs.

elide-core/src/main/java/com/yahoo/elide/graphql/operations/UpdateOperation.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import static com.yahoo.elide.graphql.ModelBuilder.ARGUMENT_DATA;
+import static com.yahoo.elide.graphql.ModelBuilder.ARGUMENT_ID;
+
+/**
+ * Encapsulate the GraphQL update logic.
+ */
+public class UpdateOperation {
```
aklish: This class has a different structure than the other operations.  It would be nice if there was a common design pattern here.

elide-core/src/main/java/com/yahoo/elide/graphql/operations/UpdateOperation.java
```diff
@@ -0,0 +1,216 @@
...
+        entityClass = (isRoot)
+                ? dictionary.getEntityClass(field.getName())
+                : dictionary.getParameterizedType(((PersistentResource) source).getObject(), field.getName());
+        idFieldName = dictionary.getIdFieldName(entityClass);
+    }
+
+    public Object execute() {
+        Collection<PersistentResource> objectsToUpdate = fetchResourceCollection();
+
+        if (outputType instanceof GraphQLList) {
```
aklish: The outputType could be a single GraphQLOutputObject.  This assumption is not correct.

elide-core/src/main/java/com/yahoo/elide/graphql/operations/UpdateOperation.java
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    private void processActions(Collection<PersistentResource> objectsToUpdate, Map<String, Object> input) {
+        String entityId = (String) input.get(idFieldName);
+        boolean foundEntityId = entityId != null;
+        if (!foundEntityId) {
+            entityId = UUID.randomUUID().toString();
+        }
+        final String entityIdCapture = entityId;
+        input.entrySet().stream()
```
aklish: Regular for each is more readable I would say here.

elide-core/src/main/java/com/yahoo/elide/graphql/operations/UpdateOperation.java
```diff
@@ -0,0 +1,216 @@
...
+
+    public UpdateOperation(Environment environment) {
+        field = environment.field;
+        requestScope = environment.requestScope;
+        dictionary = requestScope.getDictionary();
+        id = environment.id.orElse(null);
+        source = environment.source;
+        outputType = environment.outputType;
+        requestData = environment.data;
+        isRoot = !(source instanceof PersistentResource);
```
aklish: Environment has an isRoot method.

elide-core/src/main/java/com/yahoo/elide/graphql/operations/UpdateOperation.java
```diff
@@ -0,0 +1,216 @@
...
+            boolean replaceRoot = checkIsReplacingRoot();
+            for (Map<String, Object> input : requestData) {
+                // Find the id's that we're trying to update
+                Set<String> objectIds = input.entrySet().stream()
+                        .filter(entry -> ARGUMENT_ID.equalsIgnoreCase(entry.getKey()))
+                        .map(entry -> (String) entry.getValue())
+                        .collect(Collectors.toSet());
+                if (replaceRoot) {
+                    // Delete all the things we don't want to keep...
+                    PersistentResource.loadRecords(entityClass, requestScope).stream()
```
aklish: We need to make sure this loadRecords is filtering out the records that cannot be read (ReadPermission security check).

elide-core/src/main/java/com/yahoo/elide/graphql/operations/UpdateOperation.java
```diff
@@ -0,0 +1,216 @@
...
+        idFieldName = dictionary.getIdFieldName(entityClass);
+    }
+
+    public Object execute() {
+        Collection<PersistentResource> objectsToUpdate = fetchResourceCollection();
+
+        if (outputType instanceof GraphQLList) {
+            GraphQLObjectType objectType = (GraphQLObjectType) ((GraphQLList) outputType).getWrappedType();
+            Set<PersistentResource> container = new HashSet<>();
+            boolean replaceRoot = checkIsReplacingRoot();
```
aklish: This feels like a strange way to determine if we are replacing a collection (or just modifying an element within a collection).  Maybe we need to tweak the data model to differentiate more explicitly?

GraphQL.md
clayreimann: I think this is a typo. Should read `Remove single by id.`

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -0,0 +1,65 @@
...
+        if (environment.getFields().size() != 1) {
+            throw new WebApplicationException("Resource fetcher contract has changed");
+        }
+        Map<String, Object> args = environment.getArguments();
+
+        requestScope = (RequestScope) environment.getContext();
```
clayreimann: In a refactor we should probably make `RequestScope` a `ThreadLocal` object, we currently pass it as an argument in a thousand and one places but we never alter it during the course of a request.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
clayreimann: Actually we don't. The way the datafetcher works we will get a call to `get(DataFetchingEnvironment)` for each attribute that is passed in

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
clayreimann: Probably a bug.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
clayreimann: It's just a proxy for `Map::get`

elide-core/src/main/java/com/yahoo/elide/graphql/operations/UpdateOperation.java
clayreimann: We were moving to this style to clean up the `PersistentResourceFetcher`, but the effort is young

elide-core/src/main/java/com/yahoo/elide/graphql/operations/UpdateOperation.java
clayreimann: Operations could (probably should) have a common superclass to encapsulate shared functions.

DennisMcWherter:
Is this branch stale now?

aklish:
Yes.  We can remove.

PR #480, opened 2017-06-16
Replaced ADD operator with UPSERT. Replaced id argument with ids.

aklish:


elide-core/src/test/java/com/yahoo/elide/graphql/ModelBuilderTest.java
```diff
@@ -33,6 +33,7 @@
     private EntityDictionary dictionary;
 
     private static final String ID = "id";
```
DeathByTape: Is this still in use? It appears as though `idArgument` has been renamed to `ids` everywhere. Is there a singular `id` remaining?

aklish: It is also an attribute for the @id field in the response.

PR #485, opened 2017-06-29
Add both GraphQL docs

javeria2:
This PR contains both the docs, `graphql-elide.md` and `graphql-elide-design.md`,
The former is the readme for the graphql-elide integration while the latter is the design doc for graphql-elide. A few comments - 
1) The doc has a couple _QUESTIONS_ which I had from an understanding perspective. (I'll remove these later before primetime)
2) Somethings (like `testcases`) haven't been fully filled yet. I'll do that once I come to them after finishing the current tasks.


graphql-elide-design.md
```diff
@@ -0,0 +1,411 @@
+# GraphQL Support for Elide
+
+## Overview
+
+The next development iteration for Elide focuses on seamless conversion from REST to a graphQL endpoint by simply mounting the provided endpoint to a specified path. Using GraphQL can be a more relevant approach in cases when, say, the user API is not using REST specific services like HATEOAS. Among others, it provides the following key advantages -
```
aklish: This implies we are doing away with JSON-API which is not true.  We are going to offer both.

graphql-elide-design.md
```diff
@@ -0,0 +1,411 @@
+# GraphQL Support for Elide
+
+## Overview
+
+The next development iteration for Elide focuses on seamless conversion from REST to a graphQL endpoint by simply mounting the provided endpoint to a specified path. Using GraphQL can be a more relevant approach in cases when, say, the user API is not using REST specific services like HATEOAS. Among others, it provides the following key advantages - 
+- Single endpoint to access all the data
```
aklish: I wouldn't qualify the differences here.  GraphQL has some pros & cons in comparison.

graphql-elide-design.md
```diff
@@ -0,0 +1,411 @@
...
+- No need to tailor specific endpoints for each view/collection
+- Flexibility to access only the data the client needs in a single request
```
aklish: Both JSON-API and GraphQL provide this.  Again, I wouldn't get into the tradeoffs in this doc.

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -0,0 +1,73 @@
...
+public class Environment {
+    public final RequestScope requestScope;
+    public final List<String> ids;
+    public final Optional<String> sort;
+    public final List<Map<String, Object>> data;
+    public final Optional<String> filters;
+    public final Optional<String> offset;
+    public final Optional<String> first;
+
+    public final Object source;
```
aklish: What's the difference between the 'source' and the 'parentResource'?

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Optional;
+
+/**
+ * Encapsulates GraphQL's DataFetchingEnvironment
+ */
+public class Environment {
+    public final RequestScope requestScope;
+    public final List<String> ids;
+    public final Optional<String> sort;
+    public final List<Map<String, Object>> data;
```
aklish: Why not just a list of GraphQLInputObjects?

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+/**
+ * Encapsulates GraphQL's DataFetchingEnvironment
+ */
+public class Environment {
+    public final RequestScope requestScope;
+    public final List<String> ids;
```
aklish: ids are optional

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
aklish: data is optional

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
aklish: Do we need source?  Can we just use parentResource?

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -0,0 +1,73 @@
...
+        List<Map<String, Object>> data = (List<Map<String, Object>>) args.get(ModelBuilder.ARGUMENT_DATA);
+        if(data != null) {
+            this.data = ImmutableList.copyOf(data);
+        } else {
+            this.data = ImmutableList.of();
+        }
+    }
+
+    public boolean isRoot() {
+        return source instanceof RequestScope;
```
aklish: Wouldn't this make more sense to parentResource is null or parentResource.isEmpty()?

elide-core/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -149,7 +149,7 @@ public  GraphQLList classToQueryMap(Class<?> keyClazz,
     }
 
     /**
-     * Creates a GraphQL Map Input type.  GraphQL doesn't support this natively.  We mimic
+     * Creates a GraphQL Map Input type.  PersistentResourceFetcher doesn't support this natively.  We mimic
```
aklish: The original comment was correct.  GraphQL doesn't support maps.

elide-core/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -121,7 +121,7 @@ public GraphQLEnumType classToEnumType(Class<?> enumClazz) {
     }
 
     /**
-     * Creates a GraphQL Map Query type.  GraphQL doesn't support this natively.  We mimic
+     * Creates a GraphQL Map Query type.  PersistentResourceFetcher doesn't support this natively.  We mimic
```
aklish: The original comment was correct.  GraphQL doesn't support maps.

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
aklish: We should consolidate source with parentResource.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -0,0 +1,93 @@
...
+     * @param environment environment encapsulating graphQL's request environment
+     */
+    private void logContext(RelationshipOp operation, Environment environment) {
+        List<Field> children = environment.field.getSelectionSet() != null
+                ? (List) environment.field.getSelectionSet().getChildren()
+                : new ArrayList<>();
+        String requestedFields = environment.field.getName() + (children.size() > 0
+                ? "(" + children.stream().map(Field::getName).collect(Collectors.toList()) + ")" : "");
+        GraphQLType parent = environment.parentType;
+        log.debug("{} {} fields for {} with parent {}<{}>",
```
aklish: We don't need to both source, and the parent details.  They are the same thing.

elide-core/src/main/java/com/yahoo/elide/graphql/RelationshipOp.java
```diff
@@ -12,6 +12,7 @@
 public enum RelationshipOp {
     FETCH,
     DELETE,
-    ADD,
-    REPLACE;
+    UPSERT,
```
aklish: We should merge with my other PR there.

graphql-elide-design.md
```diff
@@ -0,0 +1,406 @@
+# GraphQL Support for Elide
```
aklish: Let's remove this file.

PR #489, opened 2017-07-05
changed the way changeSpec is created

suriyasundar92:
Changed the arguments passed for constructing changeSpec in the method delFromCollection in PersistentResource. This is required for conforming to the general convention of representing a ToOne relationship with an object and a ToMany relationship with a collection.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -1358,21 +1358,22 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
             // Compute the ChangeSpec for the inverse relation and check whether or not we have access
             // to apply this change to that field.
             final Object originalValue = toDelete.getValueUnchecked(inverseField);
-            final Collection originalBidirectional;
+            final Object originalBidirectional;
+            final Object removedBidirectional;
 
             if (originalValue instanceof Collection) {
                 originalBidirectional = copyCollection((Collection) originalValue);
+                removedBidirectional = CollectionUtils
+                        .disjunction(Collections.singleton(this.getObject()), (Collection) originalBidirectional);
             } else {
-                originalBidirectional = Collections.singleton(originalValue);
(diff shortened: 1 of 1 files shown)


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1358,21 +1358,22 @@ protected void delFromCollection(
...
+            final Object removedBidirectional;
 
             if (originalValue instanceof Collection) {
                 originalBidirectional = copyCollection((Collection) originalValue);
+                removedBidirectional = CollectionUtils
+                        .disjunction(Collections.singleton(this.getObject()), (Collection) originalBidirectional);
             } else {
-                originalBidirectional = Collections.singleton(originalValue);
+                originalBidirectional = originalValue;
+                removedBidirectional = null;
```
DeathByTape: Why `null`? Can this be `Collections.emptySet()`?

clayreimann:
What is the motivation for this change? IIRC we model all relationships as sets (with `ToOne` collections being [`SingleElementSet`](https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/jsonapi/models/SingleElementSet.java)s)

suriyasundar92:
This was required for the usage of a a new permission added in elide-annotation.
https://github.com/yahoo/elide/blob/master/elide-annotations/src/main/java/com/yahoo/elide/security/checks/prefab/Common.java#L43

This issue could have been solved in the check logic by checking if the modified value is an empty set. However I got an impression that we tend to represent ToOne relations as a object and not a collection when I referred to the updateRelation method in PersistentResource.
https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java#L403
So I thought it makes sense to make it consistent.

The check was not working because of the way the relationships are represented in changeSpec specifically. If SingleElementSet is the way we would want the ToOne relationships has to be represented then we can make the changes in both the check object in dbAccessLayer and PersistentResource in elide.

PR #493, opened 2017-07-06
FETCH working minus pagination/filtering/sorting

javeria2:
In this PR you will find - 
1) Test cases for FETCH (taken from the `add-data-fetcher` branch), `Author`, `Book`, `Publisher` model also borrowed from `add-data-fetcher` branch. 
2) All the functionality for FETCH based on our revised model
3) TODOs marked for pagination/filtering/sorting
5) changes to `logback.xml` are there to display debug logs on the console 

All test cases pass except those dealing with pagination/filtering/sorting.


elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -64,6 +93,17 @@ public Object get(DataFetchingEnvironment environment) {
     }
 
     /**
+     * Checks whether sort/filter/pagination params are passed w unsupported operation
+     * @param environment environment encapsulating graphQL's request environment
+     */
+    private void filterSortPaginateSanityCheck(Environment environment) {
+        if(environment.filters.isPresent() || environment.sort.isPresent() || environment.offset.isPresent()
+                || environment.first.isPresent()) {
+            throw new IllegalArgumentException("Pagination/Filtering/Sorting is only supported with FETCH operation");
```
aklish: Down below we throw BadRequestException.  Here we throw IllegalArgumentException.  Let's be consistent.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +119,117 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+     * @return list of persistent resource objects
+     */
+    private Object fetchObjects(Environment context) {
+        /* sanity check for data argument w FETCH */
+        if(context.data.isPresent()) {
+            throw new BadRequestException("FETCH must not include data");
+        }
+
+        /* check whether current object has a parent or not */
+        if (context.isRoot() || context.outputType instanceof GraphQLList) {
```
aklish: I think the logic should be more like:
````
if (context.isRoot()) {
   fetch root object
} else {
   if (field is attribute) {
      fetch attribute
   } else {
      fetch relationship
   }
}```

File: elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java [hunk 3]
```diff
@@ -79,6 +119,117 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+
+            /* handle empty list of ids */
+            if(idList.isEmpty()) {
+                throw new IllegalArgumentException("Empty list passed to ids");
+            }
+
+            /* access records from internal db and return */
+            HashSet recordSet = new HashSet();
+            for (String id : idList) {
+                recordSet.add(PersistentResource.loadRecord(recordType, id, requestScope));
```
aklish: This will result in a bunch of round trip calls to the datastore/DB.  A better approach would be to add the list of IDS as filters and then just call loadRecords with the list of filters.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +119,117 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+     * @param first pagination first argument
+     * @param filters filter params
+     * @return list of persistent resource objects
+     */
+    private Object fetchObject(RequestScope requestScope, PersistentResource parentResource, Field field,
+                               GraphQLType outputType, Optional<List<String>> ids,
+                               Optional<String> sort, Optional<String> offset,
+                               Optional<String> first, Optional<String> filters) {
+
+        //TODO: Do we need P/F/S? If so, do it!
```
aklish: Yes - do it.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +119,117 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+
+        if(outputType instanceof GraphQLObjectType) {
+            DataStoreTransaction tx = requestScope.getTransaction();
+            Object obj =
+                    tx.getRelation(tx, parentResource.getObject(), field.getName(),
+                            null, null, null, requestScope);
+            return new PersistentResource(obj, parentResource, requestScope.getUUIDFor(obj), requestScope);
+        }
+
+        else if(outputType instanceof GraphQLScalarType || outputType instanceof GraphQLEnumType) {
```
aklish: Instead of checking the types, you should just check to see if this is an attribute (see above comment).  Regardless of type, we should just call parentResource.getAttribute(if this is an attribute)

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +119,117 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+                               GraphQLType outputType, Optional<List<String>> ids,
+                               Optional<String> sort, Optional<String> offset,
+                               Optional<String> first, Optional<String> filters) {
+
+        //TODO: Do we need P/F/S? If so, do it!
+
+        if(outputType instanceof GraphQLObjectType) {
+            DataStoreTransaction tx = requestScope.getTransaction();
+            Object obj =
+                    tx.getRelation(tx, parentResource.getObject(), field.getName(),
```
aklish: See earlier comment about logic..  The flow here is not correct.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -289,17 +289,26 @@ public boolean matchesId(String checkId) {
...
 
         if (shouldSkipCollection(loadClass, ReadPermission.class, requestScope)) {
             return Collections.emptySet();
         }
 
         Iterable list;
-        Optional<FilterExpression> filterExpression = requestScope.getLoadFilterExpression(loadClass);
+        Optional<FilterExpression> filter = filterExpression;
+        if(filterExpression.isPresent()) {
+            if(requestScope.getLoadFilterExpression(loadClass).isPresent()) {
```
aklish: getLoadFilterExpression is not a free operation.  Let's do it once and use the result in all these if/else checks.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +118,157 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+            Class parentClass = context.parentResource.getResourceClass();
+            String fieldName = context.field.getName();
+
+            if(dictionary.isAttribute(parentClass, fieldName)) { /* fetch attribute properties */
+                return fetchAttribute(context.parentResource, context.field);
+
+            } else if(dictionary.isRelation(parentClass, fieldName)){ /* fetch relationship properties */
+                return fetchRelationship(context);
+            } else {
+                throw new IllegalStateException("Unrecognized object type: " + context.outputType.getClass().getName());
```
aklish: Commonize with BadRequestException?

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +118,157 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+
+    /**
+     * Fetches an entity, can be root or a nested entity
+     * @param context environment encapsulating graphQL's request environment
+     * @return persistent resource object(s)
+     */
+    private Object fetchEntity(Environment context) {
+        Class recordType = getRecordType(context.requestScope, context.outputType);
+        return fetchObject(context.requestScope, recordType, context.ids, context.outputType,
+                context.requestScope.getDictionary(), context.sort,
```
aklish: I don't think you need to pass in both the request scope and the dictionary (since one can be obtained from the other).  Just pass in the request scope.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +118,157 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+     * @return persistence resource object(s)
+     */
+    private Object fetchRelationship(Environment context) {
+        EntityDictionary dictionary = context.requestScope.getDictionary();
+        Class parentClass = context.parentResource.getResourceClass();
+        String fieldName = context.field.getName();
+
+        /* check for toOne relationships */
+        if(dictionary.getRelationshipType(parentClass, fieldName).isToOne()) {
+            DataStoreTransaction tx = context.requestScope.getTransaction();
```
aklish: You need to go through the persistent resource (instead of the data store), to ensure security is invoked.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +118,157 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+
+            //TODO: paginate/filter/sort
+            return records;
+
+        } else { /* fetching by id(s) */
+            List<String> idList = ids.get();
+
+            /* handle empty list of ids */
+            if(idList.isEmpty()) {
+                throw new IllegalArgumentException("Empty list passed to ids");
```
aklish: We need to be more consistent with exceptions.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +118,157 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+            }
+
+            /* access records from internal db and return */
+            Optional<FilterExpression> filterExpression;
+            GraphQLObjectType graphQLType = (GraphQLObjectType) ((GraphQLList) outputType).getWrappedType();
+            String entityType = graphQLType.getName();
+            Class entityClass = dictionary.getEntityClass(entityType);
```
aklish: Isn't this the same thing as 'recordType' that was passed in?

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +118,157 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+
+            Class<?> idType = dictionary.getIdType(entityClass);
+
+            String idField = dictionary.getIdFieldName(entityClass);
+
+            String relation = dictionary.getJsonAliasFor(entityClass);
```
aklish: A better name here would be entityTypeName (instead of relation)

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -289,17 +289,26 @@ public boolean matchesId(String checkId) {
...
         Iterable list;
-        Optional<FilterExpression> filterExpression = requestScope.getLoadFilterExpression(loadClass);
+        Optional<FilterExpression> filter = filterExpression;
+        Optional<FilterExpression> loadFilterExpression = requestScope.getLoadFilterExpression(loadClass);
+        if(filterExpression.isPresent()) {
+            if(loadFilterExpression.isPresent()) {
+                filter = Optional.of(new AndFilterExpression(filterExpression.get(), loadFilterExpression.get()));
+            }
+        } else {
+            filter = requestScope.getLoadFilterExpression(loadClass);
```
aklish: filter = loadFilterExpression

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +117,151 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+     * @return persistence resource object(s)
+     */
+    private Object fetchRelationship(Environment context) {
+        EntityDictionary dictionary = context.requestScope.getDictionary();
+        Class parentClass = context.parentResource.getResourceClass();
+        String fieldName = context.field.getName();
+        String ID = "id";
+
+        /* check for toOne relationships */
+        if(dictionary.getRelationshipType(parentClass, fieldName).isToOne()) {
```
aklish: You don't need to differentiate the persistentResource call based on the relationshipType.

There is a single method which you should call for both toMany and toOne relationships.  That call will always return a Set.

What you can do instead is look at the relationship type to determine whether to return the single instance from that set or to return the set.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -79,6 +117,151 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+     * @param requestScope request scope
+     * @param recordType record type
+     * @param ids list of ids (can be NULL)
+     * @param sort sort by ASC/DESC
+     * @param offset pagination offset argument
+     * @param first pagination first argument
+     * @param filters filter params
+     * @return list of persistent resource objects
+     */
+    private Object fetchObject(RequestScope requestScope, Class recordType, Optional<List<String>> ids,
```
aklish: The naming of fetchEntity and fetchObject is inconsistent.  They are fetching the same thing - so I don't think we should name the function differently.

PR #496, opened 2017-07-13
Refactor multi edit transactions

aklish:


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -983,9 +984,20 @@ protected boolean checkRelation(String relationName) {
                 filterExpression = permissionFilter;
         }
 
-        Object val = transaction.getRelation(transaction, obj, relationName, filterExpression,
-                sortedAndPaginated ? sorting : Optional.empty(),
-                sortedAndPaginated ? pagination.map(p -> p.evaluate(relationClass)) : Optional.empty(), requestScope);
+        Object val;
+        if (requestScope.isMutatingMultipleEntities()) {
```
wcekan: Could this if and the above be unified?

PR #497, opened 2017-07-14
Patch extension read permission

wcekan:
Not all FilterExpressionChecks are able to run in memory for newly created objects during a Patch Extension create.  This is an attempt to avoid the failure in filter for these newly created objects.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -1039,8 +1039,10 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
 
         InMemoryFilterVisitor inMemoryFilterVisitor = new InMemoryFilterVisitor(requestScope);
         Predicate inMemoryFilterFn = filterExpression.get().accept(inMemoryFilterVisitor);
+        // NOTE: We can safely _skip_ tests on NEWLY created objects.
+        // We assume a user can READ their object they are allowed to create.
         return (Collection) collection.stream()
-                .filter(e -> inMemoryFilterFn.test(e))
+                .filter(e -> requestScope.isNewResource(e) || inMemoryFilterFn.test(e))
                 .collect(Collectors.toList());
     }
 
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -565,7 +565,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
     public boolean clearRelation(String relationName) {
         Set<PersistentResource> mine = filter(ReadPermission.class,
                 getRelationUncheckedUnfiltered(relationName), false);
-        checkFieldAwarePermissions(UpdatePermission.class, relationName, Collections.emptySet(),
+        checkFieldAwareDeferPermissions(UpdatePermission.class, relationName, Collections.emptySet(),
                 mine.stream().map(PersistentResource::getObject).collect(Collectors.toSet()));
 
         if (mine.isEmpty()) {
@@ -636,7 +636,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
             );
         }
 
-        checkFieldAwarePermissions(UpdatePermission.class, fieldName, modified, original);
+        checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, modified, original);
 
         if (relation instanceof Collection) {
             if (!((Collection) relation).contains(removeResource.getObject())) {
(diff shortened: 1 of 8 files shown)


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -819,13 +819,14 @@ public PersistentResource getRelation(String relation, String id) {
...
         if (requestScope.isMutatingMultipleEntities()
                         && requestScope.getObjectById(dictionary.getJsonAliasFor(entityType), id) != null) {
-            filterExpression = Optional.empty();
-            // NOTE: We can safely _skip_ tests here since we are only skipping READ checks on
-            // NEWLY created objects. We assume a user can READ their object in the midst of creation.
-            // Imposing a constraint to the contrary-- at this moment-- seems arbitrary and does not
-            // reflect reality (i.e. if a user is creating an object with values, he/she knows those values
-            // already).
-            skipNew = true;
+            if (!this.checkRelation(relation)) {
```
aklish: Let's remove skipNew as this is the only place this is being set.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -819,13 +819,14 @@ public PersistentResource getRelation(String relation, String id) {
...
+                throw new ForbiddenAccessException("oops");
```
aklish: Better message?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -819,13 +819,14 @@ public PersistentResource getRelation(String relation, String id) {
...
+            }
+            Object entity = requestScope.getObjectById(dictionary.getJsonAliasFor(entityType), id);
+            PersistentResource resource = requestScope.getNewPersistentResources().stream()
+                    .filter(r -> r.getObject() == entity).findFirst().orElse(null);
```
aklish: orElse we should throw an exception.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1040,7 +1041,7 @@ protected boolean checkRelation(String relationName) {
         InMemoryFilterVisitor inMemoryFilterVisitor = new InMemoryFilterVisitor(requestScope);
         Predicate inMemoryFilterFn = filterExpression.get().accept(inMemoryFilterVisitor);
         return (Collection) collection.stream()
-                .filter(e -> inMemoryFilterFn.test(e))
+                .filter(e -> requestScope.isNewResource(e) || inMemoryFilterFn.test(e))
```
aklish: Add some comments.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -245,10 +245,15 @@ protected RequestScope(String path, JsonApiDocument jsonApiDocument, RequestScop
         this.mutatingMultipleEntities = outerRequestScope.mutatingMultipleEntities;
     }
 
+    @Override
     public Set<com.yahoo.elide.security.PersistentResource> getNewResources() {
         return (Set<com.yahoo.elide.security.PersistentResource>) (Set) newPersistentResources;
     }
 
+    public boolean isNewResource(Object entity) {
+        return newPersistentResources.stream().filter(r -> r.getObject() == entity).findAny().orElse(null) != null;
```
aklish: isPresent?

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -116,6 +116,16 @@ public ActivePermissionExecutor(boolean verbose, final com.yahoo.elide.core.Requ
         };
 
         Function<Expression, ExpressionResult> expressionExecutor = (expression) -> {
+            // for newly created object in PatchRequest limit to User checks
```
aklish: I would say something like:

We only run User checks during patch extension newly created object because there is a good chance
that inline checks may fail (relationships are not yet fixed up).  If the user checks return deferred, we will defer all of the inline checks to commit phase.  It is safe to defer inline checks in this case because the
object was newly created and all values excluded computed attributes were provided by the client.

wcekan: Consolidated in function below.

elide-integration-tests/src/test/java/example/Child.java
```diff
@@ -152,4 +165,11 @@ public FilterExpression getFilterExpression(Class<?> entityClass, RequestScope r
             return new FilterPredicate(new PathElement(Child.class, "child", Long.class, "id"), Operator.NOTNULL);
         }
     }
+
+    static public class FailOpPatchExtension extends OperationCheck<Child> {
```
aklish: Let's make this FailOperationCheck

PR #499, opened 2017-07-17
Add UPSERT (Creating new objects)

javeria2:
Here you will find - 
- create part of _UPERT_ operation
- TODOs for update and fixing _FETCH_ for fetching nested entities by id 
- Test cases for _UPSERT_, will add more to this later on

### DIFF
--- a/elide-core/pom.xml
+++ b/elide-core/pom.xml
@@ -172,7 +172,12 @@
         </dependency>
         <dependency>
             <groupId>com.yahoo.elide</groupId>
-            <artifactId>dropwizard-elide-example</artifactId>
+            <artifactId>elide-core</artifactId>
+            <version>3.0.11</version>
+        </dependency>
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-core</artifactId>
             <version>3.0.7-SNAPSHOT</version>
         </dependency>
     </dependencies>
--- a/elide-core/pom.xml
+++ b/elide-core/pom.xml
@@ -170,6 +170,11 @@
             <artifactId>groovy-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>dropwizard-elide-example</artifactId>
+            <version>3.0.7-SNAPSHOT</version>
+        </dependency>
     </dependencies>
 
     <build>
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
@@ -46,8 +46,8 @@ import java.util.Queue;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArrayList;
-import java.util.stream.Collectors;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 /**
  * Entity Dictionary maps JSON API Entity beans to/from Entity type names.
@@ -862,6 +862,20 @@ public class EntityDictionary {
         }
 
         return Collections.emptyList();
+    }
+
+    /**
+     * Returns the parameterized type or the binding class for the given {@param identifier}.
+     * @param entityClass the entity class
+     * @param identifier the identifier
+     * @return the binding class
+     */
+    public Class<?> getLoadClass(Class<?> entityClass, String identifier) {
+        if(entityClass != null) {
+            return getParameterizedType(entityClass, identifier);
+        } else {
+            return getEntityClass(identifier);
+        }
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -112,7 +112,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
      * @param parent - The immediate ancestor in the lineage or null if this is a root.
      * @param entityClass the entity class
      * @param requestScope the request scope
-     * @param UUID the (optional) uuid
+     * @param uuid the (optional) uuid
      * @param <T> object type
      * @return persistent resource
      */
@@ -121,14 +121,14 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
             PersistentResource<?> parent,
             Class<T> entityClass,
             RequestScope requestScope,
-            Optional<String> UUID) {
+            Optional<String> uuid) {
 
         //instead of calling transcation.createObject, create the new object here.
         T obj = requestScope.getTransaction().createNewObject(entityClass);
 
-        String uuid = UUID.orElse(null);
+        String id = uuid.orElse(null);
 
-        PersistentResource<T> newResource = new PersistentResource<>(obj, parent, uuid, requestScope);
+        PersistentResource<T> newResource = new PersistentResource<>(obj, parent, id, requestScope);
 
         // Keep track of new resources for non shareable resources
         requestScope.getNewPersistentResources().add(newResource);
@@ -139,7 +139,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
         requestScope.queueTriggers(newResource, CRUDAction.CREATE);
 
         String type = newResource.getType();
-        requestScope.setUUIDForObject(type, uuid, newResource.getObject());
+        requestScope.setUUIDForObject(type, id, newResource.getObject());
 
         // Initialize null ToMany collections
         requestScope.getDictionary().getRelationships(entityClass).stream()
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -112,7 +112,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
      * @param parent - The immediate ancestor in the lineage or null if this is a root.
      * @param entityClass the entity class
      * @param requestScope the request scope
-     * @param uuid the uuid
+     * @param UUID the (optional) uuid
      * @param <T> object type
      * @return persistent resource
      */
@@ -121,10 +121,12 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
             PersistentResource<?> parent,
             Class<T> entityClass,
             RequestScope requestScope,
-            String uuid) {
+            Optional<String> UUID) {
 
         //instead of calling transcation.createObject, create the new object here.
         T obj = requestScope.getTransaction().createNewObject(entityClass);
+
+        String uuid = UUID.orElse(null);
 
         PersistentResource<T> newResource = new PersistentResource<>(obj, parent, uuid, requestScope);
 
@@ -161,7 +163,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
      */
     @Deprecated
     public static <T> PersistentResource<T> createObject(Class<T> entityClass, RequestScope requestScope, String uuid) {
-        return createObject(null, entityClass, requestScope, uuid);
+        return createObject(null, entityClass, requestScope, Optional.ofNullable(uuid));
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -159,7 +159,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
      * @param <T> type of resource
      * @return persistent resource
      * @deprecated Will be removed in Elide 4. Instead use
-     *  {@link PersistentResource#createObject(PersistentResource, Class, RequestScope, String)}
+     *  {@link PersistentResource#createObject(PersistentResource, Class, RequestScope, Optional<String>)}
      */
     @Deprecated
     public static <T> PersistentResource<T> createObject(Class<T> entityClass, RequestScope requestScope, String uuid) {
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -163,11 +163,11 @@ public class PersistentResourceFetcher implements DataFetcher {
      * @param idFieldName id field name
      */
     private static void setId(Map<String, Object> input, String idFieldName) {
-            String id = UUID.randomUUID().toString()
-                    .replaceAll("[^0-9]", "")
-                    .substring(0, 3); //limit the number of digits to prevent InvalidValueException in PersistentResource.createObject()
-            //TODO: this is hacky, ask for a workaround for this.
-            input.put(idFieldName, id);
+        String id = UUID.randomUUID().toString()
+                .replaceAll("[^0-9]", "")
+                .substring(0, 3); //limit the number of digits to prevent InvalidValueException in PersistentResource.createObject()
+        //TODO: this is hacky, ask for a workaround for this.
+        input.put(idFieldName, id);
     }
 
     /**
@@ -409,7 +409,7 @@ public class PersistentResourceFetcher implements DataFetcher {
                 Boolean isToOne = newParent.getRelationshipType(relationship.getKey()).isToOne();
                 List entryToAdd;
                 if(isToOne) {
-                     entryToAdd = new ArrayList<>();
+                    entryToAdd = new ArrayList<>();
                     entryToAdd.add(relationship.getValue());
                 } else {
                     entryToAdd = (List) relationship.getValue();
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -316,12 +316,13 @@ public class PersistentResourceFetcher implements DataFetcher {
                                 PersistentResource parentSource, Field field, GraphQLType outputType,
                                 RequestScope requestScope) {
         /* fetch id value fed to 'data' argument */
-        Optional<Map.Entry<String, Object>> id = input.entrySet().stream()
+        Optional<Object> id = input.entrySet().stream()
                 .filter(entry -> idFieldName.equalsIgnoreCase(entry.getKey()))
+                .map(Map.Entry::getValue)
                 .findFirst();
         Optional<String> idField = Optional.empty();
         if(id.isPresent()) {
-            idField = Optional.ofNullable((String) id.get().getValue());
+            idField = Optional.of((String) id.get());
         }
 
         Set<Object> fetchEntries;
@@ -340,7 +341,7 @@ public class PersistentResourceFetcher implements DataFetcher {
     }
 
     /**
-     * creates or updates a new/existing entity
+     * creates or updates a new/existing root-level entity
      * @param idFieldName field name
      * @param input input data object
      * @param requestScope request scope
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -73,14 +73,14 @@ public class PersistentResourceFetcher implements DataFetcher {
             case FETCH:
                 return fetchObjects(context);
 
+            case UPSERT:
+                return upsertObjects(context);
+
             case DELETE:
                 return deleteObjects(context);
 
             case REMOVE:
                 return removeObjects(context);
-
-            case UPSERT:
-                return upsertObjects(context);
 
             case REPLACE:
                 return replaceObjects(context);
@@ -290,7 +290,7 @@ public class PersistentResourceFetcher implements DataFetcher {
                 String idFieldName = dictionary.getIdFieldName(dictionary.getEntityClass(context.field.getName()));
 
                 upsertedObjects.add((PersistentResource) upsertObject(idFieldName, input, context.requestScope,
-                        context.field, context.outputType));
+                        context.outputType));
             } else {
                 String idFieldName = dictionary.getIdFieldName(context.parentResource.getResourceClass());
 
@@ -347,7 +347,7 @@ public class PersistentResourceFetcher implements DataFetcher {
      * @return list/set of persistent resource objects
      */
     private Object upsertObject(String idFieldName, Map<String, Object> input,
-                                RequestScope requestScope,  Field field, GraphQLType outputType) {
+                                RequestScope requestScope, GraphQLType outputType) {
         /* fetch id values fed to 'data' argument */
         Optional<Map.Entry<String, Object>> id = input.entrySet().stream()
                 .filter(entry -> idFieldName.equalsIgnoreCase(entry.getKey()))
@@ -471,13 +471,13 @@ public class PersistentResourceFetcher implements DataFetcher {
             throw new BadRequestException("REPLACE must include ids argument");
         }
 
-        Set<PersistentResource> toDelete = (Set<PersistentResource>) fetchObjects(context);
+        Set<PersistentResource> toRemove = (Set<PersistentResource>) fetchObjects(context);
         if(!context.isRoot()) { /* has parent */
-            toDelete.forEach(item -> context.parentResource.removeRelation(context.field.getName(), item));
+            toRemove.forEach(item -> context.parentResource.removeRelation(context.field.getName(), item));
         } else { /* is root */
             deleteObjects(context);
         }
-        return toDelete;
+        return toRemove;
     }
 
     /** stub code **/
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -130,8 +130,9 @@ public class PersistentResourceFetcher implements DataFetcher {
      * @return entityClass
      */
     private static Class<?> getEntityClass(Optional<PersistentResource> parent, EntityDictionary dictionary, String fieldName) {
-        return !parent.isPresent() ? getRecordType(dictionary, fieldName) :
-                dictionary.getParameterizedType(parent.get().getResourceClass(), fieldName);
+        return parent
+                .<Class<?>>map(persistentResource -> dictionary.getParameterizedType(persistentResource.getResourceClass(), fieldName))
+                .orElseGet(() -> getRecordType(dictionary, fieldName));
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -456,9 +456,30 @@ public class PersistentResourceFetcher implements DataFetcher {
         return toDelete;
     }
 
+    /**
+     * Removes a relationship, or deletes a root level resource
+     * @param context environment encapsulating graphQL's request environment
+     * @return set of removed persistent resource objects
+     */
+    private Object removeObjects(Environment context) {
+        /* sanity check for id and data argument w REPLACE */
+        if(context.data.isPresent()) {
+            throw new BadRequestException("REPLACE must not include data argument");
+        }
+
+        if(!context.ids.isPresent()) {
+            throw new BadRequestException("REPLACE must include ids argument");
+        }
+
+        Set<PersistentResource> toDelete = (Set<PersistentResource>) fetchObjects(context);
+        if(!context.isRoot()) { /* has parent */
+            toDelete.stream().forEach(item -> context.parentResource.removeRelation(context.field.getName(), item));
+        } else { /* is root */
+            deleteObjects(context);
+        }
+        return toDelete;
+    }
+
     /** stub code **/
     private Object replaceObjects(Environment context) { return null; }
-
-    private Object removeObjects(Environment context) { return null; }
-
 }
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -328,11 +328,11 @@ public class PersistentResourceFetcher implements DataFetcher {
             fetchEntries = (Set) fetchObject(dictionary, parentSource, field, outputType.getClass().getName());
         } catch (BadRequestException e) {
             log.debug(e.toString());
-            return createObject(input, outputType, requestScope, idField, field, parentSource);
+            return createObject(input, requestScope, idField, field, parentSource);
         }
 
         if(fetchEntries == null) {
-            return createObject(input, outputType, requestScope, idField, field, parentSource);
+            return createObject(input, requestScope, idField, field, parentSource);
         } else {
             return updateObject(fetchEntries, input);
         }
@@ -364,11 +364,11 @@ public class PersistentResourceFetcher implements DataFetcher {
                     Optional.empty(), Optional.empty(), Optional.empty());
         } catch(BadRequestException e) { /* no id(s) provided, must create new */
             log.debug(e.toString());
-            return createObject(input, outputType, requestScope, Optional.empty(), field, null);
+            return createObject(input, requestScope, Optional.empty(), outputType);
         }
 
         if(fetchEntries.isEmpty()) { /* empty set returned, must create new */
-            return createObject(input, outputType, requestScope, Optional.ofNullable(idField), field, null);
+            return createObject(input, requestScope, Optional.ofNullable(idField), outputType);
         } else { /* must update object */
             return updateObject(fetchEntries, input);
         }
@@ -382,51 +382,78 @@ public class PersistentResourceFetcher implements DataFetcher {
      * @param uuid uuid
      * @return persistent resource object
      */
-    private PersistentResource createObject(Map<String, Object> input, GraphQLType outputType,
-                                RequestScope requestScope, Optional<String> uuid,
-                                Field field, PersistentResource parent) {
+    private Object createObject(Map<String, Object> input, RequestScope requestScope,
+                                Optional<String> uuid, GraphQLType outputType) {
         GraphQLObjectType objectType = (GraphQLObjectType) ((GraphQLList) outputType).getWrappedType();
         EntityDictionary dictionary = requestScope.getDictionary();
-
-        Class<?> entityClass;
-        if(parent == null) {
-            entityClass = dictionary.getEntityClass(objectType.getName());
-        } else {
-            entityClass = dictionary.getParameterizedType(parent.getResourceClass(), field.getName());
-        }
-
-        PersistentResource toCreate = PersistentResource.createObject(parent, entityClass, requestScope, uuid);
-
+        Class<?> entityClass = dictionary.getEntityClass(objectType.getName());
+        Object toCreate = PersistentResource.createObject(null, entityClass, requestScope, uuid);
         input.entrySet().stream()
                 .filter(entry -> dictionary.isAttribute(entityClass, entry.getKey()))
-                .forEach(entry -> toCreate.updateAttribute(entry.getKey(), entry.getValue()));
+                .forEach(entry -> ((PersistentResource) toCreate).updateAttribute(entry.getKey(), entry.getValue()));
 
-        //add new relation between parent, updated object
-        if(parent != null) {
-            parent.addRelation(field.getName(), toCreate);
-        }
-
-//        return updateObject(toCreate, input); TODO: think of splitting this method into two based on parent == null or not
-        return toCreate;
+        return updateObject(new HashSet<>(Arrays.asList(toCreate)), input);
     }
 
     /**
+     * Creates a new "non root-level" resource in the database
+     * @param input individual input data objects
+     * @param requestScope request scope
+     * @param uuid uuid
+     * @param field field name
+     * @param parent parent resource
+     * @return persistent resource object
+     */
+    private Object createObject(Map<String, Object> input, RequestScope requestScope,
+                                Optional<String> uuid, Field field, PersistentResource parent) {
+        EntityDictionary dictionary = requestScope.getDictionary();
+        Class<?> entityClass = dictionary.getParameterizedType(parent.getResourceClass(), field.getName());
+        Object toCreate = PersistentResource.createObject(parent, entityClass, requestScope, uuid);
+        input.entrySet().stream()
+                .filter(entry -> dictionary.isAttribute(entityClass, entry.getKey()))
+                .forEach(entry -> ((PersistentResource) toCreate).updateAttribute(entry.getKey(), entry.getValue()));
+
+        //add new relation between parent, updated object
+        parent.addRelation(field.getName(), (PersistentResource) toCreate);
+
+        return updateObject(new HashSet<>(Arrays.asList(toCreate)), input);
+    }
+
+    /**
+     * Updates an object
      * @param toUpdate entities to update
      * @param input input data to update entities with
-     * @return
+     * @return updated object
      */
     private Object updateObject(Set<Object> toUpdate, Map<String, Object> input) {
+        Object obj = toUpdate.iterator().next();
         for(String key : input.keySet()) { /* iterate through each attribute provided */
-            /* safe loop over objects, in practice this will still be O(n) since toUpdate is always a single (set) object */
-            for(Object object : toUpdate) {
-                ((PersistentResource) object).updateAttribute(key, input.get(key));
-            }
+            ((PersistentResource) obj).updateAttribute(key, input.get(key));
         }
 
         //TODO: update relationships
-        
-        //toUpdate is always singleton since we're looping over each in upsertObjects()
-        return toUpdate.iterator().next();
+
+        return obj;
+    }
+
+    /**
+     * Deletes a resource
+     * @param context environment encapsulating graphQL's request environment
+     * @return set of deleted persistent resource objects
+     */
+    private Object deleteObjects(Environment context) {
+        /* sanity check for id and data argument w DELETE */
+        if(context.data.isPresent()) {
+            throw new BadRequestException("DELETE must not include data argument");
+        }
+
+        if(!context.ids.isPresent()) {
+            throw new BadRequestException("DELETE must include ids argument");
+        }
+
+        Set<PersistentResource> toDelete = (Set<PersistentResource>) fetchObjects(context);
+        toDelete.stream().forEach(item -> item.deleteResource());
+        return toDelete;
     }
 
     /** stub code **/
@@ -434,5 +461,4 @@ public class PersistentResourceFetcher implements DataFetcher {
 
     private Object removeObjects(Environment context) { return null; }
 
-    private Object deleteObjects(Environment context) { return null; }
 }
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -131,7 +131,7 @@ public class PersistentResourceFetcher implements DataFetcher {
                     context.offset, context.first, context.filters);
 
         } else { /* fetch attribute or relationship */
-            //TODO: fix fetch multiple nested entities by id
+            //TODO: fix "fetch" nested entities by id(s)
             return fetchObject(context.requestScope.getDictionary(), context.parentResource,
                     context.field, context.outputType.getClass().getName());
         }
@@ -269,9 +269,9 @@ public class PersistentResourceFetcher implements DataFetcher {
      * @return list of persistent resource objects
      */
     private Object upsertObjects(Environment context) {
-        /* sanity check for id and data argument w FETCH */
+        /* sanity check for id and data argument w UPSERT */
         if(context.ids.isPresent()) {
-            throw new BadRequestException("FETCH must not include ids");
+            throw new BadRequestException("UPSERT must not include ids");
         }
 
         if(!context.data.isPresent()) {
@@ -404,7 +404,4 @@ public class PersistentResourceFetcher implements DataFetcher {
     private Object removeObjects(Environment context) { return null; }
 
     private Object deleteObjects(Environment context) { return null; }
-    //TODO: next steps: implement UPSERT for non-root entities (see stub method above),
-    //TODO: implement create for non-root entities,
-    //TODO: handle update for root/non-root at the very end
 }
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -129,7 +129,7 @@ public class PersistentResourceFetcher implements DataFetcher {
      * @param fieldName graphql field name
      * @return entityClass
      */
-    private Class<?> getEntityClass(Optional<PersistentResource> parent, EntityDictionary dictionary, String fieldName) {
+    private static Class<?> getEntityClass(Optional<PersistentResource> parent, EntityDictionary dictionary, String fieldName) {
         return !parent.isPresent() ? getRecordType(dictionary, fieldName) :
                 dictionary.getParameterizedType(parent.get().getResourceClass(), fieldName);
     }
@@ -199,7 +199,7 @@ public class PersistentResourceFetcher implements DataFetcher {
      * @param fieldName field name
      * @return record type
      */
-    private Class<?> getRecordType(EntityDictionary dictionary, String fieldName) {
+    private static Class<?> getRecordType(EntityDictionary dictionary, String fieldName) {
         return dictionary.getEntityClass(fieldName);
     }
 
@@ -518,11 +518,8 @@ public class PersistentResourceFetcher implements DataFetcher {
                                 RequestScope requestScope, Optional<String> uuid,
                                             Optional<PersistentResource> parent) {
         EntityDictionary dictionary = requestScope.getDictionary();
-
         Class<?> entityClass = getEntityClass(parent, dictionary, fieldName);
-
         PersistentResource toCreate = PersistentResource.createObject(parent.orElse(null), entityClass, requestScope, uuid);
-
         //add new relation between parent and the updated object
         if(parent.isPresent()) {
             parent.get().addRelation(fieldName, toCreate);
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -26,10 +26,14 @@ import lombok.extern.slf4j.Slf4j;
 import javax.ws.rs.BadRequestException;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -131,12 +135,13 @@ public class PersistentResourceFetcher implements DataFetcher {
 
         /* check whether current object has a parent or not */
         if (context.isRoot()) {
-            return fetchObject(context.requestScope, context.outputType, context.ids, context.sort,
+            Class<?> loadClass = getRecordType(context.requestScope, context.outputType);
+            return fetchObject(context.requestScope, loadClass, context.ids, context.sort,
                     context.offset, context.first, context.filters);
 
         } else { /* fetch attribute or relationship */
             return fetchObject(context.requestScope.getDictionary(), context.parentResource,
-                    context.field, context.outputType.getClass().getName());
+                    context.field.getName(), context.outputType.getClass().getName());
         }
     }
 
@@ -148,11 +153,11 @@ public class PersistentResourceFetcher implements DataFetcher {
      */
     private Class getRecordType(RequestScope requestScope, GraphQLType outputType) {
         GraphQLObjectType graphQLType = (GraphQLObjectType) ((GraphQLList) outputType).getWrappedType();
+        String entityName = graphQLType.getName();
 
-        String entityType = graphQLType.getName();
-        Class recordType = requestScope.getDictionary().getEntityClass(entityType);
+        Class recordType = requestScope.getDictionary().getEntityClass(entityName);
         if (recordType == null) {
-            throw new UnknownEntityException(entityType);
+            throw new UnknownEntityException(entityName);
         } else {
             return recordType;
         }
@@ -161,7 +166,7 @@ public class PersistentResourceFetcher implements DataFetcher {
     /**
      * Fetches a root-level entity
      * @param requestScope request scope
-     * @param outputType GraphQL output type
+     * @param loadClass load class
      * @param ids list of ids (can be NULL)
      * @param sort sort by ASC/DESC
      * @param offset pagination offset argument
@@ -169,15 +174,14 @@ public class PersistentResourceFetcher implements DataFetcher {
      * @param filters filter params
      * @return {@link PersistentResource} object(s)
      */
-    private Object fetchObject(RequestScope requestScope, GraphQLType outputType, Optional<List<String>> ids,
-                               Optional<String> sort, Optional<String> offset,
-                               Optional<String> first, Optional<String> filters) {
-        Class recordType = getRecordType(requestScope, outputType);
-
+    private Set<PersistentResource> fetchObject(RequestScope requestScope, Class loadClass,
+                                                Optional<List<String>> ids, Optional<String> sort,
+                                                Optional<String> offset, Optional<String> first,
+                                                Optional<String> filters) {
         /* fetching a collection */
         if(!ids.isPresent()) {
 
-            Set<PersistentResource> records = PersistentResource.loadRecords(recordType, requestScope, Optional.empty());
+            Set<PersistentResource> records = PersistentResource.loadRecords(loadClass, requestScope, Optional.empty());
 
             //TODO: paginate/filter/sort
             return records;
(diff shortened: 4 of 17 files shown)


elide-core/pom.xml
```diff
@@ -170,6 +170,11 @@
             <artifactId>groovy-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>dropwizard-elide-example</artifactId>
+            <version>3.0.7-SNAPSHOT</version>
```
aklish: You'll want 3.0.11 to get the changes we talked about.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +269,145 @@ private Object fetchRelationship(Environment context) {
...
+            throw new BadRequestException("UPSERT must not include ids");
+        }
+
+        if(!context.data.isPresent()) {
+            throw new BadRequestException("UPSERT must include data argument");
+        }
+
+        Set<PersistentResource> upsertedObjects = new HashSet<>();
+        for(Map<String, Object> input : context.data.get()) {
+            EntityDictionary dictionary = context.requestScope.getDictionary();
```
aklish: You can do this outside the loop

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +269,145 @@ private Object fetchRelationship(Environment context) {
...
+
+
+            if(context.isRoot()) {
+                /* get internal reference idFieldName, 'id' in this case */
+                String idFieldName = dictionary.getIdFieldName(dictionary.getEntityClass(context.field.getName()));
+
+                upsertedObjects.add((PersistentResource) upsertObject(idFieldName, input,
```
aklish: Should upsertObject return a PersistentResource so you don't have to cast here?

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +265,144 @@ private Object fetchRelationship(Environment context) {
...
+     * @param parentSource parent persistent resource object
+     * @param field graphql-java field type
+     * @param outputType graphql-java output type
+     * @return list/set of persistent resource objects
+     */
+    private PersistentResource upsertObject(String idFieldName, Map<String, Object> input, EntityDictionary dictionary,
+                                PersistentResource parentSource, Field field, GraphQLType outputType,
+                                RequestScope requestScope) {
+        /* fetch id valued fed to 'data' argument */
+        Set<String> ids = input.entrySet().stream().filter(entry -> idFieldName.equalsIgnoreCase(entry.getKey()))
```
aklish: There should be a single ID field - so instead of collecting the results in a set, you could just findFirst after the filter.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +265,144 @@ private Object fetchRelationship(Environment context) {
...
+                .map(entry -> (String) entry.getValue())
+                .collect(Collectors.toSet());
+
+        try {
+            Object fetchEntries = fetchObject(dictionary, parentSource, field, outputType.getClass().getName());
```
aklish: findFirst returns an optional.  If that is empty, you can just create the object.  Otherwise proceed as below.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +265,144 @@ private Object fetchRelationship(Environment context) {
...
+            if(fetchEntries == null) {
+                return createObject(input, outputType, requestScope, Optional.of(ids), field, parentSource);
+            } else {
+                return updateObject(); //TODO: implement update
+            }
+        } catch (BadRequestException e) {
```
aklish: Close the try block right after fetch so it is explicit this exception handling only applies for the fetch.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +265,144 @@ private Object fetchRelationship(Environment context) {
...
+            String idFieldName = dictionary.getIdFieldName(dictionary.getEntityClass(field.getName()));
+            entityClass = dictionary.getParameterizedType(parent.getResourceClass(), field.getName());
+            Optional<String> parentId = Optional.ofNullable(String.valueOf(PersistentResource.getValue(parent.getObject(),
+                    idFieldName, requestScope)));
+            toCreate = PersistentResource.createObject(parent, entityClass,  requestScope, parentId);
+        }
+
+        input.entrySet().stream()
+                .filter(entry -> dictionary.isAttribute(entityClass, entry.getKey()))
+                .forEach(entry -> toCreate.updateAttribute(entry.getKey(), entry.getValue()));
```
aklish: We should explicitly add the object to its parent relationship.  We should also update the relationships as well that were passed in the data input object.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +265,144 @@ private Object fetchRelationship(Environment context) {
...
+        PersistentResource toCreate;
+
+        if(parent == null) {
+            entityClass = dictionary.getEntityClass(objectType.getName());
+            toCreate = PersistentResource.createObject(null, entityClass, requestScope, uuid);
+
+        } else {
+            String idFieldName = dictionary.getIdFieldName(dictionary.getEntityClass(field.getName()));
+            entityClass = dictionary.getParameterizedType(parent.getResourceClass(), field.getName());
+            Optional<String> parentId = Optional.ofNullable(String.valueOf(PersistentResource.getValue(parent.getObject(),
```
aklish: PersistentResource has a method called getId.  You can just call that here instead.

aklish: In fact, you don't need to fetch the parentId.  The parentId parameter should just be the uuid

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +265,144 @@ private Object fetchRelationship(Environment context) {
...
+                    idFieldName, requestScope)));
+            toCreate = PersistentResource.createObject(parent, entityClass,  requestScope, parentId);
```
aklish: replace parentId with uuid.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +265,144 @@ private Object fetchRelationship(Environment context) {
...
+        GraphQLObjectType objectType = (GraphQLObjectType) ((GraphQLList) outputType).getWrappedType();
+        EntityDictionary dictionary = requestScope.getDictionary();
+
+        Optional<String> uuid = ids.map(strings -> strings.iterator().next());
+        Class<?> entityClass;
+        PersistentResource toCreate;
+
+        if(parent == null) {
+            entityClass = dictionary.getEntityClass(objectType.getName());
+            toCreate = PersistentResource.createObject(null, entityClass, requestScope, uuid);
```
aklish: There is a function in PersistentResource that doesn't require a parent.  You can just remove the first null param.

javeria2: That method is deprecated in Elide 4, java-docs suggest to use the one with parent param instead.

javeria2: I took the call to `createObject()` outside the if-else block so I can just specify the `parent` param as-is without having to use the other method.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,254 @@ private Object fetchRelationship(Environment context) {
...
+
+        Set<PersistentResource> upsertedObjects = new HashSet<>();
+        EntityDictionary dictionary = context.requestScope.getDictionary();
+
+        for(Map<String, Object> input : context.data.get()) {
+            if(context.isRoot()) {
+                /* get internal reference idFieldName, 'id' in this case */
+                String idFieldName = dictionary.getIdFieldName(dictionary.getEntityClass(context.field.getName()));
+
+                upsertedObjects.add((PersistentResource) upsertObject(idFieldName, input, context.requestScope,
```
aklish: For every function in this file that returns object (excluding the @Override function for the DataFetcher), why not return a PersistentResource or Set of them where appropriate?  Then you don't need to cast in places like this.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,254 @@ private Object fetchRelationship(Environment context) {
...
+     * @param parentSource parent {@link PersistentResource} object
+     * @param field graphql-java field type
+     * @param outputType graphql-java output type
+     * @return list/set of {@link PersistentResource} objects
+     */
+    private Object upsertObject(String idFieldName, Map<String, Object> input, EntityDictionary dictionary,
+                                PersistentResource parentSource, Field field, GraphQLType outputType,
+                                RequestScope requestScope) {
+        /* fetch id value fed to 'data' argument */
+        Optional<Map.Entry<String, Object>> id = input.entrySet().stream()
```
aklish: Could this be simplified like:
```
Optional<String> idField = input.enrtySet().stream().filter(...).map(EntrySet::getValue).findFirst()

Comment (javeria2) [inline elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java]: Umm, that would return an `Optional<Object>` type. To get to `Optional<String>`, I'd still have to check if this value is present if so, extract it out and cast as a `String`. So the if-else block right after this line would still be needed. Is there any other way of simplifying without going through all of this?

File: elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java [hunk 13]
```diff
@@ -263,12 +267,255 @@ private Object fetchRelationship(Environment context) {
...
+     * @param outputType graphql-java output type
+     * @return list/set of {@link PersistentResource} objects
+     */
+    private Object upsertObject(String idFieldName, Map<String, Object> input, EntityDictionary dictionary,
+                                PersistentResource parentSource, Field field, GraphQLType outputType,
+                                RequestScope requestScope) {
+        /* fetch id value fed to 'data' argument */
+        Optional<Object> id = input.entrySet().stream()
+                .filter(entry -> idFieldName.equalsIgnoreCase(entry.getKey()))
+                .map(Map.Entry::getValue)
```
aklish: You can do the case to (String) here - and then the return type can be ```Optional<String>```

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,255 @@ private Object fetchRelationship(Environment context) {
...
+                .findFirst();
+        Optional<String> idField = Optional.empty();
+        if(id.isPresent()) {
+            idField = Optional.of((String) id.get());
+        }
+
+        Set<Object> fetchEntries;
+        try {
+            fetchEntries = (Set<Object>) fetchObject(dictionary, parentSource, field, outputType.getClass().getName());
+        } catch (BadRequestException e) {
```
aklish: I wonder if this is too broad.  If the ID doesn't exist, shouldn't this throw a NotFoundException?  BadRequest encompasses a whole swath of issues.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,255 @@ private Object fetchRelationship(Environment context) {
...
+     * @param idFieldName field name
+     * @param input input data object
+     * @param requestScope request scope
+     * @param outputType graphql-java output type
+     * @return list/set of {@link PersistentResource} objects
+     */
+    private Object upsertObject(String idFieldName, Map<String, Object> input,
+                                RequestScope requestScope, GraphQLType outputType) {
+        /* fetch id values fed to 'data' argument */
+        Optional<Map.Entry<String, Object>> id = input.entrySet().stream()
```
aklish: You can simplify this in the same way as you did for the other upsert function.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,255 @@ private Object fetchRelationship(Environment context) {
...
+            idField = (String) id.get().getValue();
+        }
+
+        Set<Object> fetchEntries;
+        try {
+            fetchEntries = (Set) fetchObject(requestScope, outputType,
+                    Optional.ofNullable(Arrays.asList(idField)), Optional.empty(),
+                    Optional.empty(), Optional.empty(), Optional.empty());
+        } catch(BadRequestException e) { /* no id(s) provided, must create new */
+            log.debug(e.toString());
```
aklish: I'm not sure we need to recover from a BadRequest here.  It is probably simpler to just propagate the exception.

aklish: You can instead just create the object if the ID is missing in the provided data.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,255 @@ private Object fetchRelationship(Environment context) {
...
+        Class<?> entityClass = dictionary.getParameterizedType(parent.getResourceClass(), field.getName());
+        Object toCreate = PersistentResource.createObject(parent, entityClass, requestScope, uuid);
+        input.entrySet().stream()
+                .filter(entry -> dictionary.isAttribute(entityClass, entry.getKey()))
+                .forEach(entry -> ((PersistentResource) toCreate).updateAttribute(entry.getKey(), entry.getValue()));
+
+        //add new relation between parent, updated object
+        parent.addRelation(field.getName(), (PersistentResource) toCreate);
+
+        return updateObject(new HashSet<>(Arrays.asList(toCreate)), input);
```
aklish: This looks redundant to the above updateAttribute loop.

aklish: You need to update the relationships next.  To do that, you'll need to extract the value of the data object and recurse by calling upsert again

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,255 @@ private Object fetchRelationship(Environment context) {
...
+    }
+
+    /**
+     * Updates an object
+     * @param toUpdate entities to update
+     * @param input input data to update entities with
+     * @return updated object
+     */
+    private Object updateObject(Set<Object> toUpdate, Map<String, Object> input) {
```
aklish: I don't think you need this.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,255 @@ private Object fetchRelationship(Environment context) {
...
+        return deleteObjects(toDelete);
+    }
+
+    /**
+     * Utility method to delete all {@link PersistentResource} objects in set {@param toDelete} from data-store
+     * @param toDelete set of {@link PersistentResource} objects to delete
+     * @return deleted {@link PersistentResource} objects
+     */
+    private Object deleteObjects(Set<PersistentResource> toDelete) {
+        toDelete.forEach(PersistentResource::deleteResource);
```
aklish: I would just fold this into the above function.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,255 @@ private Object fetchRelationship(Environment context) {
...
+
+        if(!context.ids.isPresent()) {
+            throw new BadRequestException("REPLACE must include ids argument");
+        }
+
+        Set<PersistentResource> toRemove = (Set<PersistentResource>) fetchObjects(context);
+        if(!context.isRoot()) { /* has parent */
+            toRemove.forEach(item -> context.parentResource.removeRelation(context.field.getName(), item));
+        } else { /* is root */
+            return deleteObjects(toRemove);
```
aklish: Expand deleteObjects as I mentioned above.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
javeria2: changed it such that there's no further bubbling of exceptions.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
javeria2: Yup, moved call to `updateAttribute()` to `updateObjects()` so that any call to updates happen only in one method.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,382 @@ private Object fetchRelationship(Environment context) {
...
+     * Strips out the attributes from the {@code element} singleton list and returns just the relationships, if present.
+     * @param element data input
+     * @param entityClass entity class
+     * @param dictionary entity dictionary
+     * @return relationship map
+     */
+    private Map<String, Object> stripAttributes(List<Map<String, Object>> element, Class<?> entityClass,
+                                                EntityDictionary dictionary) {
+        Map<String, Object> relationshipsOnly = new HashMap<>();
+        Map<String, Object> entry = element.get(0); /* 'element' will always be singleton */
```
aklish: At this level of function (low level helper), it doesn't have the context to know that the list will only have a single element.  That decision should happen somewhere else.  I would change this function to not take a list of maps, but rather just a single map.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,382 @@ private Object fetchRelationship(Environment context) {
...
+    }
+
+    /**
+     * Strips out the relationships from the {@code element} singleton list and returns just the attributes, if present.
+     * @param element data input
+     * @param entityClass entity class
+     * @param dictionary entity dictionary
+     * @return attributes map
+     */
+    private Map<String, Object> stripRelationships(List<Map<String, Object>> element, Class<?> entityClass,
```
aklish: I think this should take a Map instead of a List of maps.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,382 @@ private Object fetchRelationship(Environment context) {
...
+    }
+
+    /**
+     * Strips out the attributes from the {@code element} singleton list and returns just the relationships, if present.
+     * @param element data input
+     * @param entityClass entity class
+     * @param dictionary entity dictionary
+     * @return relationship map
+     */
+    private Map<String, Object> stripAttributes(List<Map<String, Object>> element, Class<?> entityClass,
```
aklish: Method can be static

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
aklish: Method can be static

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,382 @@ private Object fetchRelationship(Environment context) {
...
+     * @param toUpdate entities to update
+     * @param input input data to update entities with
+     * @return updated object
+     */
+    private PersistentResource updateObject(Set<PersistentResource> toUpdate, Map<String, Object> input,
+                                String fieldName, RequestScope requestScope,
+                                            PersistentResource parent) {
+        EntityDictionary dictionary = requestScope.getDictionary();
+        Class<?> entityClass;
+        if(parent == null) {
```
aklish: Use Optional instead of null here.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +267,382 @@ private Object fetchRelationship(Environment context) {
...
+
+        String idFieldName = dictionary.getIdFieldName(entityClass);
+        PersistentResource obj = toUpdate.iterator().next();
+        for(Map.Entry<String, Object> row : input.entrySet()) {
+            if(dictionary.isAttribute(entityClass, row.getKey())) { /* iterate through each attribute provided */
+                obj.updateAttribute(row.getKey(), row.getValue());
+            } else if(Objects.equals(row.getKey(), idFieldName)) { /* update 'id' attribute */
+                obj.updateAttribute(row.getKey(), row.getValue());
+            } else {
+                throw new BadRequestException("Unrecognized attribute passed to 'data': " + row.getKey());
```
aklish: I would throw an IllegalStateException.   This should never happen (and would require a bug in your code somewhere).  A BadRequestException is meant to be propagated back to the client as a 400 error (they did something wrong).  Instead, this should return a 500 error (we did something wrong).

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+     * @param uuid uuid
+     * @return {@link PersistentResource} object
+     */
+    private PersistentResource createObject(Map<String, Object> input, String fieldName,
+                                RequestScope requestScope, Optional<String> uuid,
+                                GraphQLType outputType, PersistentResource parent) {
+        EntityDictionary dictionary = requestScope.getDictionary();
+
+        Class<?> entityClass;
+        if(parent == null) {
```
aklish: Use Optional instead of null.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+            GraphQLObjectType objectType = (GraphQLObjectType) ((GraphQLList) outputType).getWrappedType();
```
aklish: Can we do the same thing we did for updateObject where we use the fieldName to extract the entityClass:
```
entityClass = dictionary.getEntityClass(fieldName);
```

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+        } else {
+            entityClass = dictionary.getParameterizedType(parent.getResourceClass(), fieldName);
+        }
+        PersistentResource toCreate = PersistentResource.createObject(parent, entityClass, requestScope, uuid);
+
+        //add new relation between parent and the updated object
+        if(parent != null) {
+            parent.addRelation(fieldName, toCreate);
+        }
+        return updateObject(new HashSet<>(Arrays.asList(toCreate)), input, fieldName, requestScope, parent);
```
aklish: You need to strip the relationships don't you?

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+        Set<PersistentResource> existingObjects = (Set<PersistentResource>) fetchObjects(context);
+        Set<PersistentResource> upsertedObjects = upsertObjects(context);
+        Set<PersistentResource> toDelete = Sets.difference(existingObjects, upsertedObjects);
+
+        if(!context.isRoot()) { /* has parent */
+            toDelete.forEach(item -> context.parentResource.removeRelation(context.field.getName(), item));
+        } else { /* is root */
+            toDelete.forEach(PersistentResource::deleteResource);
+        }
+        return toDelete;
```
aklish: This should return what remains - not what was deleted.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+     * updates or creates existing/new entities
+     * @param input data input
+     * @param id id
+     * @param requestScope request scope
+     * @param outputType graphql output type
+     * @param fieldName field name
+     * @param parent parent resource
+     * @return {@link PersistentResource} object
+     */
+    private PersistentResource updateOrCreateAttributes(Map<String, Object> input, Optional<String> id,
```
aklish: The name here is odd - we never 'create' attributes.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+        try {
+            if(parent == null) {
+                fetchEntries = fetchObject(requestScope, loadClass,
+                        Optional.of(Arrays.asList(id.orElse(null))), Optional.empty(),
+                        Optional.empty(), Optional.empty(), Optional.empty());
+            } else {
+                fetchEntries = (Set<PersistentResource>) fetchRelationship(parent, fieldName);
+            }
+        } catch (BadRequestException e) {
+            throw e;
```
aklish: YOu can remove this try/catch block.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+        }
+
+        Set<PersistentResource> fetchEntries;
+        try {
+            if(parent == null) {
+                fetchEntries = fetchObject(requestScope, loadClass,
+                        Optional.of(Arrays.asList(id.orElse(null))), Optional.empty(),
+                        Optional.empty(), Optional.empty(), Optional.empty());
+            } else {
+                fetchEntries = (Set<PersistentResource>) fetchRelationship(parent, fieldName);
```
aklish: This seems odd to me.  If we are doing a fetch, we should be doing it by ID.  Not fetching the entire collection.  We should be calling a fetch function that only returns a single item (in both cases)

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+    /**
+     * handles both updating/creating by first handling the attributes and then the relationships, queuing up
+     * any further nested relationships
+     * @param input data input
+     * @param parent parent
+     * @param requestScope request scope
+     * @param fieldName field name
+     * @return set of persistent resource objects
+     */
+    private Set<PersistentResource> inputObjectToUpsertOrUpdateOperations(Map<String, Object> input, Optional<String> id,
```
aklish: They cyclomatic complexity of this function is 9.  It might make sense to break this up a bit.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+            /* loop through each relation */
+            for(Map.Entry<String, Object> relationship : relationshipsOnly.entrySet()) {
+                Object value = relationship.getValue();
+                List<Map<String, Object>> currentRelationship;
+                Class<?> resourceClass = triplet.parentResource == null ?
+                        dictionary.getParameterizedType(entityClass, relationship.getKey()) :
+                        dictionary.getParameterizedType(triplet.parentResource.getResourceClass(), relationship.getKey());
+
+                /* handle toOne relationship */
+                if(!(value instanceof List)) {
```
aklish: Use the entity dictionary instead of instanceof to determine toOne vs toMany

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+            PersistentResource newParent = updateOrCreateAttributes(attributesOnly, id, requestScope, entityClass,
+                    outputType, triplet.fieldName, triplet.parentResource);
+
+            returnOps.add(newParent);
+
+            /* loop through each relation */
+            for(Map.Entry<String, Object> relationship : relationshipsOnly.entrySet()) {
+                Object value = relationship.getValue();
+                List<Map<String, Object>> currentRelationship;
+                Class<?> resourceClass = triplet.parentResource == null ?
```
aklish: Let's add a function to the entityDictionary that returns the entity Type of a relationship.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
+
+            /* new parent will be the resource we create from the stripped attributes */
+            PersistentResource newParent = updateOrCreateAttributes(attributesOnly, id, requestScope, entityClass,
+                    outputType, triplet.fieldName, triplet.parentResource);
+
+            returnOps.add(newParent);
+
+            /* loop through each relation */
+            for(Map.Entry<String, Object> relationship : relationshipsOnly.entrySet()) {
+                Object value = relationship.getValue();
```
aklish: This function can be greatly simplified by simply adding the relationship entity to the toVisit list.  You don't need to do anything else inside this loop.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
aklish: See my later comment about simplifying the relationship logic.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +266,382 @@ private Object fetchRelationship(Environment context) {
...
 
-    private Object deleteObjects(Environment context) { return null; }
-}
+        Set upsertOps = new HashSet();
+        for(Map<String, Object> input : context.data.get()) {
+            String idFieldName = context.isRoot() ?
+                    dictionary.getIdFieldName(dictionary.getEntityClass(context.field.getName())) :
+                    dictionary.getIdFieldName(context.parentResource.getResourceClass());
+
+            /* fetch id value fed to 'data' argument */
```
aklish: This doesn't match the pseudocode we discussed earlier.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -213,48 +198,66 @@ private Object fetchObject(RequestScope requestScope, Class recordType, Optional
...
      * Fetches a relationship for a top-level entity
-     * @param context environment encapsulating graphQL's request environment
+     * @param parentResource parent object
      * @return persistence resource object(s)
      */
-    private Object fetchRelationship(Environment context) {
-        String fieldName = context.field.getName();
-
-        Set<PersistentResource> relations = context.parentResource.getRelationCheckedFiltered(fieldName);
+    private Object fetchRelationship(PersistentResource parentResource, String fieldName) {
```
aklish: It is odd for a function to return a Set in one instance and an single Object in another.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -131,21 +135,13 @@ private Object fetchObjects(Environment context) {
 
         /* check whether current object has a parent or not */
         if (context.isRoot()) {
-            return fetchObject(context);
+            Class<?> loadClass = getRecordType(context.requestScope, context.outputType);
+            return fetchObject(context.requestScope, loadClass, context.ids, context.sort,
```
aklish: Why did we switch out the old code back to this?

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -157,46 +153,35 @@ private Object fetchObjects(Environment context) {
      */
     private Class getRecordType(RequestScope requestScope, GraphQLType outputType) {
         GraphQLObjectType graphQLType = (GraphQLObjectType) ((GraphQLList) outputType).getWrappedType();
+        String entityName = graphQLType.getName();
```
aklish: This logic (using the name of the graphQL type to get the underlying class name) will only work for GraphQLOutputType.  Change the outputType parameter to match this constraint.

javeria2: Hmm, this was hacky. I changed it to fetch the `entityClass` using just the field name provided by graphql under `field.getName()`. Makes it more concise.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
javeria2: Made this change to pass two fewer parameters to `fetchObject()`, as they were simply being propagated further down to fetch the `loadClass` from `getRecordType()`.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -121,11 +121,13 @@ public String toString() {
             PersistentResource<?> parent,
             Class<T> entityClass,
             RequestScope requestScope,
-            String uuid) {
+            Optional<String> UUID) {
```
aklish: All caps is reserved for static finals usually.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +256,375 @@ private Object fetchRelationship(Environment context) {
...
+    /**
+     * Forms the graph from data {@param input} and upsert's {@param function} all the nodes
+     * @param input input data
+     * @param function function to process nodes
+     * @param parent parent resource
+     * @param requestScope request scope
+     * @param fieldName graphql field name
+     * @return set of {@link PersistentResource} objects
+     */
+    private Set<PersistentResource> upsertGraphWalker(Map<String, Object> input, UpsertInterface function,
```
aklish: Let's consolidate these two walkers into a single walker function.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +305,321 @@ private Object fetchRelationship(Environment context) {
...
```
aklish: Let's rename the function to just graphWalker.  The function param type should also be a different name.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +305,321 @@ private Object fetchRelationship(Environment context) {
...
+            if(visited.contains(triplet)) {
+                continue;
+            } else {
+                visited.add(triplet);
+            }
+            Class<?> entityClass = getEntityClass(triplet.parentResource, dictionary, triplet.fieldName);
+            String idFieldName = getIdFieldName(dictionary, entityClass);
+            Optional<String> id = getId(input, idFieldName);
+            if(!id.isPresent()) {
+                setId(input, idFieldName);
```
aklish: The while loop should never directly reference the input parameter passed into the function.  It should purely operate from the triplet it pops from the queue.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +305,321 @@ private Object fetchRelationship(Environment context) {
...
+     * @param input input data
+     * @param function function to process nodes
+     * @param parent parent resource
+     * @param requestScope request scope
+     * @param fieldName graphql field name
+     * @return set of {@link PersistentResource} objects
+     */
+    private Set<PersistentResource> upsertGraphWalker(Map<String, Object> input, UpsertInterface function,
+                                                      Optional<PersistentResource> parent, RequestScope requestScope,
+                                                      String fieldName) {
```
aklish: Replace fieldName with entityClass.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +305,321 @@ private Object fetchRelationship(Environment context) {
...
+
+        while(!toVisit.isEmpty()) {
+            Triplet triplet = toVisit.remove();
+
+            if(visited.contains(triplet)) {
+                continue;
+            } else {
+                visited.add(triplet);
+            }
+            Class<?> entityClass = getEntityClass(triplet.parentResource, dictionary, triplet.fieldName);
```
aklish: Remove dictionary from triplet

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +305,321 @@ private Object fetchRelationship(Environment context) {
...
+            Map<String, Object> relationshipsOnly = stripAttributes(triplet.input.iterator().next(), entityClass,
+                    dictionary);
+            for(Map.Entry<String, Object> relationship : relationshipsOnly.entrySet()) {
+                Boolean isToOne = newParent.getRelationshipType(relationship.getKey()).isToOne();
+                List entryToAdd;
+                if(isToOne) {
+                    entryToAdd = new ArrayList<>();
+                    entryToAdd.add(relationship.getValue());
+                } else {
+                    entryToAdd = (List) relationship.getValue();
```
aklish: Triplet should contain a single item (not a list).

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +305,321 @@ private Object fetchRelationship(Environment context) {
...
+        }
+
+        return upsertOps;
+    }
+
+    /**
+     * utility class to support storing a triplet on each node of the queue used while upserting
+     */
+    private class Triplet {
+        private List<Map<String, Object>> input;
```
aklish: Should just Map<String, Object> input.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +305,321 @@ private Object fetchRelationship(Environment context) {
...
+        private Optional<PersistentResource> parentResource;
+        private String fieldName;
```
aklish: Replace this with the input Class.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +305,321 @@ private Object fetchRelationship(Environment context) {
...
+                                RequestScope requestScope, Optional<String> uuid,
+                                            Optional<PersistentResource> parent) {
+        EntityDictionary dictionary = requestScope.getDictionary();
+        Class<?> entityClass = getEntityClass(parent, dictionary, fieldName);
+        PersistentResource toCreate = PersistentResource.createObject(parent.orElse(null), entityClass, requestScope, uuid);
+        //add new relation between parent and the updated object
+        if(parent.isPresent()) {
+            parent.get().addRelation(fieldName, toCreate);
+        }
+        return updateAttribute(toCreate, entityClass, input, requestScope);
```
aklish: Should rename to updateAttributes.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +306,319 @@ private Object fetchRelationship(Environment context) {
...
+            if(visited.contains(triplet)) {
+                continue;
+            } else {
+                visited.add(triplet);
+            }
+            Class<?> entityClass = getEntityClass(triplet.parentResource, dictionary, triplet.fieldName);
+            String idFieldName = getIdFieldName(dictionary, entityClass);
+            Optional<String> id = getId(triplet.input, idFieldName);
+            if(!id.isPresent()) {
+                setId(triplet.input, idFieldName);
```
aklish: setting the ID is business logic and can be done outside the graphWalker.  It probably belongs inside createObject.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -131,76 +184,48 @@ private Object fetchObjects(Environment context) {
...
-    /**
-     * Fetches an entity, can be root or a nested entity
-     * @param context environment encapsulating graphQL's request environment
-     * @return persistent resource object(s)
-     */
-    private Object fetchObject(Environment context) {
-        Class recordType = getRecordType(context.requestScope, context.outputType);
-        return fetchObject(context.requestScope, recordType, context.ids, context.sort,
-                context.offset, context.first, context.filters);
+    private static Class<?> getRecordType(EntityDictionary dictionary, String fieldName) {
```
aklish: This single line function can just be folded into the calling code.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -119,9 +122,59 @@ private void logContext(RelationshipOp operation, Environment environment) {
...
+                .orElseGet(() -> getRecordType(dictionary, fieldName));
+    }
+
+    /**
+     * Utility method to get id field name associated with {@param entityClass} of a resource
+     * @param dictionary entity dictionary
+     * @param entityClass entity class
+     * @return idFieldName
+     */
+    private static String getIdFieldName(EntityDictionary dictionary, Class<?> entityClass) {
```
aklish: This single line function can be folded into the calling code.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -213,48 +198,66 @@ private Object fetchObject(RequestScope requestScope, Class recordType, Optional
...
+    /**
+     * fetches a non-root level attribute or relationship
+     * @param dictionary entity dictionary
+     * @param parentResource parent resource
+     * @param fieldName field type
+     * @return attribute or relationship object
+     */
+    private Object fetchObject(EntityDictionary dictionary, PersistentResource parentResource,
+                               String fieldName) {
+        Class parentClass = parentResource.getResourceClass();
```
aklish: Can we move this functionality into the function which calls this?

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -213,48 +238,66 @@ private Object fetchObject(RequestScope requestScope, Class recordType, Optional
...
      * FETCH attributes of top level entity
      * @param parentResource parent object
-     * @param field Field type
-     * @return list of persistent resource objects
+     * @param fieldName Field type
+     * @return list of {@link PersistentResource} objects
      */
-    private Object fetchAttribute(PersistentResource parentResource, Field field) {
-        return parentResource.getAttribute(field.getName());
+    private Object fetchAttribute(PersistentResource parentResource, String fieldName) {
```
aklish: Single line function can be collapsed/folded into calling code.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +306,319 @@ private Object fetchRelationship(Environment context) {
...
+        /* sanity check for id and data argument w UPSERT */
+        if(context.ids.isPresent()) {
+            throw new BadRequestException("UPSERT must not include ids");
+        }
+
+        if(!context.data.isPresent()) {
+            throw new BadRequestException("UPSERT must include data argument");
+        }
+
+        Set upsertOps = new HashSet();
```
aklish: This should be ```HashSet<PersistentResource>```

aklish: upsertOps was intended to be upserOperations.   Since these are not operations, but rather create objects, let's rename this variable.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +306,319 @@ private Object fetchRelationship(Environment context) {
...
+            this.fieldName = fieldName;
+        }
+    }
+
+    /**
+     * A function to handle upserting (update/create) objects.
+     */
+    @FunctionalInterface
+    private interface Executor<T> {
+        T execute(Map<String, Object> input, String id,
```
aklish: id should not be part of the triplet.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +306,319 @@ private Object fetchRelationship(Environment context) {
...
+            /* first execute the given function with input data and then add the remaining relationships back to queue */
+            PersistentResource newParent = (PersistentResource) function.execute(triplet.input, getId(triplet.input,
+                    idFieldName).get(), requestScope, entityClass, triplet.fieldName, triplet.parentResource);
+            toReturn.add(newParent);
+            Map<String, Object> relationshipsOnly = stripAttributes(triplet.input, entityClass,
+                    dictionary);
+            /* loop over relationships */
+            for(Map.Entry<String, Object> relationship : relationshipsOnly.entrySet()) {
+                Boolean isToOne = newParent.getRelationshipType(relationship.getKey()).isToOne();
+                List<Map<String, Object>> entryToAdd;
```
aklish: This section could be rewritten simpler like:

List<Map<String, Object>> entryToAdd = new ArrayList<>();
if (isToOne) {
    entryToAdd.add((Map<String, Object>) relationship.getValue());
} else {
    entryToAdd.addAll((List) relationship.getValue());
}
```

File: elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java [hunk 59]
```diff
@@ -263,12 +306,319 @@ private Object fetchRelationship(Environment context) {
...
+     * @param fieldName graphql field name
+     * @param parent parent resource
+     * @return {@link PersistentResource} object
+     */
+    private PersistentResource updateRelationship(Map<String, Object> dataField, String id,
+                                            RequestScope requestScope, Class<?> entityClass,
+                                            String fieldName, Optional<PersistentResource> parent) {
+        PersistentResource resource = PersistentResource.loadRecord(entityClass, id, requestScope);
+        parent.ifPresent(persistentResource ->
+                persistentResource.updateRelation(fieldName, Collections.singleton(resource)));
```
aklish: This function cannot be called a single item at a time.  It is intended to be like our REPLACE operator.    It replaces what's there with what you are providing.  Each time you call this,  it will replace the previous element you put here with this new one.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +306,319 @@ private Object fetchRelationship(Environment context) {
...
+     * @param fieldName field name
+     * @param parent parent resource
+     * @return {@link PersistentResource} object
+     */
+    private PersistentResource upsertObject(Map<String, Object> dataField, String id,
+                                             RequestScope requestScope, Class<?> entityClass,
+                                             String fieldName, Optional<PersistentResource> parent) {
+        Map<String, Object> input = stripRelationships(dataField, entityClass, requestScope.getDictionary());
+
+        Optional<PersistentResource> fetchEntry = fetchObject(requestScope, entityClass,
```
aklish: What happened to the create part of the upsert?  If the object cannot be fetched it must be created.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +306,319 @@ private Object fetchRelationship(Environment context) {
...
+                        Optional.of(Arrays.asList(id)), Optional.empty(),
+                        Optional.empty(), Optional.empty(), Optional.empty()).stream().findFirst();
+        return fetchEntry
+                .map(persistentResource -> updateAttributes(persistentResource, entityClass, input, requestScope))
+                .orElseGet(() -> createObject(input, fieldName, requestScope, Optional.of(id), parent));
```
aklish: When you create the object here, you are not updating the attributes.  This logic should be more like:
```
resource = fetchObject(...)
if (resource.isEmpty()) {
  resource = createObject(...)
}
updateAttributes.
```
This form is also easier to follow than the functional style above.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +306,319 @@ private Object fetchRelationship(Environment context) {
...
-    private Object removeObjects(Environment context) { return null; }
+    /**
+     * Creates a new "root-level" resource in the database
+     * @param input individual input data objects
+     * @param fieldName graphql-java field type
+     * @param requestScope request scope
+     * @param uuid uuid
+     * @return {@link PersistentResource} object
+     */
+    private PersistentResource createObject(Map<String, Object> input, String fieldName,
```
aklish: This function should be called createAndUpdate since it does both.  Either that or don't do the update here (instead call the update elsewhere in the calling code).

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +306,319 @@ private Object fetchRelationship(Environment context) {
...
+    /**
+     * Forms the graph from data {@param input} and upsert's {@param function} all the nodes
+     * @param input input data
+     * @param function function to process nodes
+     * @param parent parent resource
+     * @param requestScope request scope
+     * @param fieldName graphql field name
+     * @return set of {@link PersistentResource} objects
+     */
+    private Set<PersistentResource> graphWalker(Map<String, Object> input, Executor function,
```
aklish: This will be hard to explain in comments, but there is a much simpler way to do this.  This walker is attempting to  walk each relationship build tuples like (E1, E2, R) where E1 and E2 are parent and child entities and R represents a relationship between them.  Trying to fixup the object tree this way is problematic because it is doing it at too fine a grain.

What you really need is no tuple at all.   The walker should queue up just the entity (E1).

The driver function will call the walker twice.  The first time, the execute function will just
create the entity and set its attributes.  IT WILL NOT try to set any relationships (including fixing up its parent).  

The second walker also visits each entity.  For each entity, it will navigate the provided relationships.  For each relationship, it will attempt to load all the objects provided in the data.  It will then attempt to REPLACE the entire relationship.

There is one final steps, the driver function still needs to link the top level objects to its parent.  That doesn't happen in the walker execute functions though.  It happens in the driver.

elide-core/src/test/java/com/yahoo/elide/graphql/GraphQLTest.java
```diff
@@ -24,7 +26,11 @@
 
     @BeforeClass
     public void init() {
-        dictionary = new EntityDictionary(Collections.EMPTY_MAP);
+        Map<String, Class<? extends Check>> checks = new HashMap<>();
+        checks.put("allow all", Book.class);
+        checks.put("allow all", Author.class);
```
aklish: This should be the check class.

elide-core/src/test/java/com/yahoo/elide/books/Author.java
```diff
@@ -30,7 +32,7 @@
         operation = 10,
         logStatement = "{0}",
         logExpressions = {"${author.name}"})
-public class Author {
+public class Author extends UserCheck {
```
aklish: Let's not mingle the test classes with the check classes.

elide-core/src/test/java/com/yahoo/elide/books/Book.java
```diff
@@ -43,7 +45,7 @@
         operation = 10,
         logStatement = "{0}",
         logExpressions = {"${book.title}"})
-public class Book {
+public class Book extends UserCheck {
```
aklish: YOu can remove this and instead link "allow all" to 
```com.yahoo.elide.security.checks.prefab.ALL```

File: elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java [hunk 67]
```diff
@@ -865,6 +865,20 @@ public String getId(Object value) {
     }
 
     /**
+     * Returns the parameterized type or the binding class for the given {@param identifier}.
+     * @param entityClass the entity class
+     * @param identifier the identifier
+     * @return the binding class
+     */
+    public Class<?> getLoadClass(Class<?> entityClass, String identifier) {
```
aklish: I don't think we'll need this after the refactor.

elide-core/src/test/java/com/yahoo/elide/core/LifeCycleTest.java
```diff
@@ -152,7 +152,7 @@ public void testCreate() {
         DataStoreTransaction tx = mock(DataStoreTransaction.class);
         when(tx.createNewObject(Book.class)).thenReturn(book);
         RequestScope scope = new RequestScope(null, null, tx, new User(1), null, getElideSettings(null, dictionary, MOCK_AUDIT_LOGGER));
-        PersistentResource resource = PersistentResource.createObject(null, Book.class, scope, "uuid");
+        PersistentResource resource = PersistentResource.createObject(null, Book.class, scope, java.util.Optional.of("uuid"));
```
aklish: You should import java.util.Optional.

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherUpsertTest.java
```diff
@@ -0,0 +1,347 @@
...
+                    "\"author\":[{" +
+                        "\"id\":\"1\"," +
+                        "\"name\":\"abc\"" +
+                    "}]" +
+                "}";
+
+        assertQueryEquals(graphQLRequest, expectedResponse);
+
+        // Ensure we don't accidentally null values
+        graphQLRequest =
```
aklish: We should remove everything after (and including) this line.

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherUpsertTest.java
```diff
@@ -0,0 +1,347 @@
...
+                "}";
+
+        assertQueryEquals(graphQLRequest, expectedResponse);
+    }
+
+    @Test
+    public void testRootSingleWithList() throws JsonProcessingException {
+        String graphQLRequest =
+                "mutation { " +
+                    "book(op:UPSERT, data: [{id: \"1\", title: \"abc\"}, {id: \"2\", title: \"xyz\"}]) { " +
```
aklish: Let's add a comment that books 1 & 2 already exist.

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherUpsertTest.java
```diff
@@ -0,0 +1,347 @@
...
+        assertQueryEquals(graphQLRequest, expectedResponse);
+    }
+
+    @Test
+    public void testRootCollectionMixedIds() throws JsonProcessingException {
+        // Update 1, create for id 42, create new book with title "abc"
+        String graphQLRequest =
+                "mutation { " +
+                    "book(op:UPSERT, data: [{id: \"1\", title: \"my id\"}, {id: \"42\", title: \"xyz\"}, {title: \"abc\"}]) { " +
+                        "title " +
```
aklish: Can we return the IDs here too?

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherUpsertTest.java
```diff
@@ -0,0 +1,347 @@
...
+
+        assertQueryEquals(graphQLRequest, expectedResponse);
+    }
+
+    @Test
+    public void testUpsertOnCollection() {
+        String graphQLRequest =
+                "mutation { " +
+                    "author(op:UPSERT, data: {name: \"John Snow\", books: [{id: \"1\", title: \"my id\"}, {title: \"abc\"}]}) { " +
+                        "name " +
```
aklish: Can we return the author id as well?

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherUpsertTest.java
```diff
@@ -0,0 +1,347 @@
...
+
+    @Test
+    public void testNonCreatedIdReferenceCollection() {
+        String graphQLRequest =
+                "mutation { " +
+                    "author(op:UPSERT, data: {id: \"1\", name: \"John Snow\", books: [{id: \"3\", title: \"updated title\"}, {title: \"abc\"}, {id: \"2\", title: \"new title\"}]}) { " +
+                        "id " +
+                        "name " +
+                        "books { " +
+                            "title " +
```
aklish: Return the books title

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherUpsertTest.java
```diff
@@ -0,0 +1,347 @@
...
+                        "},{" +
+                            "\"title\":\"updated title\"" +
+                        "},{" +
+                            "\"title\":\"abc\"" +
+                        "}]" +
+                    "}]" +
+                "}";
+        assertQueryEquals(graphQLRequest, expectedResponse);
+    }
+
```
aklish: You should add a test case like:
```
books(op:UPSERT, data: some data containing authors)
   authors(op: UPSERT, data: some different data)
```
Verify this query works as expected (the final result is the list of authors updated by the final mutation).

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherDeleteTest.java
```diff
@@ -0,0 +1,142 @@
...
+                "} " +
+            "}";
+        assertQueryFails(graphQLRequest);
+    }
+
+    @Test
+    public void testRootIdNoData() throws JsonProcessingException {
+        String graphQLRequest =
+            "mutation { " +
+                "author(op:DELETE, ids: [\"1\"]) { " +
```
aklish: Do a fetch following the delete to ensure that it was actually deleted.

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherDeleteTest.java
```diff
@@ -0,0 +1,142 @@
...
+            "mutation { " +
+                "book(op:DELETE, ids: [\"1\", \"2\"]) { " +
+                    "id " +
+                    "title " +
+                "} " +
+            "}";
+        String expectedResponse =
+            "{" +
+                "\"book\":[{" +
+                    "\"id\":\"1\"," +
```
aklish: Same comment.

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherDeleteTest.java
```diff
@@ -0,0 +1,142 @@
...
+                        "title " +
+                    "} " +
+                "} " +
+            "}";
+        String expectedResponse =
+            "{" +
+                "\"author\":[{" +
+                    "\"books\":[{" +
+                        "\"id\":\"1\"," +
+                        "\"title\":\"Libro Uno\"" +
```
aklish: Same comment.

elide-core/src/test/java/com/yahoo/elide/graphql/FetcherDeleteTest.java
```diff
@@ -0,0 +1,142 @@
...
+                "author(ids: [\"1\"]) { " +
+                    "books(op:DELETE, ids: [\"1\", \"2\"]) { " +
+                        "id " +
+                        "title " +
+                    "} " +
+                "} " +
+            "}";
+        String expectedResponse =
+            "{\"author\":[{" +
+                "\"books\":[{" +
```
aklish: Same comment.

PR #502, opened 2017-07-21
Add "NOT" op to FilterExpressionCheck 

tarrantzhang:
This PR allows NOT operator in writing FilterExpressionCheck.
Given the dictionary:
- "user has all access": Role.ALL
- "IN predicate": Author.books.title IN [Harry Potter]
- "LE predicate": Company.revenue LE [3000]
- "user has no access": Role.NONE

Here are some examples how such FilterExpressionChecks evaluate:
1. "NOT (IN predicate)" => Author.books.title NOT [Harry Potter]
2. "NOT (LE predicate)" => Company.revenue GT [3000]
3. "NOT (IN predicate AND LE predicate)" => "(NOT IN predicate) OR (NOT LE predicate)" => Author.books.title NOT [Harry Potter] OR Company.revenue GT [3000]
4. "NOT (IN predicate OR LE predicate)" => "(NOT IN predicate) AND (NOT LE predicate)" => Author.books.title NOT [Harry Potter] AND Company.revenue GT [3000]
5. "NOT (IN predicate OR (LE predicate AND (NOT IN predicate))) =>
(NOT IN predicate) AND (NOT LE predicate OR IN predicate) =>
Author.books.title NOT [Harry Potter] AND (Company.revenue GT [3000] OR Author.books.title IN [Harry Potter])
6. Especially a non-predicate-filterExpression generated by userCheck, for example "NO_EVALUATION_EXPRESSION" will evaluate to itself since it does not have negation.


elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidOperatorNegationException.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright 2016, Yahoo Inc.
```
DeathByTape: We should update this (and checkstyle if necessary).

Year: 2017
Company: Oath Inc.

We should be adding Oath copyright headers now, right?

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
```diff
@@ -157,4 +163,42 @@ private Operator operator(FilterExpression expression) {
...
+            }
+        } else if (fe instanceof AndFilterExpression) {
+            return new AndFilterExpression(
+                    normalizeNegation(((AndFilterExpression) fe).getLeft()),
+                    normalizeNegation(((AndFilterExpression) fe).getRight()));
+        } else if (fe instanceof OrFilterExpression) {
+            return new OrFilterExpression(
+                    normalizeNegation(((OrFilterExpression) fe).getLeft()),
+                    normalizeNegation(((OrFilterExpression) fe).getRight()));
+        } else if (fe instanceof FilterPredicate) {
```
DeathByTape: This conditional is unnecessary. Just let it fall through to the final return statement since they're the same.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
```diff
@@ -157,4 +163,42 @@ private Operator operator(FilterExpression expression) {
     public FilterExpression visitPAREN(ExpressionParser.PARENContext ctx) {
         return visit(ctx.expression());
     }
+
+    /**
+     * normalizeNegation function serves to push down NotFilterexpression to the predicate level.
+     * @param fe FilterExpression that need to normalize
+     * @return A normalized version of input
+     */
+    public static FilterExpression normalizeNegation(FilterExpression fe) {
```
DeathByTape: I'm having a hard time understanding why we make this a `public static` in this class. It doesn't seem like this class needs to ever perform this operation. Should we move this to the `PermissionExpressionBuilder` instead?

tarrantzhang:
I made changes according to @DeathByTape's comment, not sure why it still appears unsatisfied.

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/FilterExpressionVisitor.java
```diff
@@ -0,0 +1,14 @@
...
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core.filter.expression;
+
+/**
+ * Visitor for filterExpression .
+ * @param <T> The return type of the visitor
+ */
+public interface FilterExpressionVisitor<T> extends Visitor<T> {
```
aklish: You shouldn't need to define this.  You can call FilterExpression.accept on the visitor you created below to start the visitation.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/FilterExpressionNormalizationVisitor.java
```diff
@@ -0,0 +1,62 @@
...
+            return nfe;
+        } else if (nfe instanceof NotFilterExpression) {
+            return (((NotFilterExpression) nfe).getNegated()).accept(this);
+        } else {
+            return nfe;
+        }
+    }
+
+    @Override
+    public FilterExpression visitCustomizedExpression(FilterExpression expression) {
```
aklish: can remove.

elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java
```diff
@@ -39,7 +41,7 @@
     public static final FilterExpression NO_EVALUATION_EXPRESSION = new FilterExpression() {
         @Override
         public <T> T accept(Visitor<T> visitor) {
-            return null;
+            return (T) ((FilterExpressionVisitor) visitor).visitCustomizedExpression(NO_EVALUATION_EXPRESSION);
```
aklish: Could this instead be:
```return this;```

File: elide-core/src/main/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitor.java [hunk 7]
```diff
@@ -50,6 +52,9 @@ public String toString() {
     public static final FilterExpression FALSE_USER_CHECK_EXPRESSION = new FilterExpression() {
         @Override
         public <T> T accept(Visitor<T> visitor) {
+            if (visitor instanceof FilterExpressionVisitor) {
+                return (T) ((FilterExpressionVisitor) visitor).visitCustomizedExpression(FALSE_USER_CHECK_EXPRESSION);
```
aklish: Could this instead be:
```return this```

File: elide-core/src/test/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitorTest.java [hunk 8]
```diff
@@ -308,6 +356,47 @@ public boolean ok(Model object, RequestScope requestScope, Optional<ChangeSpec>
...
 
+    //Test NOT with "IN" FilterPredicate
+    @Entity
+    @Include
+    @ReadPermission(expression = "NOT (owner is user)")
+    static class GOOD11 {
+    }
+
+    //Test NOT with "LT" FilterPredicate
+    @Entity
```
aklish: These classes should be embedded inside their own test functions.  Each of these tests should be a separate TestNG function.

elide-core/src/test/java/com/yahoo/elide/parsers/expression/PermissionToFilterExpressionVisitorTest.java
```diff
@@ -308,6 +356,47 @@ public boolean ok(Model object, RequestScope requestScope, Optional<ChangeSpec>
...
+    @Include
+    @ReadPermission(expression = "NOT (owner is user OR less than predicate)")
+    static class GOOD14 {
+    }
+
+    //Test Not with nested FilterExpression
+    @Entity
+    @Include
+    @ReadPermission(expression = "NOT (owner is user OR (less than predicate AND (NOT owner is user)))")
```
aklish: A more readable way to do this might be to have two classes in a single test function.  One untransformed - and one transformed.  Extract the filter expressions for both and see if they are identical.


So you would compare the above ReadPermission with:
```"(owner is NOT user AND (greater than equal to predicate OR (owner is user))"```

File: elide-core/src/main/java/com/yahoo/elide/core/filter/expression/FilterExpressionVisitor.java (hunk 4 above)
Comment (tarrantzhang) [inline elide-core/src/main/java/com/yahoo/elide/core/filter/expression/FilterExpressionVisitor.java]: Yes, you were right. This class was created merely to deal with anonymous FilterExpression classes. Turns out it is not necessary.

File: elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidOperatorNegationException.java [hunk 10]
```diff
@@ -0,0 +1,18 @@
...
+import com.yahoo.elide.core.HttpStatus;
+
+/**
+ * Invalid predicate negation error.
+ *
+ * {@link com.yahoo.elide.core.HttpStatus#SC_FORBIDDEN invalid}
+ */
+public class InvalidOperatorNegationException extends HttpStatusException {
+    public InvalidOperatorNegationException() {
+        super(HttpStatus.SC_FORBIDDEN, "Invalid Usage of NOT in FilterExpression.");
```
aklish: Should be 500 error.

PR #503, opened 2017-07-25
Throw 400 instead of 500 for invalid filtering input.

DeathByTape:


elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/filter/HQLFilterOperation.java
```diff
@@ -50,14 +54,29 @@ public String apply(FilterPredicate filterPredicate, boolean prefixWithAlias) {
             case PREFIX:
                 return String.format("%s LIKE CONCAT(:%s, '%%')", fieldPath, alias);
             case PREFIX_CASE_INSENSITIVE:
+                if (fieldPath == null || fieldPath.isEmpty()) {
```
wcekan: would using Guava look cleaner?
```
if (Strings. isNullOrEmpty(fieldPath))
```

Issue #491, opened 2017-07-06
"Totals" not work with "Joined Filter Syntax"


songoo:
Hello when i use "Joined Filter Syntax" which is enabled with "withJoinFilterDialect", filters works as expected but it get crashed when i add "totals" in pagination settings :

works:
`page: {size:10, number: this.get("page")}`  
does not work :
`page: {size:10, number: this.get("page"), totals: true},`

Resulting in error :
```
{
  "errors":[
    "InternalServerErrorException: org.hibernate.QueryException: illegal attempt to dereference collection [document0_.ticket_id.staffs] with element property reference [id] [SELECT COUNT(*) FROM models.Document Document WHERE (Document.documentType.id IN (:documentType_id_41c7708c) AND (Document.staffs.id IN (:staffs_id_8526eb6c) OR Document.managers.id IN (:managers_id_2a48f153)))]"
  ]
}
```

Here are query parameters :

```
filter:documentType.id==1;(staffs.id==116,managers.id==116)
page[size]:10
page[number]:1
page[totals]:true
sort:-createdAt
```

DeathByTape:
I can look into this soon, but I may defer to @aklish in the meantime as he has the most experience working with the `pageTotals` code.

aklish:
Without getting into the details, it looks like this particular query is attempting to paginate a collection (Document) while also joining (via filters) to another, toMany collection (staffs).  Does that sound correct?

There is no easy way in Hibernate to paginate across a join with a toMany relationship.   The result set that comes back will duplicate document rows (one row per staff) - and Hibernate doesn't handle that correctly.   As a result, Elide doesn't support that scenario with its Hibernate DataStore. 

That said, the count query shouldn't blow in this way.  We should think about how Elide should handle joins across toMany relationship - should we allow them at all?  Should we allow them but not allow pagination?

aklish:
I'm working on a fix that will allow totals to work in conjunction with joined filters.  Caveat though - there still is no workaround to get Hibernate to paginate correctly through a toMany relationship.

aklish:
I'm closing the issue for now.  If there is still a problem, please reopen.

songoo:
Its working, thnx much ^__^

Issue #415, opened 2017-02-28
Upgrade hibernate5 and testng


DennisMcWherter:
During our great dependency upgrade of 2017 (see: https://github.com/yahoo/elide/pull/414), we upgraded most dependencies except hibernate5 and testng. We should upgrade to the latest version. It should be noted that this will take a bit of effort, however. 

At time of writing, for hibernate5 the jump between 5.0.2 to 5.2.8 has many deprecated methods. In fact, the code may still compile in many cases but will now throw deprecated exceptions :( This is a bit sad (as it would be better to know at compile time something is broken), but it will take a bit of digging into to fully upgrade.

For testng we noticed sporadic test failures (perhaps related to `priority` settings?) that appeared to be a result of things running out of order. This should also be further investigated.

aklish:
Hibernate has been upgraded to 5.2.15.  Testng still needed.

PR #506, opened 2017-08-02
Replace criteria with jpql in Hibernate stores.

aklish:


aklish:
This Fixes #491 (to the extent it can be fixed).

I'm going to add unit tests in another commit (coming in a day or two).

aklish:
Also completes ~90% of the work for #415

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/PredicateExtractionVisitor.java
```diff
@@ -8,39 +8,54 @@
...
-    public Set<FilterPredicate> visitPredicate(FilterPredicate filterPredicate) {
+    public Collection<FilterPredicate> visitPredicate(FilterPredicate filterPredicate) {
         filterPredicates.add(filterPredicate);
         return filterPredicates;
     }
 
     @Override
-    public Set<FilterPredicate> visitAndExpression(AndFilterExpression expression) {
-        filterPredicates.addAll(expression.getLeft().accept(this));
-        filterPredicates.addAll(expression.getRight().accept(this));
```
clayreimann: Why are we not adding these to the collection any more? Is it because they get added at the leaf?

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/AbstractHQLQueryBuilder.java
```diff
@@ -0,0 +1,205 @@
...
+            );
+            if (!validSortingRules.isEmpty()) {
+                final List<String> ordering = new ArrayList<>();
+                // pass over the sorting rules
+                validSortingRules.entrySet().stream().forEachOrdered(entry ->
+                        ordering.add(entry.getKey() + " " + (entry.getValue().equals(Sorting.SortOrder.desc)
+                                ? "desc"
+                                : "asc"))
+                );
+                sortingRules = " order by " + StringUtils.join(ordering, ",");
```
clayreimann: `JOIN` is in caps, do we want to put `ORDER BY`, `ASC`, and `DESC` in caps too?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -238,21 +196,48 @@ public Object getRelation(
...
+                    public Object getParent() {
+                        return entity;
+                    }
+
+                    @Override
+                    public Collection getChildren() {
+                        return filteredVal;
+                    }
+                };
+
```
clayreimann: Is there a reason to use an inline anonymous class here? It would probably be cleaner to have a this trivial class generated by Lombok (e.g. `@Data`)

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -231,21 +195,48 @@ public Object getRelation(
...
+                    public Object getParent() {
+                        return entity;
+                    }
+
+                    @Override
+                    public Collection getChildren() {
+                        return filteredVal;
+                    }
+                };
+
```
clayreimann: This could use the same class as the Hibernate3 version

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/PredicateExtractionVisitor.java
aklish: Yes.  This was a bug in the old code.  When we switched from a Set to a more general Collections, I saw lists had too many predicates in them.

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/AbstractHQLQueryBuilder.java
```diff
@@ -0,0 +1,218 @@
...
+                return joinClause;
+            }
+
+            String alias = typeAlias + UNDERSCORE + fieldName;
+
+            String joinFragment;
+
+            //This is the first path element
+            if (previousAlias == null) {
+                joinFragment = JOIN + typeAlias + PERIOD + fieldName + SPACE + alias + SPACE;
```
DeathByTape: No need to join `ON` here? Or is that just HQL magic?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -238,21 +197,28 @@ public Object getRelation(
                 @SuppressWarnings("unchecked")
                 Class<?> relationClass = dictionary.getParameterizedType(entity, relationName);
 
+                RelationshipImpl relationship = new RelationshipImpl();
```
DeathByTape: Can we just give this an `@AllArgsConstructor`? These things should either be fully constructed or not at all, correct?

We could also then make it immutable.

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -231,21 +196,28 @@ public Object getRelation(
                 @SuppressWarnings("unchecked")
                 Class<?> relationClass = dictionary.getParameterizedType(entity, relationName);
 
+                RelationshipImpl relationship = new RelationshipImpl();
```
DeathByTape: Same comment as above.

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/RootCollectionPageTotalsQueryBuilder.java
```diff
@@ -0,0 +1,96 @@
...
+            //Build the JOIN clause
+            joinClause =  getJoinClauseFromFilters(filterExpression.get());
+
+        } else {
+            predicates = new HashSet();
+            filterClause = "";
+            joinClause = "";
+        }
+
+        Query query = session.createQuery("SELECT COUNT(DISTINCT "
```
DeathByTape: Do we really want `DISTINCT` for totals?

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/AbstractHQLQueryBuilder.java
aklish: Magic

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/RootCollectionPageTotalsQueryBuilder.java
aklish: Yes.  Otherwise we count rows instead of objects.

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/AbstractHQLQueryBuilder.java
```diff
@@ -0,0 +1,218 @@
...
+
+    public abstract Query build();
+
+    public AbstractHQLQueryBuilder withPossibleFilterExpression(Optional<FilterExpression> filterExpression) {
+        this.filterExpression = filterExpression;
+        return this;
+    }
+
+    public AbstractHQLQueryBuilder withFilterExpression(FilterExpression filterExpression) {
+        return withPossibleFilterExpression(Optional.of(filterExpression));
```
aklish: Given that there is another interface that takes the Optional, I think it is ok to require a non-null value here.

wcekan:
@aklish This appears to be a somewhat breaking change.  My subclass of HibernateStore fails now.
```java
        /**
         * Override Elide Hibernate handling to add max results.
         */
        @Override
        public Iterable loadObjects(Class<?> loadClass, Criteria criteria, Optional<Set<Order>> sortingRules,
                                    Optional<Pagination> pagination, RequestScope scope) {
```

PR #509, opened 2017-08-15
Add sorting joins

aklish:


elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -68,20 +64,34 @@ public Sorting(final Map<String, SortOrder> sortingRules) {
...
-     *
-     * @return a map of the sorting rules to be applied
-     */
-    public Map<String, SortOrder> getSortingRules() {
-        return this.sortRules;
+            //Creating a path validates that the dot separated path is valid.
+            Path path = new Path(entityClass, dictionary, dotSeparatedPath);
+
+            //Validate that none of the relationships are to-many
+            for (Path.PathElement pathElement : path.getPathElements()) {
```
DennisMcWherter: Can we break this out to its own function for readability? i.e. `isValidSortingRule(Path)`. This could then be reused in the logic above.

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/RootCollectionFetchQueryBuilder.java
```diff
@@ -61,7 +61,7 @@ public Query build() {
                         + SPACE
                         + filterClause
                         + SPACE
-                        + getSortClause(sorting, entityClass, false)
+                        + getSortClause(sorting, entityClass, true)
```
wcekan: The only problem with true/false parameters, you have to check the doc on the method to see what this means.  But that is not new to this PR.

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -50,13 +50,9 @@ public Sorting(final Map<String, SortOrder> sortingRules) {
...
 
-        final List<String> entities = dictionary.getAttributes(entityClass);
-        sortRules.keySet().stream().forEachOrdered(sortRule -> {
-            if (!entities.contains(sortRule) && !JSONAPI_ID_KEYWORD.equals(sortRule)) {
-                throw new InvalidValueException(entityClass.getSimpleName()
-                        + " doesn't contain the field " + sortRule);
-            }
-        });
+        //This validates the paths for the given entityClass or throws InvalidValueException
+        getValidSortingRules(entityClass, dictionary);
```
DennisMcWherter: At a glance, it's not obvious why this method performs a `getValidSortingRules` and doesn't use the result. I do see that this functionality is embedded in the logic of `getValidSortingRules`, but you have to dig to find it. Can we restructure the code a bit so it's clearer?

elide-core/src/main/java/com/yahoo/elide/core/Path.java
```diff
@@ -0,0 +1,106 @@
...
+                Class<?> relationClass = dictionary.getParameterizedType(currentClass, fieldName);
+                pathElements.add(new PathElement(currentClass, currentClassName, relationClass, fieldName));
+                currentClass = relationClass;
+                currentClassName = fieldName;
+            } else if (dictionary.isAttribute(currentClass, fieldName)
+                    || fieldName.equals(dictionary.getIdFieldName(entityClass))) {
+                Class<?> attributeClass = dictionary.getType(currentClass, fieldName);
+                pathElements.add(new PathElement(currentClass, currentClassName, attributeClass, fieldName));
+            } else {
+                throw new InvalidValueException(entityClass.getSimpleName()
```
wcekan: Is `SimpleName` really what we want here?  Should we use type name (which can be different)

elide-core/src/main/java/com/yahoo/elide/core/sort/Sorting.java
```diff
@@ -68,20 +64,34 @@ public Sorting(final Map<String, SortOrder> sortingRules) {
...
-    public <T> Map<String, SortOrder> getValidSortingRules(final Class<T> entityClass,
-                                                           final EntityDictionary dictionary)
+    public <T> Map<Path, SortOrder> getValidSortingRules(final Class<T> entityClass,
+                                                         final EntityDictionary dictionary)
             throws InvalidValueException {
-        hasValidSortingRules(entityClass, dictionary);
-        return replaceIdRule(dictionary.getIdFieldName(entityClass));
-    }
+        Map<Path, SortOrder> returnMap = new LinkedHashMap<>();
+        for (Map.Entry<String, SortOrder> entry : replaceIdRule(dictionary.getIdFieldName(entityClass)).entrySet()) {
```
DennisMcWherter: If we're concerned about performance of code restructuring, we could add one more private method here;

`Stream<Path> getSortingRules(Class<T>, EntityDictionary)`

This way we can stay DRY on our `replaceIdRule` code (and lazily evaluate) and still only have to iterate once since streams are lazy.

PR #511, opened 2017-08-16
Add Entity class

javeria2:
Previous code + Contains a refactor fleshing out a new class called `Entity` encapsulating `entityClass`, `parent` and `data` for each entity.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
@@ -9,9 +9,10 @@ import com.yahoo.elide.core.EntityDictionary;
 import com.yahoo.elide.core.PersistentResource;
 import com.yahoo.elide.core.RequestScope;
 
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -20,8 +21,11 @@ import java.util.UUID;
 
 public class Entity {
     private Optional<Entity> parentResource;
-    private Optional<Map<String, Object>> data;
+    private Map<String, Object> data;
     private Class<?> entityClass;
+    private RequestScope requestScope;
+    private Set<Attribute> attributes;
+    private Set<Relationship> relationships;
 
     /**
      * Class constructor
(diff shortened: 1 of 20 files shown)


elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,152 @@
...
+     * Getter for binding entity class
+     */
+    public Class<?> getEntityClass() {
+        return this.entityClass;
+    }
+
+    /**
+     * Getter for entity data
+     */
+    public Map<String, Object> getData() {
```
aklish: Instead of having this method, why don't we have something like getAttributes and getRelationships

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,152 @@
...
+    public Optional<Entity> getParentResource() {
+        return parentResource;
+    }
+
+    /**
+     * Get the id of the entity
+     * @param requestScope Request scope
+     * @return the optional id
+     */
+    public Optional<String> getId(RequestScope requestScope) {
```
aklish: Since you aren't using the entire RequestScope, you should change this to just dictionary.

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,152 @@
...
+        } else {
+            return Optional.empty();
+        }
+    }
+
+    /**
+     * Set the id of the entity if it doesn't have one already
+     * @param requestScope Request scope
+     */
+    public void setId(RequestScope requestScope) {
```
aklish: Same comment.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +246,261 @@ private Object fetchRelationship(Environment context) {
...
+
+    /**
+     * Forms the graph from data {@param input} and upsert's {@param function} all the nodes
+     * @param input Input data
+     * @param function Function to process nodes
+     * @param entityClass Entity class
+     * @param requestScope Request scope
+     * @return set of {@link PersistentResource} objects
+     */
+    private Set<PersistentResource> graphWalker(Map<String, Object> input, Executor function,
```
aklish: graphWalker should take an Entity instead of the map.  In general, nothing should deal with the Map except the code that initially converts that to an Entity (that can probably be done in one place).

aklish: Have this function return void for now.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +246,261 @@ private Object fetchRelationship(Environment context) {
...
+            graphWalker(input, this::updateRelationship, entityClass, context.requestScope);
+        }
 
-    private Object removeObjects(Environment context) { return null; }
+        /* add relation between parent and nested entity */
+        if(!context.isRoot()) {
+            for(PersistentResource obj : upsertedObjects) {
+                context.parentResource.addRelation(context.field.getName(), obj);
+            }
+        }
```
aklish: return entities.stream.map(Entity::toPersistentResource)

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,233 @@
...
+
+        public Set<Entity> getValue() {
+            return value;
+        }
+    }
+
+    /**
+     * Extract the attributes of the entity
+     */
+    public void setAttributes() {
```
aklish: shoudl be private.

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,233 @@
...
+            if(Objects.equals(entry.getKey(), idFieldName)) {
+                this.attributes.add(new Attribute(entry.getKey(), entry.getValue()));
+            }
+        }
+    }
+
+    /**
+     * Extract the relationships of the entity
+     */
+    public void setRelationships() {
```
aklish: Should be private

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,233 @@
...
+                     Class<?> entityClass, RequestScope requestScope) {
+        this.parentResource = parentResource;
+        this.data = data;
+        this.entityClass = entityClass;
+        this.requestScope = requestScope;
+        setAttributes();
+        setRelationships();
+    }
+
+    class Attribute {
```
aklish: You can use lombok for these inner classes to reduce boilerplate.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -213,48 +178,66 @@ private Object fetchObject(RequestScope requestScope, Class recordType, Optional
...
+     * fetches a non-root level attribute or relationship
+     * @param dictionary entity dictionary
+     * @param ids list of ids
+     * @param parentResource parent resource
+     * @param fieldName field type
+     * @return attribute or relationship object
      */
-    private Object fetchAttribute(PersistentResource parentResource, Field field) {
-        return parentResource.getAttribute(field.getName());
+    private Object fetchObject(EntityDictionary dictionary, Optional<List<String>> ids,
```
aklish: It looks like this is only called in one place.  I think we should fold this into the calling function.  It is a function that does two things and takes arguments that don't apply for one of those two things.

elide-core/src/test/java/com/yahoo/elide/graphql/GraphQLTest.java
```diff
@@ -24,7 +26,11 @@
 
     @BeforeClass
     public void init() {
-        dictionary = new EntityDictionary(Collections.EMPTY_MAP);
+        Map<String, Class<? extends Check>> checks = new HashMap<>();
+        checks.put("allow all", com.yahoo.elide.security.checks.prefab.Role.ALL.class);
+        checks.put("allow all", com.yahoo.elide.security.checks.prefab.Role.ALL.class);
```
aklish: You can remove one of these two lines.

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,176 @@
...
+        @Getter private String name;
+        @Getter private Object value;
+
+        public Attribute(String name, Object value) {
```
aklish: You can use ```@AllArgumentConstructor``` from lombok and don't need to specify the constructor.

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,176 @@
...
+            this.name = name;
+            this.value = value;
+        }
+    }
+
+    class Relationship {
+        @Getter private String name;
+        @Getter private Set<Entity> value;
+
+        public Relationship(String name, Set<Entity> value) {
```
aklish: Same comment.

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,176 @@
...
+                .map(e -> (String) e.value)
+                .findFirst();
+    }
+
+    /**
+     * Set the id of the entity if it doesn't have one already
+     */
+    public void setId() {
+        EntityDictionary dictionary = this.requestScope.getDictionary();
+        String idFieldName = dictionary.getIdFieldName(this.entityClass);
```
aklish: These two statements can be moved inside the if statement.

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,176 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
+
+public class Entity {
+    @Getter private Optional<Entity> parentResource;
+    @Getter private Map<String, Object> data;
```
aklish: Does this need a getter?  It doesn't look like it is used.

elide-core/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -0,0 +1,176 @@
...
+     */
+    public void setId() {
+        EntityDictionary dictionary = this.requestScope.getDictionary();
+        String idFieldName = dictionary.getIdFieldName(this.entityClass);
+        if(!getId().isPresent()) {
+            String uuid = UUID.randomUUID().toString()
+                    .replaceAll("[^0-9]", "")
+                    .substring(0, 3); //limit the number of digits to prevent InvalidValueException in PersistentResource.createObject()
+            //TODO: this is hacky, ask for a workaround for this.
+            this.data.put(idFieldName, uuid);
```
aklish: Can we just remove it then?  data is private to Entity (or at least can be made so), and I don't think we need to fetch the ID from the data once it is set.

elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -263,12 +235,234 @@ private Object fetchRelationship(Environment context) {
...
+        if(context.isRoot()) {
+            entityClass = dictionary.getEntityClass(context.field.getName());
+        } else {
+            entityClass = dictionary.getParameterizedType(context.parentResource.getResourceClass(), context.field.getName());
+        }
+
+        /* form entities */
+        Set<Entity> entitySet = new HashSet<>();
+        for(Map<String, Object> input : context.data.get()) {
+            entitySet.add(new Entity(Optional.empty(), input, entityClass, context.requestScope));
```
aklish: I still think you should put the parent here (if it exists) for correctness.  Otherwise, this might be a latent bug.

PR #512, opened 2017-08-16
Type is Entity name if Include not defined.

wcekan:


elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/SubCollectionPageTotalsQueryBuilder.java
```diff
@@ -104,7 +104,7 @@ public Query build() {
             predicates.stream().forEach(predicate -> {
                 predicate.getPath().add(0, new FilterPredicate.PathElement(
                         parentType,
-                        parentType.getSimpleName(),
+                        dictionary.getEntityFor(parentType),
```
aklish: There are lots of other places that build predicates.  We should be consistent and use this new method everywhere we build a FilterPredicate.PathElement OR a new Path.

wcekan: This argument has been removed.  TypeName can be looked up from Type class.

PR #513, opened 2017-08-17
Cleanup equals

wcekan:
Cleanup code style to be more readable.


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1360,7 +1360,7 @@ protected boolean addToCollection(Collection collection, String collectionName,
         if (collection == null) {
             collection = Collections.singleton(toAdd.getObject());
             Object value = getValueUnchecked(collectionName);
-            if ((value == null && toAdd.getObject() != null) || (value != null && !value.equals(toAdd.getObject()))) {
```
wcekan: @DennisMcWherter Please review this is still working as intended.

PR #514, opened 2017-08-18
Fix id extractor.

DennisMcWherter:


elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -241,14 +241,12 @@ public void setAttribute(Object entity, String attributeName, Object attributeVa
     private Serializable extractId(FilterExpression filterExpression,
                                    String idFieldName,
                                    Object parent,
-                                   Class<?> relationClass,
                                    EntityDictionary dictionary) {
         Collection<FilterPredicate> predicates = filterExpression.accept(new PredicateExtractionVisitor());
         for (FilterPredicate predicate : predicates) {
             List<Object> values = predicate.getValues();
             Class<?> entityClass = dictionary.getParameterizedType(parent, predicate.getField());
-            if (relationClass == entityClass
```
aklish: Removing this makes the logic incorrect IMO.  What if the client had a filter predicate on two different entity types (that also had the same ID field name).  We might grab the wrong ID here.

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexTransaction.java
```diff
@@ -240,13 +240,11 @@ public void setAttribute(Object entity, String attributeName, Object attributeVa
...
                                    String idFieldName,
-                                   Object parent,
-                                   Class<?> relationClass,
-                                   EntityDictionary dictionary) {
+                                   Class<?> relationClass) {
         Collection<FilterPredicate> predicates = filterExpression.accept(new PredicateExtractionVisitor());
         for (FilterPredicate predicate : predicates) {
             List<Object> values = predicate.getValues();
-            Class<?> entityClass = dictionary.getParameterizedType(parent, predicate.getField());
+            Class<?> entityClass = predicate.getEntityType();
```
aklish: Should the line below be relationClass.equals(entityClass)?

DennisMcWherter: Since `Class` doesn't override `.equals`, `Class` actually falls back on `Object.equals`. i.e.:

```
public boolean equals(Object obj) {
        return (this == obj);
    }
```

So they're equivalent. If we think `.equals` is more readable we can use that. Otherwise, shouldn't make a difference :)

PR #518, opened 2017-08-24
Add deferred ID

javeria2:
This PR contains the code for lazy loading of the `id` field from `PersistentResource` by delayed jackson serialization. 
IIRC, we're now left with the following TODOs -   
- Pagination/filtering/sorting for `FETCH`.
- Passing filter expression to `getRelation()` to reduce number of sql queries while fetching a relationship.
- The actual `/graphql` endpoint. 
- A lot of IT/unit tests.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/DeferredId.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/DeferredId.java
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2017, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+
+package com.yahoo.elide.graphql;
+
+import com.yahoo.elide.core.PersistentResource;
+
+public class DeferredId {
+    private PersistentResource resource;
+
+    public DeferredId(PersistentResource resource) {
+        this.resource = resource;
+    }
+
+    @Override
+    public String toString() {
+        return resource.getId();
+    }
+}
--- a/elide-core/src/main/java/com/yahoo/elide/graphql/DeferredId.java
+++ b/elide-core/src/main/java/com/yahoo/elide/graphql/DeferredId.java
@@ -6,17 +6,28 @@
 
 package com.yahoo.elide.graphql;
 
+import com.fasterxml.jackson.core.JsonGenerator;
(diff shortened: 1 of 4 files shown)


elide-core/src/main/java/com/yahoo/elide/graphql/DeferredId.java
```diff
@@ -0,0 +1,33 @@
...
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.yahoo.elide.core.PersistentResource;
+import lombok.Getter;
+
+import java.io.IOException;
+
+@JsonSerialize(using = SerializeId.class)
+public class DeferredId {
```
aklish: Add a comment to explain what this class is about.

elide-core/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -204,10 +206,27 @@ private GraphQLObjectType buildQueryObject(Class<?> entityClass) {
         builder.name(entityName);
 
         String id = dictionary.getIdFieldName(entityClass);
+        GraphQLScalarType customIdType = new GraphQLScalarType(id, "custom id type", new Coercing() {
```
aklish: Add a comment explaining that our ID types are expected to be DefferedIds

elide-core/src/main/java/com/yahoo/elide/graphql/DeferredId.java
```diff
@@ -0,0 +1,39 @@
...
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.yahoo.elide.core.PersistentResource;
+import lombok.Getter;
+
+import java.io.IOException;
+
+/**
+ * Simple encapsulation to serialize the id field from a {@link PersistentResource} object.
```
aklish: Add the following text.

The id for any given entity might be populated at transaction commit (as opposed to inline with the data fetch).  This class wraps a PersistentResource and allows deferred deserialization of the ID field until when it is populated and when the GraphQL response is generated.

PR #523, opened 2017-09-06
Fix broken graph ql tests

aklish:


elide-core/src/main/java/com/yahoo/elide/core/datastore/inmemory/InMemoryTransaction.java
```diff
@@ -136,19 +137,21 @@ public Object loadObject(Class<?> entityClass, Serializable id,
...
-
-        // Support for filtering
-        if (filterExpression.isPresent()) {
-            Predicate predicate = filterExpression.get().accept(new InMemoryFilterVisitor(scope));
-            return (Collection<Object>) data.values().stream()
-                    .filter(predicate::test)
-                    .collect(Collectors.toList());
-        }
+        Map<String, Object> data = dataStore.get(entityClass);
+        synchronized (data) {
```
DennisMcWherter: We need this to be synchronized every where it's used (i.e. `loadObject` and `commit`). Would it be easier to wrap `LinkedHashMap` and synchronize the entire structure? Or do we want to do the bookkeeping here?

elide-core/src/main/java/com/yahoo/elide/core/datastore/inmemory/InMemoryTransaction.java
```diff
@@ -35,17 +34,17 @@
...
+    private final Map<Class<?>, Map<String, Object>> dataStore;
     private final List<Operation> operations;
     private final EntityDictionary dictionary;
+    private final Map<Class<?>, AtomicLong> typeIds;
 
-    public InMemoryTransaction(ConcurrentHashMap<Class<?>, ConcurrentHashMap<String, Object>> dataStore,
-                               EntityDictionary dictionary) {
+    public InMemoryTransaction(Map<Class<?>, Map<String, Object>> dataStore,
+                               EntityDictionary dictionary, Map<Class<?>, AtomicLong> typeIds) {
         this.dataStore = dataStore;
```
wcekan: Since we are worried about synchronization below, this should be
```
this.dataStore = Collections.synchronizedMap(dataStore);
```

Issue #521, opened 2017-09-03
How to set up an Initializer?


Brutus5000:
Hi there,

as there is no example for implementing an initializer, I just gave it a try.
Unfortunately my initializer is not used, whether I put it into the entityDictionary of the DataStore or of the ElideSettingsBuilder.

I have tried GETing an existing collection and POSTing a new entity.
Can you give me a hint what I am doing wrong?
Another question: When is the Initializer supposed to be fired? On loading of each entity or only on creation/commit?

Below you can find my Configuration:

```
@Configuration
@Slf4j
public class ElideConfig {
    @Bean
    public Elide elide(EntityManagerFactory entityManagerFactory) {
        ConcurrentHashMap<String, Class<? extends Check>> checks = new ConcurrentHashMap<>();
//        checks.put("user is a superuser", AdminCheck.class);

        EntityDictionary entityDictionary = new EntityDictionary(checks);
        RSQLFilterDialect rsqlFilterDialect = new RSQLFilterDialect(entityDictionary);

        entityDictionary.bindInitializer(new Initializer<ProductClass>() {
            @Override
            public void initialize(ProductClass productClass) {
                log.info("initializing ProductClass "+productClass);
            }
        }, ProductClass.class);

        HibernateStore hibernateStore = new HibernateStore.Builder(entityManagerFactory.unwrap(SessionFactory.class)).build();
        hibernateStore.populateEntityDictionary(entityDictionary);

        return new Elide(new ElideSettingsBuilder(hibernateStore)
                .withAuditLogger(new Slf4jLogger())
                .withEntityDictionary(entityDictionary)
                .withJoinFilterDialect(rsqlFilterDialect)
                .withSubqueryFilterDialect(rsqlFilterDialect)
                .build());
    }
}

```

aklish:
Did you call ```entityDictionary.bindEntity``` on ProductClass somewhere?

aklish:
Here is another more complicated example that we use in our production servers.  We use Jersey (which has HK2 for dependency injection).  We bind HK2's ```ServiceLocator::inject``` method as our initializer for every class:

To do this, we override the ```HibernateStore``` and do our own entity bindings as follows:
```
    @Override
    public void populateEntityDictionary(EntityDictionary dictionary) {
        /* bind all entities */
        for (ClassMetadata meta : getFactory().getAllClassMetadata().values()) {
            Class mappedClass = meta.getMappedClass(EntityMode.POJO);

[...]

PR #524, opened 2017-09-06
Fixes Issue #523

aklish:


aklish:
Fixes issue #521

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -676,6 +679,10 @@ public boolean isShareable(Class<?> entityClass) {
      * @param cls Entity bean class
      */
     public void bindEntity(Class<?> cls) {
+        if (entityBindings.containsKey(cls)) {
```
wcekan: This should be
```java
        entityBindings.containsKey(lookupEntityClass(cls))
```

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -657,6 +657,9 @@ public String getNameFromAlias(Object entity, String alias) {
      * @param cls         Class to bind initialization
      */
     public <T> void bindInitializer(Initializer<T> initializer, Class<T> cls) {
+        if (! entityBindings.containsKey(cls)) {
```
wcekan: This should call
```java
     entityBindings.containsKey(lookupEntityClass(cls))
```

PR #527, opened 2017-09-07
GraphQL Sorting & Pagination

DennisMcWherter:


elide-core/src/main/java/com/yahoo/elide/core/pagination/Pagination.java
```diff
@@ -81,11 +81,46 @@ private Pagination(Map<PaginationKey, Integer> pageData, int defaultMaxPageSize,
...
 
     /**
+     * Given an offset and first parameter from GraphQL, generate page and pageSize values.
+     *
+     * @param offsetString Provided offset string
+     * @param firstString Provided first string
+     * @param elideSettings Elide settings object containing default pagination values
+     * @return The new Pagination object.
+     */
+    public static Pagination fromOffsetAndFirst(String offsetString, String firstString, ElideSettings elideSettings) {
```
aklish: I would make this take an Optional<Long> for both offset and first params.

PR #528, opened 2017-09-11
GraphQL Filtering

DennisMcWherter:
Support for filtering in GraphQL. Builds off of: https://github.com/yahoo/elide/pull/527


elide-core/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -531,4 +545,22 @@ private Object replaceObjects(Environment context) {
...
         return sort.map(Sorting::parseSortRule);
     }
+
+    private Optional<FilterExpression> buildFilter(String typeName,
+                                                   Optional<String> filter,
+                                                   RequestScope requestScope) {
+        return filter.map(filterStr -> {
+            MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>() {
+                {
+                    put("filter[" + typeName + "]", Arrays.asList(filterStr));
```
aklish: Can we refactor the filterDialect so that it just takes a String instead of a map of query params.

PR #529, opened 2017-09-15
Fix build

DennisMcWherter:


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -782,56 +781,98 @@ public boolean isIdGenerated() {
...
             String idField = dictionary.getIdFieldName(entityType);
 
-            filterExpression = Optional.of(new FilterPredicate(
-                    new FilterPredicate.PathElement(
-                            entityType,
-                            idType,
-                            idField),
-                    Operator.IN,
-                    Collections.singletonList(idVal)));
+            if (!ids.isEmpty()) {
```
aklish: We need to split the list of IDs into those that already exist and those that are new.  We should build ID filters for the existing and query the data store.  For the IDs that are new, we should get those from the Request scope.

The union of these two sets is what should be returned.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1672,27 +1712,13 @@ protected void addInverseRelation(String relationName, Object inverseObj) {
...
-     * @param resources  the resources
-     * @param skipNew whether or not objects created this transaction should be skipped
-     * @return Filtered set of resources
-     */
-    protected static Set<PersistentResource> filter(Class<? extends Annotation> permission,
-                                                    Set<PersistentResource> resources, boolean skipNew) {
         Set<PersistentResource> filteredSet = new LinkedHashSet<>();
         for (PersistentResource resource : resources) {
             try {
+                boolean skipNew = resource.getRequestScope().isMutatingMultipleEntities();
```
aklish: This variable name doesn't match what we are checking.  Why not call it isMutation or something like that?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
DennisMcWherter: Done. See below.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -782,56 +781,126 @@ public boolean isIdGenerated() {
...
-                filter(ReadPermission.class,
-                        (Set) getRelationChecked(relation, filterExpression, Optional.empty(), Optional.empty()),
-                        skipNew);
-
-        for (PersistentResource childResource : resources) {
-            if (childResource.matchesId(id)) {
-                return childResource;
+            if (!ids.isEmpty()) {
+                String typeAlias = dictionary.getJsonAliasFor(entityType);
+                Map<String, PersistentResource> newResourceMap = requestScope.getNewPersistentResources().stream()
```
aklish: Can we simplify these two lines like:
```
newResources = requestScope.getNewPersistentResources().stream()
    .filter(resource -> typeAlias.equals(resource.getType()))
    .filter(resource -> ids.contains(resource.getUUID())
    .collect(Collectors.toList()));

File: elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java [hunk 4]
```diff
@@ -782,56 +781,126 @@ public boolean isIdGenerated() {
...
+
+        Set<PersistentResource> existingResources = filter(ReadPermission.class,
+            (Set) getRelationChecked(relation, Optional.ofNullable(filterExpression), sorting, pagination));
+
+        // TODO: Sort again in memory now that two sets are glommed together?
+
+        Set<PersistentResource> allResources = Sets.union(newResources, existingResources);
+
+        String missedIds = ids.stream()
+                .filter(id -> allResources.stream().noneMatch(p -> p.matchesId(id)))
```
aklish: Simpler to do:
```
List<String> allIds = allResources.stream().map(PersistentResoruce::getUUID())
List<String> missingIds = Sets.difference(ids, allIds)
if (!missedIds.isEmpty()) {
   throw new InvalidObjectIdentifierException(missedIds, relation);
}
```

DennisMcWherter: Still a problem here: `matchesId` does more than a simple comparison against UUID. Since this list isn't _exclusively_ UUID's, I think this would likely be wrong.

PR #530, opened 2017-09-15
Simplify skipNew.

DennisMcWherter:


DennisMcWherter:
Closing since we've decided not to port this to elide 3.x

PR #532, opened 2017-09-19
Add standalone application.

DennisMcWherter:
MVP of Elide standalone application.

### DIFF
--- a/elide-standalone/src/main/java/com/yahoo/elide/standalone/interfaces/UserExtractionFunctionProvider.java
+++ b/elide-standalone/src/main/java/com/yahoo/elide/standalone/interfaces/UserExtractionFunctionProvider.java
@@ -1,0 +1,21 @@
+/*
+ * Copyright 2017, Oath Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.standalone.interfaces;
+
+import com.yahoo.elide.resources.DefaultOpaqueUserFunction;
+
+/**
+ * User extraction function provider.
+ */
+public interface UserExtractionFunctionProvider {
+
+    /**
+     * User extraction function provider.
+     *
+     * @return User extraction function.
+     */
+    DefaultOpaqueUserFunction getUserExtractionFunction();
+}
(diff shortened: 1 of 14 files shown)


elide-standalone/src/main/java/com/yahoo/elide/standalone/interfaces/UserExtractionFunctionProvider.java
```diff
@@ -0,0 +1,21 @@
...
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.standalone.interfaces;
+
+import com.yahoo.elide.resources.DefaultOpaqueUserFunction;
+
+/**
+ * User extraction function provider.
+ */
+public interface UserExtractionFunctionProvider {
```
aklish: I wonder if example would be simpler if we just directly used elide builder rather than di to setup elide?

elide-standalone/example-app/src/main/java/com/yourcompany/elide/filters/UserAuthFilter.java
```diff
@@ -0,0 +1,58 @@
...
+            // Fetch a single account by user name
+            Account account = (Account) store.getSession()
+                .createCriteria(Account.class)
+                .add(Restrictions.eq("name", userName))
+                .uniqueResult();
+            // Validate password hash
+            boolean success = account != null && BCrypt.checkpw(plaintextPassword, account.getPassword());
+            requestContext.setSecurityContext(
+                new AuthSecurityContext(requestContext.getSecurityContext(), new UserPrincipal(account, success)));
+            tx.commit(null);
```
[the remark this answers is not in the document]
DennisMcWherter: It is in hibernate5. Just used for example. We need to commit before we close.

elide-standalone/src/main/java/com/yahoo/elide/standalone/interfaces/UserExtractionFunctionProvider.java
DennisMcWherter: We do have this capability. We have 2 modes. In short, you have mechanisms to use ElideBuilder (full customizability, but need to setup everything still), or you give us a list of your security checks and user extraction function and we set everything else up with Hibernate5.

PR #537, opened 2017-09-21
Allow for arbitrary relationships that aren't transitive.

cythrawll:
The main use case here is allowing for hibernate's `@Any` annotation.
This allows for polymorphic relationship to unrelated models.
Elide gave me two challenges.

1.) Needed annotations outside of JPA so Elide recognized them as relationships
2.) Needed the EntityDictionary to allow Interfaces (+2 squashed commits)
Squashed commits:

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
@@ -117,8 +117,10 @@ public class EntityDictionary {
     }
 
     protected EntityBinding getEntityBinding(Class<?> entityClass) {
-        Class<?> entitySearched = isMappedInterface(entityClass) ? entityClass : lookupEntityClass(entityClass);
-        return entityBindings.getOrDefault(entitySearched, EntityBinding.EMPTY_BINDING);
+        if (isMappedInterface(entityClass)) {
+            return EntityBinding.EMPTY_BINDING;
+        }
+        return entityBindings.getOrDefault(lookupEntityClass(entityClass), EntityBinding.EMPTY_BINDING);
     }
 
     public boolean isMappedInterface(Class<?> interfaceClass) {
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
@@ -121,7 +121,7 @@ public class EntityDictionary {
         return entityBindings.getOrDefault(entitySearched, EntityBinding.EMPTY_BINDING);
     }
 
-    protected boolean isMappedInterface(Class<?> interfaceClass) {
+    public boolean isMappedInterface(Class<?> interfaceClass) {
         return interfaceClass.isInterface() && interfaceClass.isAnnotationPresent(MappedInterface.class);
     }
 
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
@@ -12,6 +12,7 @@ import com.yahoo.elide.annotation.ComputedAttribute;
 import com.yahoo.elide.annotation.ComputedRelationship;
 import com.yahoo.elide.annotation.Exclude;
 import com.yahoo.elide.annotation.Include;
+import com.yahoo.elide.annotation.MappedInterface;
 import com.yahoo.elide.annotation.SharePermission;
 import com.yahoo.elide.core.exceptions.DuplicateMappingException;
 import com.yahoo.elide.security.checks.Check;
@@ -116,7 +117,12 @@ public class EntityDictionary {
     }
 
     protected EntityBinding getEntityBinding(Class<?> entityClass) {
-        return entityBindings.getOrDefault(lookupEntityClass(entityClass), EntityBinding.EMPTY_BINDING);
+        Class<?> entitySearched = isMappedInterface(entityClass) ? entityClass : lookupEntityClass(entityClass);
+        return entityBindings.getOrDefault(entitySearched, EntityBinding.EMPTY_BINDING);
+    }
+
+    protected boolean isMappedInterface(Class<?> interfaceClass) {
+        return interfaceClass.isInterface() && interfaceClass.isAnnotationPresent(MappedInterface.class);
     }
 
     /**
--- a/elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
+++ b/elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
@@ -16,6 +16,9 @@ public class AnyPolymorphismIT {
     private static final String JSONAPI_CONTENT_TYPE = "application/vnd.api+json";
     private final JsonParser jsonParser = new JsonParser();
 
+    private final int horsepower = 102;
+
+
     @BeforeClass
     public void setUp() {
         //Create a tractor
@@ -49,7 +52,6 @@ public class AnyPolymorphismIT {
         final String includedType = "included[0].type";
         final String includedId = "included[0].id";
         final String includedSize = "included.size()";
-        final int horsepower = 102;
 
         RestAssured
                 .given()
@@ -120,7 +122,6 @@ public class AnyPolymorphismIT {
                         "included[0].attributes.type", equalTo("android"),
                         includedSize, equalTo(1));
 
-
         RestAssured
                 .given()
                 .contentType(JSONAPI_CONTENT_TYPE)
@@ -130,6 +131,20 @@ public class AnyPolymorphismIT {
                 .assertThat()
                 .statusCode(HttpStatus.SC_OK)
                 .body("data.size()", equalTo(2));
+    }
+
+    public void testAnySubpaths() {
+        final String propertyPath = "/property";
+
+        RestAssured
+                .given()
+                .contentType(JSONAPI_CONTENT_TYPE)
+                .accept(JSONAPI_CONTENT_TYPE)
+                .body(jsonParser.getJson("/AnyPolymorphismIT/AddTractorProperty.json"))
+                .post(propertyPath)
+                .then()
+                .assertThat()
+                .statusCode(HttpStatus.SC_CREATED);
 
         RestAssured
                 .given()
@@ -150,7 +165,7 @@ public class AnyPolymorphismIT {
                 .assertThat()
                 .statusCode(HttpStatus.SC_OK)
                 .body("data.attributes.horsepower", equalTo(horsepower),
-                         "data.id", equalTo("1"));
+                        "data.id", equalTo("1"));
 
         //single entity so no page appropriate stuff
         RestAssured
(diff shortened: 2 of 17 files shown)


yahoocla:

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
```diff
@@ -0,0 +1,122 @@
...
+                .contentType(JSONAPI_CONTENT_TYPE)
+                .accept(JSONAPI_CONTENT_TYPE)
+                .body("{\"data\": {\"type\": \"smartphone\", \"attributes\": {\"type\": \"android\" }}}")
+                .post("/smartphone")
+                .then()
+                .statusCode(HttpStatus.SC_CREATED);
+    }
+
+    @Test
+    public void testAny() {
```
aklish: There are 4 different ways Elide generates HQL:
```
/someRootEntity - a query against a root entity
/someRootEntity?page[totals] - a query including page totals
/someRootEntity/24/anotherEntity - a query against a subcollection
/someRootEntity/24/anotherEntity?page[totals] - a query including page totals against a subcollection.
```
To really make sure this works, we should have a test for all of the above.

cythrawll: Tests have been added. They did show one issue which I fixed.

aklish:
A few other questions/comments.  Is this new mechanism only intended to work for reads?

We should test create, delete, update as well - and make sure the appropriate thing happens.  Maybe the appropriate thing is to return a 400 error to the caller.  I suspect things will blow up instead.

For reads, we need to add tests for things liked sorting, filters, pagination, & computed attributes - making sure everything still works as expected.

Finally, since this now changes the JSON-API state machine, we'll need another PR against the elide-4.x branch to make any necessary changes to graphQL.

cythrawll:
The current scope of this is to support the @ToOne and @Any with Creates and Reads, I can test delete and update as well and make sure they do the appropriate thing.  

I do have a separate branch to work on supporting @ManyToAny and @ToMany, in which we do need to test things like sorting, filters, pagination.  You are correct that they currently do break pretty badly.  Which so far I've had to return 400 for a few of them due to hibernate limitations on creating filtered queries on them.

So since for what I need it for, the @ToOne and @Any use case that already works well that meets my immediate needs. Followed up better support for the @ToMany and @ManyToAny in a future PR.

cythrawll:
I think I ran into a bug with codacy, or am I missing something?

cythrawll:
Ready for review, let me know if there is more here.  The ManyToAny use case will come in a following PR.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -116,7 +117,12 @@ public static Method findMethod(Class<?> entityClass, String name, Class<?>... p
     }
 
     protected EntityBinding getEntityBinding(Class<?> entityClass) {
-        return entityBindings.getOrDefault(lookupEntityClass(entityClass), EntityBinding.EMPTY_BINDING);
+        Class<?> entitySearched = isMappedInterface(entityClass) ? entityClass : lookupEntityClass(entityClass);
```
aklish: Would the intent be clearer here if we explicitly return EntityBinding.EMPTY_BINDING if the interface is mapped?

```
if (isMappedInterface(entityClass) {
    return EntityBindings.EMPTY_BNDING;
}
```

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
```diff
@@ -0,0 +1,269 @@
...
+import static org.hamcrest.Matchers.greaterThan;
+
+public class AnyPolymorphismIT extends AbstractApiResourceInitializer {
+    private static final String JSONAPI_CONTENT_TYPE = "application/vnd.api+json";
+    private final JsonParser jsonParser = new JsonParser();
+
+    private final String sep = "/";
+
+    private final int horsepower = 102;
+    private final String one = "1";
```
aklish: I realize codacy probably made you do this, but I personally find the tests harder to read with this indirection.

aklish: This applies for all of these string constants.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
```diff
@@ -0,0 +1,269 @@
...
+        //single entity so no page appropriate stuff
+        RestAssured
+                .given()
+                .contentType(JSONAPI_CONTENT_TYPE)
+                .accept(JSONAPI_CONTENT_TYPE)
+                .get(propertyPath + sep + id + "/myStuff?page[totals]")
+                .then()
+                .assertThat()
+                .statusCode(HttpStatus.SC_OK)
+                .body("meta", equalTo(null));
```
aklish: This doesn't seem right to me.  Why would meta be null?

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
```diff
@@ -0,0 +1,269 @@
...
+
+        RestAssured
+                .given()
+                .contentType(JSONAPI_CONTENT_TYPE)
+                .accept(JSONAPI_CONTENT_TYPE)
+                .get(propertyPath + "?page[totals]")
+                .then()
+                .assertThat()
+                .statusCode(HttpStatus.SC_OK)
+                .body("meta.page.totalRecords", greaterThan(0));
```
aklish: Can you use an exact comparison here to make sure it equals what you expect?

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
```diff
@@ -0,0 +1,269 @@
...
+                .body(jsonParser.getJson(tractorAsPropertyFile))
+                .post(propertyPath)
+                .then()
+                .assertThat()
+                .statusCode(HttpStatus.SC_CREATED)
+                .body(relationshipType, equalTo(tractorType))
+                .extract()
+                .path(idPath);
+
+        String prefix = "{\"data\": {\"id\": \"" + id;
```
aklish: Same comment about readability.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
```diff
@@ -0,0 +1,269 @@
...
+                .given()
+                .contentType(JSONAPI_CONTENT_TYPE)
+                .accept(JSONAPI_CONTENT_TYPE)
+                .get(propertyPath + sep + id)
+                .then()
+                .assertThat()
+                .statusCode(HttpStatus.SC_OK)
+                .body("data.relationships.myStuff.data", equalTo(null));
+    }
+
```
aklish: I'm curious to see what would happen if you tried to query /device

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
cythrawll: I 100% agree, is there a way to get codacy to like it?   Or do we ignore these in codacy?

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
cythrawll: Because this relationship is a single entity so it doesn't paginate (a one to one), the pagination is ignored here.  I found this to be consistent with the rest of elide.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
clayreimann: Codacy should now ignore files in `*/src/test/*`

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
cythrawll: Yeah this is a subjective one. I actually had that at first, but after refactoring it a few times.   I thought this way ended up being more clear on "why we're doing this".  But I might be coming from a bad perspective.  So input would be appreciated.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
cythrawll: It returns a 404 like it should.  added test to show this.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
```diff
@@ -216,41 +195,57 @@ public void testAnySubpaths() {
...
+                .statusCode(HttpStatus.SC_NOT_FOUND)
+                .body("errors[0]", containsString("Unknown collection"));
 
         //single entity so no page appropriate stuff
         RestAssured
                 .given()
                 .contentType(JSONAPI_CONTENT_TYPE)
                 .accept(JSONAPI_CONTENT_TYPE)
-                .get(propertyPath + sep + id + "/myStuff?page[totals]")
+                .get("/property/" + id + "/myStuff?page[totals]")
```
aklish: I still don't think "meta" should be null below.  This looks like a bug.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
```diff
@@ -216,41 +195,57 @@ public void testAnySubpaths() {
...
-                        idPath, equalTo(one));
+                .body("data.attributes.horsepower", equalTo(102),
+                        "data.id", equalTo("1"));
+
+        //show that you can't get a relation by the interface type
+        RestAssured
+                .given()
+                .contentType(JSONAPI_CONTENT_TYPE)
+                .accept(JSONAPI_CONTENT_TYPE)
+                .get("/property/" + id + "/device")
```
aklish: I really meant to test /device.

Since it is a mapped interface, I wonder how loading a root level collection for a mapped interface might behave.

cythrawll: It 404s the exact same way as this path does.  Which is what I'd expect because the mappings belong to the entities. a MappedInterface isn't an entity.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/AnyPolymorphismIT.java
cythrawll: this means that meta isn't a valid jsonpath. (meta doesn't exist in the response), not that meta: null

PR #539, opened 2017-09-21
Allow case insensitive comparisons.

### LINKED ISSUE #387: Support case insensitive filtering (RSQL)
RSQL should support a case insensitive filtering syntax. A ticket has been [opened here](https://github.com/jirutka/rsql-parser/issues/25) with the RSQL repository, but we can also consider contributing our own scheme back to avoid stalling too much on this feature.

**NOTE:** After further investigating the [FIQL Spec](https://tools.ietf.org/html/draft-nottingham-atompub-fiql-00) (a subset of RSQL), it looks like searches should generally be case _insensitive_. Consequently, we maybe should update our RSQL implementation and look into case **sensitive** support.

clayreimann:
RSQL == and =out= will now behave in a case insensitive manner.

Closes #387

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
@@ -16,6 +16,7 @@ import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.NonNull;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.tuple.Pair;
 
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -23,6 +24,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 /**
  * Predicate class.
@@ -88,6 +91,13 @@ public class FilterPredicate implements FilterExpression, Function<RequestScope,
      */
     public String getParameterName() {
         return getFieldPath().replace(PERIOD, UNDERSCORE) + UNDERSCORE + Integer.toHexString(hashCode());
+    }
+
+    public List<Pair<String, Object>> getNamedParameters() {
+        String baseName = getParameterName() + UNDERSCORE;
+        return IntStream.range(0, values.size())
+                .mapToObj(idx -> Pair.of(String.format("%s%d", baseName, idx), values.get(idx)))
+                .collect(Collectors.toList());
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
@@ -38,7 +38,6 @@ public class FilterPredicate implements FilterExpression, Function<RequestScope,
     @Getter @NonNull private Path path;
     @Getter @NonNull private Operator operator;
     @Getter @NonNull private List<Object> values;
-//    @Getter @NonNull private String fieldPath;
 
     public static boolean toManyInPath(EntityDictionary dictionary, Path path) {
         return path.getPathElements().stream()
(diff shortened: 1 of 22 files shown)


elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
```diff
@@ -7,108 +7,109 @@
...
         this.operator = op;
-        this.values = values;
-    }
+        this.values = new ArrayList<>(values);
 
-    public FilterPredicate(FilterPredicate copy) {
-        this.path = new Path(new ArrayList<>(copy.getPath().getPathElements()));
-        this.operator = copy.getOperator();
-        this.values = new ArrayList<>(copy.getValues());
+        // TODO: figure out why this cannot be here :(
```
DennisMcWherter: Should we resolve this before merging?

Also, what problems are we seeing with this?

elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
```diff
@@ -197,4 +193,22 @@ public void negate() {
...
+    public static class FilterParameter {
+        @Getter private String name;
+        @Getter private Object value;
+
+        public String getPlaceholder() {
+            return ":" + name;
+        }
+
+        public String escapeMatching() {
+            // It is unclear why replaceAll here breaks our tests
```
DennisMcWherter: Should we investigate this? We wouldn't want to accidentally introduce an injection attack.

aklish:
Let's update the change log for elide 4.x

elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
```diff
@@ -5,110 +5,109 @@
...
+    public List<FilterParameter> getParameters() {
+        String baseName = String.format("%s_%s_",
+                getFieldPath().replace(PERIOD, UNDERSCORE),
+                Integer.toHexString(hashCode()));
+        return IntStream.range(0, values.size())
+                .mapToObj(idx -> new FilterParameter(String.format("%s%d", baseName, idx), values.get(idx)))
+                .collect(Collectors.toList());
+    }
+
+    public FilterPredicate scopedBy(PathElement scope) {
```
aklish: Maybe a comment here about what this is doing?  Could this be package and not public?

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -169,20 +185,21 @@ public static Operator fromString(final String string) {
...
-            Object val = getFieldValue(entity, field, requestScope);
+            Class type = requestScope.getDictionary().getType(entity, field);
+            if (!type.isAssignableFrom(String.class)) {
+                throw new IllegalStateException("Cannot case insensitive compare non-string values");
+            }
 
-            return val == null || values.stream()
-                    .map(v -> CoerceUtil.coerce(v, val.getClass()))
-                    .noneMatch(val::equals);
+            String val = transform.apply((String) getFieldValue(entity, field, requestScope));
```
aklish: I'm wondering if it is a good idea to assume everything must be string assignable for the in operator.

clayreimann: I just followed the convention elsewhere in this file to have a specialized version of the function share the same name. In reality this `in` is only called for case insensitive in operations. I'll add clarifying comments though

elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
clayreimann: Not the same package, so can't change the access level. I'll add a comment though.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -322,17 +324,18 @@ private FilterExpression equalityExpression(String argument, Path path, List<Obj
          * @return Returns Predicate for '=isnull=' case depending on its arguments.
          */
         private FilterExpression buildIsNullOperator(Path path, List<String> arguments) {
-            Operator elideOP;
+            String arg = arguments.get(0);
```
aklish: Is it required that there is an argument here?  This will NPE if no argument is provided.

PR #542, opened 2017-09-25
Evaluate SharePermission falls back to ReadPermission

tarrantzhang:
Some changes to SharePermission:

1. An entity with no `SharePermission` is not shareable now. 
2. Whenever SharePermission is evaluated, falls back to ReadPermission of the entity.
3. Remove `expression` from SharePermission since it's useless now.
4. Add a `boolean isShareable()` to SharePermission to allow a way of override package level `SharePermission`.

### DIFF
--- a/elide-contrib/dropwizard-elide/src/test/java/com/yahoo/elide/contrib/dropwizard/elide/Author.java
+++ b/elide-contrib/dropwizard-elide/src/test/java/com/yahoo/elide/contrib/dropwizard/elide/Author.java
@@ -22,7 +22,7 @@ import java.util.Collection;
  * Model for authors
  */
 @Entity
-@SharePermission()
+@SharePermission
 @Table(name = "author")
 @Include(rootLevel = true)
 public class Author {
--- a/elide-contrib/dropwizard-elide/src/test/java/com/yahoo/elide/contrib/dropwizard/elide/Author.java
+++ b/elide-contrib/dropwizard-elide/src/test/java/com/yahoo/elide/contrib/dropwizard/elide/Author.java
@@ -22,7 +22,7 @@ import java.util.Collection;
  * Model for authors
  */
 @Entity
-@SharePermission(expression = "Prefab.Role.All")
+@SharePermission()
 @Table(name = "author")
 @Include(rootLevel = true)
 public class Author {
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
@@ -669,10 +669,8 @@ public class EntityDictionary {
      * @return true if entityClass is shareable.  False otherwise.
      */
     public boolean isShareable(Class<?> entityClass) {
-        return entityHasChecksForPermission(
-                entityClass,
-                SharePermission.class
-        ) && getAnnotation(entityClass, SharePermission.class).sharable();
+        return getAnnotation(entityClass, SharePermission.class) != null
+                && getAnnotation(entityClass, SharePermission.class).sharable();
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
@@ -669,7 +669,10 @@ public class EntityDictionary {
      * @return true if entityClass is shareable.  False otherwise.
      */
     public boolean isShareable(Class<?> entityClass) {
-        return getAnnotation(entityClass, SharePermission.class) != null;
+        return entityHasChecksForPermission(
+                entityClass,
+                SharePermission.class
+        ) && getAnnotation(entityClass, SharePermission.class).sharable();
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
@@ -75,15 +75,17 @@ public class EntityPermissions implements CheckInstantiator {
                              Collection<AccessibleObject> fieldOrMethodList)  {
         this.dictionary = dictionary;
         for (Class<? extends Annotation> annotationClass : PERMISSION_ANNOTATIONS) {
-            ParseTree classPermission = bindClassPermissions(cls, annotationClass);
             final Map<String, ParseTree> fieldPermissions = new HashMap<>();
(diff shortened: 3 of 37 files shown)


elide-contrib/dropwizard-elide/src/test/java/com/yahoo/elide/contrib/dropwizard/elide/Author.java
```diff
@@ -22,7 +22,7 @@
  * Model for authors
  */
 @Entity
-@SharePermission(expression = "Prefab.Role.All")
+@SharePermission()
```
DennisMcWherter: No need for `()` at the end.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -669,7 +669,10 @@ public String getNameFromAlias(Object entity, String alias) {
      * @return true if entityClass is shareable.  False otherwise.
      */
     public boolean isShareable(Class<?> entityClass) {
-        return getAnnotation(entityClass, SharePermission.class) != null;
+        return entityHasChecksForPermission(
```
DennisMcWherter: Does the first part of this statement make sense anymore or must everything be explicitly _marked_ shareable now?

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
```diff
@@ -101,6 +104,9 @@ private void bindMemberPermissions(Map<String, ParseTree> fieldPermissions,
 
     private ParseTree getPermissionExpressionTree(Class<? extends Annotation> annotationClass, Annotation annotation) {
         try {
+            if (annotationClass == SharePermission.class) {
+                return null;
```
DennisMcWherter: This should never happen, should it? `ReadPermission` should now be evaluated instead?

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
tarrantzhang: So basically entity with `@SharePermission` is shareable, **No SharePermission**=**Not sharable**, also `@SharePermission(false)`=not shareable

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
```diff
@@ -75,12 +75,17 @@ public EntityPermissions(EntityDictionary dictionary,
...
         for (Class<? extends Annotation> annotationClass : PERMISSION_ANNOTATIONS) {
-            ParseTree classPermission = bindClassPermissions(cls, annotationClass);
             final Map<String, ParseTree> fieldPermissions = new HashMap<>();
             fieldOrMethodList.stream()
                     .forEach(member -> bindMemberPermissions(fieldPermissions, member, annotationClass));
-            if (classPermission != null || !fieldPermissions.isEmpty()) {
-                bindings.put(annotationClass, new AnnotationBinding(classPermission, fieldPermissions));
+            if (annotationClass == SharePermission.class
+                    && EntityDictionary.getFirstAnnotation(cls, Arrays.asList(annotationClass)) != null) {
+                bindings.put(SharePermission.class, new AnnotationBinding(null, fieldPermissions));
```
aklish: Should we be binding the read permissions here?

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -111,7 +111,14 @@ public ActivePermissionExecutor(boolean verbose, final RequestScope requestScope
                                                                    ChangeSpec changeSpec) {
         Supplier<Expression> expressionSupplier = () -> {
             if (SharePermission.class == annotationClass) {
-                return expressionBuilder.buildSharePermissionExpressions(resource);
+                if (requestScope.getDictionary().isShareable(resource.getResourceClass())) {
+                    return expressionBuilder.buildAnyFieldExpressions(resource, ReadPermission.class, changeSpec);
+                } else {
+                    ForbiddenAccessException e = new ForbiddenAccessException("Missing SharePermission for "
```
aklish: So the expressionSupplier just builds an expression - it doesn't actually execute it.  Inline with this code, this could build an expression that explicitly denies permission.

Later - executing that expression will throw the appropriate exception.

elide-integration-tests/src/test/java/example/NoShareBiDirectional.java
```diff
@@ -17,7 +17,7 @@
 
 @Entity
 @Include(rootLevel = true, type = "noShareBid")
-@SharePermission(expression = "deny all")
+@SharePermission
```
aklish: Shouldn't this be ```SharePermission(shareable=false)```

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
tarrantzhang: If I understand it right, I think we still bind SharePermission here to indicate if the entity is shareable or not, and when SharePermission evaluation is trigger in `PersistenceResource` we then fall back to evaluate the ReadPermission of entity.

elide-core/src/main/java/com/yahoo/elide/security/permissions/PermissionExpressionBuilder.java
```diff
@@ -97,7 +97,10 @@ public PermissionExpressionBuilder(ExpressionResultCache cache, EntityDictionary
      *
      * @param resource        Resource
      * @return Commit and operation expressions
+     *
+     * @deprecated Trying to evaluate SharePermission should fall to ReadPermission in elide 4.0
      */
+    @Deprecated
```
aklish: Since this is Elide 4.0 - you can just remove this now.

elide-core/src/main/java/com/yahoo/elide/core/EntityPermissions.java
aklish: Can we just skip the binding in this case (call next on SharePermission).

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
aklish: How about something like:
```
Annotation shareAnnotation = getAnnotation(entityClass, SharePermission.class)
return (shareAnnotation !=null && shareAnnotation.shareable())
```

PR #544, opened 2017-09-26
Use CreatePermission to set field on newly created resources

suriyasundar92:


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -389,12 +396,22 @@ public boolean updateRelation(String fieldName, Set<PersistentResource> resource
...
-                    UpdatePermission.class,
-                    fieldName,
-                    resourceIdentifiers.stream().map(PersistentResource::getObject).collect(Collectors.toList()),
-                    resources.stream().map(PersistentResource::getObject).collect(Collectors.toList())
-            );
+            if (isNewlyCreated()) {
+                checkFieldAwareDeferPermissions(
+                        CreatePermission.class,
+                        fieldName,
+                        resourceIdentifiers.stream().map(PersistentResource::getObject).collect(Collectors.toList()),
```
DennisMcWherter: Can we consolidate this input and the one following?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -543,8 +564,14 @@ protected boolean updateToOneRelation(String fieldName,
      */
     public boolean clearRelation(String relationName) {
         Set<PersistentResource> mine = filter(ReadPermission.class, getRelationUncheckedUnfiltered(relationName));
-        checkFieldAwareDeferPermissions(UpdatePermission.class, relationName, Collections.emptySet(),
-                mine.stream().map(PersistentResource::getObject).collect(Collectors.toSet()));
+
+        if (isNewlyCreated()) {
+            checkFieldAwareDeferPermissions(CreatePermission.class, relationName, Collections.emptySet(),
+                    mine.stream().map(PersistentResource::getObject).collect(Collectors.toSet()));
```
DennisMcWherter: Can we compute this once for both calls?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1435,12 +1495,22 @@ protected void delFromCollection(
             final Collection removedBidrectional = CollectionUtils
                     .disjunction(Collections.singleton(this.getObject()), originalBidirectional);
 
-            toDelete.checkFieldAwareDeferPermissions(
-                    UpdatePermission.class,
-                    inverseField,
-                    removedBidrectional,
-                    originalBidirectional
-            );
+            if (toDelete.isNewlyCreated()) {
```
DennisMcWherter: This should never happen afaik. @aklish while this sort of thing would be _technically_ possible, should we disallow it outright?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
tarrantzhang: Can we refactor the following segment even more into it's own function? Seen a lot of occurences in persistence resource:
```
if(isNewlyCreated()) {
    checkFieldAwareDeferPermissions( CreatePermission.class, ...);
} else {
    checkFieldAwareDeferPermissions( UpdatePermission.class, ...);
}
```

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1330,7 +1366,12 @@ protected void nullValue(String fieldName, PersistentResource oldValue) {
         }
         String inverseField = getInverseRelationField(fieldName);
         if (!inverseField.isEmpty()) {
-            oldValue.checkFieldAwareDeferPermissions(UpdatePermission.class, inverseField, null, getObject());
+            if (isNewlyCreated()) {
```
tarrantzhang: Should we check `oldValue.isNewlyCreated()` instead of `this.isNewlyCreated()` since we are calling `oldValue.checkFieldAwareDeferPermissions` below?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -224,6 +224,9 @@ protected PersistentResource(T obj, PersistentResource<?> parent, RequestScope r
         this(obj, parent, requestScope.getUUIDFor(obj), requestScope);
     }
 
+    protected boolean isNewlyCreated() {
```
tarrantzhang: A function level documentation is appreciated since it might be called by another class within the same package.  Also unluckily another [variable](https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java#L1459) in the same file has declared the name `isNewlyCreated`, you might consider changing either of their name to avoid confusion.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
aklish: Maybe make a function for this.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
clayreimann: Actually, I think this function should be used in place of the variable that @tarrantzhang is referring to.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
clayreimann: This definitely wants to be a function.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
clayreimann: Yes.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1375,11 +1416,20 @@ protected Object getValueUnchecked(String fieldName) {
...
-        checkFieldAwareDeferPermissions(
-                UpdatePermission.class,
-                collectionName,
-                CollectionUtils.union(CollectionUtils.emptyIfNull(collection), singleton),
-                original);
+        if (isNewlyCreated()) {
+            checkFieldAwareDeferPermissions(
+                    CreatePermission.class,
+                    collectionName,
+                    CollectionUtils.union(CollectionUtils.emptyIfNull(collection), singleton),
```
clayreimann: Again, can we compute this once for both cases?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1412,12 +1462,22 @@ protected void delFromCollection(
...
-                UpdatePermission.class,
-                collectionName,
-                CollectionUtils.disjunction(collection, Collections.singleton(toDelete.getObject())),
-                original
-        );
+        if (isNewlyCreated()) {
+            checkFieldAwareDeferPermissions(
+                    CreatePermission.class,
+                    collectionName,
+                    CollectionUtils.disjunction(collection, Collections.singleton(toDelete.getObject())),
```
clayreimann: Ditto.

PR #545, opened 2017-09-28
Remove more deprecated code

clayreimann:


DennisMcWherter:
Looks good, but doesn't build.

PR #548, opened 2017-09-29
@ManyToAny support for polymorphic relations

cythrawll:
Other than writing tests, the biggest changes were making sure that the system sent 400s for filtering that's not supported on this @ManyToAny (pagination and filtering are not supported due to limitations on hibernate's HQL generator)

The biggest change was to the orderedById tree, when serializing the model, this tree ended up squashing objects with the same id together, even though they had different types.  So that is why you see me adding the type to the key.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -14,7 +14,6 @@ import com.yahoo.elide.annotation.DeletePermission;
 import com.yahoo.elide.annotation.OnReadPreSecurity;
 import com.yahoo.elide.annotation.ReadPermission;
 import com.yahoo.elide.annotation.SharePermission;
-import com.yahoo.elide.annotation.ToMany;
 import com.yahoo.elide.annotation.UpdatePermission;
 import com.yahoo.elide.audit.InvalidSyntaxException;
 import com.yahoo.elide.audit.LogMessage;
@@ -932,9 +931,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
         }
 
         final Class<?> relationClass = dictionary.getParameterizedType(obj, relationName);
-        boolean isAnyToMany = dictionary.isMappedInterface(relationClass)
-                && dictionary.getAttributeOrRelationAnnotation(obj.getClass(), ToMany.class, relationName) != null;
-        if (isAnyToMany || (! requestScope.getPagination().isDefaultInstance()
+        if (dictionary.isMappedInterface(relationClass) || (! requestScope.getPagination().isDefaultInstance()
                 && !CanPaginateVisitor.canPaginate(relationClass, dictionary, requestScope))) {
             throw new InvalidPredicateException(String.format("Cannot paginate %s",
                     dictionary.getJsonAliasFor(relationClass)));
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -14,6 +14,7 @@ import com.yahoo.elide.annotation.DeletePermission;
 import com.yahoo.elide.annotation.OnReadPreSecurity;
 import com.yahoo.elide.annotation.ReadPermission;
 import com.yahoo.elide.annotation.SharePermission;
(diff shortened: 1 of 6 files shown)


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -927,8 +932,10 @@ private boolean interfaceHasFilterExpression(Class<?> entityInterface) {
         }
 
         final Class<?> relationClass = dictionary.getParameterizedType(obj, relationName);
-        if (! requestScope.getPagination().isDefaultInstance()
-                && !CanPaginateVisitor.canPaginate(relationClass, dictionary, requestScope)) {
+        boolean isAnyToMany = dictionary.isMappedInterface(relationClass)
+                && dictionary.getAttributeOrRelationAnnotation(obj.getClass(), ToMany.class, relationName) != null;
```
DennisMcWherter: Are there any ways for us to use `MappedInterface`s and not have them marked `AnyToMany`? If so, this could be an overly restrictive check, right?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -833,6 +834,10 @@ public PersistentResource getRelation(String relation, String id) {
             Object idVal = CoerceUtil.coerce(id, idType);
             String idField = dictionary.getIdFieldName(entityType);
 
+            if (dictionary.isMappedInterface(entityType) && idField == null) {
```
DennisMcWherter: This has some unfortunate consequences. Is the goal that polymorphic collections are read-only at the collection level? I am happy accepting that for now until we have use cases for more in the future, but want to ensure this is what we're going for.

cythrawll: Kinda, the problem is adding filters to @ManyToAny queries in hibernate isn't supported.  The only way to support that as of now is to write your own HQL manually.

You can still create and update. But filtering is a no go.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
cythrawll: This is correct, I will fix it.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
DennisMcWherter: As a corollary, one could not access any individual element within that collection either (i.e. filter to a single element with id). I suspect this is alright and intended though, based on what you've said.

cythrawll: That is correct.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -927,8 +931,8 @@ private boolean interfaceHasFilterExpression(Class<?> entityInterface) {
         }
 
         final Class<?> relationClass = dictionary.getParameterizedType(obj, relationName);
-        if (! requestScope.getPagination().isDefaultInstance()
-                && !CanPaginateVisitor.canPaginate(relationClass, dictionary, requestScope)) {
+        if (dictionary.isMappedInterface(relationClass) || (! requestScope.getPagination().isDefaultInstance()
+                && !CanPaginateVisitor.canPaginate(relationClass, dictionary, requestScope))) {
```
aklish: Why cannot hibernate paginate the mapped interfaces?  What breaks?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
aklish: Can you provide details why this isn't supported - maybe link to hibernate docs or stack overflow?

cythrawll:
Yes, otherwise i'm gonna be stuck in 3.2 :|

DennisMcWherter:
@cythrawll would you be able to detail the information requested by @aklish above?

cythrawll:
Yeah i'm going to have to dig it up again. The documentation isn't clear on it. and I only ran across it after trying and failing and looking up the exceptions and what the hibernate authors had to say about it.  I think most of it was concluded by unclosed bug reports.  I'll see if I can dig em up.

aklish:
No longer needed.

PR #549, opened 2017-09-29
Update Pagination Support in GraphQL

DennisMcWherter:
This PR updates our pagination scheme for our GraphQL support to be [Relay compatible](https://facebook.github.io/relay/graphql/connections.htm).

Notable changes:

  * Objects (rootable and relationships) are now `connections` (i.e. `edges` and `node` syntax)
  * Support the `pageInfo` object under any connection object
  * Change pagination keywords. We were using the originally proposed `first` and `offset` keywords, but have since moved to `first` and `after`.
    * **NOTE:** Only _forward_ pagination is currently supported.

Many of the code changes here are to tests. Any integration style test (i.e. send a request, expect a response) I have separated into their own files and load those for the test to improve request/response readability.

There is additional cleanup I would like to perform, but I feel bad making this PR much longer. That said, the core code changes are minimal. There are a few critical sections modified in the `ModelBuilder` and `EntityDictionary`, but most of the changes in the fetcher-- for instance-- are superficial.


elide-core/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -230,9 +212,11 @@ public String parseLiteral(Object o) {
                 .dataFetcher(dataFetcher)
                 .type(customIdType));
 
+        // Additional special fields
         builder.field(newFieldDefinition()
-                    .name("__meta")
-                    .type(metaObject));
+                .name(getTotalRecordKey(entityName))
```
aklish: We should have one of these for each relationship.

aklish: Actually - we may need to support something more like Relay's style of pagination

DennisMcWherter:
Looks like `GraphQLIT` is failing, but I don't see how it would have ever been working? I should have a chance to fix this soon.

elide-core/src/main/java/com/yahoo/elide/graphql/containers/ConnectionContainer.java
```diff
@@ -0,0 +1,24 @@
...
+import java.util.Set;
+
+/**
+ * Container representing a GraphQL "connection" object.
+ */
+@AllArgsConstructor
+public class ConnectionContainer implements NestedContainer {
+    @Getter private final Set<PersistentResource> persistentResources;
+    @Getter private final Optional<Pagination> pagination;
+    @Getter private final String typeName;
```
aklish: Add a comment: "refers to the type of the persistentResources"

elide-core/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -168,27 +169,57 @@ public GraphQLSchema build() {
...
-        dictionary.walkEntityGraph(rootClasses, this::buildQueryObject);
+        dictionary.walkEntityGraph(rootClasses, this::buildConnectionObject);
 
         /* Construct the schema */
         GraphQLSchema schema = GraphQLSchema.newSchema()
                 .query(root)
                 .mutation(root)
                 .build(new HashSet<>(CollectionUtils.union(
-                        queryObjectRegistry.values(),
+                        connectionObjectRegistry.values(),
```
aklish: I'm not sure this is right.  The registry is meant to be a container for types.  It should include the query Objects

elide-core/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -168,27 +169,57 @@ public GraphQLSchema build() {
...
+     */
+    private GraphQLObjectType buildConnectionObject(Class<?> entityClass) {
+        if (connectionObjectRegistry.containsKey(entityClass)) {
+            return connectionObjectRegistry.get(entityClass);
+        }
+
+        String relationName = dictionary.getJsonAliasFor(entityClass);
+
+        GraphQLObjectType connectionObject = newObject()
+                .name(relationName)
```
aklish: I think this needs to be unique.

elide-core/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -168,27 +169,57 @@ public GraphQLSchema build() {
...
+     *
+     * @param entityClass The class to use to construct the output object
+     * @return The GraphQL object.
+     */
+    private GraphQLObjectType buildConnectionObject(Class<?> entityClass) {
+        if (connectionObjectRegistry.containsKey(entityClass)) {
+            return connectionObjectRegistry.get(entityClass);
+        }
+
+        String relationName = dictionary.getJsonAliasFor(entityClass);
```
aklish: Rename this.

elide-core/src/main/java/com/yahoo/elide/graphql/GraphQLRequestScope.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Persistent state across GraphQL requests. This object is stored in the environment context.
+ */
+public class GraphQLRequestScope extends RequestScope {
+    @Getter private final Map<String, Long> totalRecordCounts = new HashMap<>();
+
+    public GraphQLRequestScope(DataStoreTransaction transaction,
+                               User user,
+                               MultivaluedHashMap<String, String> queryParams,
```
aklish: Is this needed?  THe endpoint passes in null for this.

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -36,19 +38,26 @@
...
 
         outputType = environment.getFieldType();
 
-        parentResource = environment.getSource() instanceof RequestScope ? null
-                : (PersistentResource) environment.getSource();
+        rawSource = environment.getSource();
+
+        if (rawSource instanceof PersistentResourceContainer) {
+            parentResource = ((PersistentResourceContainer) rawSource).getPersistentResource();
+        } else if (rawSource instanceof PersistentResource) {
```
aklish: Will this ever be true anymore?

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -59,6 +68,6 @@ public Environment(DataFetchingEnvironment environment) {
     }
 
     public boolean isRoot() {
-        return parentResource == null;
+        return parentResource == null && !(rawSource instanceof GraphQLContainer);
```
aklish: Do we really need both of these checks?

elide-core/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
```diff
@@ -73,23 +77,49 @@ public GraphQLEndpoint(
...
+
+            String query = jsonDocument.get(QUERY).asText();
+
+            String operationName = null;
+            if (jsonDocument.has(OPERATION_NAME)) {
+                operationName = jsonDocument.get(OPERATION_NAME).asText();
+            }
+
+            ExecutionResult result;
+            if (jsonDocument.has(VARIABLES)) {
```
aklish: Do we have tests for these new options?

elide-core/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
DennisMcWherter: Connection objects retain the top-level name for types. i.e. `book` or `author`. This registry is required for GraphQLTypeReferences. I don't think we would ever reference a `__node__author`.

elide-core/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
DennisMcWherter: It is. Renaming.

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
DennisMcWherter: I thought there were certain situations, but maybe not. I'll go through and double check.

elide-core/src/main/java/com/yahoo/elide/graphql/GraphQLRequestScope.java
DennisMcWherter: I left it open for the future-- presumably we'll want to provide access to headers in the future.

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
DennisMcWherter: We can probably safely stick with the `instanceof`.

elide-core/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
DennisMcWherter: These aren't new. They existed before, but were not always required. The GraphQL endpoint update is being done in a different PR and can add tests there.

elide-core/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
aklish: 👍

elide-core/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
aklish: 👍

elide-core/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -50,11 +52,10 @@ public Environment(DataFetchingEnvironment environment) {
         outputType = environment.getFieldType();
 
         rawSource = environment.getSource();
+        container = (rawSource instanceof GraphQLContainer) ? (GraphQLContainer) rawSource : new RootContainer();
```
aklish: This could be folded into the checks below.

PR #551, opened 2017-09-29
Allowing CreatePermissions to be specified on attributes and relations

suriyasundar92:
This PR is an update to a previous PR https://github.com/yahoo/elide/pull/544
It was closed as I tried to rename the branch

The changes include the following:
1. Changed PersistenceResource and ActivePermissionExecutor to support the change
2. Added tests in for the same, with some refactoring as we couldn't add more tests to PersistenceResourceTests

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -1791,6 +1791,11 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
                 ? new ChangeSpec(this, fieldName, original, modified)
                 : null;
 
+        if (UpdatePermission.class.isAssignableFrom(annotationClass)
+                && this.requestScope.isNewResource(this)) {
+            return requestScope.getPermissionExecutor()
+                    .checkSpecificFieldPermissions(this, changeSpec, CreatePermission.class, fieldName);
+        }
         return requestScope.getPermissionExecutor()
                 .checkSpecificFieldPermissions(this, changeSpec, annotationClass, fieldName);
     }
@@ -1803,6 +1808,11 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
                 ? new ChangeSpec(this, fieldName, original, modified)
                 : null;
 
+        if (UpdatePermission.class.isAssignableFrom(annotationClass)
+                && this.requestScope.isNewResource(this.getObject())) {
+            return requestScope.getPermissionExecutor()
+                    .checkSpecificFieldPermissionsDeferred(this, changeSpec, CreatePermission.class, fieldName);
+        }
         return requestScope
                 .getPermissionExecutor()
                 .checkSpecificFieldPermissionsDeferred(this, changeSpec, annotationClass, fieldName);
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -1764,7 +1764,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
         for (String field : fields) {
             try {
                 if (checkIncludeSparseField(requestScope.getSparseFields(), type, field)) {
-                    checkFieldAwarePermissions(ReadPermission.class, field, (Object) null, (Object) null);
+                    checkFieldAwareReadPermissions(field);
                     filteredSet.add(field);
                 }
             } catch (ForbiddenAccessException e) {
@@ -1783,21 +1783,9 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
         return requestScope.getPermissionExecutor().checkPermission(annotationClass, this);
     }
 
-    private <A extends Annotation> ExpressionResult checkFieldAwarePermissions(Class<A> annotationClass,
-                                                                   String fieldName,
-                                                                   Object modified,
-                                                                   Object original) {
-        ChangeSpec changeSpec = (UpdatePermission.class.isAssignableFrom(annotationClass))
-                ? new ChangeSpec(this, fieldName, original, modified)
-                : null;
-
-        if (UpdatePermission.class.isAssignableFrom(annotationClass)
-                && this.requestScope.isNewResource(this)) {
-            return requestScope.getPermissionExecutor()
-                    .checkSpecificFieldPermissions(this, changeSpec, CreatePermission.class, fieldName);
-        }
+    private <A extends Annotation> ExpressionResult checkFieldAwareReadPermissions(String fieldName) {
         return requestScope.getPermissionExecutor()
-                .checkSpecificFieldPermissions(this, changeSpec, annotationClass, fieldName);
+                .checkSpecificFieldPermissions(this, null, ReadPermission.class, fieldName);
     }
 
     private <A extends Annotation> ExpressionResult checkFieldAwareDeferPermissions(Class<A> annotationClass,
@@ -1808,11 +1796,6 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
                 ? new ChangeSpec(this, fieldName, original, modified)
                 : null;
 
-        if (UpdatePermission.class.isAssignableFrom(annotationClass)
-                && this.requestScope.isNewResource(this.getObject())) {
-            return requestScope.getPermissionExecutor()
-                    .checkSpecificFieldPermissionsDeferred(this, changeSpec, CreatePermission.class, fieldName);
-        }
         return requestScope
                 .getPermissionExecutor()
                 .checkSpecificFieldPermissionsDeferred(this, changeSpec, annotationClass, fieldName);
--- a/elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
+++ b/elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
@@ -5,10 +5,11 @@
  */
 package com.yahoo.elide.security.executors;
 
-import com.yahoo.elide.annotation.DeletePermission;
+import com.yahoo.elide.annotation.CreatePermission;
 import com.yahoo.elide.annotation.ReadPermission;
-import com.yahoo.elide.annotation.SharePermission;
 import com.yahoo.elide.annotation.UpdatePermission;
+import com.yahoo.elide.annotation.SharePermission;
+import com.yahoo.elide.annotation.DeletePermission;
 import com.yahoo.elide.core.RequestScope;
 import com.yahoo.elide.core.exceptions.ForbiddenAccessException;
 import com.yahoo.elide.core.filter.expression.FilterExpression;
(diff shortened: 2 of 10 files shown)


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1791,6 +1791,11 @@ protected void addInverseRelation(String relationName, Object inverseObj) {
                 ? new ChangeSpec(this, fieldName, original, modified)
                 : null;
 
+        if (UpdatePermission.class.isAssignableFrom(annotationClass)
+                && this.requestScope.isNewResource(this)) {
+            return requestScope.getPermissionExecutor()
+                    .checkSpecificFieldPermissions(this, changeSpec, CreatePermission.class, fieldName);
```
aklish: I think create checks must always be deferred.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -76,11 +77,45 @@ public ActivePermissionExecutor(boolean verbose, final RequestScope requestScope
...
 
+    //The method returns the annotation that has to be used to evaluate the specific action for the resource
+    //
+    private Class<? extends Annotation> getSecurityAnnotationType(EntityAction action) {
+        switch (action) {
+            case CREATE:
+                return CreatePermission.class;
+            case READ:
+                return ReadPermission.class;
+            case SHARE:
```
tarrantzhang: We already have a PR merged for falling back share permission. Why are we implementing this again here?

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -106,33 +170,37 @@ public ActivePermissionExecutor(boolean verbose, final RequestScope requestScope
...
-            return expressionBuilder.buildAnyFieldExpressions(resource, annotationClass, changeSpec);
-        };
+
+        if (action == EntityAction.SHARE && !requestScope.getDictionary().isShareable(resource.getResourceClass())) {
+            ForbiddenAccessException e = new ForbiddenAccessException(
+                    SharePermission.class.getSimpleName(),
+                    expressionBuilder.FAIL_EXPRESSION,
+                    Expression.EvaluationMode.ALL_CHECKS);
+            log.trace("{}", e.getLoggedMessage());
+            throw e;
```
tarrantzhang: Could we simply return `expressionBuilder.FAIL_EXPRESSION` here instead of throwing error. Logically they might make no difference but this could change logging behavior.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
```diff
@@ -89,14 +124,43 @@ public ActivePermissionExecutor(boolean verbose, final RequestScope requestScope
...
+        }
+
+        Class annotation = getSecurityAnnotationType(action);
+
+        Supplier<Expression> expressionSupplier =
+                () -> expressionBuilder.buildAnyFieldExpressions(resource, annotation, changeSpec);
+
+        Function<Expression, ExpressionResult> expressionExecutor = (expression) -> {
```
tarrantzhang: Lambda block can be made inline.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1737,10 +1700,7 @@ protected void addInverseRelation(String relationName, Object inverseObj) {
                 // retrieve the object id to return to the caller. If no fields on the object are readable by the caller
                 // then they will be filtered out and only the id is returned. Similarly, all future requests to this
                 // object will behave as expected.
-                boolean isMutation = resource.getRequestScope().isMutatingMultipleEntities();
-                if (!isMutation || !resource.getRequestScope().getNewResources().contains(resource)) {
-                    resource.checkFieldAwarePermissions(permission);
-                }
+                resource.checkReadPermissionsAllowNewEntities();
```
tarrantzhang: `filter()` use to call `checkFieldAwarePermissions()` which check permission on field level. Is it enough that `checkReadPermissionsAllowNewEntities()` only check permission entity on entity level?

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
suriyasundar92: We were using the annotation class objects to represent the action that we intend to perform on the object. We have seen some cases recently where the annotation expression that we try to evaluate is different than the annotation that we specify as the argument. e.g. We evaluate @ReadPermission whenever we pass @SharePermission and We evaluate @CreatePermission when we call @UpdatePermission on newly created entities. 
So I felt like having a different set of enums for representing the actions would signify the difference between actions on entities and the security expressions that we evaluate.
When I refactored the the code around this idea, I had the method to map UPDATE action to @CreatePermission for newly created entities. So the action of mapping SHARE action to @ReadPermission seemed similar, so I moved the logic to the same method.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
suriyasundar92: I had it done that way initially, but when we happen to have some user check on the @ReadPermission we seem to be fetching the cached expression and never seem to evaluate the FAIL_EXPRESSION that we pass. This is happening inside the checkPermissions().
https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java#L289

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
suriyasundar92: The method checkFieldAwarePermissions(permission) in turn calls only the checkPermission() method which adds only entity level checks.

elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java
tarrantzhang: I guess this is a side effect of mapping Annotation to EntityAction. [L143](https://github.com/yahoo/elide/blob/0f9a32893ad2394c37568a787e0da2de12df94f4/elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java#L143) get rid of annotation in the calling chain but we are still using `annotation` to check the cache in [L361](https://github.com/yahoo/elide/blob/0f9a32893ad2394c37568a787e0da2de12df94f4/elide-core/src/main/java/com/yahoo/elide/security/executors/ActivePermissionExecutor.java#L361). We should either pass the `annotation` along with `EntityAction` to preserve information loss OR change the `checkCache` to map by `EntityAction`. So that `@SharePermission` could be match to cache itself instead of `@ReadPermission`.

suriyasundar92: Yes I think we should make changes to the caching methodology if we are going to change the model of using the PermissionExecutor. I have reverted the changes as they require better analysis with respect to caching and deferred checks. As you said this should go in a separate PR.

elide-core/src/test/java/com/yahoo/elide/core/UpdateOnCreateTest.java
```diff
@@ -0,0 +1,242 @@
...
+        created.getRequestScope().getPermissionExecutor().executeCommitChecks();
+        Assert.assertThrows(ForbiddenAccessException.class, () -> {
+            created.updateAttribute("type", UpdateAndCreate.AuthorType.CONTRACTED);
+            created.getRequestScope().getPermissionExecutor().executeCommitChecks();
+        });
+        created.updateAttribute("alias", "");
+        created.getRequestScope().getPermissionExecutor().executeCommitChecks();
+        created.addRelation("books", loadedBook);
+        created.getRequestScope().getPermissionExecutor().executeCommitChecks();
+        Assert.assertThrows(ForbiddenAccessException.class, () -> {
```
tarrantzhang: Having multiple assertions in a single test is probably a indication that you should **separate it into multiple tests**, so that one test only fail for one reason.

elide-core/src/test/java/com/yahoo/elide/core/UpdateOnCreateTest.java
```diff
@@ -0,0 +1,242 @@
...
+        PersistentResource<Book> loadedBook = PersistentResource.loadRecord(Book.class,
+                "1",
+                userOneScope);
+        PersistentResource<Author> loadedAuthor = PersistentResource.loadRecord(Author.class,
+                "1",
+                userOneScope);
+
+        Assert.assertThrows(ForbiddenAccessException.class, () -> {
+            loaded.updateAttribute("name", "");
+            loaded.getRequestScope().getPermissionExecutor().executeCommitChecks();
```
tarrantzhang: Both `updateAttribute()` and `executeCommitChecks()` throws `ForbiddenAccessException`, so I guess you should make two assertions to check which one is actually throwing. Same to all following.

elide-core/src/test/java/example/UpdateAndCreate.java
```diff
@@ -0,0 +1,106 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Table;
+import javax.persistence.Id;
+import javax.persistence.OneToOne;
+import javax.persistence.OneToMany;
+import java.util.ArrayList;
+import java.util.Collection;
+
+/**
+ * Model for authors.
```
tarrantzhang: Model for `UpdateAndCreate`?

elide-core/src/test/java/example/UpdateAndCreate.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+@Entity
+@Table(name = "author")
```
tarrantzhang: `@Table(name = "author")` this is not needed.

elide-integration-tests/src/test/java/example/CreateButNoUpdate.java
```diff
@@ -27,6 +27,7 @@
     public Long id;
     public String textValue;
 
+    @CreatePermission(expression = "deny all")
```
aklish: This permission doesn't seem to jive with the class name.

elide-core/src/test/java/example/UpdateAndCreate.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public String getName() {
+        return name;
+    }
+
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    @CreatePermission(expression = "isUserOne OR isUserTwo OR isUserThree")
```
aklish: Why not use spaces "isUserOne" - "Principal is user one"

elide-core/src/test/java/example/UpdateAndCreate.java
aklish: Comment is wrong.

elide-core/src/test/java/com/yahoo/elide/core/UpdateOnCreateTest.java
```diff
@@ -0,0 +1,232 @@
...
+                any(RequestScope.class)
+        )).thenReturn(book);
+        when(tx.loadObject(eq(Author.class),
+                eq((Serializable) CoerceUtil.coerce(1, Long.class)),
+                eq(Optional.empty()),
+                any(RequestScope.class)
+        )).thenReturn(author);
+    }
+
+    //----------------------------------------- ** Entity Creation ** -------------------------------------------------
```
aklish: Can you add a few more comments in each section explaining the purpose of the group of tests.

elide-integration-tests/src/test/java/example/CreateButNoUpdate.java
suriyasundar92: We wanted to deny the update permission to the field be it on create or not. Thats why we wanted a @CreatePermission("deny all") expression for forbidding updates on the field specifically.

PR #553, opened 2017-10-02
Test computed attributes in GraphQL

DennisMcWherter:
Additional tests for GraphQL around computed attributes.

Requires: https://github.com/yahoo/elide/pull/549

### DIFF
--- a/elide-core/src/test/java/com/yahoo/elide/books/Author.java
+++ b/elide-core/src/test/java/com/yahoo/elide/books/Author.java
@@ -6,16 +6,11 @@
 package com.yahoo.elide.books;
 
 import com.yahoo.elide.annotation.Audit;
+import com.yahoo.elide.annotation.ComputedAttribute;
 import com.yahoo.elide.annotation.Include;
 import com.yahoo.elide.annotation.SharePermission;
 
-import javax.persistence.Entity;
-import javax.persistence.GeneratedValue;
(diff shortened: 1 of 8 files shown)


elide-core/src/test/java/com/yahoo/elide/books/Author.java
```diff
@@ -93,6 +89,17 @@ public void setPenName(Pseudonym penName) {
         this.penName = penName;
     }
 
+    // TODO: Eventually we should support multiple argument computed attributes to be used as API _functions_
+    @ComputedAttribute
+    @Transient
+    public String getSpecialOrEcho() {
```
aklish: Minor comment but I think it makes the tests more readable.  We have a lot of classes like Left, Right and FunWithComputedAttributes - basically meaningless things that we test.  

I think our model for graphQL tests should make sense.  Maybe this should be a computed attribute like fullName or something along those lines.

DennisMcWherter: Good idea

DennisMcWherter:
**NOTE:** There is a bit of additional work here. Since `ComputedAttributes` can only be modified right now by `UPSERT`ing, that data doesn't seem to be coming back properly. I feel like we can punt on this issue though until we tackle the "computed attributes as functions" problem.

Issue #99, opened 2016-01-13
Return less revealing status codes


clayreimann:
I would expect the following behavior:
- [ ] when querying a root collection I have access to I should always receive `403 Forbidden` (and never `404 Not Found`) so that I cannot discover the collection size.
- [ ] when querying a nested collection I should always receive `404 Not Found` (and never `403 Forbidden`) when I do not have access so that I cannot discover:
  1. what collections exist.
  2. the potential contents of collections I know about.

I am currently receiving both error codes for each scenario.

clayreimann:
Talked with @aklish and we decided that it makes sense to return `404 Not Found` in all cases where you reference an object by ID and `403 Forbidden` when you try to access a collection.

Examples:

Request | Permitted | Exists | Response
----------- | -------- | ---------- | ---------------
`/book` | Yes | Yes | `200 OK`
`/book/1` | Yes | Yes | `200 OK`
`/book/2` | No | Yes | `404 Not Found`
`/book/666` | No | No | `404 Not Found`
`/book/1/publishers` | No | Yes | `403 Forbidden`
`/book/1/conspiracies` | No | No | `403 Forbidden` (aha! They *are* out to get me)
`/book/1/authors` | Yes | Yes | `200 OK`
`/book/1/authors/1` | Yes | Yes | `200 OK`
`/book/1/authors/2` | No | Yes | `404 Not Found`
`/book/1/authors/99` | No | No | `404 Not Found`

PR #555, opened 2017-10-03
Less revealing status code

tarrantzhang:
Addressing issue #99

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -24,7 +24,7 @@ import com.yahoo.elide.core.exceptions.InvalidAttributeException;
 import com.yahoo.elide.core.exceptions.InvalidEntityBodyException;
 import com.yahoo.elide.core.exceptions.InvalidObjectIdentifierException;
 import com.yahoo.elide.core.exceptions.InvalidPredicateException;
-import com.yahoo.elide.core.exceptions.EntityForbiddenException;
+import com.yahoo.elide.core.exceptions.ForbiddenEntityException;
 import com.yahoo.elide.core.filter.FilterPredicate;
 import com.yahoo.elide.core.filter.Operator;
 import com.yahoo.elide.core.filter.expression.AndFilterExpression;
@@ -275,7 +275,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
             obj = tx.loadObject(loadClass, (Serializable) CoerceUtil.coerce(id, idType),
                     permissionFilter, requestScope);
             if (obj == null) {
-                throw new EntityForbiddenException();
+                throw new ForbiddenEntityException();
             }
         }
 
@@ -286,7 +286,7 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
                 resource.checkFieldAwarePermissions(ReadPermission.class);
             }
         } catch (ForbiddenAccessException e) {
-            throw new EntityForbiddenException();
+            throw new ForbiddenEntityException();
         }
 
         return resource;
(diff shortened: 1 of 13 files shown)


elide-integration-tests/src/test/resources/ResourceIT/failRootCollectionId.json
```diff
@@ -1 +1 @@
-{"errors":["InvalidObjectIdentifierException: Unknown identifier '6789' for parent"]}
+{"errors":["NonRevealingForbiddenAccessException: Entity not found"]}
```
clayreimann: This response body is incorrect.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -855,7 +860,7 @@ public PersistentResource getRelation(String relation, String id) {
         Set<String> missedIds = Sets.difference(new HashSet<>(ids), allExpectedIds);
 
         if (!missedIds.isEmpty()) {
-            throw new InvalidObjectIdentifierException(missedIds.toString(), relation);
+            throw new EntityForbiddenException();
```
aklish: If we are not going to return the missing IDs, we can remove the calculation of allExpectedIds and missedIds above.

tarrantzhang: The `missedIds` is still used to check if queried id matches ant existent entity id.
```
if (!missedIds.isEmpty()) {
...

```

elide-core/src/main/java/com/yahoo/elide/parsers/state/StartState.java
```diff
@@ -68,7 +70,7 @@ public void handle(StateContext state, RootCollectionRelationshipContext ctx) {
         String id = ctx.entity().id().getText();
         Class<?> entityClass = dictionary.getEntityClass(entityName);
         if (entityClass == null || !dictionary.isRoot(entityClass)) {
-            throw new InvalidCollectionException(entityName);
+            throw new ForbiddenEntityException();
```
suriyasundar92: Should we be throwing ForbiddenAccessException here since we are in the process of fetching a relationship?

suriyasundar92: Should we also rethrow the InvalidCollectionException below with a ForbiddenAccessException, for the same reason.

tarrantzhang: Yes indeed, good catch.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
aklish: You can remove if you change the check to:
```
if (ids.size() != allResources.size()) {
```

tarrantzhang:
```
if (ids.size() != allResources.size()) {
```
I think this only checks their cardinality. Even the sizes are the same, their id could be mismatched. Say a query `/book/1000` to a entity `author` who has one book with id 1 will not trigger the above statement.

elide-core/src/main/java/com/yahoo/elide/parsers/state/StartState.java
aklish: This still needs an update.

elide-core/src/main/java/com/yahoo/elide/parsers/state/RecordState.java
```diff
@@ -40,7 +41,7 @@ public void handle(StateContext state, SubCollectionReadCollectionContext ctx) {
         try {
             RelationshipType type = dictionary.getRelationshipType(resource.getObject(), subCollection);
             if (type == RelationshipType.NONE) {
-                throw new InvalidCollectionException(subCollection);
+                throw new ForbiddenAccessException(subCollection);
```
aklish: There is a missed case below in the other handle function

aklish: Also, the other place where we throw InvalidCollectionException needs to be mapped as well.

tarrantzhang: Addressed.

elide-core/src/main/java/com/yahoo/elide/parsers/state/StartState.java
tarrantzhang: Addressed.

PR #559, opened 2017-10-12
Fix NPE in path expansion

wcekan:
If filter path had a null value, the next get could NPE.
For example, `chapter.book.publisher` when chapter.book is null.


wcekan:
Will address better in 4.x

PR #560, opened 2017-10-15
Relationship update returns correct response code

QubitPi:
Resolves issue https://github.com/yahoo/elide/issues/397

I assume no unexpected/unrequested modification is possible on to-many relationships. It only happens when we are `add`ing to a to-one relationship

I'll add tests if implementation looks good to reviewers. :)


QubitPi:
@DennisMcWherter Why is it closed?

DennisMcWherter:
This was probably auto-closed. We merged Elide 4 to master, so the elide 4 branch no longer exists (which is why we can't reopen PR).

Can you rebase off master and reopen?

PR #564, opened 2017-11-13
Handle Forbidden for Include

wcekan:
Uncaught exception caused Include to fail when not allowed.

By returning empty set, all the other allowed fields are returned instead of having to guess which are visible.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,7 @@
 # Change Log
 ## 3.1.4
 **Fixes**
- * properly handle ForbiddenAccess for Include
+ * Instead of ForbiddenAccess for denied Include, filter to empty set.
 
 ## 3.1.3
  * Add support for @Any relationships through a @MappedInterface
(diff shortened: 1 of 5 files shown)


changelog.md
```diff
@@ -1,4 +1,8 @@
 # Change Log
+## 3.1.4
+**Fixes**
+ * properly handle ForbiddenAccess for Include
```
DennisMcWherter: This description is too vague I think since "proper" may be-- more or less-- a matter of opinion. Maybe something like:

```
Changed handling of including forbidden collections. Instead of throwing a ForbiddenAccess exception, we filter the included collection.
```

wcekan: Of course only my opinion matters 😄   But will change the text.

changelog.md
```diff
@@ -1,4 +1,8 @@
 # Change Log
+## 3.1.4
+**Fixes**
+ * Instead of ForbiddenAccess for denied Include, filter to empty set.
```
clayreimann: Missing note about bailing when unable to parse permission strings

wcekan: Added.

PR #566, opened 2017-11-16
Handle ValidationException and throw HttpStatusException

niketh:
@DennisMcWherter @wcekan Can you review

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -44,6 +44,7 @@ import org.apache.commons.lang3.tuple.Pair;
 
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MultivaluedMap;
+import javax.xml.bind.ValidationException;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -244,6 +245,14 @@ protected ElideResponse handleRequest(boolean isReadOnly, Object opaqueUser,
             return buildErrorResponse(new InvalidURLException(e), isVerbose);
 
         } catch (Exception e) {
+            if (e instanceof ValidationException) {
+                return buildErrorResponse(
+                    new HttpStatusException(HttpStatus.SC_BAD_REQUEST, e.getMessage()) {
```
DennisMcWherter: I would check first if there exists `statusCode` on the `ValidationException` and return that as an HTTP response?

DennisMcWherter:
Any update on this?

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -244,6 +245,14 @@ protected ElideResponse handleRequest(boolean isReadOnly, Object opaqueUser,
             return buildErrorResponse(new InvalidURLException(e), isVerbose);
 
         } catch (Exception e) {
+            if (e instanceof ValidationException) {
```
wcekan: Why is this not
```java
catch (ValidationException e)
```

DennisMcWherter:
I'm going to close for now due to inactivity. Feel free to reopen!

PR #567, opened 2017-11-18
Elide 4 fixups

clayreimann:


elide-annotations/pom.xml
```diff
@@ -9,7 +9,7 @@
     <parent>
         <groupId>com.yahoo.elide</groupId>
         <artifactId>elide-parent-pom</artifactId>
-        <version>3.1.3-SNAPSHOT</version>
+        <version>4.0-alpha</version>
```
DennisMcWherter: Don't bump the version yet. First, let's rebase or merge master -> elide-4.x. This will save us some merge conflict pain.

DennisMcWherter: We'll probably also want this to be `4.0-alpha-1` to remain compliant with maven versioning standards.

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/filter/HQLFilterOperation.java
```diff
@@ -178,41 +163,60 @@ public String apply(FilterExpression filterExpression, boolean prefixWithAlias)
...
+            return apply(filterPredicate, prefixWithAlias);
         }
 
         @Override
         public String visitAndExpression(AndFilterExpression expression) {
-            String left = expression.getLeft().accept(this);
-            String right = expression.getRight().accept(this);
-            query = "(" + left + " AND " + right + ")";
-            return query;
+            FilterExpression left = expression.getLeft();
```
aklish: Let's make sure that the DataStore transactions only receive FilterExpressions that never have a FALSE_USER_CHECK_EXPRESSION or a NO_EVALUATION_EXPRESSION.  There will be better separation of concerns.

PR #573, opened 2017-11-27
GraphQL: Add remove test case

aklish:


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -308,34 +308,69 @@ public boolean matchesId(String checkId) {
...
+
+            // Combine filters if necessary
+            filterExpression = filter
+                    .map(fe -> (FilterExpression) new AndFilterExpression(idExpression, fe))
+                    .orElse(idExpression);
+        } else {
+            filterExpression = filter.orElse(null);
+        }
+
+        Set<PersistentResource> existingResources = filter(ReadPermission.class,
```
DennisMcWherter: General comment: I know we're going to push out the refactor to clean all of this up, but we probably want to get to this as soon as we can after getting elide 4.0 into good shape.

PR #576, opened 2017-11-29
Optimize copy constructors of immuatables

clayreimann:


clayreimann:
Premature optimization.

PR #577, opened 2017-11-29
Simplied response for delete & remove so they return the current stat…

aklish:
…e of the relationship

### DIFF
--- a/elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -30,6 +30,7 @@ import javax.ws.rs.core.MultivaluedMap;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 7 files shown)


elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -433,7 +433,12 @@ private Object deleteObjects(Environment context) {
         ConnectionContainer connection = (ConnectionContainer) fetchObjects(context);
         Set<PersistentResource> toDelete = connection.getPersistentResources();
         toDelete.forEach(PersistentResource::deleteResource);
-        return new ConnectionContainer(toDelete, Optional.empty(), connection.getTypeName());
+
+        return new ConnectionContainer(
+                Sets.difference(connection.getPersistentResources(), toDelete),
```
DennisMcWherter: This doesn't look correct. I think your tests work because you've deleted everything. However,

```
toDelete = connection.getPersistentResources()
```

`Sets.difference()` will always be empty. If we want to return the _current_ state, we'll need to go fetch all of the objects without the id fields first.

elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -459,7 +464,12 @@ private Object removeObjects(Environment context) {
         } else { /* is root */
             toRemove.forEach(PersistentResource::deleteResource);
         }
-        return new ConnectionContainer(toRemove, Optional.empty(), connection.getTypeName());
+
+        return new ConnectionContainer(
+                Sets.difference(connection.getPersistentResources(), toRemove),
```
DennisMcWherter: Same comment as above.

DennisMcWherter:
We decided to always return empty set.

The rationale is that we believe these requests should return the **state of the world** after the operation.

In this case, since `DELETE` and `REMOVE` are id-scoped, the state of the world after these operations (scoped to these id's) is always empty.

PR #580, opened 2017-12-01
Add GraphQL initial endpoint test.

DennisMcWherter:
GraphQL endpoint tests.

### DIFF
--- a/elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
+++ b/elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
@@ -32,6 +32,7 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.core.SecurityContext;
 
 import java.io.IOException;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.function.Function;
 
@@ -83,6 +84,7 @@ public class GraphQLEndpoint {
             JsonNode jsonDocument = mapper.readTree(graphQLDocument);
             final User user = tx.accessUser(getUser.apply(securityContext));
             GraphQLRequestScope requestScope = new GraphQLRequestScope(tx, user, elide.getElideSettings());
+            isVerbose = requestScope.getPermissionExecutor().isVerbose();
 
             if (!jsonDocument.has(QUERY)) {
                 return Response.status(400).entity("A `query` key is required.").build();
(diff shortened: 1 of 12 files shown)


elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
```diff
@@ -103,23 +105,42 @@ public Response post(
                 result = api.execute(query, operationName, requestScope);
             }
 
-            if (query.startsWith(MUTATION)) {
+            tx.preCommit();
+            requestScope.runQueuedPreSecurityTriggers();
+            requestScope.getPermissionExecutor().executeCommitChecks();
```
aklish: This throws an exception and doesn't return a GraphQL error object.

elide-graphql/src/test/java/graphqlEndpointTestModels/security/CommitChecks.java
```diff
@@ -0,0 +1,25 @@
...
+import com.yahoo.elide.security.RequestScope;
+import com.yahoo.elide.security.checks.CommitCheck;
+
+import java.security.Principal;
+import java.util.Optional;
+
+public abstract class CommitChecks {
+    public static final String IS_NOT_USER_3 = "isnt user three";
+
+    public static class IsUser3 extends CommitCheck {
```
aklish: SHould this class be IsNotUser3?

elide-graphql/src/test/java/graphqlEndpointTestModels/Book.java
```diff
@@ -31,8 +33,9 @@
 @Include(rootLevel = true)
 @Entity
 @CreatePermission(expression = UserChecks.IS_USER_1)
-@ReadPermission(expression = UserChecks.IS_USER_1 + " OR " + UserChecks.IS_USER_2)
-@UpdatePermission(expression = UserChecks.IS_USER_1)
+@ReadPermission(expression = UserChecks.IS_USER_1 + " OR " + UserChecks.IS_USER_2 + " OR NOT "
+        + CommitChecks.IS_NOT_USER_3)
+@UpdatePermission(expression = CommitChecks.IS_NOT_USER_3 + " AND NOT " + UserChecks.IS_USER_2)
```
aklish: It seems like this would short circuit and never hit the commit check.

PR #583, opened 2017-12-05
Fix id assignments

aklish:


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -131,6 +132,8 @@ public String toString() {
 
         PersistentResource<T> newResource = new PersistentResource<>(obj, parent, id, requestScope);
 
+        assignId(newResource, id);
```
DennisMcWherter: Just add comment around this.

PR #585, opened 2017-12-05
Pushed down security filters for root level loads.

aklish:


aklish:
Added a test.  We already have tests, but I wanted to make it a bit more explicit.

PR #586, opened 2017-12-06
Add additional logging around caught exceptions.

DennisMcWherter:
This adds additional logging around our exception handling. `debug` for "expected" exceptions and `error` for response codes which are more likely to be server-related issues.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,8 @@
 # Change Log
+## 3.2.1
+**Fixes**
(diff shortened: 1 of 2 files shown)


clayreimann:
Well, I've been skeptical of this in the past but now I truly believe that Codacy is broken. How does adding 6 lines of code reduce the complexity of a file by 26 whatevers?

PR #588, opened 2017-12-06
Log exceptions elide4

DennisMcWherter:
Elide 4 version of: https://github.com/yahoo/elide/pull/586

### DIFF
--- a/elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
+++ b/elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
@@ -135,8 +135,10 @@ public class GraphQLEndpoint {
 
             return Response.ok(mapper.writeValueAsString(result)).build();
         } catch (JsonProcessingException e) {
+            log.debug("Invalid json body provided to GraphQL", e);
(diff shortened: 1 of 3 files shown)


elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
```diff
@@ -135,8 +135,10 @@ public Response post(
 
             return Response.ok(mapper.writeValueAsString(result)).build();
         } catch (JsonProcessingException e) {
+            log.debug("Invalid json body provided to GraphQL", e);
             return buildErrorResponse(new InvalidEntityBodyException(graphQLDocument), isVerbose);
         } catch (IOException e) {
+            log.error("Uncaught IO Exception by Elide in GraphQL", e);
```
aklish: Shouldn't we log the ForbiddenAccessExceptions here (or are these already logged internally)?

DennisMcWherter: Good question. We do log forbidden access elsewhere, but we can throw in a debug here too just for completeness. Adding now

DennisMcWherter:
Rebased to update changelog versioning. Will merge after build passes.

PR #589, opened 2017-12-06
Avoid throwing jax-rs exception directly

clayreimann:


DennisMcWherter:
@clayreimann can you update changelog?

PR #591, opened 2017-12-06
Update elide standalone to support GraphQL.

DennisMcWherter:
This allows developers to enable GraphQL without any friction. If they don't want GraphQL included in their project, there is only one step:

_In their ElideStandaloneSettings_
```java
boolean enableGraphQL() {
    return false;
}
```

That being said, this approach still will include the Elide graphql code. If they don't want any trace of GraphQL in their project, it's easy to exclude in their dependency management by simply excluding the `com.yahoo.elide:elide-graphql` transitive dependency that is brought in by `elide-standalone`.

### DIFF
--- a/elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
+++ b/elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
@@ -108,6 +108,33 @@ public interface ElideStandaloneSettings {
     }
 
     /**
+     * API root path specification for the GraphQL endpoint. Namely, this is the root uri for GraphQL.
+     *
+     * @return Default: /graphql/api/v1
+     */
+    default String getGraphQLApiPathSepc() {
+        return "/graphql/api/v1";
+    }
+
+    /**
+     * Enable the JSONAPI endpoint. If false, the endpoint will be disabled.
+     *
+     * @return Default: True
+     */
+    default boolean enableJSONAPI() {
+        return true;
+    }
+
+    /**
+     * Enable the GraphQL endpoint. If false, the endpoint will be disabled.
+     *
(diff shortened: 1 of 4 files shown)


DennisMcWherter:
Yes, I will add one

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -107,6 +107,33 @@ default String getJsonApiPathSpec() {
         return "/api/v1/*";
     }
 
+    /**
+     * API root path specification for the GraphQL endpoint. Namely, this is the root uri for GraphQL.
+     *
+     * @return Default: /graphql/api/v1
+     */
+    default String getGraphQLApiPathSepc() {
+        return "/graphql/api/v1";
```
clayreimann: `/v1`? I thought the point of graphql was that it's not versioned?

clayreimann:
In other news I think we should consider replacing codacy with something else. Their complexity metrics appear totally FUBAR–adding two conditionals decreases complexity by 1???

DennisMcWherter:
Agreed :)

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
DennisMcWherter: Yeah, we can probably drop the versioning by default if we want. I'm happy with that.

aklish: I think all APIs should be versioned - even GraphQL.

DennisMcWherter:
Rebased to resolve conflicts. Merging.

PR #593, opened 2017-12-08
Delay OnCreatePreSecurity until fields are populated.

wcekan:
Restored lifecycle hook for create pre-security to occur after fields are populated.

Expanded test as well.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
 **Fixes**
  * Root collection loads now push down security filter predicates.
  * Avoid throwing exceptions that must be handled by the containing application, instead throw exceptions that will be handled directly within Elide.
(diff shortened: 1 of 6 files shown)


wcekan:
@DennisMcWherter @aklish Is it sufficient to restore only the `OnCreatePreSecurity` since it defers all inline checks to commit time?  Or should we restore the `OnUpdatePreSecurity` and `OnDeletePreSecurity` as well?

DennisMcWherter:
Should get consensus with @aklish, but `DELETE` shouldn't need to be restored since the user would never be providing additional data to the model before removing (it's just not how the system works).

We should restore `Update` as well though. If we don't restore update, then the hooks would not be operating on the _latest_ data (i.e. the fields the user has already updated).

PR #594, opened 2017-12-08
Add UPDATE operation to GraphQL

aklish:
UPDATE operation behaves like UPSERT but fails if the _data_ parameter is missing an ID or the ID doesn't exist (rather than creating a new entity).

The same behavior is replicated for nested objects within _data_.

Doc PR is also coming.


changelog.md
```diff
@@ -1,4 +1,8 @@
 # Change Log
+## 4.0.3-beta
```
DennisMcWherter: We're moving to 4.0.2-beta. Please update this label :)

elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -390,6 +410,28 @@ private PersistentResource upsertObject(Environment context, Entity entity) {
...
+            throw new BadRequestException("UPDATE data objects must include ids");
+        } else {
+            Set<PersistentResource> loadedResource = fetchObject(context, requestScope, entity.getEntityClass(),
+                Optional.of(Arrays.asList(id.get())),
+                Optional.empty(),
+                Optional.empty(),
+                Optional.empty(),
+                Optional.empty(),
+                false).getPersistentResources();
+            updatedResource = loadedResource.iterator().next();
```
DennisMcWherter: what happens if this list is empty?

aklish: The underlying call to fetchObject will throw an exception (InvalidObjectIdentifierException) so that will never happen.

PR #596, opened 2017-12-15
OnUpdate life cycle hooks now accept an optional ChangeSpec

aklish:


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1486,13 +1488,18 @@ protected void setValue(String fieldName, Object value) {
...
+        requestScope.queueTriggers(this, fieldName,
+                (isNewlyCreated) ? CRUDAction.CREATE : CRUDAction.UPDATE, Optional.of(changeSpec));
         requestScope.queueTriggers(this, (isNewlyCreated) ? CRUDAction.CREATE : CRUDAction.UPDATE);
         auditField(new ChangeSpec(this, fieldName, original, value));
     }
 
-    <A extends Annotation> void runTriggers(Class<A> annotationClass, String fieldName) {
+    <A extends Annotation> void runTriggers(Class<A> annotationClass,
+                                            String fieldName,
+                                            Optional<ChangeSpec> changeSpec) {
```
clayreimann: Much as it pains me to say, might it be worth while to provide an overloaded `runTriggers(Class, String)` that passes `Optional.empty()` to the full method since most places don't have a `ChangeSpec` to pass?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1364,10 +1366,10 @@ protected Object getValueChecked(String fieldName) {
      */
     protected Object getValueUnchecked(String fieldName) {
         requestScope.queueTriggers(this, CRUDAction.READ);
-        requestScope.queueTriggers(this, fieldName, CRUDAction.READ);
+        requestScope.queueTriggers(this, fieldName, CRUDAction.READ, Optional.empty());
         // Run the pre-security checks:
-        runTriggers(OnReadPreSecurity.class, "");
-        runTriggers(OnReadPreSecurity.class, fieldName);
+        runTriggers(OnReadPreSecurity.class, "", Optional.empty());
```
clayreimann: Maybe we should extract this empty string into a constant like `CLASS_NO_FIELD = ""`?

PR #604, opened 2017-12-21
Upgraded graphql-java.

DennisMcWherter:
This PR upgrades us to the latest graphql-java. A lot has been fixed and changed in the library since we first pulled in version 2.2.0. I have made sure our code uses only non-deprecated interfaces and no longer relies on bugs that existed in the old version.

A light summary of the changes:
  * Add variables example
  * Call into modern code
  * Remove reliance on old graphql-java bugs (i.e. dup type names)


elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -133,19 +152,49 @@ public  GraphQLList classToQueryMap(Class<?> keyClazz,
...
                                         ConversionLedger ledger) {
-        return new GraphQLList(
+        String mapName = toValidNameName(keyClazz.getName() + valueClazz.getCanonicalName() + MAP);
+
+        if (mapConversions.containsKey(mapName)) {
+            return mapConversions.get(mapName);
+        }
+
+        GraphQLOutputType keyType;
+        if (keyClazz.isEnum()) {
```
aklish: This if/else statement is repeated 4 times in this file.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -133,19 +152,49 @@ public  GraphQLList classToQueryMap(Class<?> keyClazz,
...
+            keyType = classToEnumType(keyClazz);
+        } else {
+            keyType = fetchScalarOrObjectOutput(keyClazz, ledger, fetcher);
+            if (keyType == null) {
```
aklish: This if statement will never be hit.  fetchScalarOrObjectOutput already does this check.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -133,19 +152,49 @@ public  GraphQLList classToQueryMap(Class<?> keyClazz,
...
+                keyType = classToQueryObject(keyClazz, ledger, fetcher);
+            }
+        }
+
+        GraphQLOutputType valueType;
+        if (valueClazz.isEnum()) {
+            valueType = classToEnumType(valueClazz);
+        } else {
+            valueType = fetchScalarOrObjectOutput(valueClazz, ledger, fetcher);
+            if (valueType == null) {
```
aklish: Remove this if statement.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -159,17 +208,47 @@ public  GraphQLList classToQueryMap(Class<?> keyClazz,
...
+        if (mapConversions.containsKey(mapName)) {
+            return mapConversions.get(mapName);
+        }
+
+        GraphQLInputType keyType;
+        if (keyClazz.isEnum()) {
+            keyType = classToEnumType(keyClazz);
+        } else {
+             keyType = fetchScalarOrObjectInput(keyClazz, ledger);
+            if (keyType == null) {
```
aklish: Remove this if statement.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -159,17 +208,47 @@ public  GraphQLList classToQueryMap(Class<?> keyClazz,
...
+                keyType = classToInputObject(keyClazz, ledger);
+            }
+        }
+
+        GraphQLInputType valueType;
+        if (valueClazz.isEnum()) {
+            valueType = classToEnumType(valueClazz);
+        } else {
+            valueType = fetchScalarOrObjectInput(valueClazz, ledger);
+            if (valueType == null) {
```
aklish: Remove this if statement

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -246,7 +335,13 @@ protected GraphQLOutputType attributeToQueryObject(Class<?> parentClass,
             }
 
             // If this is a collection of a boxed type scalar, we want to unwrap it properly
-            GraphQLOutputType outputType = fetchScalarOrObjectOutput(listType, ledger, fetcher);
+            GraphQLOutputType outputType;
+
+            if (listType.isEnum()) {
```
aklish: Instead of all these checks for enums sprinkled throughout the file, we can maybe move the check for enum inside the fetchScalarOrObjectOutput (same comment applies for input objects as well).

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -44,6 +46,11 @@
     protected NonEntityDictionary nonEntityDictionary = new NonEntityDictionary();
     protected EntityDictionary entityDictionary;
 
+    private final Map<Class, GraphQLObjectType> outputConversions = new HashMap<>();
```
aklish: Maybe we should move these maps into the conversion ledger?

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLScalars.java
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Additional scalar/serializers for built-in graphql types.
+ */
+@Slf4j
+public class GraphQLScalars {
+    private static final String ERROR_BAD_EPOCH = "Date must be provided as epoch millis";
+    private static final String ERROR_BAD_EPOCH_TYPE = "Date must be provided as string or integral in epoch millis";
+
+    public static final GraphQLScalarType GRAPHQL_DATE_TYPE = new GraphQLScalarType(
```
aklish: Are there tests for this class?

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -394,14 +395,23 @@ private GraphQLInputType buildInputObjectStub(Class<?> clazz) {
...
-                            ((GraphQLInputObjectType) attributeType).getDescription(),
-                            ((GraphQLInputObjectType) attributeType).getFields()
-                        );
-                attributeType = wrappedType;
-
+                String objectName = attributeType.getName() + ARGUMENT_INPUT;
+                if (!convertedInputs.containsKey(objectName)) {
+                    MutableGraphQLInputObjectType wrappedType =
+                            new MutableGraphQLInputObjectType(
+                                    entityName + "__" + attributeType.getName() + ARGUMENT_INPUT,
```
aklish: This seems wrong to me.  If the type is _Address_.  I would expect the name to not be "author__address" or "publisher__address".  The name should be linked to the type - not the attribute name.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -65,6 +65,8 @@
     private Map<Class<?>, GraphQLObjectType> connectionObjectRegistry;
     private Set<Class<?>> excludedEntities;
 
+    private HashMap<String, GraphQLInputType> convertedInputs = new HashMap<>();
```
aklish: Why do we need this?

elide-graphql/src/test/java/example/Book.java
```diff
@@ -51,6 +54,10 @@
     private long publishDate = 0;
     private Collection<Author> authors = new ArrayList<>();
     private Publisher publisher = null;
+    private Date publicationDate = null;
+    private Date lastPurchasedDate = null;
+    private Author.AuthorType authorType;
```
aklish: This member doesn't make sense for this model.  Books have a collection of authors but a single authorType?

DennisMcWherter: We should update this. I was working to force the `ModelBuilder` to go through a use case where it was blowing up. We need the same enum in multiple models.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
DennisMcWherter: I agree with this. I'll have to revisit. The types are mostly important for schema viewing and query variables. The case for query variables, however, is far more important imo and I believe we should optimize for those.

Before we merge this PR I should have several `GraphQLIT` tests that have query mutation variables so we can have a nice example of how it works.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
DennisMcWherter: We want to reuse the same types. If we blindly call `newObject()` the updated GraphQL fails; it doesn't allow us to have multiple types with the same name (even if those types are actually identical).

Overall, I think this is beneficial. In fixing this I have uncovered several additional bugs with how we were re-writing before.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLScalars.java
DennisMcWherter: No, this was the first pass. While I convert things internally to GraphQL I have been making my changes and pushing them to ensure I'm "doing the right thing." We'll definitely want tests for this before we merge.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
DennisMcWherter: I agree. It would be worth refactoring to consolidate this logic.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
DennisMcWherter: that may be appropriate. I'll have to look it over again; I thought we actually needed to look at some of this stuff outside of the conversion ledger. In any case, refactoring to combine the two makes sense.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -44,39 +43,15 @@
...
+    private final Map<Class, GraphQLInputObjectType> inputConversions = new HashMap<>();
+    private final Map<Class, GraphQLEnumType> enumConversions = new HashMap<>();
```
aklish: Do we need a separate map for enumConversions?  Can we just use outputConversions?

DennisMcWherter: `GraphQLEnumType`s are both `GraphQLInputObjectType` and `GraphQLOutputObjectType`. As a result, we can share this across inputs and outputs. In fact, we may have to so we avoid naming collisions.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -239,7 +244,7 @@ private GraphQLObjectType buildQueryObject(Class<?> entityClass) {
         String id = dictionary.getIdFieldName(entityClass);
 
         /* our id types are DeferredId objects (not Scalars.GraphQLID) */
-        GraphQLScalarType customIdType = new GraphQLScalarType(id, "custom id type", new Coercing() {
+        GraphQLScalarType customIdType = new GraphQLScalarType("__id__" + entityName, "custom id type", new Coercing() {
```
aklish: Can we have a single one of these?

Issue #611, opened 2018-01-08
Envers not supported since 3.x version


agea:
It appears that Envers (http://hibernate.org/orm/envers/) is not supported.

To reproduce, just add envers as a dependency in pom.xml

```xml
<dependency>
    <groupId>org.hibernate</groupId>
    <artifactId>hibernate-entitymanager</artifactId>
    <version>5.0.2.Final</version>
</dependency>
```

and annotate an entity as `@Audited` (this is the User entity from the Example):

```java
@Entity
@Table(name = "user")
@Include(rootLevel = true)
@SharePermission
@Audited
//@CreatePermission(expression = "Prefab.Role.All")
public class User {
    private long id;
    private String name;
    private Role role;
```

When you start the application you get this error:

```
Caused by: java.lang.IllegalArgumentException: Unknown Entity interface java.util.Map
        at com.yahoo.elide.core.EntityDictionary.lookupEntityClass(EntityDictionary.java:936)
        at com.yahoo.elide.core.EntityDictionary.bindEntity(EntityDictionary.java:715)
        at com.yahoo.elide.datastores.hibernate5.HibernateEntityManagerStore.populateEntityDictionary(HibernateEntityManagerStore.java:61)
```

DennisMcWherter:
Sorry for the delay, I haven't used envers but will look into this shortly. I don't see any `Map` objects in your code, so I wonder if envers is putting something in. I'll let you know what I come up with.

Thanks for the report!

DennisMcWherter:
I haven't forgotten about this; I've just been a bit busy tying up some loose ends elsewhere. Hopefully I'll have a minute to investigate this later this evening.

DennisMcWherter:
I finally had a chance to look at this. I wasn't able to reproduce your exact error (with the `java.util.Map`) but I _was_ able to get an analogous error:

```
Caused by: java.lang.IllegalArgumentException: Unknown Entity class org.hibernate.envers.DefaultRevisionEntity
	at com.yahoo.elide.core.EntityDictionary.lookupEntityClass(EntityDictionary.java:936)
	at com.yahoo.elide.core.EntityDictionary.bindEntity(EntityDictionary.java:715)
	at com.yahoo.elide.datastores.hibernate5.AbstractHibernateStore.populateEntityDictionary(AbstractHibernateStore.java:105)
```

After digging a bit deeper I think my original hunch was correct. It looks like envers is actually modifying the classes when you use the `@Audited` annotation.

[...]

PR #615, opened 2018-01-12
Ignore non-entity types in hibernate class metadata.

DennisMcWherter:
This is a proposed fix for https://github.com/yahoo/elide/issues/611. In short, with tools like envers hibernate can legitimately store non-entity types in its class metadata. As a result, we do not want to bind these objects in Elide. This allows us to first check if the class we're trying to bind is an entity. If it's not, we simply ignore the class otherwise we proceed as usual.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 4.0-beta-5
+**Fixes**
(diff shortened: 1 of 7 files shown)


DennisMcWherter:
After further investigation, it appears that hibernate3 doesn't suffer from this issues.

Issue #603, opened 2017-12-20
Support GraphQL Request Batching


DennisMcWherter:
We should investigate supporting batched requests in GraphQL. In particular, a sane first approach may be to adopt apollo's mechanism. It simply sends an array of GraphQL objects to be processed. This directly relates to https://github.com/yahoo/elide/issues/592.

If we support this mode of request batching, we could likely consider multiple "root" elements in the same query as occurring within the same transaction. Each separate object in the array would have its own, isolated transaction.

PR #617, opened 2018-01-19
Add support for GraphQL query batching

DennisMcWherter:
This PR adds support for GraphQL query batching and additional tests.

Addresses: https://github.com/yahoo/elide/issues/603


elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEndpoint.java
```diff
@@ -180,4 +235,8 @@ private Response buildErrorResponse(HttpStatusException error, boolean isVerbose
                 .entity(isVerbose
                         ? error.getVerboseErrorResponse().getRight() : error.getErrorResponse().getRight()).build();
     }
+
+    private static JsonNode extractResponse(ArrayNode arrayNode, Response response) {
+        return (JsonNode) response.getEntity();
+    }
```
wcekan: Is this used?

PR #620, opened 2018-01-31
Fix id type wrong due to no sorted methods from Class.getMethods

hugogu:
The Class.getMethods returns undetermined ordering in OpenJDK which may break the idType.


yahoocla:

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -188,6 +188,11 @@ private void bindEntityId(Class<?> cls, String type, AccessibleObject fieldOrMet
         String fieldName = getFieldName(fieldOrMethod);
         Class<?> fieldType = getFieldType(fieldOrMethod);
 
+        if (idType != null && fieldType.isAssignableFrom(idType)) {
```
clayreimann: If ordering is undefined in the OpenJDK, how can we be sure that we always want the _first_ `idType`? 

This looks like it will cause us to only bind the first field or method annotated with `@Id` we find and not necessarily the `@Id` that is closest to the class we are binding.

DennisMcWherter:
Reviewing the [JPA spec](http://download.oracle.com/otn-pub/jcp/persistence-2_1-fr-eval-spec/JavaPersistence.pdf?AuthParam=1517498828_13162372f4ef76159a4af7d687b722a3) (**warning:** Oracle PDF link), it may be worth considering the information in **Section 2.4**. Id's are actually a bit complicated and we have not implemented the full spec here (mainly due to lack of practical use rather than lack of will). Namely, behavior related to `@IdClass` is likely not spec compliant today. However, this seems mostly related to information about _derived_ entities (section 2.4.1).

@hugogu can you give an example of what exactly this is trying to solve?

If you are extending and using a `@MappedSuperclass` (this is distinct from a section 2.4.1 derived class), section 2.11 demonstrates it is not explicitly supported to override attributes through typical Java inheritance (and in fact, may be disallowed):

> When applied to the subclasses, the inherited mappings will apply in the context of the subclass tables. Mapping information can be overridden in such subclasses by using the AttributeOverride and AssociationOverride annotations or corresponding XML elements.

In short, it looks like you would want to use [@AttributeOverride](https://javaee.github.io/javaee-spec/javadocs/javax/persistence/AttributeOverride.html) to make this stick.

Now, to be clear, I am pretty sure we do not support this today. However, implementing the override in the `EntityBinding` should be a straightforward implementation.

hugogu:
@clayreimann @DennisMcWherter , 

Thank you all for the review comments. It sounds like there is a more appropriate way to achieve the same purpose. I'll do some more investigation as per your suggestions and get back with more details. 

Hugo

DennisMcWherter:
Thanks! Looking forward to seeing the update! :)

zhangzhenhuajack:
1：When MyEntity.class extend AbstractPersistable .  This problem while reshow.
2：Because  EntityBinding source code  cls.getMethods()  while get two getId() Methods，have the some ’AnnotationPresent(Id.class)‘，but One is 
the interface Persistable<T>‘s Method and return Type is Obejct。
3：May be have better method ?

clayreimann:
Could you provide a small project that highlights this issue? That way we can experiment and write tests to ensure we don’t break this use case in the future.

PR #622, opened 2018-02-03
Update PersistentResourceFetcher.java

### LINKED ISSUE #621: PersistentResourceFetcher constructor should be public
I am trying to implement a GraphQL API executor bean in Kotlin, like this:
```kotlin
@Bean
fun elideGraphQlApi(elide: Elide): GraphQL {
    val fetcher = PersistentResourceFetcher(elide.elideSettings)
    val builder = ModelBuilder(elide.elideSettings.dictionary, fetcher)
    return GraphQL(builder.build())
}
```
Since `PersistentResourceFetcher` constructor is not strictly marked as public, it is not accessible for Kotlin(nor Java) compiler.

Serg-de-Adelantado:
Change class constructor visibility to public to avoide Kotlin compilation error: `Kotlin: Cannot access '<init>': it is public/*package*/` while using this class. This should fix #621.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 4.0.1
+**Fixes**
+ * Change `PersistentResourceFetcher` constructor visibility to public in order to allow this class instantiation outside of the elide-graphql.
+
 ## 4.0.0
 
 See: 4.0-beta-5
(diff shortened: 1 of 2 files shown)


yahoocla:

Serg-de-Adelantado:
@aklish I've already did it: 
image

DennisMcWherter:
Before merging, can you please just update the changelog to note that we made this public? This would be the first entry in the `4.0.1` release notes.

DennisMcWherter:
Great, thanks for the fix! 👍

DennisMcWherter:
Just an update: I'll be releasing 4.0.1 with just this fix right now as it's potentially important to get going.

PR #636, opened 2018-03-19
Register life cycle hook functions

aklish:


elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -361,12 +363,37 @@ private void bindTriggerIfPresent(Class<? extends Annotation> annotationClass, A
...
+                    } else {
+                        method.invoke(entity);
+                    }
+                } catch (ReflectiveOperationException e) {
+                    Throwables.propagateIfPossible(e.getCause());
+                    throw new IllegalArgumentException(e);
+                }
+            };
+
+            fieldsToTriggers.put(Pair.of(annotationClass, value), callback);
```
DennisMcWherter: Can we just call `bindTrigger` here?

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -361,12 +363,37 @@ private void bindTriggerIfPresent(Class<? extends Annotation> annotationClass, A
...
         }
     }
 
-    public <A extends Annotation> Collection<Method> getTriggers(Class<A> annotationClass, String fieldName) {
-        Collection<Method> methods = fieldsToTriggers.get(Pair.of(annotationClass, fieldName));
+    public void bindTrigger(Class<? extends Annotation> annotationClass,
+                            String fieldOrMethodName,
+                            LifeCycleHook callback) {
+        fieldsToTriggers.put(Pair.of(annotationClass, fieldOrMethodName), callback);
```
DennisMcWherter: How does `put` behave for MultivaluedMap? Will it fallback to `putSingle` behavior? We probably want to merge rather than overwrite in general, right? If so, we might consider [MutliValuedMap#add](https://docs.oracle.com/javaee/7/api/javax/ws/rs/core/MultivaluedMap.html#add-K-V-).

elide-annotations/src/main/java/com/yahoo/elide/functions/LifeCycleHook.java
```diff
@@ -0,0 +1,28 @@
+/*
+ * Copyright 2018, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.functions;
+
+import com.yahoo.elide.security.ChangeSpec;
+import com.yahoo.elide.security.RequestScope;
```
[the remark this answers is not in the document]
wcekan: @DennisMcWherter No.  This package knows nothing about core.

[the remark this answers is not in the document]
DennisMcWherter: Didn't realize this was in Elide annotations 👍

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
aklish:
```    /**
     * Adds a key-value mapping to this multi-valued map.
     * <p>
     * Unlike a normal {@code Map} the previous value is not replaced.
     * Instead the new value is added to the collection stored against the key.
     * Depending on the collection type used, duplicate key-value mappings may
     * be allowed.
     * <p>
     * The method will return {@code true} if the size of the multi-valued map
     * has been increased because of this operation.
     *
     * @param key  the key to store against
     * @param value  the value to add to the collection at the key
     * @return true if the map changed as a result of this put operation, or false
     *   if the map already contained the key-value mapping and the collection
     *   type does not allow duplicate values, e.g. when using a Set
     * @throws UnsupportedOperationException if the put operation is not supported by
     *   this multi-valued map, e.g. if it is unmodifiable
     * @throws NullPointerException if the key or value is null and null is invalid (optional)
     * @throws IllegalArgumentException if some aspect of the specified key or value prevents
     *   it from being stored in this multi-valued map
     */
    boolean put(K key, V value);```

File: elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java [hunk 4]
```diff
@@ -361,12 +363,37 @@ private void bindTriggerIfPresent(Class<? extends Annotation> annotationClass, A
...
+            Method method = (Method) fieldOrMethod;
+            LifeCycleHook callback = (entity, scope, changes) -> {
+                try {
+                    if (changes.isPresent() && method.getParameterCount() == 2
+                            && method.getParameterTypes()[0].isInstance(scope)
+                            && method.getParameterTypes()[1].isInstance(changes.get())) {
+                        method.invoke(entity, scope, changes.get());
+                    } else if (method.getParameterCount() == 1 && method.getParameterTypes()[0].isInstance(scope)) {
+                        method.invoke(entity, scope);
+                    } else {
```
clayreimann: /nit: maybe extract `method.getParameterCount()` and `method.getParameterTypes()` out into `bindTriggerIfPresent` and close over the extracted values so we're not redoing it every time we attempt to invoke this hook

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
wcekan: For `MultivalueMap` a `put` acts like an `add` to the list of values for the given key.

elide-annotations/src/main/java/com/yahoo/elide/functions/LifeCycleHook.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+@FunctionalInterface
+public interface LifeCycleHook<T> {
+    /**
+     * Run for a lifecycle event
+     * @param elideEntity The entity that triggered the event
+     * @param requestScope The request scope
+     * @param changes Optionally, the changes that were made to the entity
+     */
+    public abstract void execute(T elideEntity,
```
wcekan: _Style_:  methods should be declared after field declarations.

elide-core/src/test/java/com/yahoo/elide/core/LifeCycleTest.java
```diff
@@ -39,8 +52,17 @@
  * Tests the invocation & sequencing of DataStoreTransaction method invocations and life cycle events.
  */
 public class LifeCycleTest {
+
+    public class MockCallback<T> implements LifeCycleHook<T> {
```
wcekan: Style: inner class should be declared after field declarations.

PR #637, opened 2018-03-21
Updated when pre-security hooks run for create.

DennisMcWherter:
This changes when pre-security hooks run for commit-time checks. In general:

`OnCreatePreSecurity` actually runs before the security check is run but _after_ the values on the object are set. Since create checks are all deferred to commit-time to look at the final state of the object, this allows the `PreSecurity` hooks to be useful for creates. If we don't do this, the object is totally empty at time of `PreSecurity`.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -1337,8 +1337,11 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
         if (!isNewlyCreated) {
             runTriggers(OnUpdatePreSecurity.class, CLASS_NO_FIELD);
             runTriggers(OnUpdatePreSecurity.class, fieldName, Optional.of(spec));
-            checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, newValue, existingValue);
         }
+
+        // TODO: Need to refactor this logic. For creates this is properly converted in the executor. This logic
+        // should be explicitly encapsulated here, not there.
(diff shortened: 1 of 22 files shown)


DennisMcWherter:
**NOTE:** I want to add some tests before we merge. So please **DO NOT MERGE** until I have added those so we can avoid regressions in the future.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1335,12 +1334,12 @@ protected void setValueChecked(String fieldName, Object newValue) {
         ChangeSpec spec = new ChangeSpec(this, fieldName, existingValue, newValue);
         boolean isNewlyCreated = requestScope.getNewPersistentResources().contains(this);
 
-        Class<? extends Annotation> annotationClass = (isNewlyCreated)
-                ? OnCreatePreSecurity.class  : OnUpdatePreSecurity.class;
+        if (!isNewlyCreated) {
+            runTriggers(OnUpdatePreSecurity.class, CLASS_NO_FIELD);
+            runTriggers(OnUpdatePreSecurity.class, fieldName, Optional.of(spec));
+            checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, newValue, existingValue);
```
aklish: This looks correct but is actually not right.  It gets remapped to create permission in ActivePermissionExectutor.

DennisMcWherter:
Tests have been updated to catch this, please review. We were calling the wrong variant of `setValue` (we should've run `setValueChecked`) in our tests.

PR #639, opened 2018-03-22
Upgrade hibernate 5 to 5.2.15

aklish:


elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -82,8 +82,8 @@ public void flush(RequestScope requestScope) {
         try {
             deferredTasks.forEach(Runnable::run);
             deferredTasks.clear();
-            FlushMode flushMode = session.getFlushMode();
-            if (flushMode != FlushMode.COMMIT && flushMode != FlushMode.MANUAL && flushMode != FlushMode.NEVER) {
```
DennisMcWherter: `FlushMode.NEVER` is only deprecated, right? Should we still support it if a user overrides and uses it?

PR #641, opened 2018-03-30
Log runtime exception as error

jkusa:
- Log runtime exception as error

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,9 @@
 # Change Log
+
+## TBD
+**Fixes**
+ * Log runtime exception as error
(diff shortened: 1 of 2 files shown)


changelog.md
```diff
@@ -1,4 +1,9 @@
 # Change Log
+
+## TBD
```
DennisMcWherter: `4.2.1`

jkusa: I talked to @aklish about this. Since this can wait, we can set the version once we are ready to cut a release. Maybe it will go into `4.3.0`.

PR #643, opened 2018-04-11
Add fetch joins for to-one relationships to improve HQL performance

aklish:
Hibernate will lazily fetch to-one relationships leading to N+1 problem during hydration.  This PR should help mitigate.

### DIFF
--- a/elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/AbstractHQLQueryBuilder.java
+++ b/elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/AbstractHQLQueryBuilder.java
@@ -182,6 +182,12 @@ public abstract class AbstractHQLQueryBuilder {
         return joinClause.toString();
     }
 
+    /**
+     * Builds a JOIN clause that eagerly fetches to-one relationships that Hibernate needs to hydrate.
+     * @param entityClass The entity class that is being queried in the HQL query.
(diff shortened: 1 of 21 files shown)


elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/AbstractHQLQueryBuilder.java
```diff
@@ -42,6 +45,7 @@
     protected static final String FROM = " FROM ";
     protected static final String JOIN = " JOIN ";
     protected static final String LEFT = " LEFT";
+    protected static final String FETCH = " FETCH ";
     protected static final String SELECT = "SELECT ";
     protected static final String AS = " AS ";
```
clayreimann: Do we want to inline these since we've ditched codacity? I think it'd make building these join clauses clearer

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/RootCollectionFetchQueryBuilder.java
```diff
@@ -74,6 +75,8 @@ public Query build() {
                     + AS
                     + entityAlias
                     + SPACE
+                    + extractToOneMergeJoins(entityClass, entityAlias)
+                    + SPACE
```
clayreimann: Ditto above

PR #649, opened 2018-04-19
Add support for custom error objects

olOwOlo:
- Add `CustomErrorException` extend the `HttpStatusException`
- Add `ErrorObjects` to help build error objects

For example, people build a `ErrorObjects` and provides it to the `CustomErrorException` (or its subclass) and throw it in the lifecycle trigger funtions. Then the error in the response is what they want.

Is this a valid use case ?

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/ErrorObjects.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/ErrorObjects.java
@@ -17,7 +17,7 @@ import java.util.Objects;
  * Error Objects, see http://jsonapi.org/format/#error-objects. <br><br>
  * Builder example: <br>
  * <pre><code>
- *   ErrorObjects errorObjects = ErrorObjects.create()
+ *   ErrorObjects errorObjects = ErrorObjects.builder()
  *       .addError()
  *           .withDetail("first error message")
  *       .addError()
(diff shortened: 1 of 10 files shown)


yahoocla:

elide-core/src/main/java/com/yahoo/elide/core/ErrorObjects.java
```diff
@@ -0,0 +1,123 @@
...
+            }
+
+            public ErrorObjectBuilder put(String key, Object value) {
+                errorObject.put(key, value);
+                return this;
+            }
+
+            public ErrorObjectBuilder addError() {
+                if (errorObject.isEmpty()) {
+                    throw new IllegalArgumentException("empty error object is not allow");
```
clayreimann: `Errors must contain at least one key-value pair`

elide-core/src/main/java/com/yahoo/elide/core/ErrorObjects.java
```diff
@@ -0,0 +1,123 @@
...
+                }
+                return rootBuilder.addError();
+            }
+
+            public ErrorObjects build() {
+                if (errorObject.isEmpty()) {
+                    throw new IllegalArgumentException("empty error object is not allow");
```
clayreimann: `Errors must contain at least one key-value pair`

elide-core/src/main/java/com/yahoo/elide/core/ErrorObjects.java
```diff
@@ -0,0 +1,123 @@
...
+ * </code></pre>
+ */
+public class ErrorObjects {
+    @Getter private final List<Map<String, Object>> errors;
+
+    public ErrorObjects(List<Map<String, Object>> errors) {
+        this.errors = Objects.requireNonNull(errors, "errors must not be null");
+    }
+
+    public static ErrorObjectsBuilder create() {
```
clayreimann: This should be a static method inside `ErrorObjectsBuilder`, it also feels weird to create and `ErrorObjects` in the builder before we call `build()`.

elide-core/src/main/java/com/yahoo/elide/core/ErrorObjects.java
```diff
@@ -0,0 +1,123 @@
...
+
+        public ErrorObjectBuilder addError() {
+            Map<String, Object> errorObject = new HashMap<>();
+            errors.add(errorObject);
+            return new ErrorObjectBuilder(this, errorObject);
+        }
+
+        public ErrorObjects build() {
+            if (errors.isEmpty()) {
+                throw new IllegalArgumentException("at least one error object");
```
clayreimann: `At least one error is required`

elide-core/src/main/java/com/yahoo/elide/core/ErrorObjects.java
```diff
@@ -0,0 +1,123 @@
...
+                errorObject.put("title", title);
+                return this;
+            }
+
+            public ErrorObjectBuilder withDetail(String detail) {
+                errorObject.put("detail", detail);
+                return this;
+            }
+
+            public ErrorObjectBuilder put(String key, Object value) {
```
clayreimann: Let's call this `with(String key, Object value)` to make it more consistent with the other methods. The other, more specific, methods should call this method. e.g.

```java
public ErrorObjectBuilder withId(String id) {
    return with("id", id);
}
```

elide-core/src/main/java/com/yahoo/elide/core/ErrorObjects.java
clayreimann: This check probably also wants to be refactored into a separate function like `public void validate() throws IllegalArgumentException`. That way if we impose further restrictions in the future (required detail or status code) we only have to write the check once.

elide-core/src/main/java/com/yahoo/elide/core/ErrorObjects.java
```diff
@@ -0,0 +1,123 @@
...
+            if (errors.isEmpty()) {
+                throw new IllegalArgumentException("at least one error object");
+            }
+            return container;
+        }
+
+        /**
+         * ErrorObjectBuilder.
+         */
+        public static class ErrorObjectBuilder {
```
clayreimann: I appreciate how your two builders chain together nicely, but that makes me think that they're really just two parts of the same logical unit. I think we should combine them and just have the `ErrorObjectsBuilder` keep a pointer to the error we are currently building.

elide-core/src/main/java/com/yahoo/elide/core/exceptions/CustomErrorException.java
```diff
@@ -0,0 +1,62 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.Objects;
+
+/**
+ * Define your business exception extend this.
+ */
+public class CustomErrorException extends HttpStatusException {
```
clayreimann: As you mention in #650 both the JSON API and GraphQL specs call for errors to be presented in this object format. Perhaps we should fold this into the `HttpStatusException` class so that any error thrown by Elide conforms to the spec.

olOwOlo:
Hi @clayreimann 

I made some changes

- Refactor `ErrorObjects`
- Fold `CustomErrorException` into `HttpStatusException`

    I am assuming you mean to change the behavior of `HttpStatusException`.
    
    However, it seems that subclasses of `HttpStatusException` do not make use of the `customErrors`. Maybe keeping a separate `CustomErrorException` and simply changing the behavior of `HttpStatusException` is better?

- Return GraphQL error description with key `message`

    > Every error must contain an entry with the key `message` with a string description of the error

olOwOlo:
@aklish 

How about this way? If `returnErrorObjects` is set to `true`, the [`buildErrorResponse`](https://github.com/yahoo/elide/blob/06ff5463a8f4296d7bb2df04547a1f54fb3c5641/elide-core/src/main/java/com/yahoo/elide/Elide.java#L299) method will return error objects .

clayreimann:
@olOwOlo Looks good, thanks again for the contribution. I had one minor comment based on the code Lombok is generating, feel free to ignore it though.

olOwOlo:
Added test. 
Need I use lowercase package name ?

PR #651, opened 2018-04-25
Implement JPA DataStore

belovaf:
DataStore with only javax.persistence dependency.
Can be used with any JEE container.
Supports both JTA and Non-JTA transactions.


yahoocla:

aklish:
I just looked over the PR.  Do you use EclipseLink or Hibernate for your JPA provider?  

As far as tests, we would need to wire in the elide-integration-tests dependency and the maven-failsafe-plugin to run them.  The integration tests need a DataStoreSupplier property defined that 
builds a configured data store that can talk to the test database.  It would then run a whole bunch of tests.

I think there would be a lot of value in having support for both Hibernate and EclipseLink.

We would also need to think about the best way to structure this.  The datastores today come bundled with their own datastore supplier.  However, this store (JpaDataStore) is unique in that it has to be injected with an EntityManager and ideally would not have any dependency on a specific JPA provider.  

My initial thought is that we could update Hibernate5 store to include a Supplier that supplies the new JpaDataStore.  However, I would love to see the tests also run against EclipseLink.

I will play with the branch a bit and we'll discuss some options for structuring.  Definitely open to suggestions there.

belovaf:
We have many separate teams. Each team can use any provider, but in general they use EclipseLink and Hibernate. I test this DataStore on both providers.
Actually it can work with any provider without any additional work except one moment:
AbstractJpaTransaction#getRelation linke 184:
`if (IS_PERSISTENT_COLLECTION.test(filteredVal)) {`
Can we replace this checking with something not depending on provider internal collection names?

Supplier for this DataStore on JEE environment looks very simple:
```
@ApplicationScoped
public class DataStoreProducer {
    @Inject private EntityManager em;

    @Produces @ApplicationScoped
    public DataStore getContainerManagedJpaDataStore() {
        return new JpaDataStore(() -> em, JtaTransaction::new);
    }
}
```

belovaf:
I rework `PersistenceCollectionChecker`.
Now it checks collection by package.

aklish:
Quick update.  I pulled down your branch and hacked in our integration tests.  A number of tests were failing including some deadlocks (some cases where transactions didn't either commit or rollback).  I've been slowly fixing some of the issues as I've had time.  It might take me a bit to get (just based on available time) everything to a working state with all tests passing.  If you want, I can submit a PR against your repo with the changes thus far if you want to play with it on your end.  Alternatively, I can slowly plug away at this.  

Once this is working though, I still think we'll want to restructure the existing data store directories.  Our current thinking is we would create separate maven modules under the jpa store for
 - Hibernate3
 - Hibernate5
 - Eclipselink (If you have something here, that would be great).

Each submodule would run the integration test suite.   Also, I'm currently testing with the non-jta transaction version.  I think we should probably test both to make sure they can both pass the test suite.

That's my update for now.  Definitely interested in seeing this work part of Elide!

belovaf:
Hello! Thank you for feedback. It will be great if you can submit PR. I try to fix failed tests.

By the way, this implementation of transaction seems error prone for me. It would be great if I could register Elide.class as bean and just add @Transactional or @TransactionAttribute annotation.
I create new issue for this https://github.com/yahoo/elide/issues/652.

aklish:
I was able to get the tests to pass (although there will still need to be restructuring of the tests).  

I found a few problems in the JpaDatastore, a number of problems with the tests and one bug in Elide's graphQL code.

I'm going to first submit separate PRs for the test fixes and Elide bug.

aklish:
I think I closed this by accident.  Still working on it.  First getting in test fixes though.

aklish:
Starting a new branch to continue this work.

PR #657, opened 2018-05-03
Correctly flushing the transaction between two GraphQL root document …

aklish:
…fetches in the same request

I found this testing the JpaDataStore (GraphQLIT).

### DIFF
--- a/elide-graphql/src/main/java/com/yahoo/elide/graphql/Environment.java
+++ b/elide-graphql/src/main/java/com/yahoo/elide/graphql/Environment.java
@@ -57,9 +57,7 @@ public class Environment {
 
         if (isRoot()) {
             // Flush (but don't commit) between root queries
-            requestScope.getDirtyResources().forEach(
-                    resource -> requestScope.getTransaction().save(resource.getObject(), requestScope)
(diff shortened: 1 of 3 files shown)


elide-graphql/src/main/java/com/yahoo/elide/graphql/Environment.java
```diff
@@ -57,9 +57,7 @@ public Environment(DataFetchingEnvironment environment) {
 
         if (isRoot()) {
             // Flush (but don't commit) between root queries
-            requestScope.getDirtyResources().forEach(
-                    resource -> requestScope.getTransaction().save(resource.getObject(), requestScope)
-            );
+            requestScope.saveOrCreateObjects();
```
DennisMcWherter: This seems sane to me. Do you recall why we were deferring saves here?

aklish: Saves are deferred to prevent cascades with uninitialized objects.  Elide first initializes everything in memory - and then triggers persist/merge - which then causes cascades.  That's my understanding.

Issue #634, opened 2018-03-05
@JsonFormat doesn't work


belovaf:
`@JsonFormat` on `Date` field doesn't work.
```
public class Entity {
    // Id and other fields declaration ...

    @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'", timezone = "GMT")
    public Date someTimestamp;
}
```
Can you suggest any workaround?

DennisMcWherter:
This is a known issue unfortunately-- our normalization step actually removes some of this metadata. It is something we do need to fix.

However, as a suggested work around, you can use your own custom serializer/deserializer to the object mapper. See this ticket for more information: https://github.com/yahoo/elide/issues/613

Let me know if that doesn't solve the issue!

belovaf:
Thank you for your suggestion.
With configuration from #613 Elide returns formatted date, but accepts numeric timestamps.
Unfortunately, custom deserializer for Date is not called in this case.

DennisMcWherter:
Sorry @belovaf, I was just referring to that as an example of using a custom serializer/deserializer. For your use case, I would expect you should be able to modify that code to write a new string out. Depending what kind of `Date` object you have, you should be able to do something like:

```java
...
new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'").format(someTimestamp);
...
```

Is that not sufficient for some reason?

belovaf:
@DennisMcWherter, thank you for response. Sure I'm able format `Date` to String. :)
The problem is that if I register serializer/deserializer, only serializer method executes.
```
mapper.getObjectMapper().registerModule(
  new SimpleModule("dateConverter", Version.unknownVersion())
    .addDeserializer(Date.class, new JsonDeserializer<Date>() { /* doesn't invoked on jsonpatch request */ })
    .addSerializer(Date.class, new JsonSerializer<Date>() { /* invoked on get request */ })
```

DennisMcWherter:
I understand the problem now; I will have to investigate this a bit deeper when I have a chance. At the very least, serializers/deserializers should work. Hopefully can report back soon.

sdhawley:
Having same problem, deserializer is not being called to convert json to Date.

Any ideas what is going on here?

aklish:
@DennisMcWherter Did you ever get a chance to investigate this issue?

aklish:
[...]

PR #672, opened 2018-07-06
Elide supporting ISO8601 Dates

aklish:
… provides example for IS08601 dates

Provides work around for issue #634.


elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
```diff
@@ -76,13 +89,11 @@ private static void setup() {
...
                     return TO_ENUM_CONVERTER;
-                } else if (targetType == UUID.class) {
-                    return TO_UUID_CONVERTER;
-                }
-                else if (Map.class.isAssignableFrom(sourceType)) {
+                } else if (Map.class.isAssignableFrom(sourceType)) {
                     return FROM_MAP_CONVERTER;
                 } else if ((String.class.isAssignableFrom(sourceType) || Number.class.isAssignableFrom(sourceType))
-                        && ClassUtils.isAssignable(targetType, Date.class)) {
+                        && ClassUtils.isAssignable(targetType, Date.class)
```
DennisMcWherter: It seems like we should be able to handle this in the lookup. Should our lookup map be keyed by a ‘(source, target)’ tuple (null target could imply a default mapping)? This way we wouldn’t have to be smart and the user could define the mapping from ‘String -> Date’ or (ideally) ‘X -> Y’. 

This would be a larger refactor than what we’re probably going for in here, but maybe something to think about.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -58,8 +73,50 @@ protected void configure() {
...
+                                            DateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'");
+                                            // You can set timezone however you wish
+                                            df.setTimeZone(TimeZone.getDefault());
+                                            jsonGenerator.writeObject(df.format(date));
+                                        }
+                                    }
+                                })
+                );
+
+                CoerceUtil.register(new Converter() {
```
DennisMcWherter: Is this spelled out somewhere in the docs that a user may need to implement a converter? It’s unfortunate that we have to do this 2x (once for us and once for Jackson)

clayreimann: Maybe we should wrap this in an interface we own? 

```java
interface Serializer<T> {
    public T convert(Class<T> aClass, Object o);
    public void serialize(T toWrite, JsonGenerator jsonGenerator, SerializerProvider serializerProvider);
} 
```

That way clients can just write the bits they care about and we can deal with registering it however we need to.

aklish:
Addressed some of the comments.  Still need tests.

elide-core/src/main/java/com/yahoo/elide/jsonapi/JsonApiMapper.java
```diff
@@ -29,6 +37,23 @@
      */
     public JsonApiMapper(EntityDictionary dictionary) {
         mapper = new ObjectMapper();
+
+        Serde<?, Date> serde = CoerceUtil.lookup(Date.class);
+
+        mapper.registerModule(
+                new SimpleModule("isoDate", Version.unknownVersion())
+                        .addSerializer(Date.class, new JsonSerializer<Date>() {
```
aklish: Other Date subclasses might also need registration here.

elide-core/src/main/java/com/yahoo/elide/ElideSettingsBuilder.java
```diff
@@ -159,6 +171,22 @@ public ElideSettingsBuilder withUseFilterExpressions(boolean useFilterExpression
         return this;
     }
 
+    public ElideSettingsBuilder withISO8601Dates(String dateFormat, TimeZone tz) {
```
aklish: I'm not exposing Serde's directly here as that would require more testing and the issue right now is Date formatting.  However, the plumbing is in place to expose Serde's at some future point.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -59,6 +60,7 @@ default ElideSettings getElideSettings(ServiceLocator injector) {
                 .withEntityDictionary(dictionary)
                 .withJoinFilterDialect(new RSQLFilterDialect(dictionary))
                 .withSubqueryFilterDialect(new RSQLFilterDialect(dictionary))
+                .withISO8601Dates("yyyy-MM-dd'T'HH:mm'Z'", TimeZone.getTimeZone("UTC"))
```
clayreimann: Do we want to turn this on by default in elide-standalone?

clayreimann: I think it's fine, it just doesn't seem to jibe with your other change here.
```java 
default boolean enableIS06081Dates() {
        return false;
    }
```

elide-core/src/main/java/com/yahoo/elide/jsonapi/JsonApiMapper.java
```diff
@@ -22,14 +30,49 @@
 public class JsonApiMapper {
     private final ObjectMapper mapper;
 
+    /**
+     * Instantiates a new Json Api Mapper.
+     */
+    public JsonApiMapper() {
+        this.mapper = new ObjectMapper();
+
+        Serde<?, Date> serde = CoerceUtil.lookup(Date.class);
```
wcekan: Should this be defined only for the default JsonApiMapper?  Or should we do
```java
 public JsonApiMapper() {
    this(new ObjectMapper());
}
```

PR #674, opened 2018-07-11
Make windows build work

jkusa:
- Make jcabi-mysql-maven-plugin work with Windows OS
  - Don't assume `/tmp` directory exists
  - Use a version of jcabi-mysql-maven-plugin that works with Windows OS
- Make path normalization portable
  - The previous implementation did not work on Windows OS


elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -272,17 +273,35 @@ protected ElideResponse handleRequest(boolean isReadOnly, Object opaqueUser,
...
+    }
+
     /**
      * Compile request to AST.
      *
      * @param path request
      * @return AST parse tree
      */
     public static ParseTree parse(String path) {
-        String normalizedPath = Paths.get(path).normalize().toString().replace(File.separatorChar, '/');
```
aklish: File.separatorChar is os specific.  We still need to do this step.

elide-core/src/test/java/com/yahoo/elide/ElideTest.java
```diff
@@ -0,0 +1,28 @@
...
+import static org.testng.Assert.assertEquals;
+
+public class ElideTest {
+
+    @Test
+    public void testPathNormalization() {
+
+        String normalizedPath;
+
+        normalizedPath = Elide.parse("/books").getText();
```
aklish: Add the exact same tests you have but invert "/" so it is "\".  I imagine these would break Elide.

elide-core/src/main/java/com/yahoo/elide/Elide.java
aklish: Nevermind - the Paths.get is the issue.  This looks correct.

aklish:
Needs an updated changelog.

wcekan:
@jkusa there appear to be other locations.  Should we have common code?

https://github.com/yahoo/elide/search?q=%22+Paths.get%22+normalize

jkusa:
@wcekan good call. Let me see if I can DRY this up.

jkusa:
Consolidated the path normalization to `JsonApiParser.normalizePath`.

@aklish @wcekan please check out the updates.

PR #690, opened 2018-08-08
Expose failure reason in ElideResponse

clayreimann:
- Provide hook for building custom responses in JsonApiEndpoint


elide-core/src/main/java/com/yahoo/elide/resources/JsonApiEndpoint.java
```diff
@@ -121,7 +121,10 @@ public Response delete(
         return build(elide.delete(path, jsonApiDocument, getUser.apply(securityContext)));
     }
 
-    private static Response build(ElideResponse response) {
-        return Response.status(response.getResponseCode()).entity(response.getBody()).build();
+    protected Response build(ElideResponse response) {
```
DennisMcWherter: Where does the `failureReason` get used? I see it being set, but never logged or added to a (presumably verbose?) response.

DennisMcWherter: Or is the plan that arbitrary endpoints could leverage this functionality?

clayreimann: We'll be using it internally. Right now error cause is opaque to consumers of Elide unless you override a bunch of stuff.

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -212,7 +212,7 @@ protected ElideResponse handleRequest(boolean isReadOnly, Object opaqueUser,
             }
             tx.flush(requestScope);
 
-            ElideResponse response = buildResponse(responder.get());
+            ElideResponse response = buildResponse(responder.get(), null);
```
aklish: Maybe use an Optional here.

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -276,19 +276,24 @@ protected ElideResponse buildErrorResponse(HttpStatusException error, boolean is
...
+    protected ElideResponse buildResponse(Pair<Integer, JsonNode> response, Throwable failureReason) {
         try {
             JsonNode responseNode = response.getRight();
             Integer responseCode = response.getLeft();
             String body = responseNode == null ? null : mapper.writeJsonApiDocument(responseNode);
-            return new ElideResponse(responseCode, body);
+            return new ElideResponse(responseCode, body, failureReason);
         } catch (JsonProcessingException e) {
-            return new ElideResponse(HttpStatus.SC_INTERNAL_SERVER_ERROR, e.toString());
+            return new ElideResponse(HttpStatus.SC_INTERNAL_SERVER_ERROR, e.toString(), e);
```
aklish: What about graphQL?

aklish:
Would be nice to have a single test for both JSON-API and GraphQL.

elide-core/src/main/java/com/yahoo/elide/ElideResponse.java
```diff
@@ -13,15 +13,17 @@
...
+    @Getter private final Throwable failureReason;
 
     /**
      * Constructor.
      *
      * @param responseCode HTTP response code
      * @param body returned body string
      */
-    public ElideResponse(int responseCode, String body) {
+    public ElideResponse(int responseCode, String body, Throwable failureReason) {
```
wcekan: I would keep the original public constructor too and deprecate it.

elide-core/src/main/java/com/yahoo/elide/Elide.java
clayreimann: There is no real analog to `ElideResponse` in GraphQL because we're leveraging GraphQLJava. The `Elide` class drives JsonAPI and is ignored by GraphQL, also the `GraphQLEndpiont` can't be customized in the way that `JsonApiEndpoint` can.

Issue #689, opened 2018-08-08
Error in type converter CoerceUtil


Stanyslav:
When using UUID as the primary key of the entity, when requesting data from the database, throws an exception:
```
java.lang.IllegalArgumentException: Parameter value [35bd1fa2-3829-11e8-8258-071bcf8a71f4] did not match expected type [java.util.UUID (n/a)]
	at org.hibernate.query.spi.QueryParameterBindingValidator.validate(QueryParameterBindingValidator.java:54) ~[hibernate-core-5.3.4.Final.jar:5.3.4.Final]
	at org.hibernate.query.spi.QueryParameterBindingValidator.validate(QueryParameterBindingValidator.java:27) ~[hibernate-core-5.3.4.Final.jar:5.3.4.Final]
	at org.hibernate.query.internal.QueryParameterBindingImpl.validate(QueryParameterBindingImpl.java:90) ~[hibernate-core-5.3.4.Final.jar:5.3.4.Final]
	at org.hibernate.query.internal.QueryParameterBindingImpl.setBindValue(QueryParameterBindingImpl.java:55) ~[hibernate-core-5.3.4.Final.jar:5.3.4.Final]
	at org.hibernate.query.internal.AbstractProducedQuery.setParameter(AbstractProducedQuery.java:493) ~[hibernate-core-5.3.4.Final.jar:5.3.4.Final]
	at org.hibernate.query.internal.AbstractProducedQuery.setParameter(AbstractProducedQuery.java:106) ~[hibernate-core-5.3.4.Final.jar:5.3.4.Final]
	at com.yahoo.elide.datastores.hibernate5.porting.QueryWrapper.setParameter(QueryWrapper.java:39) ~[elide-datastore-hibernate5-4.2.6.jar:na]
	at com.yahoo.elide.core.hibernate.hql.AbstractHQLQueryBuilder.lambda$supplyFilterQueryParameters$0(AbstractHQLQueryBuilder.java:108) ~[elide-datastore-hibernate-4.2.6.jar:na]
	at java.util.ArrayList.forEach(ArrayList.java:1249) ~[na:1.8.0_101]
	at com.yahoo.elide.core.hibernate.hql.AbstractHQLQueryBuilder.supplyFilterQueryParameters(AbstractHQLQueryBuilder.java:107) ~[elide-datastore-hibernate-4.2.6.jar:na]
[...]

PR #693, opened 2018-08-15
Add support for multiple classloaders for CoerceUtil.

DennisMcWherter:
Addresses issues with: https://github.com/yahoo/elide/issues/689

In summary, `BeanUtilsBean#getInstance` [returns a pseudo-singleton](https://commons.apache.org/proper/commons-beanutils/apidocs/org/apache/commons/beanutils/BeanUtilsBean.html#getInstance--). Namely, it's based on the `Thread.currentThread().getContextClassLoader()`.

This PR sets up CoerceUtil for each new classloader as necessary.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 4.2.7
+**Fixes**
+ * Add support for multiple classloaders when using CoerceUtils.
+
 ## 4.2.6
 **Fixes**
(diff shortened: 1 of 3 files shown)


wcekan:
@DennisMcWherter Perhaps we should not hold on to a `Classloader` reference indefinitely.  Could you instead stick that into a WeakReference<ClassLoader>?

Issue #673, opened 2018-07-10
@OnUpdatePostCommit does not trigger correct java method on HTTP PATCH request


vmironichev:
I have simple User entity that has method annotated with OnUpdatePostCommit annotation that accepts two args:
```java 
@OnUpdatePostCommit
public void onUpdateUserDetails(RequestScope requestScope, ChangeSpec changeSpec) {
  logAction(requestScope, changeSpec, ActionType.UPDATE);
}
```

after executing HTTP patch request for given entity I'm receiving exception that caused by wrong method invocation via reflection
```
curl -X PATCH \
  http://localhost:8080/api/v1/user/eb17f9b5-3762-4178-9c68-9c0de1ca4e5f \
  -H 'Content-Type: application/vnd.api+json' \
  -d '{
    "data": {
        "id": "eb17f9b5-3762-4178-9c68-9c0de1ca4e5f",
        "type": "user",
        "attributes": {
            "name": "Foo Bar"
        }
    }
}'
```

Code where exception is thrown: 
https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java#L381

probably this part of code should be executed instead:
https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java#L377
```

java.lang.IllegalArgumentException: wrong number of arguments
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at com.yahoo.elide.core.EntityBinding.lambda$bindTriggerIfPresent$3(EntityBinding.java:381)
	at com.yahoo.elide.core.PersistentResource.runTriggers(PersistentResource.java:1542)
	at com.yahoo.elide.core.RequestScope.lambda$null$3(RequestScope.java:423)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at com.yahoo.elide.core.RequestScope.runQueuedTriggers(RequestScope.java:396)
	at com.yahoo.elide.core.RequestScope.runQueuedPostCommitTriggers(RequestScope.java:380)
	at com.yahoo.elide.Elide.handleRequest(Elide.java:224)
[...]

PR #695, opened 2018-08-22
Issue673

aklish:
Fixes Issue #673


elide-core/src/main/java/com/yahoo/elide/core/LifecycleHookInvoker.java
```diff
@@ -0,0 +1,65 @@
...
+
+        //Collect all the hooks that are keyed on a specific field.
+        Collection<LifeCycleHook> hooks = dictionary.getTriggers(
+                event.getResource().getResourceClass(),
+                this.annotation,
+                event.getFieldName());
+
+        //Collect all the hooks that are keyed on any field.
+        if (!event.getFieldName().isEmpty()) {
+            hooks.addAll(dictionary.getTriggers(event.getResource().getResourceClass(), this.annotation));
```
DennisMcWherter: is it safe to modify the returned collection like this? if it changes to be immutable or changes from list to set, for instance this would alter the semantics of the `addAll` logic below. Should we create a new one with this?

elide-annotations/src/main/java/com/yahoo/elide/annotation/OnCreatePostCommit.java
```diff
@@ -21,4 +21,12 @@
 @Target({ElementType.METHOD})
 @Retention(RetentionPolicy.RUNTIME)
 public @interface OnCreatePostCommit {
+    /**
+     * Field name on which the annotated method is only triggered if that field is modified.
+     * If value is empty string, then trigger once when the object is created.
+     * If value is "*", then trigger for all field modifications.
```
clayreimann: Do you mean `If the value is "*", then this method will be triggered once for each field that the user sent in the creation request`?

Same question for the other annotations.

aklish:
Code doesn't handle exceptions thrown by life cycle hooks yet.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -471,16 +480,17 @@ private String getInheritanceKey(String subClass, String superClass) {
...
 
         this.distinctLifecycleEvents
                 .filter(CRUDEvent::isUpdateEvent)
-                .subscribe(new LifecycleHookInvoker(dictionary, OnUpdatePreSecurity.class));
+                .subscribeWith(new LifecycleHookInvoker(dictionary, OnUpdatePreSecurity.class, true));
 
         this.distinctLifecycleEvents
                 .filter(CRUDEvent::isDeleteEvent)
-                .subscribe(new LifecycleHookInvoker(dictionary, OnDeletePreSecurity.class));
+                .subscribeWith(new LifecycleHookInvoker(dictionary, OnDeletePreSecurity.class, true));
```
clayreimann: Why do pre-security hooks not `throwOnError`?

aklish: JavaRX has different exception handling for a normal subject and a replay subject when the Observer blows up.  The latter swallows the exception and invokes the uncaught exception handler (not what we want) whereas the former propagates the exception to the `onNext` caller of the subject.  

`onError` is only called if the Observable had an exception.  It doesn't seem to address the case where the Observer blows up.

I'm not sure if I'm misapplying reactiveX here - or reactiveX just has strange semantics.  In either case, this is the best I could come up with.  Open to suggestions.

Issue #644, opened 2018-04-12
Can't use @Id on private field


jstemen:
I've noticed when I try to annotate a field with @id, Elide isn't able to find it.  Is this intentional?  It seems like Elide should scan both public and private fields when looking for id annotations.

I haven't tracked down the issue entirely, but I believe part of this behavior is because of this line:

https://github.com/yahoo/elide/blob/dc214abddde61de80e4f56ceb2ec9651f5a878df/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java#L119

The line should be like this instead:
`        fieldOrMethodList.addAll(Arrays.asList(cls.getDeclaredFields())`

getFields() only returns public fields.  getDeclaredFields() returns public and private fields.

aklish:
It is not intentional.  Elide supports JPA annotations on properties.  In theory, there is no reason it couldn't also support annotations on fields.  However, there are no tests for this. 

Properties over fields is just a legacy aspect of the models we use in production.  We would fully support a PR to support field level access as well.

Most of this work however would be augmenting the tests to make sure everything was covered.  I suspect the non-test code would be minimal.

Mishanya840:
+++

aklish:
High priority issue.  We should plan on fixing in the next month or so.

PR #697, opened 2018-08-29
Issue691

### LINKED ISSUE #691: enable to put aanotation on private field
my entity class is 

```Java
package com.axelor.app.db;

import java.util.List;

import javax.persistence.Access;
import javax.persistence.AccessType;
import javax.persistence.CascadeType;
import javax.persistence.Entity;
import javax.persistence.FetchType;
import javax.persistence.GeneratedValue;
import javax.persistence.GenerationType;
import javax.persistence.Id;
import javax.persistence.ManyToOne;
import javax.persistence.OneToMany;
import javax.persistence.SequenceGenerator;
import javax.persistence.Table;
import javax.validation.constraints.NotNull;

@Entity
@Table(name = "contact")
public class Contact {

	private Long id;

	private String firstName;

	private String lastName;

	private String email;

	private String phone;
	
	private Address mainAddress;
	
	private Address shipAddress;

	private List<Address> addresses;

	public Contact() {
	}

	public Contact(String firstName, String lastName) {
		this.firstName = firstName;
		this.lastName = lastName;
	}

	@Id
	@GeneratedValue(strategy = GenerationType.SEQUENCE, generator = "CONTACT_SEQ")
	@SequenceGenerator(name = "CONTACT_SEQ", sequenceName = "CONTACT_SEQ", allocationSize = 1)
	public Long getId() {
		return id;
	}

	public void setId(Long id) {
		this.id = id;
	}

	public String getFirstName() {
		return firstName;
	}

	public void setFirstName(String firstName) {
		this.firstName = firstName;
	}

	public String getLastName() {
		return lastName;
	}

	public void setLastName(String lastName) {
		this.lastName = lastName;
	}

	public String getEmail() {
		return email;
	}

	public void setEmail(String email) {
		this.email = email;
	}

	public String getPhone() {
		return phone;
	}

	public void setPhone(String phone) {
		this.phone = phone;
	}

	@ManyToOne(fetch = FetchType.LAZY, cascade = { CascadeType.PERSIST, CascadeType.MERGE })
	public Address getMainAddress() {
		return mainAddress;
	}
	
	public void setMainAddress(Address mainAddress) {
		this.mainAddress = mainAddress;
	}
	
	@ManyToOne(fetch = FetchType.LAZY, cascade = { CascadeType.PERSIST, CascadeType.MERGE })
	public Address getShipAddress() {
		return shipAddress;
	}

	public void setShipAddress(Address shipAddress) {
		this.shipAddress = shipAddress;
	}
	@OneToMany(fetch = FetchType.LAZY, mappedBy = "contact", cascade = CascadeType.ALL, orphanRemoval = true)
	public List<Address> getAddresses() {
		return addresses;
	}

	public void setAddresses(List<Address> addresses) {
		this.addresses = addresses;
	}

}
```
 it is working fine.
but i want annotation put on private field so it is possible ??

aklish:
For entities which annotate the primary key getter (JPA property) with @Id, the behavior should be identical to 4.2.6 and earlier.

For entities which annotate the primary key field (JPA field) with @Id, Elide will only expose fields and computed attributes/relationships (which can be properties).

Also updated the dependency on Elide to JPA 2.2.

Fixes #691 and #644

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
@@ -116,8 +116,6 @@ public class EntityBinding {
         jsonApiType = type;
         entityName = name;
 
-        accessType = AccessType.PROPERTY;
-
         // Map id's, attributes, and relationships
         List<AccessibleObject> fieldOrMethodList = new ArrayList<>();
         fieldOrMethodList.addAll(Arrays.asList(cls.getDeclaredFields())
@@ -130,6 +128,7 @@ public class EntityBinding {
             accessType = AccessType.FIELD;
 
 
+            /* Add all public methods that are computed */
             fieldOrMethodList.addAll(Arrays.asList(cls.getMethods())
                     .stream()
                     .filter((method) -> !Modifier.isStatic(method.getModifiers()))
@@ -137,8 +136,18 @@ public class EntityBinding {
                             || method.isAnnotationPresent(ComputedRelationship.class))
                     .collect(Collectors.toList()));
         } else {
+            /* Preserve the behavior of Elide 4.2.6 and earlier */
+            accessType = AccessType.PROPERTY;
+
             fieldOrMethodList.clear();
 
+            /* Add all public fields */
+            fieldOrMethodList.addAll(Arrays.asList(cls.getFields())
+                .stream()
+                .filter((field) -> ! Modifier.isStatic(field.getModifiers()))
+                .collect(Collectors.toList()));
+
+            /* Add all public methods */
             fieldOrMethodList.addAll(Arrays.asList(cls.getMethods())
                     .stream()
                     .filter((method) -> !Modifier.isStatic(method.getModifiers()))
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
@@ -119,25 +119,22 @@ public class EntityBinding {
         entityName = name;
 
         // Map id's, attributes, and relationships
-        List<AccessibleObject> fieldOrMethodList = new ArrayList<>();
-
-        fieldOrMethodList.addAll(
-                getInstanceMembers(cls.getDeclaredFields(), (field) -> ! ((Field) field).isSynthetic()));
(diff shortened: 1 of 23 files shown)


elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -116,16 +118,46 @@ public EntityBinding(EntityDictionary dictionary, Class<?> cls, String type, Str
 
         // Map id's, attributes, and relationships
         List<AccessibleObject> fieldOrMethodList = new ArrayList<>();
-        fieldOrMethodList.addAll(Arrays.asList(cls.getFields())
+        fieldOrMethodList.addAll(Arrays.asList(cls.getDeclaredFields())
```
clayreimann: `Arrays.stream(cls.getDeclaredFields())`

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -116,16 +118,46 @@ public EntityBinding(EntityDictionary dictionary, Class<?> cls, String type, Str
...
                 .stream()
-                .filter((method) -> ! Modifier.isStatic(method.getModifiers()))
+                .filter((field) -> ! Modifier.isStatic(field.getModifiers()))
                 .collect(Collectors.toList()));
 
+            /* Add all public methods */
+            fieldOrMethodList.addAll(Arrays.asList(cls.getMethods())
+                    .stream()
+                    .filter((method) -> !Modifier.isStatic(method.getModifiers()))
+                    .collect(Collectors.toList()));
```
clayreimann: I think we should extract part of this out into a function, since it appears 4 times in this diff.

```java
private <T extends AccessibleObject> List<T> getInstanceMembers(T[] objects, Predicate<T> filteredBy) {
    return Arrays.stream(objects)
        .filter(o -> !Modifier.isStatic(o.getModifiers()))
        .filter(filteredBy)
}
```

aklish: o.getModifiers won't work here because it is not a common inherited method from AccessibleObject (It belongs in both Field and Method but not in AccessibleObject).

aklish: I think I found another way though...

PR #700, opened 2018-09-01
Hold weak references to class loaders.

DennisMcWherter:
Updates the multiple classloader implementation to hold a `WeakReference<ClassLoader>` instead of a direct reference to `ClassLoader`.

@wcekan does this look better?

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
+++ b/elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
@@ -16,7 +16,9 @@ import org.apache.commons.beanutils.ConversionException;
 import org.apache.commons.beanutils.ConvertUtils;
 import org.apache.commons.beanutils.Converter;
 
+import java.lang.ref.WeakReference;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
```diff
@@ -32,7 +34,7 @@
     private static final FromMapConverter FROM_MAP_CONVERTER = new FromMapConverter();
     private static final Map<Class<?>, Serde<?, ?>> SERDES = new HashMap<>();
     private static final BeanUtilsBean BEAN_UTILS_BEAN_INSTANCE = setup();
-    private static final Set<ClassLoader> INITIALIZED_CLASSLOADERS = ConcurrentHashMap.newKeySet();
+    private static final Set<WeakReference<ClassLoader>> INITIALIZED_CLASSLOADERS = ConcurrentHashMap.newKeySet();
```
wcekan: We could also consider a synchronized `WeakHashMap` for the set.

elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
```diff
@@ -106,19 +108,41 @@ public Converter lookup(Class<?> sourceType, Class<?> targetType) {
...
+
+    /**
+     * Method to check if classloader has been loaded or not.
+     *
+     * <strong>NOTE:</strong> The number of class loaders is expected to be <em>small enough</em>, that this
+     * scan per request is negligible.
+     *
+     * @return True if classloader is loaded, false otherwise.
+     */
+    private static boolean isCurrentClassLoaderInitialized() {
```
wcekan: This would all be handled automatically if we used `WeakHashMap` and be more efficient since the Reference queue would be polled for any expired weak references.

wcekan: Even better
```java
    private static final Set<ClassLoader> INITIALIZED_CLASSLOADERS =
            Collections.newSetFromMap(new MapMaker().weakKeys().makeMap());
```

elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
wcekan: Even better
```java
    private static final Set<ClassLoader> INITIALIZED_CLASSLOADERS =
            Collections.newSetFromMap(new MapMaker().weakKeys().makeMap());
```

wcekan:
@DennisMcWherter I rebased your PR and made the change to Weak Set.  Please review.

DennisMcWherter:
Can't officially approve my own CR! :) 

:+1:

LGTM. Will merge

PR #706, opened 2018-09-18
Fix #696 Introduce CaseSensitivityStrategy for string comparison

### LINKED ISSUE #696: Case insensitive string search cripples performance in MySQL
Hi guys,

we finally made the transition to the Elide 4.x series. Unfortunately a lot of our users complained, that queries that took seconds before now take minutes (which makes most of them return a Gateway timeout by Nginx).

My investigation shows, that the reason is the case insensitive commit https://github.com/yahoo/elide/commit/4f00b7653ddb64a4762aea2d0de197ced68e5ec6, which wraps `==` comparisons for string columns in a `lower()` function. In our MySQL setup this leads to full table scan instead of index usage (even on columns with case insensitive collation! Ironically these already worked case insensitive).

To solve this, their should be some sort of configuration, be it in Elide configuration, on table level or on field level.

If we can agree on an approach, I'd try to implement it.

Brutus5000:
Following our discussion in #696 I implemented an additional configuration to the RSQLDialect.
I am not totally sure whether this is the correct place for a "global configuration", but this would suit the requirements in our project without changing the default Elide behavior.

Any feedback is welcome.

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
@@ -19,21 +19,10 @@ import com.yahoo.elide.parsers.JsonApiParser;
 import com.yahoo.elide.utils.coerce.CoerceUtil;
 import cz.jirutka.rsql.parser.RSQLParser;
 import cz.jirutka.rsql.parser.RSQLParserException;
-import cz.jirutka.rsql.parser.ast.AndNode;
-import cz.jirutka.rsql.parser.ast.ComparisonNode;
-import cz.jirutka.rsql.parser.ast.ComparisonOperator;
-import cz.jirutka.rsql.parser.ast.Node;
-import cz.jirutka.rsql.parser.ast.OrNode;
-import cz.jirutka.rsql.parser.ast.RSQLOperators;
-import cz.jirutka.rsql.parser.ast.RSQLVisitor;
+import cz.jirutka.rsql.parser.ast.*;
 
 import javax.ws.rs.core.MultivaluedMap;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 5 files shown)


Brutus5000:
Ok guys enlighten me what Better Code Hub is telling me. All I get is: 
`Only the repository's owners and organization members (if the repository is an organization repository) can view its Better Code Hub results.`

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -19,21 +19,10 @@
...
 import cz.jirutka.rsql.parser.RSQLParser;
 import cz.jirutka.rsql.parser.RSQLParserException;
-import cz.jirutka.rsql.parser.ast.AndNode;
-import cz.jirutka.rsql.parser.ast.ComparisonNode;
-import cz.jirutka.rsql.parser.ast.ComparisonOperator;
-import cz.jirutka.rsql.parser.ast.Node;
-import cz.jirutka.rsql.parser.ast.OrNode;
-import cz.jirutka.rsql.parser.ast.RSQLOperators;
-import cz.jirutka.rsql.parser.ast.RSQLVisitor;
+import cz.jirutka.rsql.parser.ast.*;
```
aklish: I believe our checkstyle does not like asterisk includes.

Brutus5000:
Code style was fixed.
However I would recommend to remove the Better Code Hub integration. I analyzed my fork and Elide fails on master in 50% of the disciplines. No chance that a contributor will ever get this test to pass.

aklish:
@Brutus5000 Thanks for the submission.  Changes LGTM.  I agree with BetterCodeHub removal.  We were comparing with Codacy (and hoping for something better).  I think Codacy is more useful/useable.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -28,12 +28,7 @@
 import cz.jirutka.rsql.parser.ast.RSQLVisitor;
 
 import javax.ws.rs.core.MultivaluedMap;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
```
clayreimann: @aklish is correct, we prefer to avoid wildcard imports.

clayreimann:
Acutally, @Brutus5000 can you update the change log? I'll merge and cut a new version once the change log has been updated.

Brutus5000:
@clayreimann Done. 👍

PR #708, opened 2018-09-25
Scan for @Id annotated fields in superclasses

### LINKED ISSUE #707: Field level annotations: @id not found on super classes
Hi guys,

I have refactored all of our entities to JPA annotations on private fields (generating getter/setter using lombok). As an additional level of complexity we have an an abstract entity containing the @id annotation plus a create/update time.

This scenario worked before because `Class<T>.getMethods()` returns all methods including methods of super classes. Unfortunately `Class<T>.getDeclaredFields()` only returns fields of the class, but not any fields of superclasses.

To make the scenario work for fields EntityBinding needs to be enhanced to recursively read all fields of superclasses.

Brutus5000:
(fixes #707)


Brutus5000:
@aklish Does this match with what you had in mind?

elide-integration-tests/src/test/java/com/yahoo/elide/initialization/FieldLevelParentClassIdIntegrationTestApplicationResourceConfig.java
```diff
@@ -0,0 +1,18 @@
...
+ */
+package com.yahoo.elide.initialization;
+
+import com.yahoo.elide.audit.TestAuditLogger;
+import org.glassfish.jersey.server.ResourceConfig;
+
+/**
+ * Resource config for FieldLevelParentClassId IT tests.
+ */
+public class FieldLevelParentClassIdIntegrationTestApplicationResourceConfig extends ResourceConfig {
```
aklish: I don't think you should need to create this class.  You should be able to remove it.

elide-integration-tests/src/test/java/com/yahoo/elide/initialization/FieldLevelParentClassIdStandardTestBinder.java
```diff
@@ -0,0 +1,76 @@
...
+import org.glassfish.hk2.api.Factory;
+import org.glassfish.hk2.utilities.binding.AbstractBinder;
+
+import java.util.Arrays;
+import java.util.HashMap;
+
+/**
+ * Typical-use test binder for integration test resource configs.
+ */
+public class FieldLevelParentClassIdStandardTestBinder extends AbstractBinder {
```
aklish: You shouldn't need to have to create another one of these.

elide-integration-tests/src/test/java/com/yahoo/elide/fieldLevelTest/FieldLevelParentClassIdIT.java
```diff
@@ -0,0 +1,53 @@
...
+import org.testng.annotations.Test;
+
+import static com.jayway.restassured.RestAssured.given;
+
+public class FieldLevelParentClassIdIT extends AbstractIntegrationTestInitializer {
+    private final JsonParser jsonParser = new JsonParser();
+
+    private static final String JSONAPI_CONTENT_TYPE = "application/vnd.api+json";
+
+    public FieldLevelParentClassIdIT() {
```
aklish: You can remove this constructor (rely on the default constructor).  Then you should be able to remove some of the initialization scaffolding classes below.

Brutus5000:
Yeah I didn't really get a grasp on how your integretation tests are supposed to work. And I didn't even notice my IDE telling me that the classes are not required ;) Ok should be fixed now.

elide-core/src/test/java/com/yahoo/elide/core/EntityBindingTest.java
```diff
@@ -0,0 +1,48 @@
...
+    @Test
+    public void testGetAllFields() throws Exception {
+        List<AccessibleObject> allFields = entityBinding.getAllFields();
+
+        assertEquals(allFields.size(), 2);
+        assertEquals(allFields.get(0), ChildClass.class.getDeclaredField("childField"));
+        assertEquals(allFields.get(1), ParentClass.class.getDeclaredField("parentField"));
+    }
+
+    private class ParentClass {
```
wcekan: Could we add
```java
    @Test
    public void testIdField() throws Exception {
        AccessibleObject idField = entityBinding.getIdField();
        assertEquals(idField, ParentClass.class.getDeclaredField("parentField"));
    }
```

Brutus5000:
It would be really awesome if you could create a new release as our project is really affected especially by #696 (unfortunately the project can't be build by Jitpack nor locally on a windows machine).

aklish:
Will create one tonight.

Issue #710, opened 2018-10-05
[Field level annotations] @Id in superclass shown as "null" in relationships


Brutus5000:
Hey there, I'm still trying to migrate our application to field level annotations.
Since we fixed the @Id detection in superclasses I now run into a new issue:

For all my entities that use my AbstractEntity superclass with the @Id annotation the relationship id is now resolved as "null". Entities that habe the @Id annotation in their own class are not affected.

Example:
```
{
    "data": [
        {
            "type": "map",
            "id": "1",
            "attributes": {
                "averageReviewScore": 0,
                "battleType": "skirmish",
                "createTime": "2018-10-05T21:08:29+02:00",
                "displayName": "SCMP_001",
                "mapType": "FFA",
                "numberOfReviews": 0,
                "updateTime": "2018-10-05T21:08:29+02:00"
            },
            "relationships": {
                "author": {
                    "data": {
                        "type": "player",
                        "id": "null"
                    }
                },
                "latestVersion": {
                    "data": {
                        "type": "mapVersion",
                        "id": "null"
                    }
                },
                "statistics": {
                    "data": {
                        "type": "mapStatistics",
                        "id": "1"
                    }
                },
                "versions": {
                    "data": [
                        {
                            "type": "mapVersion",
                            "id": "null"
                        }
                    ]
                }
            }
        }
    ]
}
```

MapStatic entitiy has @Id on itself.
All other entities have it in their superclass.

aklish:
Could you make a failing test case?  I eyeballed the code - but didn't see anything obvious.

aklish:
@Brutus5000 I started to investigate.  It looks like I failed to notice that your original IT test wasn't running.  This PR gets that working and fixes a bug (although I think there maybe a better way to do this) for setting private fields from inherited classes.  Can you rebase your test PR on this [PR](https://github.com/yahoo/elide/pull/712)?  Thanks.

aklish:
I believe this is now fixed.

PR #712, opened 2018-10-06
Fixed IT tests that were not running.  Fixed setting inherited fields…

aklish:
… that are private

This is a first pass towards working on issue #710


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1484,9 +1484,12 @@ protected void setValue(String fieldName, Object value) {
             throw handleInvocationTargetException(e);
         } catch (IllegalArgumentException | NoSuchMethodException noMethod) {
             try {
-                Field field = targetClass.getField(fieldName);
+                Field field = (Field) dictionary.getAccessibleObject(obj, fieldName);
```
aklish: I might do some cleanup here.  This was just to get the test to pass.

PR #728, opened 2018-12-11
Filter expression path

wcekan:
Adds the annotation `FilterExpressPath` to automatically build the path for a FilterExpressionCheck.

This way multiple beans can use the same filter while providing the unique path to the property in order for SQL to build the correct filter.

For example, added a test for
Book -> Publisher -> Editor.

Book provides the path with
```java
@FilterExpressionPath("publisher.editor")
@ReadPermission(expression = "Field path editor check")
public Author getEditor() {
```


elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -308,6 +308,9 @@ public static Operator fromString(final String string) {
             if ("this".equals(field)) {
                 continue;
             }
+            if (val == null) {
+                continue;
```
aklish: return null or break.

elide-core/src/main/java/com/yahoo/elide/security/FilterExpressionCheck.java
```diff
@@ -68,12 +70,45 @@ public final boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec
...
+            throw new IllegalStateException(e);
+        }
+    }
+
+    private static FilterExpressionPath getFilterExpressionPath(
+            Class<?> type,
+            String method,
+            EntityDictionary dictionary) throws NoSuchMethodException {
+        FilterExpressionPath path = dictionary.lookupEntityClass(type)
+                .getMethod(method)
```
aklish: This should work for fields as well.

wcekan: The only way to walk a path between beans is with a `@ComputedRelationship`

PR #735, opened 2019-01-09
Predicates

wcekan:
Created all the `XXPredicate` helper classes for each `Operation` type to make Filter expression checks cleaner to build, and provide type checking.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,8 @@
 # Change Log
+## 4.2.14
+**Features**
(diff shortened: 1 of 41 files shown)


changelog.md
```diff
@@ -1,4 +1,8 @@
 # Change Log
+## 4.2.14
```
aklish: We haven't released 4.2.13 yet.

wcekan: Updated

PR #736, opened 2019-01-09
Fixed generic field handling

### LINKED ISSUE #733: entity with generic type field
If entity has generic parameterized type field, elide create wrong mapping.

for parameterized type field, elide maps to java.lang.Object.

And on entityBinding we'll get IllegalArgumentException because I haven't entity class Object.

It happen's because getFieldType return class by getType() and getReturnType() methods, which return Object for generic types.

aklish:
Fixes #733.

There is a restructure here to move to a common artifact for both unit & integration test models (elide-example-models).

There will be another PR that adds a new DSL for Rest Assured Testing for JSON-API.  Once that is added, this PR will add integration tests based on the new DSL.

### DIFF
--- a/elide-example-models/src/main/java/com/yahoo/elide/models/generics/Employee.java
+++ b/elide-example-models/src/main/java/com/yahoo/elide/models/generics/Employee.java
@@ -11,6 +11,9 @@ import com.yahoo.elide.models.BaseId;
 
 import javax.persistence.Entity;
 
+/**
+ * Helper class to test parameterized subclass/superclass hierarchies.
+ */
 @Include(rootLevel = true)
 @Entity
 public class Employee extends BaseId {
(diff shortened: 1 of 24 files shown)


DennisMcWherter:
Looks good when tests are added :+1:

elide-example-models/src/main/java/com/yahoo/elide/models/generics/Employee.java
```diff
@@ -0,0 +1,20 @@
...
+import com.yahoo.elide.models.BaseId;
+
+import javax.persistence.Entity;
+
+/**
+ * Helper class to test parameterized subclass/superclass hierarchies.
+ */
+@Include(rootLevel = true)
+@Entity
+public class Employee extends BaseId {
```
DennisMcWherter: I really like this package. Can we add a README.md that contains a brief description? To me this looks like a more complete model rather than what we have today (a bunch of disparate models intended to test specific functionality)

PR #737, opened 2019-01-23
Added Codahale InstrumentedFilter & corresponding metrics, threads, a…

aklish:
…nd admin servlets as settings options for Elide Standalone

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,9 @@
 # Change Log
+
+## 4.2.14
+**Features**
+ * Added [Codahale InstrumentedFilter](https://metrics.dropwizard.io/3.1.0/manual/servlet/) & corresponding metrics, threads, admin servlets as a setting option for Elide Standalone.
+
 ## 4.2.13
 **Features**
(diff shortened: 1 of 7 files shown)


jkusa:
@aklish can we document this in the README? Maybe put something about `/stats/metrics` and a link to https://metrics.dropwizard.io/3.1.0/manual/servlet/.

aklish:
> Changes look good to me. General question, however. Should we provide a mechanism (similar to [ElideStandaloneSettings#getFilters](https://github.com/yahoo/elide/blob/365b635496edf6940671240189ad53690f1ec2e0/elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java#L159-L161)) to allow users to customize the `context`?

I believe this already exists:
```    /**
     * A hook to directly modify the jetty servlet context handler as necessary.
     *
     * @param servletContextHandler ServletContextHandler in use by Elide standalone.
     */
    default void updateServletContextHandler(ServletContextHandler servletContextHandler) {
        // Do nothing
    }```

PR #738, opened 2019-01-25
Replace jcabi-mysql-maven-plugin with H2 for testing

wcekan:
Replacing a deprecated `jcabi-mysql-maven-plugin` with H2 for database testing.

PR #739, opened 2019-01-26
Update surefire and failsafe versions

wcekan:
It seems Hibernate 5 tests were not running due to test version incompatibility.

```
-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Results :
Tests run: 0, Failures: 0, Errors: 0, Skipped: 0
```

Spun this instead of addressing in PR #738


elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -65,6 +64,7 @@
      */
     protected HibernateTransaction(Session session, boolean isScrollEnabled, ScrollMode scrollMode) {
         this.session = session;
+        session.setHibernateFlushMode(FlushMode.COMMIT);
```
aklish: Can we add a comment here on why.

wcekan: Done.

PR #747, opened 2019-02-14
Implement JPA DataStore

wcekan:
Rebased from #651

DataStore with only javax.persistence dependency.
Can be used with any JEE container.
Supports both JTA and Non-JTA transactions.


wcekan:
@belovaf @aklish Please review the rebase and merge.  

For the next step I would like to try to move the common code to shared common superclasses that will support any JPA library.  Hibernate 3 and 5 should simply subclass and add the required wrappers.

elide-datastore/elide-datastore-jpa/pom.xml
```diff
@@ -0,0 +1,191 @@
...
+    <modelVersion>4.0.0</modelVersion>
+    <artifactId>elide-datastore-jpa</artifactId>
+    <packaging>jar</packaging>
+    <name>Elide Data Store: JPA</name>
+    <description>Elide Data Store for JPA support</description>
+    <url>https://github.com/yahoo/elide</url>
+    <parent>
+        <groupId>com.yahoo.elide</groupId>
+        <artifactId>elide-datastore-parent-pom</artifactId>
+        <version>4.3.1-SNAPSHOT</version>
```
aklish: This needs to be updated to 4.3.3

elide-datastore/elide-datastore-jpa/pom.xml
```diff
@@ -0,0 +1,191 @@
...
+    <dependencies>
+        <!-- Elide -->
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-datastore-hibernate</artifactId>
+            <version>4.3.1-SNAPSHOT</version>
```
aklish: Same comment.

elide-datastore/elide-datastore-jpa/pom.xml
```diff
@@ -0,0 +1,191 @@
...
+            <artifactId>javax.transaction-api</artifactId>
+            <version>1.2</version>
+            <scope>provided</scope>
+            <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-integration-tests</artifactId>
+            <version>4.3.1-SNAPSHOT</version>
```
aklish: Same

elide-datastore/elide-datastore-jpa/src/test/java/com/yahoo/elide/datastores/jpa/JpaDataStoreSupplier.java
```diff
@@ -0,0 +1,97 @@
...
+    private static final String JDBC = "jdbc:h2:mem:root;IGNORECASE=TRUE";
+    private static final String ROOT = "root";
+
+    @Override
+    public DataStore get() {
+        Map<String, Object> options = new HashMap<>();
+        ArrayList<Class<?>> bindClasses = new ArrayList<>();
+
+        try {
+            bindClasses.addAll(ClassScanner.getAnnotatedClasses(Parent.class.getPackage(), Entity.class));
```
aklish: I think you'll need more than this now.  This only covers one package.  You will need to include elide-example-models as well.

aklish:
We should switch Elide-Standalone to use this along with the H2 in memory DB.

elide-datastore/elide-datastore-jpa/pom.xml
wcekan: Done in the merge.

PR #759, opened 2019-03-01
Issue758

aklish:
- For JSON-API, the preCommit life cycle hooks will now run prior to response serialization.
- This work has not been done for GraphQL.  That is a more complex change given the lack of control over the graphQL parser.
- Added a new mechanism to add dependency injection to Elide.  Ultimately, the hope is that the same mechanism can be used to inject checks (as well as models).
- Made IT tests support dependency injection
- Added a new package of example/test models for testing lifecycle hooks.  This has a canonical example of how to inject a service and invoke on an lifecycle hook.


elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -694,6 +712,8 @@ public String getNameFromAlias(Object entity, String alias) {
             Initializer<T> initializer = getEntityBinding(entity.getClass()).getInitializer();
             if (initializer != null) {
                 initializer.initialize(entity);
+            } else if (injector != null) {
```
wcekan: @aklish what do we expect if both are defined?

elide-example-models/src/main/java/com/yahoo/elide/models/triggers/Invoice.java
```diff
@@ -0,0 +1,61 @@
...
+    @Transient
+    @Exclude
+    @Inject
+    private BillingService billingService;
+
+    private boolean complete = false;
+
+    private long total = 0;
+
+
```
jkusa: `\n`

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
aklish: The initializer overrides the injector.  There is only one way to initialize something - either a general injector or a class specific initializer.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -80,13 +82,29 @@
...
+     * Instantiate a new EntityDictionary with the provided set of checks and an injection function.
+     * In addition all of the checks * in {@link com.yahoo.elide.security.checks.prefab} are mapped
+     * to {@code Prefab.CONTAINER.CHECK} * (e.g. {@code @ReadPermission(expression="Prefab.Role.All")}
+     * or {@code @ReadPermission(expression="Prefab.Common.UpdateOnCreate")})
+     * @param checks a map that links the identifiers used in the permission expression strings
+     *               to their implementing classes
+     * @param injector a function typically associated with a dependency injection framework that will
+     *                 initialize Elide models.
+     */
+    public EntityDictionary(Map<String, Class<? extends Check>> checks, Initializer injector) {
```
jkusa: Can we make this `Injector injector` to make the distinction clear.

PR #765, opened 2019-03-20
Issue763

### LINKED ISSUE #763: Support Filtering on Computed Attributes

aklish:
Adds in-memory filtering & sorting capability for computed attributes/relationships.

Fixes #763


elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -210,12 +212,17 @@ public static Operator fromString(final String string) {
...
         return (T entity) -> {
-            Class type = requestScope.getDictionary().getType(entity, field);
-            if (!type.isAssignableFrom(String.class)) {
+            Object fieldValue = getFieldValue(entity, field, requestScope);
+
+            if (fieldValue == null) {
+                return false;
+            }
+
+            if (!fieldValue.getClass().isAssignableFrom(String.class)) {
```
wcekan: Should we just use `fieldValue .toString()` the fieldValue so you can?

PR #775, opened 2019-03-29
Fixes #774

### LINKED ISSUE #774: Elide returns wrong date parsing format in 400 error for non-default SimpleDateFormat.
```
    @Override
    public Date deserialize(String val) {
        Date date;
        try {
            date = df.parse(val);
        } catch (java.text.ParseException e) {
            throw new IllegalArgumentException("Date strings must be formated as yyyy-MM-dd'T'HH:mm'Z'");
        }
```

should instead be:

```
    @Override
    public Date deserialize(String val) {
        Date date;
        try {
            date = df.parse(val);
        } catch (java.text.ParseException e) {
            throw new IllegalArgumentException("Date strings must be formated as " + df.toPattern());
        }
```

aklish:


elide-core/src/main/java/com/yahoo/elide/utils/coerce/converters/ISO8601DateSerde.java
```diff
@@ -17,7 +16,7 @@
  */
 public class ISO8601DateSerde implements Serde<String, Date> {
 
-    protected DateFormat df;
+    protected SimpleDateFormat df;
```
wcekan: This is dangerous to reuse.  `SimpleDateFormat` is not thread safe.

elide-core/src/main/java/com/yahoo/elide/utils/coerce/converters/ISO8601DateSerde.java
```diff
@@ -49,7 +48,7 @@ public Date deserialize(String val) {
         try {
             date = df.parse(val);
```
wcekan: You will need to serialize here if you want to reuse the `SimpleDateFormat`

elide-core/src/main/java/com/yahoo/elide/utils/coerce/converters/ISO8601DateSerde.java
```diff
@@ -17,15 +17,15 @@
...
+    protected FastDateFormat df;
     protected Class<? extends Date> targetType;
 
     public ISO8601DateSerde(SimpleDateFormat df) {
         this(df, Date.class);
     }
 
     public ISO8601DateSerde(SimpleDateFormat df, Class<? extends Date> targetType) {
-        this.df = df;
+        this.df = FastDateFormat.getInstance(df.toPattern(), df.getTimeZone());
```
wcekan: why not just do
```java
this(df.toPattern(), df.getTimeZone(), targetType);
```

PR #780, opened 2019-04-09
Add support for JPA @MapsId annotation

gmckerrell:
If you have a JPA model where one entity has an ID which is dictated by the value in another entity (mapped via the JPA `@MapsId` annotation) then ELIDE needs to not force a creating POST request to contain an ID.
e.g.[https://www.concretepage.com/hibernate/example-mapsid-hibernate](https://www.concretepage.com/hibernate/example-mapsid-hibernate)

This PR adds support for the `@MapsId` annotation in entities. Without this you are unable to create such a dependent entity using ELIDE (unless you provide a dummy ID which gets discarded, and which is an incorrect use of JSON:API)

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -1893,7 +1893,8 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
 
         //If id field is not a `@GeneratedValue` or mapped via a `@MapsId` attribute
         //then persist the provided id
-        if (!(persistentResource.isIdGenerated() || persistentResource.isIdMapped())) {
+        //if (!(persistentResource.isIdGenerated() || persistentResource.isIdMapped())) {
+        if (!persistentResource.isIdGenerated()) {
             if (id != null && !id.isEmpty()) {
                 persistentResource.setId(id);
             } else {
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -69,8 +69,6 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
-import javax.persistence.GeneratedValue;
-import javax.persistence.MapsId;
 import javax.ws.rs.WebApplicationException;
 
 /**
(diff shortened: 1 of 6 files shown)


yahoocla:

gmckerrell:
CLA signed

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -755,6 +756,27 @@ public boolean isIdGenerated() {
...
     }
 
+    /**
+     * Indicates if the ID is mapped to another entity or not.
+     *
+     * @return Boolean
+     */
+    public boolean isIdMapped() {
+        final Class<?> cls = getObject().getClass();
+        for (final AccessibleObject obj : dictionary.getEntityBinding(cls).getAllFields()) {
```
aklish: Rather than walking the attributes this during the request processing, I think this logic could move to the `EntityBInding` class.  The binding can store properties about the id (if it is generated or mapped) - and all of this would be computed once during server init.

The entity binding walks each field or relationship when an entity is bound.  We could check for annotations like GeneratedValue and MapsId there.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1869,8 +1891,9 @@ private static RuntimeException handleInvocationTargetException(InvocationTarget
      */
     private static void assignId(PersistentResource persistentResource, String id) {
 
-        //If id field is not a `@GeneratedValue` persist the provided id
-        if (!persistentResource.isIdGenerated()) {
+        //If id field is not a `@GeneratedValue` or mapped via a `@MapsId` attribute
+        //then persist the provided id
+       if (!(persistentResource.isIdGenerated() || persistentResource.isIdMapped())) {
```
aklish: 👍

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
gmckerrell: make sense 👍

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1869,8 +1862,9 @@ private static RuntimeException handleInvocationTargetException(InvocationTarget
      */
     private static void assignId(PersistentResource persistentResource, String id) {
 
-        //If id field is not a `@GeneratedValue` persist the provided id
-        if (!persistentResource.isIdGenerated()) {
+        //If id field is not a `@GeneratedValue` or mapped via a `@MapsId` attribute
+        //then persist the provided id
+       if (!persistentResource.isIdGenerated()) {
```
wcekan: Tab size?

PR #781, opened 2019-04-09
Fixing bug in InMemory loadObject implementation

aklish:


wcekan:
Thanks, that fixed my issue with the in-memory store.

PR #784, opened 2019-04-11
loadObject should return unique result

wcekan:
If the default datastore `loadObject` fails to filter to a single result, do not return any.

Currently the first element of the collection would erroneously be returned.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,5 @@
 # Change Log
+<<<<<<< HEAD
 
 ## 5.0.9
 **Fixes**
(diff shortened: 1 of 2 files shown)


elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -142,7 +142,10 @@ default Object loadObject(Class<?> entityClass,
                 scope);
         Iterator<Object> it = results == null ? null : results.iterator();
         if (it != null && it.hasNext()) {
-            return it.next();
+            Object obj = it.next();
+            if (!it.hasNext()) {
+              return obj;
```
aklish: Does this fix an actual issue?

aklish: Maybe throw an exception instead of returning null.

wcekan: The contract is for `loadObject` to return null if no object is loaded.

aklish (approved):

PR #785, opened 2019-04-11
Flush once for patch extension

wcekan:
Flush is required before serialization to properly populate `@Generated` fields.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,8 @@
 # Change Log
+## 4.4.1
+**Fixes**
+ * Flush once for patch extension
+
 ## 4.4.0
 **Features**
  * Issue#763 Support for filtering & sorting on computed attributes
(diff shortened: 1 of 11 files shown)


changelog.md
```diff
@@ -8,6 +8,7 @@
  * Throw proper exception on invalid PersistentResource where id=null
  * Issue#744 Elide returns wrong date parsing format in 400 error for non-default DateFormats
  * Enable RSQL filter dialect by default (in addition to the default filter dialect).
+ * Flush once for patch extension
```
aklish: Make for 4.4.1

elide-integration-tests/src/test/java/example/Generate.java
```diff
@@ -0,0 +1,43 @@
...
+
+import javax.persistence.Column;
+import javax.persistence.Entity;
+import javax.persistence.Id;
+import javax.persistence.Temporal;
+import javax.persistence.TemporalType;
+
+@Entity
+@Include(rootLevel = true)
+public class Generate {
```
aklish: Should we move this to elide-example-models?  Or do you think there is a case for keeping these kinds of tests in integration-tests?

wcekan: `@Generated` annotation is not available to ` elide-example-models`

PR #786, opened 2019-04-11
Propagate ConstraintViolationExceptions on flush for JPA Transaction.

aklish:


elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/transaction/AbstractJpaTransaction.java
```diff
@@ -92,7 +93,11 @@ public void flush(RequestScope requestScope) {
             } finally {
                 log.error("Caught entity manager exception during flush", e);
             }
-            throw new TransactionException(e);
+            if (e instanceof ConstraintViolationException) {
+                throw e;
+            } else {
+                throw new TransactionException(e);
+            }
```
wcekan:
```java
if (e instanceof ConstraintViolationException) {
    throw e;
}
throw new TransactionException(e);
```

PR #788, opened 2019-04-12
Migrated Elide Standalone to JPA Data Store

aklish:


elide-standalone/src/main/java/com/yahoo/elide/standalone/PersistenceUnitInfoImpl.java
```diff
@@ -0,0 +1,58 @@
...
+    @Override
+    public boolean excludeUnlistedClasses() {
+        return false;
+    }
+
+    public void addTransformer(ClassTransformer classTransformer) {
+        //NOOP
+    }
+
+    private ClassLoader newTempClassLoader;
```
jkusa: Is there a reason this is at the bottom of the file?

aklish: That was just the order the methods were defined that I converted to properties.  I fixed on another commit.

PR #805, opened 2019-05-07
Create ISSUE_TEMPLATE.md

jkusa:
- Create ISSUE_TEMPLATE.md


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,33 @@
+<!--- Provide a general summary of the issue in the Title above -->
+
+## Expected Behavior
+<!--- If you're describing a bug, tell us what should happen -->
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
```
clayreimann: I think we should under each of the behavior sections we should remind people to use code blocks

Issue #801, opened 2019-05-03
Filter by ID fails if the ID field name doesn't equal 'id'.


aklish:
In Elide, anything annotated with `@Id` is considered the ID of an entity - even if the name of the field is something other than 'id'.

However, the `EntityDictionary` `getType` and `getParameterizedType` methods only look at the Java field name.  

When the requested field name equals 'id', we ought to look for the field with the annotation `@Id` rather than looking by field name.

aklish:
Filtering by ID will give an error like:

```{"errors":["InvalidPredicateException: Invalid filter format: filter[book]\nInvalid query parameter: filter[book]\nInvalid filter format: filter[book]\nNo such association id for type book"]}```

aklish:
Fixed

PR #809, opened 2019-05-08
Support custom ID field name

QubitPi:
Issue
====

https://github.com/yahoo/elide/issues/801

Doc Changes
==========

https://github.com/yahoo/elide-doc/pull/82


elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -603,12 +604,19 @@ public RelationshipType getRelationshipType(Object entity, String relation) {
 
     /**
      * Get a type for a field on an entity.
+     * <p>
+     * If the field is "id" and the ID name of the identity is not "id", this method returns the type of the field that
+     * is annotated by {@link @Id}.
```
aklish: I think your comments are breaking the javadoc generation.  My guess is the second `@Id`.  Travis build fails for this PR.

QubitPi: Thank you! Let me remove it

elide-core/src/test/java/example/IdFieldAnnotations.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.persistence.Id;
+import javax.persistence.ManyToOne;
+import javax.persistence.OneToMany;
+
+/**
+ * A test entity that has an ID field whose field name is not "id".
+ */
+@Include(rootLevel = true)
+@Entity
+public class IdFieldAnnotations {
```
aklish: Two comments.  Probably only one of them needs to be addressed:
1. Can we remove all of the fields in this class that are needed for the test?
2. Instead of making a new test bean, could we use an existing one and just change the `@Id` field to a different name?

QubitPi: Sure. That sounds good.

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -327,7 +327,7 @@ private void bindAttrOrRelation(AccessibleObject fieldOrMethod) {
         String fieldName = getFieldName(fieldOrMethod);
         Class<?> fieldType = getFieldType(entityClass, fieldOrMethod);
 
-        if (fieldName == null || "id".equals(fieldName) || "class".equals(fieldName)
+        if (fieldName == null || "class".equals(fieldName)
```
QubitPi: Since we can have ID fields that are not "id", this implies that some customer could have an ID field called "jobId" and another non-ID field called "id". For that reason, I removed "id" from reserved field lisst

aklish: "id" is reserved in JSON-API.   From the JSON-API spec:
> In other words, a resource can not have an attribute and relationship with the same name, nor can it have an attribute or relationship named type or id.

Based on this, the original code is correct.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -603,16 +604,120 @@ public RelationshipType getRelationshipType(Object entity, String relation) {
 
     /**
      * Get a type for a field on an entity.
+     * <p>
+     * If this method is called on a bean such as the following
```
aklish: Very nice documentation

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -603,16 +604,120 @@ public RelationshipType getRelationshipType(Object entity, String relation) {
...
+     * </pre>
+     * If, however, the ID field is not "id" and there is a non-ID field called "id", i.e.
+     * <pre>
+     * {@code
+     * public class Address {
+     *
+     *     {@literal @}Id
+     *     private Long surrogateKey
+     *
+     *     private String id;
```
aklish: This last case is not allowed in JSON-API.   ID is a reserved keyword.  There can never be a separate attribute called ID.   In fact, we should probably throw an exception during entity binding if this case is ever found.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -603,16 +604,120 @@ public RelationshipType getRelationshipType(Object entity, String relation) {
...
+     * Returns whether or not a type-checked field is an ID of the bean.
+     * <p>
+     * The ID field of the bean is annotated by JAP {@code {@literal @}Id} annotation.
+     *
+     * @param entityClass Entity class
+     * @param identifier  Field to lookup type
+     *
+     * @return {@code true} if the field is an ID field
+     */
+    public boolean isIdType(Class<?> entityClass, String identifier) {
```
aklish: If we remove the above scenario (ID as an attribute), I believe we can revert to your earlier commit where this function didn't exist and we instead used the existing isIdType function in the EntityBinding.

elide-core/src/test/java/example/Job.java
```diff
@@ -37,4 +37,7 @@
     @MapsId
     @Getter @Setter
     private Parent parent;
+
+    @Getter @Setter
+    private String id;
```
aklish: Should not be possible.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
QubitPi: Thanks. I will do that.

I believe that previous check `!ID_IDENTIFIER.equals(getEntityBinding(entityClass).getIdFieldName())` has a bug when id is not "id" but user is actually filtering on a non-ID field. The new helper method will fix that bug

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
QubitPi: Thanks. Influenced by Rick 😉

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
QubitPi: Removed

elide-core/src/test/java/example/Job.java
QubitPi: Removed

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
QubitPi: Restored

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -603,16 +604,94 @@ public RelationshipType getRelationshipType(Object entity, String relation) {
...
+     * </pre>
+     * JSON-API spec does not allow "id" as non-ID field name. If, therefore, there is a non-ID field called "id",
+     * callling this method has undefined behavior
      *
      * @param entityClass Entity class
      * @param identifier  Field to lookup type
      * @return Type of entity
      */
     public Class<?> getType(Class<?> entityClass, String identifier) {
+        if (isIdType(entityClass, identifier)) {
```
aklish: In your first commit, you had code like:
```java
        if (!ID_IDENTIFIER.equals(getEntityBinding(entityClass).getIdFieldName())) {
            return getIdType(entityClass);
        }```

Is there a reason we have to define a new method `isIdType` or could we use the code from the first commit?

Comment (QubitPi) [reply elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java]: yes, with first commit, if `identifier` is a non-ID field and the bean ID is not "id", the first commit returns the type of ID while we are asking for the type of a non-ID field instead

The first commit also has undefined behavior in tests on Beans which do not have ID defined such as `Address.java`. The new method handles that case gracefully.

I can figure out a way to remove the new method if that's needed

Comment (aklish) [reply elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java]: How about:

```java
if (identifier.equals(ID_IDENTIFIER) {
  return getEntityBinding(entityClass).getIdFieldType();
}
```

QubitPi: Yeah looks simpler.

PR #810, opened 2019-05-08
Update README

QubitPi:

## Description

[Elide standalone startup throws unchecked exception](https://github.com/yahoo/elide/blob/master/elide-standalone/src/main/java/com/yahoo/elide/standalone/ElideStandalone.java#L81) which hasn't been captured in doc yet

A docker image has been introduced to give new users a quicker and easier experience of Elide (If core team like it)

### DIFF
--- a/elide-standalone/README.md
+++ b/elide-standalone/README.md
@@ -33,7 +33,7 @@ To actually start your Elide application, add the following to your main method:
 
 ```java
 public class Main {
-  public static void main(String[] args) {
+  public static void main(String[] args) throws Exception {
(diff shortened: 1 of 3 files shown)


elide-standalone/README.md
```diff
@@ -33,7 +33,7 @@ To actually start your Elide application, add the following to your main method:
 
 ```java
 public class Main {
-  public static void main(String[] args) {
+  public static void main(String[] args) throws Exception {
```
DennisMcWherter: Rather than adding the unchecked exception to the signature, I would suggest adding it to the javadoc.

Same comment for the others.

Issue #812, opened 2019-05-10
Support for BigDecimal in GraphQL API


arpradhan:
Hello, our team would like to enable the GraphQL on our Elide project, but one of our models has a BigDecimal field which is not supported in Elide. 

Would it be possible to add BigDecimal support in GraphQLConversionUtils#classToScalarType?
https://github.com/yahoo/elide/blob/b76f50409c253cb18f6ace450e3eae42e123da37/elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java#L61

It appears that java-graphql has support for BigDecimal: https://github.com/graphql-java/graphql-java/blob/1e12da93a692b8fff549e417906c899c9bf21345/src/main/java/graphql/Scalars.java#L528

I can make a PR to add this change if it helps.

aklish:
Happy to review the PR if you want to submit one.  It looks like this should be fairly straightforward.

PR #813, opened 2019-05-13
Add graphql support for BigDecimal

arpradhan:
Resolves https://github.com/yahoo/elide/issues/812

## Description
This change adds support for BigDecimal in the GraphQL API by mapping it to the GraphQLBigDecimal class.

## Motivation and Context
This change allows models with BigDecimal fields to be properly converted to GraphQLBigDecimal, when using the GraphQL API. Currently, BigDecimal gets converted to an Object type instead of a Scalar.

## How Has This Been Tested?
- Added BigDecimal field `weightLbs` to the Book model in `elide-graphql/src/test/java/example/Book.java`.
- Added assertion to ModelBuilderTest to assert that `weightLbs` on the Book model gets converted to a GraphQLBigDecimal class.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,8 @@
 # Change Log
+## 4.4.5
+**Features**
+* Issue#812 Add support for BigDecimal field in GraphQL. 
+
 ## 4.4.4
 **Fixes**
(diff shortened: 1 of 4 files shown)


changelog.md
```diff
@@ -1,4 +1,8 @@
 # Change Log
+## 4.4.5
```
aklish: Should add as a feature under 4.4.4.

arpradhan: ah sorry, fixing it

arpradhan: See that you fixed the changelog, thanks

Issue #800, opened 2019-05-03
Support RSQL INFIX, POSTFIX, and PREFIX filters on number types.


aklish:
Elide currently can't support filters like:
`filter[books]=id==*20*`

It will return 
`{"errors":["InvalidValueException: Invalid value: “*20*"]}`

Inside RSQLFilterDialect, Elide attempts to type coerce the arguments of the predicate to the correct type:

```Java
            //Coerce arguments to their correct types
            List<Object> values = arguments.stream()
                    .map((argument) -> (Object) CoerceUtil.coerce(argument, relationshipType))
                    .collect(Collectors.toList());
```
The type coercion logic is happening too soon.  We should first strip off any asterisk characters (depending on the operator).

QubitPi:
@aklish Stripping off any asterisk reminds me about other regex characters. Asterisk can be generalized to any character in a valid regex expression. If we agree on that, here is my strategy for fixing this issue:

I assume there is absolutely no way to coerce `*20*` to a `long` type. If we could treat `*20*` a regex, we will coerce `*20*` to `String` instead of `Long`.

Implementation-wise, we will check if `argument` is a regex string, if it is, override `relationshipType` to `String`

clayreimann:
I disagree. If the id type here is numeric then `*20*` is not a valid filter. If Id is a string type then I don’t see why this wouldn’t work. 

Are you proposing that instead we should allow `=id==*20*` to match ids like `200` and `320`?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

aklish:
The asterisk character is part of the RSQL grammar.  When the RSQLFilterDialect finds tokens like '*20*', it should remove the asterisks before extracting the value (20 in this case).  We don't need any type coercion.   We just need some special logic inside RSQLFilterDialect to handle asterisk correctly.   Does this make sense?

PR #814, opened 2019-05-13
Support RSQL INFIX, POSTFIX, and PREFIX filters on number types.

QubitPi:
Resolves https://github.com/yahoo/elide/issues/800

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,10 @@
 # Change Log
 
+## 4.4.5
+**Fixes**
+ * Support RSQL INFIX, POSTFIX, and PREFIX filters on number types: remove '*' before coercing.
+
+
 ## 4.4.4
 **Fixes**
  * When requesting an ID field whose name is not 'id', an error happens: `No such association id for type xxx`. When the requested field name equals 'id', Elide has been fixed to look for the field with the annotation @Id rather than looking by field name.
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## 4.4.5
+## 4.4.4
 **Fixes**
  * Support RSQL INFIX, POSTFIX, and PREFIX filters on number types: remove '*' before coercing.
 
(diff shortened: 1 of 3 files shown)


elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -279,6 +279,13 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
                     .map(Path.PathElement::getFieldType)
                     .orElseThrow(() -> new IllegalStateException("Path must not be empty"));
 
+            if (Number.class.isAssignableFrom(relationshipType)) {
+                //Support number types
+                for (int i = 0; i < arguments.size(); i++) {
+                    arguments.set(i, arguments.get(i).replace("*", ""));
```
aklish: Stripping this here means we lose the context of what operator is being applied.  We need to know if this is startsWith, endsWith, contains, or equals.  The position of the asterisk (beginning, end, or both beginning and end) defines which operator applies.

changelog.md
```diff
@@ -1,5 +1,10 @@
 # Change Log
 
+## 4.4.5
```
aklish: 4.4.4

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -281,6 +281,11 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
 
             //Coerce arguments to their correct types
             List<Object> values = arguments.stream()
+                    .map(argument ->
```
aklish: This doesn't look correct.  Take a look at the `equalityExpression` function below.  It needs the asterisks to determine the correct operator.  By stripping them here, when `equalityExpression` is called, it will always map to equals.

Issue #825, opened 2019-05-23
Create AggregationDataStore module


QubitPi:
We need to add scaffolding first to implement AggregationDataStore

## Expected Behavior

A sub-module in elide parent directory

## Current Behavior

N/A

## Possible Solution

Create a new maven sub-module

## Context

N/A

PR #827, opened 2019-05-23
Create AggregationDataStore module

QubitPi:
Issue https://github.com/yahoo/elide/issues/825


elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+    </scm>
+
+    <dependencies>
+        <!-- Elide (including integration tests) -->
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
```
aklish: So for this store, I think we will not want to run the stock integration tests.  The tests will be very different from other stores - so we can remove this dependency.

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+            <type>test-jar</type>
+            <scope>test</scope>
+    	</dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <!-- Integration test -->
```
aklish: Let's remove all the test dependencies to start and add them in as needed.

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -0,0 +1,123 @@
+<!--
```
aklish: You will need to update the elide-datastore/pom.xml to add a module for this new build component.

aklish:
Can you open this PR against the AggregationDataStore branch instead of master?

Issue #819, opened 2019-05-21
AggregationDataStore: Having Clause Support


aklish:
Filter expressions must be split into those that apply to metric aggregations (having clauses) and those that apply to dimensions (where clauses).

A single filter expression is passed to the datastore, but not all expression conjunctions (notably OR) can be split into a having and where clause in a meaningful way.

## Expected Behavior
An expression parser should throw an error whenever an illegal expression is encountered.  Here are examples of legal & illegal expressions.  Clauses prefixed with H are having clauses.  Clauses prefixed with W are where clauses.

### Legal
```
Expression: H1 AND W1 AND H2 
SQL: WHERE W1 HAVING (H1 AND H2)
```
```
Expression: (H1 OR H2) AND W1
SQL: WHERE W1 HAVING (H1 OR H2)
```

### Illegal
* H1 OR W1
* (W1 AND H1) OR W2

The parser should either:

- Split the filter expression into a *where* expression and a *having* expression.
- Throw an exception

## Possible Solution
Create a visitor that will walk the `FilterExpression` abstract syntax tree and split the filter expression into where and having clauses.

Prior to walking the syntax tree, it should leverage the `FilterExpressionNormalizationVisitor` to transform expressions like:
`NOT (A AND B)` into `NOT A OR NOT B`

The visitor will be leveraged by the AggregationStore to construct the JPQL query.

## Context

https://github.com/yahoo/elide/issues/804

aklish:
Instead of throwing an exception, the illegal combinations can be promoted to a having clause.

QubitPi:
@aklish When we say `Instead of throwing an exception, the illegal combinations can be promoted to a having clause.`, do we secretly throw away any `WHERE` expression?

PR #840, opened 2019-05-28
AggregationDataStore: Having Clause Support

QubitPi:
Issue https://github.com/yahoo/elide/issues/819

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
@@ -17,6 +17,28 @@ public class AndFilterExpression implements FilterExpression {
     @Getter private FilterExpression left;
     @Getter private FilterExpression right;
 
+    /**
+     * Returns a new {@link AndFilterExpression} instance with the specified null-able left and right operands.
+     * <p>
+     * If both left and right are not {@link null}, this method produces the same instance as
+     * {@link #AndFilterExpression(FilterExpression, FilterExpression)} does. If only one of them is {@link null}, the
+     * other non-null is returned with no modification. If both left and right are {@link null}, this method returns
+     * {@code null}.
+     *
+     * @param left  The provided left {@link FilterExpression}
+     * @param right  The provided right {@link FilterExpression}
+     *
+     * @return a new {@link AndFilterExpression} instance or {@code null}
+     */
+    public static FilterExpression and(FilterExpression left, FilterExpression right) {
+        if (left != null && right != null) {
+            return new AndFilterExpression(left, right);
+        } else if (left == null) {
+            return right;
+        }
+        return left;
+    }
+
     public AndFilterExpression(FilterExpression left, FilterExpression right) {
         this.left = left;
         this.right = right;
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
@@ -20,17 +20,21 @@ public class AndFilterExpression implements FilterExpression {
     /**
      * Returns a new {@link AndFilterExpression} instance with the specified null-able left and right operands.
      * <p>
-     * If both left and right are not {@link null}, this method produces the same instance as
-     * {@link #AndFilterExpression(FilterExpression, FilterExpression)} does. If only one of them is {@link null}, the
-     * other non-null is returned with no modification. If both left and right are {@link null}, this method returns
-     * {@code null}.
+     * The publication rules are
+     * <ol>
+     *     <li> If both left and right are not {@code null}, this method produces the same instance as
+     *          {@link #AndFilterExpression(FilterExpression, FilterExpression)} does,
+     *     <li> If one of them is {@code null}, the other non-null is returned with no modification,
+     *     <li> If both left and right are {@code null}, this method returns
+     *          {@code null}.
+     * </ol>
      *
      * @param left  The provided left {@link FilterExpression}
      * @param right  The provided right {@link FilterExpression}
      *
      * @return a new {@link AndFilterExpression} instance or {@code null}
      */
-    public static FilterExpression and(FilterExpression left, FilterExpression right) {
+    public static FilterExpression fromPair(FilterExpression left, FilterExpression right) {
         if (left != null && right != null) {
             return new AndFilterExpression(left, right);
         } else if (left == null) {
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/OrFilterExpression.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/OrFilterExpression.java
@@ -18,6 +18,28 @@ public class OrFilterExpression implements FilterExpression {
     @Getter private FilterExpression left;
     @Getter private FilterExpression right;
 
+    /**
+     * Returns a new {@link OrFilterExpression} instance with the specified null-able left and right operands.
+     * <p>
+     * If both left and right are not {@link null}, this method produces the same instance as
+     * {@link #OrFilterExpression(FilterExpression, FilterExpression)} does. If only one of them is {@link null}, the
+     * other non-null is returned with no modification. If both left and right are {@link null}, this method returns
+     * {@code null}.
+     *
+     * @param left  The provided left {@link FilterExpression}
+     * @param right  The provided right {@link FilterExpression}
+     *
+     * @return a new {@link OrFilterExpression} instance or {@code null}
+     */
+    public static FilterExpression or(FilterExpression left, FilterExpression right) {
+        if (left != null && right != null) {
+            return new OrFilterExpression(left, right);
+        } else if (left == null) {
+            return right;
+        }
+        return left;
+    }
+
     public OrFilterExpression(FilterExpression left, FilterExpression right) {
         this.left = left;
         this.right = right;
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/OrFilterExpression.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/expression/OrFilterExpression.java
@@ -21,17 +21,21 @@ public class OrFilterExpression implements FilterExpression {
     /**
      * Returns a new {@link OrFilterExpression} instance with the specified null-able left and right operands.
      * <p>
-     * If both left and right are not {@link null}, this method produces the same instance as
-     * {@link #OrFilterExpression(FilterExpression, FilterExpression)} does. If only one of them is {@link null}, the
-     * other non-null is returned with no modification. If both left and right are {@link null}, this method returns
-     * {@code null}.
+     * The publication rules are
+     * <ol>
+     *     <li> If both left and right are not {@code null}, this method produces the same instance as
+     *          {@link #OrFilterExpression(FilterExpression, FilterExpression)} does,
+     *     <li> If one of them is {@code null}, the other non-null is returned with no modification,
+     *     <li> If both left and right are {@code null}, this method returns
+     *          {@code null}.
+     * </ol>
      *
      * @param left  The provided left {@link FilterExpression}
      * @param right  The provided right {@link FilterExpression}
      *
      * @return a new {@link OrFilterExpression} instance or {@code null}
      */
-    public static FilterExpression or(FilterExpression left, FilterExpression right) {
+    public static FilterExpression fromPair(FilterExpression left, FilterExpression right) {
         if (left != null && right != null) {
             return new OrFilterExpression(left, right);
         } else if (left == null) {
--- a/elide-datastore/elide-datastore-aggregation/pom.xml
+++ b/elide-datastore/elide-datastore-aggregation/pom.xml
@@ -39,7 +39,6 @@
     </scm>
 
     <dependencies>
-        <!-- Elide (including integration tests) -->
         <dependency>
             <groupId>com.yahoo.elide</groupId>
             <artifactId>elide-core</artifactId>
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
@@ -13,6 +13,7 @@ import com.yahoo.elide.core.filter.expression.FilterExpressionVisitor;
 import com.yahoo.elide.core.filter.expression.NotFilterExpression;
 import com.yahoo.elide.core.filter.expression.OrFilterExpression;
 import com.yahoo.elide.datastores.aggregation.annotation.MetricAggregation;
+import com.yahoo.elide.datastores.aggregation.annotation.MetricComputation;
 import com.yahoo.elide.parsers.expression.FilterExpressionNormalizationVisitor;
 
 import org.apache.commons.lang3.tuple.Pair;
@@ -56,7 +57,7 @@ import lombok.extern.slf4j.Slf4j;
  * {@link SplitFilterExpressionVisitor} is thread-safe and can be accessed by multiple threads at the same time.
  */
 @Slf4j
-public class SplitFilterExpressionVisitor implements FilterExpressionVisitor<WhereHaving> {
+public class SplitFilterExpressionVisitor implements FilterExpressionVisitor<FilterConstraints> {
 
     @Getter(value = AccessLevel.PRIVATE)
     private final EntityDictionary entityDictionary;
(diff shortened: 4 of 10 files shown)


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,247 @@
...
+                    new AndFilterExpression(
+                            extractWhereExpression(left),
+                            extractWhereExpression(right)
+                    ),
+                    null
+            );
+        } else if (isHavingExpression(left) && isWhereExpression(right)) {
+            // pure-H AND pure-W
+            return Pair.of(
+                    extractHavingExpression(left),
```
aklish: I believe these are swapped.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,247 @@
...
+                            extractHavingExpression(right)
+                    )
+            );
+        }    }
+
+    @Override
+    public Pair<FilterExpression, FilterExpression> visitNotExpression(NotFilterExpression expression) {
+        FilterExpression normalized = getNormalizationVisitor().visitNotExpression(expression);
+
+        if (normalized instanceof AndFilterExpression) {
```
aklish: This might be missing the not expression.  Normalizing a not can still return a not.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,247 @@
...
+
+        } else if (isWhereExpression(left) && isHavingExpression(right)) {
+            // pure-W AND pure-H
+            return Pair.of(
+                    extractWhereExpression(left),
+                    extractHavingExpression(right)
+            );
+        } else {
+            return Pair.of(
+                    AndFilterExpression.withLeftAndRight(
```
aklish: Isn't this the pure having case?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,247 @@
...
+     * @param filterPredicate  The terminal filter expression to check for
+     *
+     * @return {@code true} if the {@link FilterPredicate} is a HAVING clause
+     */
+    private boolean isHavingPredicate(final FilterPredicate filterPredicate) {
+        String entityName = filterPredicate.getEntityType().getName();
+        String fieldName = filterPredicate.getField();
+
+        try {
+            return getEntityDictionary().getFieldAnnotations(entityName, fieldName).stream()
```
aklish: There is a method in EntityDictionary called `getAttributeOrRelationAnnotation` that you can use here instead.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,247 @@
...
+     * <p>
+     * A {@link FilterPredicate} corresponds to a {@code HAVING} clause iff the predicate field has
+     * {@link MetricAggregation} annotation on it.
+     *
+     * @param filterPredicate  The terminal filter expression to check for
+     *
+     * @return {@code true} if the {@link FilterPredicate} is a HAVING clause
+     */
+    private boolean isHavingPredicate(final FilterPredicate filterPredicate) {
+        String entityName = filterPredicate.getEntityType().getName();
```
aklish: Instead of the entityName, get the entityClass (and see comment below).

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,247 @@
...
+ * Pair<FilterExpression, FilterExpression> filterPair = filterExpression.accept(splitFilterExpressionVisitor);
+ *
+ * FilterExpression whereClauseFilter = filterPair.getLeft();
+ * FilterExpression havingClauseFilter = filterPair.getRight();
+ * }
+ * </pre>
+ * {@link SplitFilterExpressionVisitor} is thread-safe and can be accessed by multiple threads at the same time.
+ */
+@Slf4j
+public class SplitFilterExpressionVisitor implements FilterExpressionVisitor<Pair<FilterExpression, FilterExpression>> {
```
hellohanchen: We can create a class WhereHaving to store where and having clauses instead of using Pair.

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
```diff
@@ -17,6 +17,31 @@
...
+        if (left != null && right != null) {
+            return new AndFilterExpression(left, right);
+        } else if (left == null && right == null) {
+            return null;
+        } else if (left == null && right != null) {
+            return right;
+        } else {
+            return left;
+        }
+    }
```
hellohanchen: Can we add a new method 
```Java
public FilterExpression and(FilterExpression other) 
```
in FilterExpression class?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
QubitPi: This is pure having in some cases. There are also cases when `left` and `right` are both some random mixes of H and W (non-pure)

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1020,6 +1020,20 @@ public String getId(Object value) {
...
+     *
+     * @param entityName  The entity name
+     * @param field  The entity field name
+     *
+     * @return a Collection of Annotations
+     *
+     * @throws NoSuchFieldException if the {@code field} is not defined in the entity
+     */
+    public Collection<Annotation> getFieldAnnotations(String entityName, String field) throws NoSuchFieldException {
+        return Arrays.asList(getEntityClass(entityName).getDeclaredField(field).getDeclaredAnnotations());
```
aklish: I would make this consistent with other methods that do similar work (both in name and implementation):

```java
    public <A extends Annotation> A getAttributeOrRelationAnnotations(Class<?> entityClass,
                                         String identifier) {
        AccessibleObject fieldOrMethod = getEntityBinding(entityClass).fieldsToValues.get(identifier);
        if (fieldOrMethod == null) {
            return new ArrayList<>();
        }
        return fieldOrMethod.getDeclaredAnnotations();
    }
```

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
aklish: I like the suggested name `and` better.

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/AndFilterExpression.java
```diff
@@ -17,6 +17,31 @@
...
+     *
+     * @param left  The provided left {@link FilterExpression}
+     * @param right  The provided right {@link FilterExpression}
+     *
+     * @return a new {@link AndFilterExpression} instance or {@code null}
+     */
+    public static FilterExpression withLeftAndRight(FilterExpression left, FilterExpression right) {
+        if (left != null && right != null) {
+            return new AndFilterExpression(left, right);
+        } else if (left == null && right == null) {
```
aklish: This can be shortened like this:
```java
        if (left != null && right != null) {
            return new AndFilterExpression(left, right);
        } else if (left == null) {
             return right;
        } 
        return left;
```

elide-core/src/main/java/com/yahoo/elide/core/filter/expression/OrFilterExpression.java
```diff
@@ -18,6 +18,31 @@
...
+     * other non-null is returned with no modification. If both left and right are {@link null}, this method returns
+     * {@code null}.
+     *
+     * @param left  The provided left {@link FilterExpression}
+     * @param right  The provided right {@link FilterExpression}
+     *
+     * @return a new {@link OrFilterExpression} instance or {@code null}
+     */
+    public static FilterExpression withLeftAndRight(FilterExpression left, FilterExpression right) {
+        if (left != null && right != null) {
```
aklish: Same comments as for the AndFilterExpression.

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -44,10 +44,36 @@
             <groupId>com.yahoo.elide</groupId>
             <artifactId>elide-core</artifactId>
         </dependency>
+
+        <!-- Test -->
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-integration-tests</artifactId>
```
aklish: Remove this dependency.  We definitely don't want to run the elide-integration-tests on this store.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
aklish: I"m not following.  Pure Having should not have a Where component.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
aklish: :+1: on having a WhereHaving class.

QubitPi:
@aklish Regarding the pure having, we could have something like this

```
left = Pair[W1, H1]
right = Pair[W2, H2]
```

which gets turned into `Pair[(W1 AND W2), (H1 and H2)]`. This is not covered by all 3 `if-else` branch above because both left and right are not pure

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/WhereHaving.java
```diff
@@ -0,0 +1,108 @@
...
+     * {@code HAVING} filter expression.
+     *
+     * @param whereExpression  A pure {@code HAVING} filter expression
+     * @param havingExpression  A pure {@code WHERE} filter expression
+     *
+     * @return a new instance of {@link WhereHaving}
+     *
+     * @throws NullPointerException if the provided {@code WHERE} or {@code HAVING} filter expression is {@code null}
+     */
+    public static WhereHaving withWhereAndHaving(FilterExpression whereExpression, FilterExpression havingExpression) {
```
hellohanchen: Can we use `@NonNull` (lombok), here?
`withWhereAndHaving(@NonNull FilterExpression whereExpression, @NonNull FilterExpression havingExpression)`
Actually using a builder might be better for this class? Any opinion @aklish ?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,187 @@
...
+                ? WhereHaving.pureHaving(filterPredicate) // this filterPredicate belongs to a HAVING clause
+                : WhereHaving.pureWhere(filterPredicate); // this filterPredicate belongs to a WHERE clause
+    }
+
+    @Override
+    public WhereHaving visitAndExpression(final AndFilterExpression expression) {
+        WhereHaving left = expression.getLeft().accept(this);
+        WhereHaving right = expression.getRight().accept(this);
+
+        if (left.isWhereExpression() && right.isWhereExpression()) {
```
hellohanchen: Can we just do 
```
return new WhereHaving(and(left.getWhere, right.getWhere), and(left.getHaving, right.getHaving))
```
I think the `and()` and `WhereHaving constructor` should be able to handle `null`

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/WhereHaving.java
QubitPi: Build is not a good idea with less than 3 optional args. It's an overkill. Static factory gives better readability

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,187 @@
...
+     *
+     * @param filterPredicate  The terminal filter expression to check for
+     *
+     * @return {@code true} if the {@link FilterPredicate} is a HAVING clause
+     */
+    private boolean isHavingPredicate(final FilterPredicate filterPredicate) {
+        Class entityClass = filterPredicate.getEntityType();
+        String fieldName = filterPredicate.getField();
+
+        return getEntityDictionary()
```
aklish: This is where the schema will come in handy.  MetricComputation might also require Having (not just MetricAggregation).  

Instead of using the EntityDictionary, this could leverage the schema.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,187 @@
...
+    public WhereHaving visitNotExpression(NotFilterExpression expression) {
+        FilterExpression normalized = getNormalizationVisitor().visitNotExpression(expression);
+
+        if (normalized instanceof AndFilterExpression) {
+            return visitAndExpression((AndFilterExpression) normalized);
+        } else if (normalized instanceof OrFilterExpression) {
+            return visitOrExpression((OrFilterExpression) normalized);
+        } else if (normalized instanceof NotFilterExpression) {
+            normalized = ((NotFilterExpression) normalized).getNegated();
+            return visitNotExpression((NotFilterExpression) normalized);
```
aklish: This will effectively drop the Not.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitorTest.java
```diff
@@ -0,0 +1,155 @@
...
+import org.testng.annotations.BeforeMethod;
+import org.testng.annotations.Test;
+
+import java.util.Collections;
+
+public class SplitFilterExpressionVisitorTest {
+
+    private static final FilterExpressionNormalizationVisitor NORMALIZATION_VISITOR =
+            new FilterExpressionNormalizationVisitor();
+    private static final FilterPredicate WHERE_PREDICATE = new FilterPredicate(
```
aklish: The RSQLDialect now has a simple method to convert from a RSQL expression to a FilterExpression.  That might be easier to construct queries - rather than do it this way.

QubitPi: Hmm I like that!

I would like to keep this as static but RSQLDialect cannot be initialized without a entity dictionary which is not static. 

Well I think constructing in the current way looks very easy to me though :)

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,247 @@
...
+     * expressions like {@code NOT (A AND B)} into {@code NOT A OR NOT B}
+     *
+     * @throws NullPointerException if any one of the argument is {@code null}
+     */
+    public SplitFilterExpressionVisitor(
+            final Schema schema,
+            final FilterExpressionNormalizationVisitor normalizationVisitor
+    ) {
+        this.schema = Objects.requireNonNull(schema, "schema");
+        this.normalizationVisitor = Objects.requireNonNull(normalizationVisitor, "normalizationVisitor");
```
aklish: This could probably just be constructed in the constructor rather than passed in as an argument.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
```diff
@@ -0,0 +1,247 @@
...
+        if (normalized instanceof AndFilterExpression) {
+            return visitAndExpression((AndFilterExpression) normalized);
+        } else if (normalized instanceof OrFilterExpression) {
+            return visitOrExpression((OrFilterExpression) normalized);
+        } else if (normalized instanceof NotFilterExpression) {
+            FilterConstraints negatedConstraint = visitNotExpression((NotFilterExpression) normalized);
+
+            if (negatedConstraint.isPureWhere()) {
+                return FilterConstraints.pureWhere(new NotFilterExpression(negatedConstraint.getWhereExpression()));
+            } else if (negatedConstraint.isPureHaving()) {
```
aklish: It isn't possible to have a mixed where/having for a NotFilterExpression after normalization.  You can just do a simple if/else here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/visitor/SplitFilterExpressionVisitor.java
QubitPi: Yep.

How about putting it as a static field?

Issue #821, opened 2019-05-21
AggregationDataStore: Change JPA & Hibernate Stores to Implement BridgeableTransaction


aklish:
The data model entities managed by the AggregationStore will be fact tables that join to dimension tables.  The dimension tables will likely be managed by other DataStores (so they can be mutated).  To enable the client to request a compound document that includes fact rows and corresponding dimension rows, Elide needs to fetch data across different data stores.

## Expected Behavior

Currently, neither JPA nor Hibernate stores implement BridgeableTransaction.  

They should implement this interface to allow cross-store, compound document fetches.

## Context

https://github.com/yahoo/elide/issues/804

PR #841, opened 2019-05-28
AggregationDataStore: Change JPA & Hibernate Stores to Implement Brid…

QubitPi:
Issue https://github.com/yahoo/elide/issues/821


elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -242,6 +246,31 @@ public Object getRelation(
         return val;
     }
 
+    @Override
+    public Object bridgeableLoadObject(
```
aklish: We shouldn't need to define these.

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/BridgeableTransaction.java
```diff
@@ -54,12 +55,27 @@
...
                                 Serializable lookupId,
                                 Optional<FilterExpression> filterExpression,
-                                RequestScope scope);
+                                RequestScope scope) {
+        EntityDictionary entityDictionary = scope.getDictionary();
+
+        if (lookupId == null) {
+            // according to contract, lookupId can be null
+            // in this case, we compute the lookupId
+            Object parentInstance = muxTx.loadObject(
```
aklish: The parentInstance is already loaded here (parent parameter in the function).

elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/BridgeableTransaction.java
```diff
@@ -54,12 +55,27 @@
...
+                    entityDictionary.getEntityClass(parent.getClass().getName()),
+                    entityDictionary.getId(parent),
+                    filterExpression, scope
+            );
+            lookupId = entityDictionary.getId(entityDictionary.getField(parentInstance, relationName));
```
aklish: I understand what you are trying to do here, but I think the whole contract for store bridging is busted.  I don't know if there is a general way to get the ID from the parent.  It seems like the other store (the one we bridged from would have to supply it somehow).

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -991,6 +991,14 @@ public String getId(Object value) {
         }
     }
 
+    public AccessibleObject getField(Object value, String fieldName) {
```
hellohanchen: Can this be renamed as `getFieldValue`?

Issue #823, opened 2019-05-21
AggregationDataStore : RequestScope should include fields requested for a given entity load


aklish:
The Aggregation Data Store needs to know what attributes and relationships are being fetched so that it can construct a complete JPQL query.

While this information is available today for JSON-API (via the request headers and the list of sparse fields for a given entity model), this capability does not exist for GraphQL.

## Expected Behavior
A new field, *requestedSubgraph* will be set inside the requestScope each time `loadObjects` or `getRelationship` is called on the `DataStoreTransaction`.

This field will contains a nested map containing the entity relationship graph (entities, fields, and relationships to other entities) that is being requested starting at the current collection or entity that is being loaded.

## Possible Solution
Both JSON-API and GraphQL will require changes:

1. JSON-API can construct this graph from the `&include` and `&fields` query parameters.
2. GraphQL can look at the `SelectionSet` passed in the `DataFetchingEnvironment`.

## Context
https://github.com/yahoo/elide/issues/804

QubitPi:
@aklish I assume `requestedSubgraph` will essentially be the same thing as `RequestScope.sparseFields`, right?

QubitPi:
@aklish When, in the case of JSON-API, the graph can be constructed using `&include` and `&fields`, what will be the use of **setting the graph each time `loadObjects` or `getRelationship` is called on the `DataStoreTransaction`**?

PR #842, opened 2019-05-28
AggregationDataStore : Include fields requested for a given entity load in RequestScope

QubitPi:
Issue https://github.com/yahoo/elide/issues/823


elide-datastore/elide-datastore-jpa/pom.xml
```diff
@@ -56,7 +56,7 @@
         <dependency>
             <groupId>com.yahoo.elide</groupId>
             <artifactId>elide-datastore-multiplex</artifactId>
-            <version>4.4.5-SNAPSHOT</version>
+            <version>RELEASE</version>
```
aklish: This needs to remain as a SNAPSHOT dependency for maven release to work correctly.

elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -65,6 +65,7 @@
     @Getter private final AuditLogger auditLogger;
     @Getter private final Optional<MultivaluedMap<String, String>> queryParams;
     @Getter private final Map<String, Set<String>> sparseFields;
+    @Getter private final Map<String, Map<String, Object>> requestedSubgraph;
```
aklish: I think it would be good if we had a bit more structure to this.

Maybe we can make a class called FieldSet:

```java
class FieldSet {
  private Set<String> attributes;
  private Map<String, FieldSet> relationships;
}```

Issue #818, opened 2019-05-21
AggregationDataStore: Static Attribute Aggregation


aklish:
The AggregationDataStore needs a schema for each entity that:
- Specifies that an attribute is a metric
- Specifies that an attribute or relationship is a dimension
- Specifies that an attribute is a Data/Time column and supports grain.
- Specifies that an Attribute is a _friendlyName_ or human displayable column for that dimension.
- Specifies whether a dimension has small, medium, or large cardinality.

## Expected Behavior
New AggregationDataStore annotations should be created:

1. MetricAggregation
2. Cardinality
3. ComputedMetric
4. FriendlyName

### Examples
```java
@MetricAggregation(default = Sum.class, aggregations = {Max.class, Median.class})
```

```java
@ComputedMetric(expression = '(fieldA * fieldB) / 100')
```

```java
@Cardinality(size = MEDIUM)
```

```java
@FriendlyName
```

## Current Behavior
N/A

## Possible Solution
Elide calls `populateEntityDictionary` on each data store. 

The AggregationDataStore should only bind entities that have at least one `MetricAggregation` annotation.    For each entity bound, it should use the `EntityDictionary` to lookup other data store specific annotations and construct a schema object for that entity.  The schema will be used as foundational metadata for the store to perform its functions.

QubitPi:
@aklish I'm working on it

QubitPi:
@aklish Is `@Cardinality` is same thing as the cardinality in JPQL database? Do we simply trust user who will set whatever value it will be for `size`?

Could we have more background/business-background on the `@FriendlyName`? This helps me implement it.

aklish:
@Cardinality is a new annotation that will have a value of SMALL, MEDIUM, or LARGE.  We trust the developer to set it.  The values should be an enum.  Also, let's rename `@ComputedMetric` to `@MetricComputation`.

QubitPi:
Ok. Yeah enum is my choice, too

PR #846, opened 2019-06-04
AggregationDataStore: Schema

QubitPi:
Issue https://github.com/yahoo/elide/issues/818


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,286 @@
...
+ * {@link Schema} is thread-safe and can be accessed by multiple-threads.
+ */
+@Slf4j
+public class Schema {
+
+    @Getter
+    private final Class<?> entityClass;
+
+    @Getter
+    private final String dateTimeColumn;
```
aklish: Not every schema will have a date time column.  Let's remove this.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,286 @@
...
+
+    private final String friendlyName;
+
+    @Getter(value = AccessLevel.PRIVATE)
+    private final List<String> metrics;
```
aklish: Metrics and Dimensions should be their own classes.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,286 @@
...
+@Slf4j
+public class Schema {
+
+    @Getter
+    private final Class<?> entityClass;
+
+    @Getter
+    private final String dateTimeColumn;
+
+    private final String friendlyName;
```
aklish: Friendly name should be the property of the Dimension class.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,286 @@
...
+
+    @Getter(value = AccessLevel.PRIVATE)
+    private final List<String> metrics;
+
+    @Getter(value = AccessLevel.PRIVATE)
+    private final List<String> dimensions;
+
+    @Getter(value = AccessLevel.PRIVATE)
+    private final Map<String, DimensionSize> dimensionSizes;
```
aklish: The cardinality of a dimension should be the property of a DImension.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,286 @@
...
+    @Getter(value = AccessLevel.PRIVATE)
+    private final EntityDictionary entityDictionary;
+
+    /**
+     * Constructor.
+     *
+     * @param clz  The type of the entity, whose {@link Schema} is to be constructed
+     * @param entityDictionary  The meta info object that helps to construct {@link Schema}
+     */
+    public Schema(Class<?> clz, EntityDictionary entityDictionary) {
```
hellohanchen: Can we change "clz" to "entityClass"?

QubitPi: I'm following the codebase convention https://github.com/yahoo/elide/search?q=cls&unscoped_q=cls

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/annotation/Cardinality.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Indicates whether a dimension has small, medium, or large cardinality.
+ * <p>
+ * Example: {@literal @}Cardinality(size = {@link DimensionSize#MEDIUM}). If {@code size} is not specified,
+ * {@link DimensionSize#LARGE} is used by default. See {@link DimensionSize}.
+ * <p>
+ * To avoid double binding, {@link Cardinality} is a {@link ElementType#TYPE class-only annotation}.
+ */
+@Documented
+@Target(ElementType.TYPE)
```
QubitPi: I make Cardinality a class-level only annotation because I though it would be awkward to have a class being annotated and a second person put the same annotation in a relationship again.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/Metric.java
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Returns an {@link Optional} of the JPQL expression that represents this metric computation logic.
+     * <p>
+     * A {@link MetricAggregation simple metric} does not have a JPQL expression; in this case, this method should
+     * return {@link Optional#empty()}
+     *
+     * @return a JPQL formula for computing this {@link Metric} or {@link Optional#empty()} on a
+     * {@link MetricAggregation simple metric}.
+     */
+    Optional<String> getComputedMetricExpression();
```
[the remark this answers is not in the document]
QubitPi: This returns expanded formula. I have updated javadoc accordingly.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,175 @@
...
+     * Returns all dimension fields of the entity associated with this schema.
+     * <p>
+     * This method calls {@link #constructDimension(Class)} to obtain the concrete implementation of {@link Dimension}.
+     *
+     * @return all non-metric fields
+     */
+    protected List<Dimension> getAllDimensions() {
+        return getEntityDictionary().getAllFields(getEntityClass()).stream()
+                .filter(field -> !AbstractMetric.isMetricField(field, getEntityClass(), getEntityDictionary()))
+                .filter(field -> getEntityDictionary().isRelation(getEntityClass(), field))
```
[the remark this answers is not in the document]
QubitPi: I see. So we can have different cardinalities between a table and a table column.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,175 @@
...
+ * {@link Schema} is thread-safe and can be accessed by multiple-threads.
+ */
+@Slf4j
+public class Schema {
+
+    @Getter
+    private final Class<?> entityClass;
+
+    @Getter(value = AccessLevel.PRIVATE)
+    private final List<Metric> metrics;
```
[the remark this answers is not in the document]
QubitPi: I actually don't have preference. I chose List because it supported indexed loop, which in some cases might be helpful. But I can change to Sets

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * Returns whether or not an attribute is a dimension field.
+     *
+     * @param fieldName  The field to be checked
+     *
+     * @return {@code true} if the field is a dimension
+     */
+    public boolean isDimension(String fieldName) {
```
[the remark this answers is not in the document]
QubitPi: Right. I was wondering if it can be useful in the future. Do we still want me to remove it?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/annotation/Cardinality.java
QubitPi: Right. Based on your other comment, I'll also add `ElementType.FIELD` and `ElementType.METHOD`

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1214,6 +1214,35 @@ public boolean hasBinding(Class<?> cls) {
...
+        if (fieldOrMethods == null) {
+            return false;
+        }
+
+        AccessibleObject fieldOrMethod = fieldOrMethods.get(fieldName);
+        if (fieldOrMethod == null) {
+            return false;
+        }
+
+        return getAttributeOrRelationAnnotation(cls, annotationClass, EntityBinding.getFieldName(fieldOrMethod))
```
aklish: This method can be collapsed to just:
```java
return getAttributeOrRelationAnnotation(cls, annotationClass, EntityBinding.getFieldName(fieldOrMethod)) != null
```

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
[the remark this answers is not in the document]
aklish: List implies the same metric can exist twice.  I think a Set is a better constraint.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/AbstractMetric.java
```diff
@@ -0,0 +1,230 @@
...
+     *     <li> {@link MetricComputation}
+     * </ol>
+     *
+     * @param fieldName  The entity field
+     * @param containingCls  The entity that contains the {@code fieldName}
+     * @param entityDictionary  The auxiliary object that offers binding info used by this method
+     *
+     * @return {@code true} if the field is a metric field
+     */
+    public static boolean isMetricField(String fieldName, Class<?> containingCls, EntityDictionary entityDictionary) {
```
aklish: We should try to avoid propagating EntityDictionary to too many classes.  It should remain encapsulated inside the Schema only.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/EntityMetric.java
```diff
@@ -0,0 +1,58 @@
...
+     * Constructs a new {@link EntityMetric} instance.
+     *
+     * @param metricField  The entity field of the metric being constructed
+     * @param cls  The entity that contains the metric being constructed
+     * @param entityDictionary  The auxiliary object that offers binding info used to construct this
+     * {@link EntityMetric}
+     *
+     * @return a {@link Metric}
+     */
+    public static Metric newInstance(String metricField, Class<?> cls, EntityDictionary entityDictionary) {
```
aklish: Same comment - keep all of these construction methods in the schema or find a way to not propagate the EntityDictionary so widely.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/EntityMetric.java
```diff
@@ -0,0 +1,58 @@
...
+    /**
+     * Private constructor.
+     *
+     * @param metricIdentifier  The entity field name as the name of this metric
+     * @param defaultAggregation  The default aggregation function to apply to this {@link Metric}
+     * @param allAggregations  All supported aggregation functions that apply to this {@link Metric}
+     * @param computedMetricExpression  The JPQL expression that represents this metric computation logic; {@code null}
+     * for {@link MetricAggregation simple metric}
+     */
+    private EntityMetric(
```
aklish: Do we need this class?  Instead of having an interface, an abstract class, and a concrete class, can we collapse this to an interface and a ConcreteClass?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/dimension/EntityDimension.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Default implementation of {@link Dimension} that represents an entity as a dimension.
+ * <p>
+ * If the entity does not have {@link Cardinality} annotation on it, {@link EntityDimension} assigns a
+ * {@link #getDefaultCardinality() default cardinality}.
+ * <p>
+ * {@link EntityDimension} is thread-safe and can be accessed by multiple threads.
+ */
+public class EntityDimension extends AbstractDimension {
```
aklish: Do we need this class?  Instead of having an interface, an abstract class, and a concrete class, can we collapse this to an interface and a ConcreteClass?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/PlayerStats.java
```diff
@@ -0,0 +1,82 @@
...
+    @Id
+    public String getId() {
+        return id;
+    }
+
+    public void setId(final String id) {
+        this.id = id;
+    }
+
+    @FriendlyName
```
aklish: FriendlyName wouldn't decorate the FactTable - it would decorate a field on a relationship from the Fact Table.   It would make sense to decorate country.name with this for example.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/dimension/AbstractDimension.java
```diff
@@ -0,0 +1,110 @@
...
+            final DimensionSize cardinality
+    ) {
+        this.dimensionIdentifier = Objects.requireNonNull(dimensionIdentifier, "dimensionIdentifier");
+        this.friendlyName = friendlyName;
+        this.cardinality = Objects.requireNonNull(cardinality, "cardinality");
+    }
+
+    @Override
+    public Optional<String> getFriendlyName() {
+        return friendlyName == null ? Optional.empty() : Optional.of(friendlyName);
```
aklish: There should always be a friendly name.  It will fall into one of three categories:

1. A dimension table has a single attribute with the friendly name annotation.  That attribute is the friendly name.
2. A dimension table has no attribute with the friendly name annotation.  We should use its `@Id` field as the friendly name.
3. A fact table has a degenerate dimension (an attribute which is also a dimension).  The friendly name is the attribute in that case.  There should not be an annotation here.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/CardinalityTestBean.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+/**
+ * Test default behavior of {@link Cardinality}.
+ */
+@Entity
+@Include(rootLevel = true)
+@Cardinality
+public class CardinalityTestBean {
```
aklish: Instead of creating test beans like this, I think we should try to create a real world model that is intuitive.  Let's create a model around PlayerStats for a video game.  All the models ideally should make sense in that context.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
QubitPi: Right. To get `fieldOrMethod` used in this line, we probably need to execute extra logic above.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/EntityMetric.java
QubitPi: Yep

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/AbstractMetric.java
QubitPi: Sure.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/CardinalityTestBean.java
QubitPi: Ok. Done

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/dimension/AbstractDimension.java
QubitPi: Thanks

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,633 @@
...
+            return String.format(
+                    "%s.%s",
+                    TABLE_PREFIX,
+                    getMappedColumn(expandingField, containingCls)
+            );
+        }
+
+        // hit a computed metric field if valid. Expand
+        visited.add(expandingField);
+        String metricExpression = getEntityDictionary().getAttributeOrRelationAnnotation(
```
aklish: Maybe this could come from the schema.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1214,6 +1214,47 @@ public boolean hasBinding(Class<?> cls) {
...
+
+    /**
+     * Returns whether or not a specified field exists in an entity.
+     *
+     * @param cls  The entity
+     * @param fieldName  The provided field to check
+     *
+     * @return {@code true} if the field exists in the entity
+     */
+    public boolean isValidField(Class<?> cls, String fieldName) {
```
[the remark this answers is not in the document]
QubitPi: Sure

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,633 @@
...
+            String message = String.format(
+                    "'%s' in '%s' is not a defined as a metric",
+                    expandingField,
+                    containingCls.getName()
+            );
+            log.error(message);
+            throw new IllegalStateException(message);
+        }
+
+        if (isBaseMetric(expandingField, containingCls)) {
```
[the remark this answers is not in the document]
QubitPi: When we say computed metric will always reference the aggregations inside of them, do we mean if we have the following entity

```java
     * public class FactTable {
     *
     *     {@literal @}MetricAggregation(sum.class)
     *     Long sessions
     *
     *     @MetricAggregation(sum.class)
     *     Long timeSpent
     *
     *     @MetricComputation(expression = "timeSpent / sessions")
     *     Float timeSpentPerSession
     *
     *     @literalMetricComputation(expression = "timeSpentPerSession / gameCount")
     *     Float timeSpentPerGame
     * }
```

then will the `timeSpentPerSession` get expanded into

```
timeSpent / sessions
```

or

SUM(timeSpent) / SUM(timeSpent)

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,633 @@
...
+    /**
+     * Returns the nature of a SQL column that is backing a specified dimension field in an entity.
+     *
+     * @param dimensionField  The provided dimension field
+     * @param cls  The entity having the {@code dimensionField}
+     *
+     * @return a {@link ColumnType}, such as PK or a regular field
+     */
+    private ColumnType parseColumnType(String dimensionField, Class<?> cls) {
+        if (getEntityDictionary().attributeOrRelationAnnotationExists(cls, dimensionField, Id.class)) {
```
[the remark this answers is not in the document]
QubitPi: Are we moving this method into EntityDictionary? I was thinking this because then we would need to import aggregation module inside core module.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,633 @@
...
+     *     <li> {@link MetricComputation}
+     * </ol>
+     *
+     * @param fieldName  The entity field
+     * @param containingCls  The entity that contains the {@code fieldName}
+     *
+     * @return {@code true} if the field is a metric field
+     */
+    public boolean isMetricField(String fieldName, Class<?> containingCls) {
+        return getEntityDictionary().attributeOrRelationAnnotationExists(
```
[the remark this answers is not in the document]
QubitPi: This has been moved to a difference class - MetricUtils.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,633 @@
...
+     * <p>
+     * See {@link #TABLE_PREFIX} for more details.
+     *
+     * @param expandingField  The entity field
+     * @param containingCls  The entity that contains the {@code expandingField}
+     * @param visited  A set of already-visited expression variables, which should be skipped during expansion.
+     *
+     * @return a JPQL expression with placeholder
+     */
+    private String getExpandedMetricExpression(
```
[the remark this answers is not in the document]
QubitPi: Let't try a shorter version then. I removed some long logics there

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,633 @@
...
+     * Returns the estimated cardinality of this a dimension field.
+     * <p>
+     * {@link #getDefaultCardinality() Default} is returned when the dimension is not annotated by {@link Cardinality}.
+     *
+     * @param dimension  The dimension field
+     * @param cls  The entity name of the dimension from which estimated cardinality is being looked up
+     *
+     * @return cardinality of the dimension field
+     */
+    private CardinalitySize getEstimatedCardinality(String dimension, Class<?> cls) {
```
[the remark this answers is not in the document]
QubitPi: I've merged EntityDimension and AbstractDimension as requested

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
QubitPi: This method has been moved to ComputedMetric class, if you are ok with it

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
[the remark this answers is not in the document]
aklish: We don't want to make EntityDictionary aware of the  AggregationStore annotations.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/BaseMetric.java
```diff
@@ -0,0 +1,144 @@
...
+    private final List<Class<? extends Aggregation>> aggregations;
+
+    @Getter
+    private final String metricExpression;
+
+    public BaseMetric(String metricField, Class<?> cls, EntityDictionary entityDictionary) {
+        Meta metaData = entityDictionary.getAttributeOrRelationAnnotation(cls, Meta.class, metricField);
+        Class<?> fieldType = entityDictionary.getType(cls, metricField);
+
+        this.name = metricField;
```
aklish: I wonder if we need a base abstract class called Column that is common for both Dimension and Metric where we could consolidate the logic to parse the Meta annotation and set the basic fields?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/ComputedMetric.java
```diff
@@ -0,0 +1,209 @@
...
+     * <p>
+     * <b>It is caller's responsibility to replace the {@link #TABLE_PREFIX} with the per-user specific value at
+     * runtime.</b>
+     */
+    public static final String TABLE_PREFIX = "<PREFIX>";
+
+    private static final long serialVersionUID = -5243438667754088900L;
+
+    @Getter
+    private final String name;
```
aklish: A lot of these fields are common (with the concept of a Column described earlier).  We should move them to a common base class.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/ComputedMetric.java
```diff
@@ -0,0 +1,209 @@
...
+    private final String longName;
+
+    @Getter
+    private final String description;
+
+    @Getter
+    private final Class<?> dataType;
+
+    @Getter
+    private final List<Class<? extends Aggregation>> aggregations;
```
aklish: I'm not sure it makes sense for a computed metric to support aggregations. How do we take a SUM of SUMs?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/MetricUtils.java
```diff
@@ -0,0 +1,113 @@
...
+     *     <li> {@link MetricAggregation}
+     *     <li> {@link MetricComputation}
+     * </ol>
+     *
+     * @param fieldName  The entity field
+     * @param containingCls  The entity that contains the {@code fieldName}
+     *
+     * @return {@code true} if the field is a metric field
+     */
+    public static boolean isMetricField(String fieldName, Class<?> containingCls,  EntityDictionary entityDictionary) {
```
aklish: A method like this seems to make more sense in Schema.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/MetricUtils.java
```diff
@@ -0,0 +1,113 @@
...
+     * Parses to obtain all supported aggregation functions to apply to this {@link Metric}.
+     *
+     * @param fieldName  The entity field
+     * @param containingCls  The entity that contains the {@code fieldName}
+     *
+     * @return all available aggregation classes
+     *
+     * @throws IllegalArgumentException if the {@code fieldName} is not a metric field
+     */
+    public static List<Class<? extends Aggregation>> extractAggregations(
```
aklish: There is only really one metric type that should support aggregations - those annotated with MetricAggregation.  We can remove this function out of utils and move the logic into MetricAggregation.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/MetricUtils.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Returns whether or not an entity field is a simple metric, i.e. a metric field not decorated by
+     * {@link MetricComputation}.
+     *
+     * @param fieldName  The entity field
+     * @param containingClz  The entity containing the field
+     *
+     * @return {@code true} if the field is a metric field and also a base metric without JPQL expression annotated
+     */
+    public static boolean isBaseMetric(String fieldName, Class<?> containingClz, EntityDictionary entityDictionary) {
```
aklish: This function can be removed if you adopt my other comment on metric construction..  Then we can remove the MetricUtils class.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,223 @@
...
+     *
+     * @param metricField  The entity field of the metric being constructed
+     * @param cls  The entity that contains the metric being constructed
+     * @param entityDictionary  The auxiliary object that offers binding info used to construct this
+     * {@link Metric}
+     *
+     * @return a {@link Metric}
+     */
+    protected Metric constructMetric(String metricField, Class<?> cls, EntityDictionary entityDictionary) {
+        return MetricUtils.isBaseMetric(metricField, cls, entityDictionary)
```
aklish: I would expand this to:

```java
       if (entityDictionary.attributeOrRelationAnnotationExists(
                cls, metricField, MetricAggregation.class)) {
           return new BaseMetric(...)
       } else if(...) {
           return new ComputedMetric(...)  
       }
```

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/ComputedMetric.java
```diff
@@ -0,0 +1,209 @@
...
+        this.dataType = fieldType;
+        this.aggregations = MetricUtils.extractAggregations(metricField, cls, entityDictionary);
+
+        if (aggregations.isEmpty()) {
+            String message = String.format("'%s' in '%s' has no aggregation.", metricField, cls.getCanonicalName());
+            log.error(message);
+            throw new IllegalStateException(message);
+        }
+
+        this.metricExpression = expandMetricExpression(metricField, cls, new HashSet<>(), entityDictionary);
```
aklish: The metricExpression will be expanded at query time because it depends on the requested aggregation function.  We should not do this during construction.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1214,6 +1214,36 @@ public boolean hasBinding(Class<?> cls) {
         return bindJsonApiToEntity.contains(cls);
     }
 
+    /**
```
aklish: A few general comments (I couldn't find one specific place to put these):
1.  We should rely on the EntityDictionary less throughout.  We should build the schema and then use the schema in place of the dictionary where ever possible.
2. Reflection (checking for presence of annotations) is called repeatedly.  Reflection is pretty costly.  We should use the schema instead and its related classes (metric & dimension) to answer questions rather than going to the dictionary again and again.
3. Metric expansion has to happen at query time.  We cannot know in advance what agg function to apply during the expansion. 
4. Utils classes (IMO) are usually just misplaced functions.  I have some comments on this below, but let's find a proper home for those functions and retire the Utils class.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/ComputedMetric.java
```diff
@@ -0,0 +1,209 @@
...
+     *
+     * @return a JPQL metricExpression with placeholder
+     */
+    private static String expandMetricExpression(
+            String expandingField,
+            Class<?> containingCls,
+            Set<String> visited,
+            EntityDictionary entityDictionary
+    ) {
+        if (MetricUtils.isBaseMetric(expandingField, containingCls, entityDictionary)) {
```
aklish: We can determine if a metric is a base metric based on whether or not it has any supported aggregations.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/ComputedMetric.java
```diff
@@ -0,0 +1,209 @@
...
+    @Getter
+    private final Class<?> dataType;
+
+    @Getter
+    private final List<Class<? extends Aggregation>> aggregations;
+
+    @Getter
+    private final String metricExpression;
+
+    public ComputedMetric(String metricField, Class<?> cls, EntityDictionary entityDictionary) {
```
aklish: Can we move ComputedMetric (and related expansions) to a separate PR?  Mantis doesn't require it - so it would be good to unblock that work first.  The other reason for this - is that the whole ComputedMetric now needs to generate SQL based on table aliases based on the new Carbon design.  So the generation of the SQL is a bit different than we originally planned.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/annotation/Meta.java
```diff
@@ -0,0 +1,25 @@
...
+ * Indicates that the specified entity field has a configured long name and field description for human to read on UI.
+ */
+@Documented
+@Target({ElementType.FIELD, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Meta {
+
+    String longName() default "";
+
+    String description() default "";
```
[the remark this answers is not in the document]
QubitPi: Given that Meta annotation applies to both Dimension and Metrics, and friendly name only applies to Dimension, I personally vote for separating them

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/annotation/MetricComputation.java
```diff
@@ -0,0 +1,68 @@
...
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import javax.persistence.Transient;
+
+/**
+ * Indicates that a field is computed via a {@link #expression() JPQL metric formula expression}.
```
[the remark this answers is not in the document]
QubitPi: Sounds like we are still exploring all the possibilities, whether Calcite or something else. For that I've changed it to

```
Indicates that a field is computed via a {@link #expression() custom metric formula expression}, such as Calcite SQL.
```

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/annotation/Temporal.java
```diff
@@ -0,0 +1,38 @@
...
+
+    /**
+     * All supported units into which temporal column can be divided
+     *
+     * @return all allowed time grain of a persistent storage column
+     */
+    TimeGrain timeGrain();
+
+    /**
+     * The timezone of the column.
```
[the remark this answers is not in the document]
QubitPi: The String format can be expressed by 

* an abbreviation such as "PST", or
* a full name such as "America/Los_Angeles", or
* a custom ID such as "GMT-8:00"

as documented by https://docs.oracle.com/javase/8/docs/api/java/util/TimeZone.html#getTimeZone-java.lang.String-

`getTimeZone(String ID)` will be used to parse the timezone String. I have updated the Javadoc of this field

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/BaseMetric.java
QubitPi: That's a great idea! I'd like to do it

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/ComputedMetric.java
QubitPi: Right. I agree. Copy and paste too quickly :)

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/ComputedMetric.java
QubitPi: I've removed this class from this PR.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
QubitPi: I'll refactor according to these

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/dimension/EntityDimension.java
```diff
@@ -0,0 +1,218 @@
...
+     * @throws IllegalStateException if more than 1 fields are annotated by the {@link FriendlyName}
+     */
+    public static String getFriendlyNameField(Class<?> cls, EntityDictionary entityDictionary) {
+        List<String> singleFriendlyName = entityDictionary.getAllFields(cls).stream()
+                .filter(field -> entityDictionary.attributeOrRelationAnnotationExists(
+                        cls,
+                        field,
+                        FriendlyName.class
+                ))
+                .collect(Collectors.toList());
```
aklish: This can be simplified like this:
```java
      entityDictionary.getAllFields(cls).stream()
                .filter(field -> entityDictionary.attributeOrRelationAnnotationExists(
                        cls,
                        field,
                        FriendlyName.class
                ))
               .reduce((a, b) -> {
                     throw new IllegalStateException("Multiple @FriendlyName fields: " + a + ", " + b);
               })
              .get();
```

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/dimension/EntityDimension.java
```diff
@@ -0,0 +1,218 @@
...
+            );
+
+            if (annotation != null) {
+                return annotation.size();
+            }
+
+            // annotation is not on entity; then must be on field or method
+        }
+
+        Cardinality annotation = entityDictionary
```
aklish: This is identical to the call on 83.  I think this is not what you intended.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,290 @@
...
+     * @param dimensionField  The entity field of the dimension being constructed
+     * @param cls  The entity that contains the dimension being constructed
+     * @param entityDictionary  The auxiliary object that offers binding info used to construct this
+     * {@link Dimension}
+     *
+     * @return a {@link Dimension}
+     */
+    protected Dimension constructDimension(String dimensionField, Class<?> cls, EntityDictionary entityDictionary) {
+        Meta metaData = entityDictionary.getAttributeOrRelationAnnotation(cls, Meta.class, dimensionField);
+        Class<?> fieldType = entityDictionary.getType(cls, dimensionField);
```
aklish: There are two getType methods in the dictionary: getType and getParamterizedType.  This might not work for `Set<Player>` for example.  We should test this.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,290 @@
...
+
+    /**
+     * Returns whether or not an entity field is a simple metric, i.e. a metric field not decorated by
+     * {@link MetricComputation}.
+     *
+     * @param fieldName  The entity field
+     *
+     * @return {@code true} if the field is a simple metric
+     */
+    public boolean isBaseMetric(String fieldName) {
```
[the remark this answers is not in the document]
QubitPi: Oops. Sorry I didn't remember to refactor entityDictionary out of this method. Yeah I remember you said it is costly to do reflection on this instance method.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/dimension/EntityDimension.java
QubitPi: Ah yeah. Thanks!

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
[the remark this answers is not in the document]
aklish: I don't think my earlier comment was clear.  I don't think we need this method.  We just need a way to determine if a metric has aggregations or not.  There should be a function on Metric (the interface) called `getAggregations`.  This can return an empty list.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,290 @@
...
+        )
+                .collect(
+                        Collectors.collectingAndThen(
+                                Collectors.toList(),
+                                Collections::unmodifiableList
+                        )
+                );
+
+        // make sure we have at least 1 aggregation so we can generate static SQL functions
+        if (aggregations.isEmpty()) {
```
aklish: Based on my previous comment, this check is no longer required.  Instead, it can be moved the constructor of BaseMetric.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
aklish: I think this still might be a bug.  Does it work?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/dimension/DimensionType.java
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * A dimension backed by a column.
+     */
+    DEGENERATE,
+
+    /**
+     * A dimension backed by a table.
+     */
+    TABLE
```
aklish: maybe rename to ENTITY

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/BaseMetric.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.StringJoiner;
+import java.util.stream.Collectors;
+
+/**
+ * {@link Metric} annotated by {@link MetricAggregation}.
+ * <p>
+ * {@link BaseMetric} is thread-safe and can be accessed by multiple threads.
+ */
+@Slf4j
+public class BaseMetric extends Column implements Metric {
```
aklish: Can we rename BaseMetric to AggregatedMetric?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/BaseMetric.java
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    private static final long serialVersionUID = 3055820948480283917L;
+
+    @Getter
+    private final List<Class<? extends Aggregation>> aggregations;
+
+    @Getter
+    private final String metricExpression;
```
aklish: Instead of storing this, why don't we compute this at runtime.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/BaseMetric.java
```diff
@@ -0,0 +1,142 @@
...
+     * @param aggregations  A list of all supported aggregations on this {@link Metric}
+     * @param metricExpression  The SQL functions corresponding the default aggregation of this {@link Metric}. Note
+     * that the default aggregation is the first aggregation in this list
+     */
+    public BaseMetric(
+            String metricField,
+            Meta annotation,
+            Class<?> fieldType,
+            List<Class<? extends Aggregation>> aggregations,
+            String metricExpression
```
aklish: I would remove this - and just compute at runtime.

aklish: In fact, we might need to change the Metric.getMetricExpression interface to take an optional aggregation as an argument.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/Metric.java
```diff
@@ -0,0 +1,69 @@
...
+
+    /**
+     * Returns a metric expression that represents this metric computation logic.
+     * <p>
+     * For computed metrics, this will be the expanded expression. For metric aggregations, this will be the UDF
+     * function called against the column name.
+     *
+     * @return a arithmetic formula for computing this {@link Metric} or default aggregation UDF on a base/simple metric
+     */
+    String getMetricExpression();
```
aklish: This should take an `Optional<Aggregation>` as an argument.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/dimension/DimensionTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+        // degenerate dimension
+        Assert.assertEquals(
+                DEGENERATE_DIMENSION.toString(),
+                "DegenerateDimension[columnType=FIELD, name='overallRating', longName='overallRating', description='overallRating', dimensionType=DEGENERATE, dataType=String, cardinality=SMALL, friendlyName='overallRating']"
+        );
+
+        Assert.assertEquals(
+                TIME_DIMENSION.toString(),
+                "TimeDimension[timeZone=Pacific Standard Time, timeGrain=DAY, columnType=TEMPORAL, name='recordedTime', longName='recordedTime', description='recordedTime', dimensionType=DEGENERATE, dataType=class java.lang.Long, cardinality=LARGE, friendlyName='recordedTime']"
```
hellohanchen: The timezone "Pacific Standard Time" cause the test to fail in my local testing as I am in CST.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/BaseMetric.java
QubitPi: Sure

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/dimension/DimensionTest.java
QubitPi: Thanks. It's a bug in test setup. Try again

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
QubitPi: Oh sorry I keep forgetting this. Let me add a test to it in the next commit

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,250 @@
...
+     * @param cls  The type of the entity, whose {@link Schema} is to be constructed
+     * @param entityDictionary  The meta info object that helps to construct {@link Schema}
+     *
+     * @throws NullPointerException if anyone of the arguments is {@code null}
+     */
+    public Schema(Class<?> cls, EntityDictionary entityDictionary) {
+        this.entityClass = Objects.requireNonNull(cls, "cls");
+        this.entityDictionary = Objects.requireNonNull(entityDictionary, "entityDictionary");
+
+        this.metrics = entityDictionary.getAllFields(getEntityClass()).stream()
```
aklish: I preferred the old way where we had methods getAllMetrics and getAllDimensions.  BetterCodeHub will probably ding this for complexity and number of lines.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -0,0 +1,250 @@
...
+public class Schema {
+
+    @Getter
+    private final Class<?> entityClass;
+    @Getter
+    private final Set<Metric> metrics;
+    @Getter
+    private final Set<Dimension> dimensions;
+    @Getter
+    private final Set<Column> columns;
```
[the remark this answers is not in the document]
QubitPi: Ok. Yeah I was doing it on the fly but moved them to here to save computation, but sure

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Schema.java
```diff
@@ -281,4 +292,18 @@ protected Dimension constructDimension(String dimensionField, Class<?> cls, Enti
...
+
+    /**
+     * Returns whether or not an entity field is a toMany relationship.
+     *
+     * @param dimensionField  The entity field to check
+     * @param cls  The entity that contains the field
+     *
+     * @return {@code true} if the field is toMany
+     */
+    private boolean isToManyRelationship(String dimensionField, Class<?> cls) {
```
aklish: There already is a method in the dictionary which does this:

getRelationshipType

PR #867, opened 2019-06-26
Define QueryEngine Contract

QubitPi:
## Description

This is a base PR for

* https://github.com/yahoo/elide/pull/840
* `AggregationDataStore` & `AggregationDataStoreTransaction` implementations

## Motivation and Context

Enable richer plugability between Elide and other DB query frameworks

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -0,0 +1,161 @@
...
+ *                             |             |                     |   |                  |
+ *                             | QueryEngine +-------------------------+  JpaQueryEngine  |
+ *                             |             |                     |   |                  |
+ *                             +-------------+                     |   +------------------+
+ *                                    |                            |
+ *                                    |                            |
+ *          +--------------------------------------                |   +------------------+
+ *          |                |                    |                |   |                  |
+ * +--------v-------+   +----v----+    +----------v-----------+    +---+ DrillQueryEngine |
+ * |    Hibernate   |   |   JPA   |    |     Apache Drill     |        |                  |
```
aklish: Drill would still use a SQL driver in this diagram.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -0,0 +1,161 @@
...
+     * @param pagination  Pagination which can be pushed down to the data store.
+     * @param scope  Contains request level metadata.
+     *
+     * @return a optimized query consumed by the data store
+     */
+    default Query buildQuery(
+            Class<?> entityClass,
+            Optional<FilterExpression> filterExpression,
+            Optional<Sorting> sorting,
+            Optional<Pagination> pagination,
```
aklish: We are missing Metrics, Dimensions (for groupings), and TimeDimensions (also for grouping).

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -0,0 +1,161 @@
...
+            RequestScope scope
+    );
+
+    /**
+     * Given a {@link FilterExpression}, this method gathers all the WHERE filter clauses together into a single WHERE
+     * clause.
+     *
+     * @return a filter expression that can be executed as a "WHERE"query constraint
+     */
+    Optional<WhereFilterExpression> extractWhereExpression(Optional<FilterExpression> filterExpression);
```
aklish: I think this should be part of the base class for the aggregation data store rather than part of the QueryEngine.  QueryEngine should have a very simple contract.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
QubitPi: I've updated the diagram

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
QubitPi: Ok. I will only contract query building and query execution then

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -0,0 +1,126 @@
...
+ *     <li> a {@link Query} that can be understood by a particular persistent storage or storage client
+ *     <li> a {@link WhereFilterExpression} that maps an Elide WHERE filter clause to an equivalent query constraint,
+ *          which can be interpreted by the particular persistent storage
+ *     <li> a {@link HavingFilterExpression} that maps an Elide WHERE filter clause to an equivalent query constraint,
+ *          which can be interpreted by the particular persistent storage
+ *     <li> one or more {@link FilterExpressionVisitor}s that can split an arbitrary {@link FilterExpression} into
+ *          aforementioned {@link WhereFilterExpression} and {@link HavingFilterExpression}
+ * </ol>
+ */
+public interface QueryEngine {
```
aklish: A couple thoughts here.  I would make QueryEngine just an interface with a single method:
``` Iterable<Object> executeQuery(Query query);```.   However, I would make Query an actual object with fields and a lombok builder or something along those lines to make it easier to construct.

File: elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/filter/expression/HavingFilterExpression.java [hunk 5]
```diff
@@ -0,0 +1,21 @@
...
+import com.yahoo.elide.core.filter.expression.FilterExpressionVisitor;
+
+/**
+ * A sub-type of {@link FilterExpression} that is evaluated in-memory on top of returned data from persistent storage.
+ * <p>
+ * This is a {@link java.util.function functional interface} whose functional method is
+ * {@link #accept(FilterExpressionVisitor)}
+ */
+@FunctionalInterface
+public interface HavingFilterExpression extends FilterExpression {
```
aklish: I'm not sure these are needed.  We already have fields in the executeQuery interface that differentiates the where and having bits. I would leave those arguments as just FilterExpressions

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -44,6 +44,11 @@
             <groupId>com.yahoo.elide</groupId>
             <artifactId>elide-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-datastore-hibernate</artifactId>
```
aklish: I would rather see a dependency on the JPA store rather than hibernate.  The hibernate stores are probably going away in Elide 5.0

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -0,0 +1,126 @@
...
+ * A {@link QueryEngine} is an abstraction layer that {@link DataStoreTransaction} leverages to optimize query and send
+ * query, such as loading objects, to a persistent storage.
+ * <p>
+ * The purpose of {@link QueryEngine} is to allow a single {@link DataStore} to utilize multiple query frameworks, such
+ * as Hibernate on SQL and Drill on NoSQL shown below:
+ * <pre>
+ *                           +-----------+                      +------------------+
+ *                           |           |                      |                  |
+ *                           | DataStore |              +-------+ DrillQueryEngine |
+ *                           |           |              |       |                  |
```
aklish: I don't completely understand this diagram.  I don't understand the DrillQueryEngine and Apache Drill below.  Why are they separate boxes?  

I would probably remove the diagram as it might cause confusion.

QubitPi: Sorry I made it confusing. Originally they are separate because `DrillQueryEngine` is a `QueryEngine` implementation, while Apache Drill is out of our control.

 I've updated the diagram; however, if you still don't like it I'm happy to remove it then.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -0,0 +1,74 @@
...
+ *             |                      |                        |
+ * +-----------v----------------------v------------------------v----------+
+ * |                             SQL Driver                               |
+ * +-------------------+----------------------------+---------------------+
+ *                     |                            |
+ *               +-----v-----+                  +---v----+
+ *               |           |                  |        |
+ *               |  NoSQL DB |                  | SQL DB |
+ *               |           |                  |        |
+ *               +-----------+                  +--------+
```
QubitPi: Sorry I made it confusing. Originally they are separate because DrillQueryEngine is a QueryEngine implementation, while Apache Drill is out of our control.

I've updated the diagram; however, if you still don't like it I'm happy to remove it then.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Query.java
```diff
@@ -0,0 +1,36 @@
...
+import lombok.Data;
+
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * A {@link Query} is an object representing a query executed by {@link QueryEngine}.
+ */
+@Data
+public class Query {
```
aklish: Can you add https://projectlombok.org/features/Builder

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -0,0 +1,74 @@
...
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.datastores.aggregation;
+
+import com.yahoo.elide.core.DataStore;
+import com.yahoo.elide.core.DataStoreTransaction;
+
+/**
+ * A {@link QueryEngine} is an abstraction layer that {@link DataStoreTransaction} leverages to optimize query and send
```
aklish: I would say something like:

```A QueryEngine is an abstraction that an AggregationDataStore leverages to run analytic queries (OLAP style) against an underlying persistence layer.```

Also - why Drill?   `There will be` implies that we plan on building these things.  I think the Diagram is fine, but I wouldn't imply future support for things we don't currently plan on building.  Maybe substitute Drill with Druid.   Also, I don't think we'll need both Hibernate and JPA.  In In Elide 5.0, the explicit hibernate dependency will be greatly diminished.

PR #872, opened 2019-07-12
Add support for encoding of error responses

### LINKED ISSUE #871: Add option to enable safe HTML encoding for error messages
Elide should provide an option to allow users to automatically encode error responses to be safe for display in HTML.

## Current Behavior
Currently, error messages are not encoded. If an error occurs during a transaction, the cause of that message will be sent as a response, potentially allowing for XSS attacks.
This is true for both GraphQL and JsonApi endpoints.

### JsonApi Behavior
As an example consider the following POST request made to the JsonApi endpoint, where book.id has type long:
```
{
    "data": {
        "type": "book",
        "id": "6<script>alert(1)</script>",
        "attributes": {
        	"title": "new book",
        	"pageCount": "5"
        },
        "relationships": {
          "authors": {
        	"data": [
        		{"type": "author", "id": "2"}
        	]
          }
        }
    }
}
```

The id field for book can not be converted to long in this case, and we get the following error resposne:
```
{
    "errors": [
        "Error converting from 'String' to 'Long' For input string: \"6<script>alert(1)</script>\""
    ]
}
```
### GraphQL Behavior
A similar request being made to the GraphQL endpoint might look like:
```
{
	"query": "mutation {book(op: UPSERT, data:{id: \"6<script>alert(1)</script>\" title: \"bad book\"}){edges{node{id title}}}}"
}
```
And would produce the following error response:
```
{
    "data": null,
    "errors": [
        {
            "message": "Exception while fetching data (/book) : Invalid value: 6<script>alert(1)</script>",
            "locations": [
                {
                    "line": 1,
                    "column": 11
                }
            ],
            "path": [
                "book"
            ]
        }
    ]
}
```

A simple web service may post to Elide and return the response to the UI it is serving, which in this case would cause an alert box to pop up. The example error is only one of several possible errors that could cause a script tag to be reflected back to the user.

## Expected Behavior
`ElideSettings` should support a `.withEncodeErrorResponses(boolean encodeErrorResponses)` builder function which would enable automatic encoding of error messages using an existing encoder (such as the OWASP Java Encoder Project). This would make it easier to prevent certain types of XSS attacks, since error responses may contain things like script tags which lead to the error in the first place. These errors are essentially being reflected back at the user. A UI might want to display these error messages in HTML which would allow for a script to be run.

By doing this for users (at their request) we can improve the security of the project.

In the POST example from above, we would expect the error response to be encoded:

### JsonAPI
```
{
    "errors": [
        "Error converting from &#39;String&#39; to &#39;Long&#39; For input string: &#34;6&lt;script&gt;alert(1)&lt;/script&gt;&#34;"
    ]
}
```

### GraphQL
```
{
    "data": null,
    "errors": [
        {
            "message": "Exception while fetching data (/book) : Invalid value: 6&lt;script&gt;alert(1)&lt;/script&gt;",
            "locations": [
                {
                    "line": 1,
                    "column": 11
                }
            ],
            "path": [
                "book"
            ]
        }
    ]
}
```

If this error message is directly displayed in a UI the script tag would not activate, but would be readable as is.

## Possible Solution
Changes would need to be made to allow both JsonApi and GraphQL endpoints to optionally encode error responses.

#### HttpStatusException
In the `JsonApiEndpoint` most errors are caught during the `handleRequest(...)` call. 
The error responses are then built using that HttpStatusException (or a subclass of it), in the `Elide.buildErrorResponse(HttpStatusException error, boolean isVerbose)` function.

`HttpStatusException` currently implements `getErrorResponse()` and `getVerboseErrorResponse()`. This class and it's subclasses should implement additionally `getErrorResponse(boolean encodeResponse)` and `getVerboseErrorResponse(boolean encodeResponse)`, which provide the same error message, but with error details encoded. 

#### ErrorObjects
If `ElideSettings.isReturnErrorObjects()` is true, JsonNodes represnting the error will be generated and added to the `ElideResponse` via the object mapper. These objects may also contain text which should be encoded.

Currently this is done like:
```
ErrorObjects errors = ErrorObjects.builder().addError()
  .withDetail(isVerbose ? error.getVerboseMessage() : error.toString()).build();
```
If encoding should be performed, we can simply wrap the `error.getVerboseMessage()` and `error.toString()` results in an encoding function.

### 2) GraphQL Solution
The GraphQL solution is a little trickier, as GraphQL itself does much of the work catching errors, and some operations allow for a 200 OK status containing both data and errors if some portion of the request succeeded/failed respectively. Currently, the response of a GraphQL query is generated by the ObjectMapper:
```
return Response.ok(mapper.writeValueAsString(result.toSpecification())).build();
```
#### Normal Query Execution
`ExecutionResult.toSpecification()` is GraphQL's function to return a Map conforming to the API specifications. This includes "data", "errors", and "extensions" objects if they are present in the result. We don't want to encode the entire object, as that would include data elements as well. Ideally, we would only encode the `List<GraphQLError>`. 

To do this, it would be nice to add a serializer to the ObjectMapper which knows how to serialize GraphQLError objects with optional encoding. Unfortunately, when using `toSpecification()` the errors are propagated as a map of strings, so the ObjectMapper would not catch it.

Instead, I am proposing that we change the Response generation to:
```
return Response.ok(mapper.writeValueAsString(result)).build();
```
where `result` is an `ExecutionResult`.

A new `ExecutionResultSerializer` will be created to mimic the `toSpecification()` function, but will have the ability to encode `GraphQLError` messages. This would be added to the exiting ObjectMapper.

#### Mutation Query with Errors
In the case of a mutation query, the commit never happens, and we return a custom response object where "data" is set to null, and "errors' contains a `List<GraphQLError>`. Since we are already talking about a `ExecutionResultSerializer` which can encode GraphQLErrors, that logic could be moved to a `GraphQLErrorSerializer`. Mutation queries would then use the ObjectMapper which knows how to serialize GraphQLError objects, and only the details of the error would be encoded.

The `ExecutionResultSerializer` would then call out to the `GraphQLErrorSerializer` when it comes across GraphQLErrors.

The concern with adding these serializers to the ObjectMapper, is that the ObjectMapper can be provided in the `ElideSettings` object, allowing end-users to provide custom object mappers if they would like to. I am not sure what would happen if this proposed addition of serializers somehow interact/conflict with some custom ObjectMapper. This may be a breaking change in some sense due to this concern.

#### Exceptions
There are a few types of exceptions that get thrown, indicating that something went terribly wrong. For example, the `JsonProcessingException` can get thrown if invalid GraphQL is provided in the query.

These exceptions can be handled in the same way as in the JsonApiEndpoint proposal, as the exceptions are wrapped in an `HttpStatusException` and passed to a `buildErrorResponse()` function (analagous to the function of the same name in the JsonApiEndpoint). So these set of errors should be pretty simple to handle.

These exceptions should also handle the `ElideSettings.isReturnErrorObjects()` setting, the same way the JsonApiEndpoint proposal does.

## Your Environment
Tested with current Elide version 4.4.6-SNAPSHOT

rmihalko:
Related/Resolves #871 
Add optional support for encoding error response messages, which can be enabled through the setting `ElideSettings.encodeErrorResponses`.

This proposed implementation uses the owasp-java-encoder `Encode.forHtml(errorString)` to safely encode strings for display in HTML.

### HttpStatusException
Added two additional public functions to the `HttpStatusException` class which can optionally handle error encoding:
```
public Pair<Integer, JsonNode> getErrorResponse(boolean encodeResponse) {}
public Pair<Integer, JsonNode> getVerboseErrorResponse(boolean encodeResponse) {}
```

If `encodeResponse` is true, the error message will be encoded.

Subclasses of HttpStatusException should implement this function if they want to handle errors differently. The notable exception here is the `CustomErrorException` class, which currently does not implement the encoding function, as it is intended that anyone implementing the `CustomErrorException` as a subclass should handle encoding on their own (if even desired).

### JsonAPI
In the JsonApi endpoint (Elide.java), errors are thrown as exceptions, converted to `HttpStatusExceptions` and passed to the 
`Elide.buildErrorResponse(HttpStatusException error, boolean isVerbose)` function.

This function now additionally checks the `ElideSettings` object to see if errors should be encoded.
If so, the `HttpStatusException` get error response functions are called with the relevant boolean.

Similarly, if `ElideSettings.isReturnErrorObjects()` is true, the messages that are placed in the resulting `JsonNode` are first encoded.

### GraphQL
Two new custom serializers are added to the `ObjectMapper` which is retrieved from the `ElideSettings` object. These serializers will handle conversion of GraphQL `ExecutionResult` and `GraphQLError` objects to strings, which are sent back in the response.

In the case of most normal GraphQL queries, any errors are added to the 200 OK response, alongside any data that may also have been returned from the query. Previously, the `ExecutionResult` retrieved from the GraphQL api was converted to a map using `ExecutionResult.toSpecification()`, which only includes the members of the result object which correspond to the required spec of the GraphQL api. Unfortunately, `toSpecification()` converts the `GraphQLError` objects inside the result to strings for us. To get the serializer to properly encode `GraphQLErrors`, I am instead proposing that the `ExecutionResult` itself is sent to the ObjectMapper, rather than the `toSpecification()` result.

Mutation queries, on error, return an object containing a `List<GraphQLError>` objects. This can be easily converted by the ObjectMapper if it is using the new custom serializer.

Finally, some errors result in exceptions being thrown, which get caught, converted into `HttpStatusException` and passed into `GraphQLEndpoint.buildErrorResponse(...)` (which is analogous to the same function in Elide JsonApi). This can be handled in the same was as the JsonApi endpoint, described above since they rely on the same `HttpStatusException` classes.

## Motivation and Context
Optional encoding of error responses can help in preventing some XSS attacks.
See #871 for more detailed description of current behavior/problem.

## How Has This Been Tested?
Currently this has been tested locally using a simple Elide application with Book and Author models.
Manual testing has been done via send requests with script tags using Postman.
Pending initial discussion of whether this change should be adopted, unit testing should be added before we consider merging.

## Concerns
- Potential issues adding our own serializers to the ObjectMapper (which can be provided via ElideSettings) if the user is already doing something special with the ObjectMapper or other serializers.

## License

The `org.owasp.encoder` dependency which is being included to encode strings for HTML has a `BSD 3-Clause` license, which to my understanding should be acceptable to include in an Apache 2.0 licensed project. Link to owasp-java-encoder: https://github.com/OWASP/owasp-java-encoder

### DIFF
--- a/elide-integration-tests/src/test/resources/EncodedErrorResponsesIT/InvalidAttributeException.req.json
+++ b/elide-integration-tests/src/test/resources/EncodedErrorResponsesIT/InvalidAttributeException.req.json
@@ -1,0 +1,13 @@
+[
+  {
+    "op":"add",
+    "path":"/-",
+    "value":{
+      "type":"parent",
+      "id":"nice",
+      "attributes":{
+        "badAttr":"no val"
+      }
+    }
+  }
+]
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,9 @@
 ## 4.4.6
 **Features**
  * Issue #815.  Added the ability to customize the JPQL generation for a filter operator globally or for a specific entity attribute.
+ * Issue #871. Add ElideSettings property `encodeErrorResponses`, which when enabled will encode error messages to be safe for HTML. This works for both JSONAPI and GraphQL endpoints, with verbose errors or error object settings enabled/disabled.
+ * HttpStatusException class now supports the following additional functions: `getErrorResponse(boolean encodeResponse)` and `getVerboseErrorResponse(boolean encodeResponse)`
+ * Add `GraphQLErrorSerializer` and `ExecutionResultSerializer` which are added to the `ObjectMapper` provided by the ElideSettings. These are used to parse the GraphQL results, instead of using `ExecutionResult#toSpecification`.
 
 **Fixes**
  * Run vulnerability check during build.  Updated dependencies to fix CVE-2018-1000632, CVE-2017-15708, CVE-2019-10247
--- a/elide-core/pom.xml
+++ b/elide-core/pom.xml
@@ -144,6 +144,12 @@
             <version>0.9.11</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.owasp.encoder</groupId>
+            <artifactId>encoder</artifactId>
+            <version>1.2.2</version>
+        </dependency>
+
         <!-- Test -->
         <dependency>
             <groupId>ch.qos.logback</groupId>
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -39,6 +39,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import org.antlr.v4.runtime.misc.ParseCancellationException;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
+import org.owasp.encoder.Encode;
 
 import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
@@ -266,13 +267,20 @@ public class Elide {
         if (error instanceof InternalServerErrorException) {
             log.error("Internal Server Error", error);
         }
+
+        boolean encodeErrorResponse = elideSettings.isEncodeErrorResponses();
         if (!(error instanceof CustomErrorException) && elideSettings.isReturnErrorObjects()) {
-            ErrorObjects errors = ErrorObjects.builder().addError()
-                    .withDetail(isVerbose ? error.getVerboseMessage() : error.toString()).build();
+            String errorDetail = isVerbose ? error.getVerboseMessage() : error.toString();
+            if (encodeErrorResponse) {
+                errorDetail = Encode.forHtml(errorDetail);
+            }
+            ErrorObjects errors = ErrorObjects.builder().addError().withDetail(errorDetail).build();
             JsonNode responseBody = mapper.getObjectMapper().convertValue(errors, JsonNode.class);
             return buildResponse(Pair.of(error.getStatus(), responseBody));
         }
-        return buildResponse(isVerbose ? error.getVerboseErrorResponse() : error.getErrorResponse());
+
+        return buildResponse(isVerbose ? error.getVerboseErrorResponse(encodeErrorResponse)
+                : error.getErrorResponse(encodeErrorResponse));
     }
 
     protected ElideResponse buildResponse(Pair<Integer, JsonNode> response) {
--- a/elide-core/src/main/java/com/yahoo/elide/ElideSettings.java
+++ b/elide-core/src/main/java/com/yahoo/elide/ElideSettings.java
@@ -40,4 +40,5 @@ public class ElideSettings {
     @Getter private final int updateStatusCode;
     @Getter private final boolean returnErrorObjects;
     @Getter private final Map<Class, Serde> serdes;
+    @Getter private final boolean encodeErrorResponses;
 }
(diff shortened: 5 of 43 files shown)


elide-integration-tests/src/test/resources/EncodedErrorResponsesIT/InvalidAttributeException.req.json
```diff
@@ -0,0 +1,13 @@
+[
```
aklish: This was the old way of doing tests in Elide.  There is a JSON-API DSL that is much easier to read where the bodies remain inside the tests (rather than buried in some files).

PR #877, opened 2019-07-15
Hydrate Relationship

QubitPi:
Add on to https://github.com/yahoo/elide/branches/all?utf8=%E2%9C%93&query=CalciteQueryEngine

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLQueryEngine.java
```diff
@@ -175,10 +184,140 @@ public SQLQueryEngine(EntityManager entityManager, EntityDictionary dictionary)
...
+            String joinField = entry.getKey();
+            List<Object> joinFieldIds = entry.getValue();
+
+            List<Object> relationshipValues = getRelationshipValues(query.getEntityClass(), joinField, joinFieldIds);
+
+            for (int i = 0; i < relationshipValues.size(); i++) {
+                Object entityInstance = queryResults.get(i);
+                dictionary.setValue(entityInstance, joinField, relationshipValues.get(i));
+            }
+        }
```
QubitPi: If we would like to generalize it, we could move this part into a parent class and have `SQLQueryEngine` implement  `getRelationshipValues` below

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLQueryEngine.java
```diff
@@ -127,6 +135,7 @@ public SQLQueryEngine(EntityManager entityManager, EntityDictionary dictionary)
 
         if (!dimensionProjections.isEmpty()) {
             projectionClause = projectionClause + "," + dimensionProjections.stream()
+                .map(name -> getMappedJoinColumn(query.getEntityClass(), name))
```
aklish: The physical name can come either form `@Column` or from `@JoinColumn`.  I think my current version is correct.  This is based on my old code.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/Country.java
```diff
@@ -52,4 +55,33 @@ public String getName() {
     public void setName(final String name) {
         this.name = name;
     }
+
+    @Override
+    public boolean equals(final Object other) {
```
aklish: Instead of generating these, can we just use lombok?  We should probably do that for all the schema, dimension, metric classes as well.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLQueryEngine.java
```diff
@@ -175,10 +185,174 @@ public SQLQueryEngine(EntityManager entityManager, EntityDictionary dictionary)
...
+     * <p>
+     * If the ID list is empty or no matching objects are found, this method returns {@link Collections#emptyList()}.
+     *
+     * @param entityClass  The type of relationship
+     * @param joinField  The relationship field name
+     * @param joinFieldIds  The specified list of join ID's against the relationshiop
+     *
+     * @return a list of hydrating values
+     */
+    private List<Object> getRelationshipValues(Class<?> entityClass, String joinField, List<Object> joinFieldIds) {
```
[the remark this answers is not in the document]
aklish: We need to add more structure (in terms of classes).  Without some structure, procedural code can be hard to understand.

In coerceObjectToEntity, replace:

```java
            if (dim != null && dim.getDimensionType() == DimensionType.ENTITY) {
                //We don't hydrate relationships here.
                continue;
            }
```

With Something like this.

```java

            if (dim != null && dim.getDimensionType() == DimensionType.ENTITY) {

                stitchList.todo(entityInstance, fieldName, value)
                //We don't hydrate relationships here.
                continue;
            }
```

coerceObjectToEntity should probably to move to its own common class 

```java
class EntityHydrator { 
    private StitchList stitchList;
 
    
    public Object hydrateEntitity(...) {
        //This is where the logic to coerceObjectToEntity should live.
    }
}
```
```java
class StichList {

   class Todo {
      Object entityInstance;
      String relationshipName;
      Object foreignKey;
   }
   
   //Maps an entity class to a Map of object ID to object instance.
   private Map<Class<?>, Map<Object, Object>> objectLookups;
   
   //Maps an entity class to a Map of field names to a list of IDs.
   private List<Todo> toDoList;
   
   
   public todo(Object entityInstance, String fieldName, Object value) {
       
       // Stick this info on the todo list.
       toDoList.put(....);
   }

   public List<Object> stitch() {
      populateObjectLookupTable(...);


      for each todo {
         lookupRelationship(...)
         dictionary.setValue(...)
      }
   }

   private Set<Object> lookupById(Class<?> lookupClass, List<
}
```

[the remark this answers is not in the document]
aklish: I was trying to express the basic idea as pseudocode.  But this structure conveys meaning that is missing from large blocks of logic.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLQueryEngine.java
```diff
@@ -175,10 +185,174 @@ public SQLQueryEngine(EntityManager entityManager, EntityDictionary dictionary)
...
+        // mapping: relationship field name -> join ID's
+        Map<String, List<Object>> hydrationIdsByRelationship = new HashMap<>();
+        List<Object> queryResults = results.stream()
                 .map((result) -> { return result instanceof Object[] ? (Object []) result : new Object[] { result }; })
-                .map((result) -> coerceObjectToEntity(query, result, counter))
+                .map((result) -> coerceObjectToEntity(query, result, counter, hydrationIdsByRelationship))
                 .collect(Collectors.toList());
+
+        // hydrate each relationship
+        for (Map.Entry<String, List<Object>> entry : hydrationIdsByRelationship.entrySet()) {
```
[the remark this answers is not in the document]
QubitPi: What would make it a better choice other than consuming more memory by functional object?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/Country.java
QubitPi: Done for this class.

For others such as dimension and metric, looks like lombok is not happy with inheritance.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/StitchList.java
```diff
@@ -0,0 +1,162 @@
...
+    @Getter(AccessLevel.PRIVATE)
+    private final EntityDictionary entityDictionary;
+
+    /**
+     * Constructor.
+     *
+     * @param entityDictionary  An object that sets entity instance values and provides entity metadata info
+     */
+    public StitchList(EntityDictionary entityDictionary) {
+        this.objectLookups = new ConcurrentHashMap<>();
```
aklish: These classes will always be accessed from a single thread.  I don't think we need the overhead of a concurrent hash map.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/AbstractEntityHydrator.java
```diff
@@ -0,0 +1,198 @@
...
+
+        List<Object> queryResults = getResults().stream()
+                .map((result) -> { return result instanceof Object[] ? (Object []) result : new Object[] { result }; })
+                .map((result) -> coerceObjectToEntity(result, counter))
+                .collect(Collectors.toList());
+
+        if (getStitchList().shouldStitch()) {
+            // relationship is requested, stitch relationship then
+            populateObjectLookupTable();
+            getStitchList().stitch();
```
hellohanchen: As the `getStitchList().stitch()` always follows `populateObjectLookupTable()`, should this be moved into `populateObjectLookupTable()`?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/StitchList.java
```diff
@@ -0,0 +1,158 @@
...
+        this.entityDictionary = entityDictionary;
+    }
+
+    /**
+     * Returns whether or not the entity instances in this {@link StitchList} have relationships that are unset.
+     *
+     * @return {@code true} if the entity instances in this {@link StitchList} should be further hydrated because they
+     * have one or more relationship fields.
+     */
+    public boolean shouldStitch() {
```
hellohanchen: maybe `needPopulate/needStitch`?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/StitchList.java
```diff
@@ -0,0 +1,158 @@
...
+public final class StitchList {
+
+    /**
+     * A representation of an TODO item in a {@link StitchList}.
+     */
+    @Data
+    public static class Todo {
+        private final Object entityInstance;
+        private final String relationshipName;
+        private final Object foreignKey;
```
hellohanchen: Can we change these to `public final` as I don't think we would need to set them?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/AbstractEntityHydrator.java
```diff
@@ -0,0 +1,198 @@
...
+        return queryResults;
+    }
+
+    /**
+     * Coerces results from a {@link Query} into an Object.
+     *
+     * @param counter Monotonically increasing number to generate IDs.
+     * @return A hydrated entity object.
+     */
+    protected Object coerceObjectToEntity(Object[] result, MutableInt counter) {
```
hellohanchen: For the FiliDataStore I am working on, I would like this method to be abstract as the result fields returned from Fili might not be in the same order as the original request. The projects-result matching might not work.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/AbstractEntityHydrator.java
```diff
@@ -0,0 +1,198 @@
...
+                .map(Metric::getName)
+                .collect(Collectors.toList());
+
+        projections.addAll(query.getDimensions().stream()
+                .map(Dimension::getName)
+                .collect(Collectors.toList()));
+
+        Preconditions.checkArgument(result.length == projections.size());
+
+        SQLSchema schema = (SQLSchema) query.getSchema();
```
hellohanchen: I think we should just use `Schema` here instead of `SQLSchema`

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/StitchList.java
QubitPi: why

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/StitchList.java
QubitPi: Publisizing non-immutables is dangerous unless there is a compelling reason

aklish:
This work has already been merged into elide-5.x.

PR #878, opened 2019-07-17
SQL Query Engine 

### LINKED ISSUE #874: AggregationStore: Create Calcite QueryEngine
BLOCKED BY IT Test Refactor

JPA has no convenient mechanism that let's Elide both:
1. Define an arbitrary SQL expression that populates an entity.
2. Modify the query in some way by pushing down filter predicates or making it a subselect of another query.

Rather than only use JPA (which is limited also by JPQL), we will instead define a set of new annotations for the Elide AggregationDataStore.

### @FromTable
Elide annotation that defines the table that the Elide entity will conform to.

The generated SQL `FROM` clause will refer to this table (or joins with this table).

### @FromSubquery
Elide annotation that defines the subquery that the Elide entity will conform to.

The generated SQL `FROM` clause will refer to this subquery (or joins against it).

### @Column

Elide will reuse `javax.persistence.Column`.   The only annotation properties Elide will leverage are:
1. *name* - To extract the actual column name in the database.
2. *precision* - Only needed for big decimal attributes.
3. *scale* - Only needed for big decimal attributes.

### @ToOne @ToMany

For relationships, Elide will leverages its own annotations.

### Entity Hydration

One of the responsibilities of subclasses of the `QueryEngine` interface is to hydrate entities.

Elide will create instances of the appropriate entity via reflection (no constructor arguments).

Static entity attributes will be hydrated by calling the `set` method of the associated attribute or relationship in the entity.  The argument types are limited
to those that SQL supports (`DataType`).  The JDBC `ResultSetMetadata` will inform what type to construct.  The newly constructed instance will then be passed to
the setter method.

Static entity relationships will be hydrated by the base class AggregationStore (one of its core responsibilities).

### SQL Form

Elide will generate SQL in the following form using [Calcite's SQL Dialect](https://calcite.apache.org/apidocs/org/apache/calcite/sql/SqlNode.html#toSqlString(org.apache.calcite.sql.SqlDialect):

`SELECT ${field_list} FROM ${table_spec} WHERE ${filter_spec} GROUP BY ${dimension_list} HAVING ${having_spec} ORDER BY ${order clause} OFFSET ${offset} LIMIT ${limit}`

Not every query will have every element.

#### Table Aliases

Every table will be given an alias of the fully qualified entity name where periods are replaced with underscores.

`com.elide.Foo` would become `com_elide_Foo`.

#### Field Aliases

Every field will be given an alias which maps to the field name in the Elide entity.

For example:

```java
package com.elide;

@FromTable(name = "foo")
class Foo {

   @Column(name = "_BAR")
   private int bar;
}
```

Would expand to:

`SELECT com_elide_foo._BAR as bar FROM foo AS com_elide_foo;`

#### Table Specification

The `FROM` clause will either be:
1. A single table and its alias: `FROM ${Table Name} AS ${Table Alias}`
2. A subquery and its alias: `FROM ($Subquery) AS ${Table Alias}`
3. Either of the above left joined with one or more dimension tables that require either filtering or sorting: FROM (${Subquery}) AS ${Query Alias} LEFT JOIN Table AS ${Table Alias}`

#### WHERE Filter Expansion

Elide filters have a dot (.) separated path that uniquely qualifies a field in the entity relationship graph ('book.author.firstName').
If there is more than a single separator, this means an entity relationship must be traversed (requiring a join in the database).

Elide will support traversing to-one relationships by creating left joins in the generated SQL.  To-many relationship joins cause issues with pagination.

#### Sort Expansion

Elide sort expressions have a dot (.) separated path that uniquely qualifies a field in the entity relationship graph ('book.author.firstName').
If there is more than a single separator, this means an entity relationship must be traversed (requiring a join in the database).

Elide will support traversing to-one relationships by creating left joins in the generated SQL.  To-many relationship joins cause issues with pagination.

### SQL Dialects

It is the expectation that the SQL query in `@FromSQL` is written in the native dialect understood by the database.

Elide also generates SQL statements, but those are always in the [Calcite SQL dialect](https://calcite.apache.org/docs/reference.html).

To generate the final, native SQL, Elide will perform the following steps: 
1. Generate a template SQL query where:
  1. metric aggregations ( SUM(xyz) ) are substituted with template variable names (${metric name}).
  2. The SQL from `@FromSQL` is substituted with a template variable name (SUBQUERY).
2. The resulting SQL is parsed with Calcite's [SQLParser](https://calcite.apache.org/apidocs/org/apache/calcite/sql/parser/SqlParser.html) into an abstract syntax tree.
3. The abstract syntax tree is translated into the target dialect using [SQLNode](https://calcite.apache.org/apidocs/org/apache/calcite/sql/SqlNode.html).
4. The target SQL string template is substituted with metric aggregation fragments and any subqueries.

aklish:
Resolves #874

## Description
Implements everything in #874 except it does not use Calcite.  While Calcite may be necessary down the road, my initial thoughts is that the SQL that Elide generates is so basic that it likely doesn't need dialect translation (other than simple things like quotations). 

## Motivation and Context
Elide can then do both OLTP and OLAP style queries.

## How Has This Been Tested?
Generates real SQL queries against test data in a H2 db.

## License


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1540,24 +1435,8 @@ private boolean isValidParameterizedMap(Map<?, ?> values, Class<?> keyType, Clas
...
-                return ((Field) accessor).get(target);
-            }
-        } catch (IllegalAccessException e) {
-            throw new InvalidAttributeException(fieldName, dictionary.getJsonAliasFor(target.getClass()), e);
-        } catch (InvocationTargetException e) {
-            throw handleInvocationTargetException(e);
-        }
-        throw new InvalidAttributeException(fieldName, dictionary.getJsonAliasFor(target.getClass()));
+
+        return dictionary.getValue(target, fieldName, requestScope);
```
hellohanchen: `requestScope.getDictionary().getValue(target, fieldName, requestScope);`

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLQueryEngine.java
```diff
@@ -0,0 +1,494 @@
...
+            builder.whereClause("WHERE " + translateFilterExpression(schema, query.getWhereFilter(),
+                    this::generateWhereClauseColumnReference));
+        }
+
+        if (query.getHavingFilter() != null) {
+            builder.havingClause("HAVING " + translateFilterExpression(schema, query.getHavingFilter(),
+                    (predicate) -> { return generateHavingClauseColumnReference(predicate, query); }));
+        }
+
+        if (! query.getDimensions().isEmpty())  {
```
hellohanchen: `(!query.getDimensions().isEmpty())`, there was a space after `!`

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLQueryEngine.java
```diff
@@ -0,0 +1,494 @@
...
+    /**
+     * Converts a filter predicate into a SQL HAVING clause column reference.
+     * @param predicate The predicate to convert
+     * @return A SQL fragment that references a database column
+     */
+    private String generateHavingClauseColumnReference(FilterPredicate predicate, Query query) {
+        Path.PathElement last = predicate.getPath().lastElement().get();
+        Class<?> lastClass = last.getType();
+
+        if (! lastClass.equals(query.getSchema().getEntityClass())) {
```
rmihalko: Extra space, Han had a similar comment earlier in this PR that should be addressed as well:
```suggestion
        if (!lastClass.equals(query.getSchema().getEntityClass())) {
```

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/AggregatedMetric.java
```diff
@@ -71,12 +42,13 @@ public static String functionFormat(Class<? extends Aggregation> aggregationType
      * @param aggregations  A list of all supported aggregations on this {@link Metric}
      */
     public AggregatedMetric(
+            Schema schema,
```
rmihalko: Javadoc should be updted to include `@schema`

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Column.java
```diff
@@ -36,11 +39,12 @@
      *
      * @throws NullPointerException if {@code field} or {@code fieldType} is {@code null}
      */
-    public Column(String field, Meta annotation, Class<?> fieldType) {
+    public Column(Schema schema, String field, Meta annotation, Class<?> fieldType) {
```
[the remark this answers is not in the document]
aklish: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/dimension/DegenerateDimension.java
```diff
@@ -56,6 +57,7 @@ public static ColumnType parseColumnType(String dimensionField, Class<?> cls, En
      * @throws NullPointerException any argument, except for {@code annotation}, is {@code null}
      */
     public DegenerateDimension(
+            Schema schema,
```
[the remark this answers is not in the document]
aklish: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/dimension/EntityDimension.java
```diff
@@ -125,13 +126,14 @@ protected static CardinalitySize getDefaultCardinality() {
      * @throws NullPointerException any argument, except for {@code annotation}, is {@code null}
      */
     public EntityDimension(
+            Schema schema,
```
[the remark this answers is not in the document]
aklish: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/AggregatedMetric.java
aklish: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLQueryEngine.java
aklish: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLQueryEngine.java
aklish: 👍

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
aklish: 👍

PR #888, opened 2019-07-25
Added swagger setup in elide standalone

### LINKED ISSUE #851: Make it simpler to setup swagger with ElideStandalone
Elide Standalone doesn't have a simple mechanism to configure swagger.  

## Expected Behavior
There should be a function inside ElideStandaloneSettings:

```java    
default Map<String, Swagger> enableSwagger() {
        return new HashMap<>();
}
```

If the map is not empty, ElideStandalone should setup the swagger endpoint with the provided map.

## Current Behavior & Context
To enable this for Elide Standalone, you'll need to modify the Jetty servlet initialization. There is a hook to do that in ElideStandaloneSettings:

```java
     * A hook to directly modify the jetty servlet context handler as necessary.
     *
     * @param servletContextHandler ServletContextHandler in use by Elide standalone.
     */
    default void updateServletContextHandler(ServletContextHandler servletContextHandler) {
        // Do nothing
    }
```

You would do something like this inside that function:

```java
ServletHolder jerseyServlet = context.addServlet(ServletContainer.class, "/your/path/goes/here"); jerseyServlet.setInitOrder(0); jerseyServlet.setInitParameter("jersey.config.server.provider.packages", "com.yahoo.elide.contrib.swagger.resources"); jerseyServlet.setInitParameter("javax.ws.rs.Application", ElideResourceConfig.class.getCanonicalName());
```

There is an additional function in ElideStandaloneSettings that let's you configure the ResourceConfig (and setup dependency injection):

```java
/**
 * Supplemental resource configuration for Elide application. This should be a fully qualified class name.
 *
 * Before calling the consumer, the class will be injected by the ServiceLocator.
 *
 * @return Default: null
 */
default Consumer<ResourceConfig> getApplicationConfigurator() {
    // Do nothing by default
    return (x) -> {};
}
```

You can use this to register your own additional DependencyBinder (that Jetty will then use for DI).

This is where you would construct the swagger doc, stick it in a map, and bind it to the named parameter that the swagger endpoint uses.

Chandrasekar-Rajasekar:
Resolves #851  (if appropriate)

## Description
enableSwagger can be overridden to put swagger docs for which endpoint will be created. 

## Motivation and Context
Swagger can be enabled by configuring setting for ElideStandalone

## How Has This Been Tested?
No tests yet
## License


elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -204,6 +216,18 @@ default String getHibernate5ConfigPath() {
      * @param servletContextHandler ServletContextHandler in use by Elide standalone.
      */
     default void updateServletContextHandler(ServletContextHandler servletContextHandler) {
-        // Do nothing
+        ServletHolder jerseyServlet = servletContextHandler.addServlet(ServletContainer.class, "/swagger");
```
aklish: We should do this in the core code rather than the extension point.  Look at how we optionally configure JSON-API and GraphQL.  Swagger should be similar.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -181,11 +197,23 @@ default boolean enableServiceMonitoring() {
      *
      * Before calling the consumer, the class will be injected by the ServiceLocator.
      *
-     * @return Default: null
+     * @return Default: register the DependencyBinder that binds swagger map to named endpoint if swagger is not configured.
      */
     default Consumer<ResourceConfig> getApplicationConfigurator() {
-        // Do nothing by default
-        return (x) -> {};
+        // Do nothing if swagger map is empty
```
aklish: Let's move this logic to ElideResourceConfig.

changelog.md
```diff
@@ -3,6 +3,7 @@
 **Features**
  * Issue #815.  Added the ability to customize the JPQL generation for a filter operator globally or for a specific entity attribute.
  * Alpha release of a new Elide data store (SearchDataStore) that supports full text search on top of an existing data store.
+ * Issue #851. Added new method `enableSwagger()` in `ElideStandaloneSettings` class which allows an easier way for binding swagger docs to the given endpoint. Override this method returning the `Map<String, Swagger>` object to bind the docs.
```
aklish: Let's move to 4.5.1 (see updated changelog).

PR #893, opened 2019-07-29
Refactor EntityHydrator

hellohanchen:
- Make private fields protected so that children classes can have different constructor
- Change results from `List<Object>` to `List<Map<String, Object>>`


hellohanchen:
> I'm going to put a hold on this branch and let @aklish, who owns coereceObject, decide. This change assumes less responsibilities from query engine (@aklish).
> 
> I'll remove my block if @aklish approves it.

Thank you Jia Qi. @aklish I made this PR as the result return by fili might not match the original field order (as the request is being reconstructed) and Fili is already using the concept of **resultRow**. As we also need to construct the `projections`, I think we can just use **resultRow** in here to make both Elide and Fili happy.

PR #894, opened 2019-08-01
Elide-Testing Junit5 in elide core

hellohanchen:
- Update elide-core testing from testng to Junit 5
  - Update annotations
    - make test `setup()/init()` methods to static as Junit 5 `@BeforeAll` annotation requires static method
  - Switch order of expected value and actual value for `assertEquals()`
- Remove maven dependency in elide-core to elide-example-models so that elide-example-models can be the module that contains all testing beans for other modules.


elide-core/pom.xml
```diff
@@ -43,16 +43,15 @@
...
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>com.yahoo.elide</groupId>
             <artifactId>elide-annotations</artifactId>
         </dependency>
-        <dependency>
-            <groupId>com.yahoo.elide</groupId>
-            <artifactId>elide-example-models</artifactId>
```
aklish: The point of elide-example-models was to move the common models from core and integration tests to a central place.  This works against that.  Let's leave example models alone for now.

elide-core/src/test/java/com/yahoo/elide/audit/LogMessageTest.java
```diff
@@ -69,30 +71,34 @@ public void verifyListExpressions() {
...
+                () -> new LogMessage("{}", childRecord, expressions, 1, Optional.empty()).getMessage());
     }
 
     public static class TestLoggerException extends RuntimeException {
     }
 
     private AuditLogger testAuditLogger = new Slf4jLogger();
 
-    @Test(threadPoolSize = 10, invocationCount = 10)
+    @Test
```
aklish: The point of this tests was to test thread safety.  This is not an equivalent test.

PR #895, opened 2019-08-01
Elide-Testing Move Test Beans from Integration-Test to Example-Models

hellohanchen:
Move test beans from `elide-integration-test` module to `elide-example-models` module to release the dependencies between all elide modules to `elide-integration-test` so that `elide-integration-test` can use all other modules as dependencies. In this way, the `elide-integration-test` module can use all other modules to setup integration tests.
Current module dependency tree
- elide-annotations
- elide-core
- elide-graphql
- elide-example-models
- all other modules
- elide-integration-tests


hellohanchen:
This PR is following https://github.com/yahoo/elide/pull/894 and should be merged after that one. I don't know why the continuous-integration check fails but it works locally on my Mac. I wonder whether that's just a dependency issue about the new test-jar for `elide-example-models`

Issue #890, opened 2019-07-26
Aggregation Store: Hydrate GraphQL Schema with parameterized attributes


aklish:
The Aggregation Store requires metrics and some dimensions to take arguments.  These arguments need to be:
1. Exposed in the GraphQL Schema
2. Added to the `EntityProjection` so that the Aggregation DataStoreTransaction can leverage them when buildings its queries.

## Possible Solution

The `EntityDictionary` should have new methods for adding arguments to attributes.

The `AggregationDataStore` should populate the dictionary with these attributes in the function `populateEntityDictionary`.   The store can leverage the new Schema class to determine which attributes to add.

For now, we should add the following attributes to the schema;
1. Metrics with the `@MetricAggregation` annotation should take a parameter for the name of the aggregation to compute.
2. Time based dimensions with `@Temporal` annotation should take grain and timeZone parameters.

This work should be based on the branch AddDataCollectionClassHierarchy that includes @QubitPi changes for GraphQL.

QubitPi:
> This work should be based on the branch AddDataCollectionClassHierarchy that includes @QubitPi changes for GraphQL.

https://github.com/yahoo/elide/pull/896

PR #896, opened 2019-08-05
Transform GraphQL query to EntityProjection

QubitPi:
Resolves https://github.com/yahoo/elide/issues/890 & https://github.com/yahoo/elide/issues/823

## Description
GraphQL implementation of https://github.com/yahoo/elide/issues/890
## Motivation and Context

https://github.com/yahoo/elide/issues/823

## License


elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,205 @@
...
+ * Read-Write GraphQL query should be processed by GraphQL API instead({@link GraphQL#execute(String)}).
+ *
+ * @see EntityProjection
+ *
+ * TODO - transform sorting
+ * TODO - transform pagination
+ */
+@Slf4j
+@RequiredArgsConstructor
+public class GraphQLEntityProjectionMaker extends GraphqlBaseVisitor<Void> {
```
QubitPi: Although there are 2 approaches to map a GraphQL query to EntityProjection

1. query -> GraphQL Visitor -> Document -> EntityProjection, or
2. query -> GraphQL Visitor -> EntityProjection,

the 2nd approach is used, because

* The 1st approach has unnecessary overhead of constructing `graphql.language.Document`
* As an uncontrollable abstraction layer, `graphql.language.Document` is going to make it difficult to deeply integrate GraphQL into `EntityProjection` in the future
* If GraphQL has a breaking-change on language grammar/syntax, it is very likely that `GraphQLEntityProjectionMaker` will decompile, allowing us to make corresponding changes at compile time, rather than at runtime. This benefit is less likely by `graphql.language.Document`

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -677,6 +677,23 @@ public RelationshipType getRelationshipType(Object entity, String relation) {
...
     }
 
+    /**
+     * Get a bound type for a specific entity type name
+     *
+     * @param identifier  The entity type name in string
+     *
+     * @return Type of entity
+     */
+    public Class<?> getType(String identifier) {
```
aklish: There is already a method called getEntityClass which does the same thing.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,457 @@
...
+                || ConnectionContainer.PAGE_INFO_KEYWORD.equals(nodeName)
+        ) {
+            // do nothing on "edges", "node", or "pageInfo" field 
+            // see https://graphql.org/learn/pagination/#pagination-and-edges
+            // https://elide.io/pages/guide/11-graphql.html#relationship-metadata
+            return super.visitField(ctx);
+        }
+
+        // on relationship field, change "authors" to "author" so we can look it up in entity dictionary
+        String entityName = singularized(nodeName);
```
aklish: I don't think this it is safe to assume any modifications to a relationship name.  You can look up the type by a combination of the parent class and the relationship name.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,457 @@
...
+     * this list.
+     */
+    @Getter(AccessLevel.PRIVATE)
+    private final List<EntityProjection> allEntityProjections = new ArrayList<>();
+
+    /**
+     * A map from EntityProjection.type to the EntityProjection.
+     */
+    @Getter(AccessLevel.PRIVATE)
+    private final Map<Class<?>, EntityProjection> projectionsByType = new HashMap<>();
```
aklish: I don't think it is safe to a type will show up uniquely once in a query. We have to assume that a type can be asked for multiple times in a single request - each which a different projection.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
aklish: There is another approach too:

https://stackoverflow.com/questions/48909715/access-selection-set-of-a-graphql-query

From the Document (obtained from the parse tree), you can access the selection sets for each root query and just traverse those to build the EntityProjection.  That might be the simplest approach.

QubitPi: There will be lots if-else type casting in this way and the if-else approach still binds closely to ANTLR grammar.

aklish:
This work is already merged into elide-5.x

PR #897, opened 2019-08-08
Refactor IT Tests (ResourceIT and test infrastructure).

aklish:
## Description
1. Migrates ResourceIT to Junit 5
2. Creates a new concept called DataStoreTestHarness which is responsible for initializing a store and cleaning test data between tests.  Every store that expects IT tests to run against it needs one.
3. Refactored ResourceIT so that each test is independent.
4. Refactored ResourceIT so most tests use the JsonApiDSL to construct request and response bodies.
5. Made many fixes to HashMapDataStore so it can run most (if not all) IT tests.
6. Moved get and set value methods to entity dictionary.
7.  By default, if the system property to select a test harness has not been set, the InMemoryStoreHarness will be used.  This allows IT tests to run in an IDE without any special configuration.

## Motivation and Context
IT Tests stomp on one another non-deterministically.  TestNG is not great with a shared DB.  Elide IT tests need a major refactor.

## How Has This Been Tested?
Rerunning the rewritten IT tests.

## License


pom.xml
```diff
@@ -512,6 +512,7 @@
                 <configuration>
                     <failBuildOnCVSS>6</failBuildOnCVSS>
                     <skipSystemScope>true</skipSystemScope>
+                    <assemblyAnalyzerEnabled>false</assemblyAnalyzerEnabled>
```
wcekan: @aklish I pulled this in to speed up the build.  No need to scan .NET on Linux.

PR #899, opened 2019-08-08
To-one Relation Test

QubitPi:
💥 DO NOT MERGE THIS PR UNTIL https://github.com/yahoo/elide/pull/903 is merged

Resolves a test issue.

In a [to-one relationship test](https://github.com/yahoo/elide/blob/RefactorResourceIT/elide-contrib/elide-test-helpers/src/test/java/com/yahoo/elide/contrib/testhelpers/jsonapi/JsonApiDSLTest.java#L50), I found the relationship returns [an array](https://github.com/yahoo/elide/blob/RefactorResourceIT/elide-contrib/elide-test-helpers/src/test/java/com/yahoo/elide/contrib/testhelpers/jsonapi/JsonApiDSLTest.java#L53)  of resource identifier objects. This is because [relation()](https://github.com/yahoo/elide/blob/RefactorResourceIT/elide-contrib/elide-test-helpers/src/test/java/com/yahoo/elide/contrib/testhelpers/jsonapi/JsonApiDSLTest.java#L63) creates to-many relationship by default.

## License

### DIFF
--- a/elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/jsonapi/JsonApiDSL.java
+++ b/elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/jsonapi/JsonApiDSL.java
@@ -232,28 +232,6 @@ public class JsonApiDSL {
     }
 
     /**
-     * Creates a to-one relationship
-     *
-     * @param field           the field
-     * @param resourceLinkage the resource linkage
-     * @return the relation
-     */
-    public static Relation toOneRelation(String field, ResourceLinkage... resourceLinkage) {
-        return new Relation(field, true, resourceLinkage);
-    }
-
-    /**
-     * Creates a to-many relationship
-     *
-     * @param field           the field
-     * @param resourceLinkage the resource linkage
-     * @return the relation
-     */
-    public static Relation toManyRelation(String field, ResourceLinkage... resourceLinkage) {
-        return new Relation(field, false, resourceLinkage);
-    }
-
-    /**
      * Relation relation.
      *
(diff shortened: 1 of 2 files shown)


elide-contrib/elide-test-helpers/src/test/java/com/yahoo/elide/contrib/testhelpers/jsonapi/JsonApiDSLTest.java
```diff
@@ -60,7 +60,7 @@ public void verifyRequestWithOneToOneRelationship() {
                                 attr("title", "title")
                         ),
                         relationships(
-                                relation("author",
+                                toOneRelation("author",
```
aklish: I would replace with 
```java
relation("author", TO_ONE
```

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/jsonapi/JsonApiDSL.java
```diff
@@ -231,6 +231,28 @@ public static Relation relation(String field, ResourceLinkage... resourceLinkage
...
     }
 
+    /**
+     * Creates a to-one relationship
+     *
+     * @param field           the field
+     * @param resourceLinkage the resource linkage
+     * @return the relation
+     */
+    public static Relation toOneRelation(String field, ResourceLinkage... resourceLinkage) {
```
aklish: I had considered adding functions like this but opted instead to add the TO_ONE, TO_MANY params.  I think we should get a second opinion on which way to go.  I prefer the params and no new functions.  @jkusa ?

PR #903, opened 2019-08-09
GraphQL - Implement query DSL & Remove Groovy tests

QubitPi:

## License


elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
```diff
@@ -0,0 +1,424 @@
...
+    /**
+     * Creates a query spec that represents a object value.
+     * <p>
+     * For example, "id:\"1\", title:\"update title\"" as in "{ id:\"1\", title:\"update title\" }".
+     *
+     * @param object  A string representation of the object wrapped
+     *
+     * @return a GraphQL spec
+     */
+    public static ValueWithVariable objectValueWithVariable(String object) {
```
aklish: I wonder if we could use Jackson and lombok builders here to serialize something like:

```java
Book.builder()
   .id("123")
   .title("xyz")
   .build()
```

If we have to test very complex object graph insertion cases, it will be hard to do that with just a String.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/relayjsonapi/RelayJsonApiDSL.java
```diff
@@ -0,0 +1,47 @@
...
+import com.yahoo.elide.contrib.testhelpers.jsonapi.elements.Attributes;
+import com.yahoo.elide.contrib.testhelpers.relayjsonapi.elements.Data;
+import com.yahoo.elide.contrib.testhelpers.relayjsonapi.elements.Edges;
+import com.yahoo.elide.contrib.testhelpers.relayjsonapi.elements.Node;
+import com.yahoo.elide.contrib.testhelpers.relayjsonapi.elements.Resource;
+
+import java.util.Arrays;
+import java.util.Collections;
+
+public class RelayJsonApiDSL {
```
aklish: It looks like we have two different DSLs - one based on selections and one based on nodes and edges.  I think it would be cleaner if there was a single DSL.  I would opt to combine the two but maybe keep the node and edges because it maps more directly to the underlying graphQL.

aklish: Why is JsonApi in the name of the DSL?

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
QubitPi: Wow why didn't I came up with this great idea?

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/relayjsonapi/RelayJsonApiDSL.java
QubitPi: This class has been deleted but I gave it JsonApi because its data has the JSON format.

elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLEndpointTest.java
```diff
@@ -433,10 +430,10 @@ void testNonShareable() throws IOException, JSONException {
                                                                 argument("op", enumValue("UPSERT")),
                                                                 argument(
                                                                         "data",
-                                                                        objectValueWithVariable("id: \"123\", " +
+                                                                        enumValue("{id: \"123\", " +
                                                                                 "name: \"my new author\", \" +\n" +
                                                                                 "                " +
-                                                                                "\"noShare:{id:\"1\"}")
+                                                                                "\"noShare:{id:\"1\"}}")
```
QubitPi: I'm still thinking about how to inject alien field into Jackson serialization. Maybe I'll put it in the next PR

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/VariableDefinitions.java
```diff
@@ -0,0 +1,55 @@
...
+     * Returns a GraphQL query string that representing an set of variable definition
+     *
+     * @return a sub-string of a GraphQL query
+     */
+    public String toGraphQLSpec() {
+        return String.format(
+                "(%s)",
+                getVariableDefinitions().stream()
+                        .map(VariableDefinition::toGraphQLSpec)
+                        .collect(Collectors.joining(" "))
```
hellohanchen: Should these be joined by `,`?

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/VariableValue.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private static final long serialVersionUID = -7483470663100670977L;
+
+    @NonNull
+    @Getter(AccessLevel.PRIVATE)
+    private final String name;
+
+    @Override
+    public String toGraphQLSpec() {
+        // variable : '$' name;
```
hellohanchen: I think `'$'name` is clearer to me.

elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLEndpointTest.java
QubitPi: update: my bad. It's not a problem

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/Arguments.java
```diff
@@ -0,0 +1,74 @@
...
+     * Returns a GraphQL query string that representing an set of arguments
+     *
+     * @return a sub-string of a GraphQL query
+     */
+    public String toGraphQLSpec() {
+        return String.format(
+                "(%s)",
+                getArguments().stream()
+                        .map(Argument::toGraphQLSpec)
+                        .collect(Collectors.joining(" "))
```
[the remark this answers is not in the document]
QubitPi: [no](https://github.com/graphql-java/graphql-java/blob/master/src/main/antlr/GraphqlCommon.g4#L27)

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/VariableValue.java
QubitPi: [This is a definition](https://github.com/graphql-java/graphql-java/blob/master/src/main/antlr/GraphqlCommon.g4#L60)

elide-integration-tests/src/test/java/example/Book.java
```diff
@@ -45,6 +47,7 @@
     private String title;
     private String genre;
     private String language;
+    @JsonIgnore
```
aklish: What's this for?

elide-integration-tests/src/test/java/example/Author.java
```diff
@@ -11,6 +11,7 @@
 import com.yahoo.elide.annotation.Paginate;
 import com.yahoo.elide.annotation.SharePermission;
 
+import lombok.Builder;
```
aklish: Is this unused?

elide-graphql/src/test/java/graphqlEndpointTestModels/Book.java
```diff
@@ -15,11 +15,12 @@
 import com.yahoo.elide.annotation.OnUpdatePreSecurity;
 import com.yahoo.elide.annotation.ReadPermission;
 import com.yahoo.elide.annotation.UpdatePermission;
-import com.yahoo.elide.graphql.GraphQLEndointTest;
+import com.yahoo.elide.graphql.GraphQLEndpointTest;
 import com.yahoo.elide.security.RequestScope;
 
 import graphqlEndpointTestModels.security.CommitChecks;
 import graphqlEndpointTestModels.security.UserChecks;
+import lombok.Builder;
```
aklish: Is this Unused?

elide-graphql/src/test/java/example/Book.java
```diff
@@ -22,6 +22,11 @@
 import com.yahoo.elide.annotation.SharePermission;
 import com.yahoo.elide.security.RequestScope;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+
+import lombok.Builder;
```
aklish: Is this unused?

elide-graphql/src/test/java/example/Author.java
```diff
@@ -10,6 +10,8 @@
 import com.yahoo.elide.annotation.Include;
 import com.yahoo.elide.annotation.SharePermission;
 
+import lombok.Builder;
```
aklish: Is this unused?

elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLEndpointTest.java
```diff
@@ -0,0 +1,858 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.JsonNodeFactory;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.collect.Sets;
+
+import org.json.JSONException;
+import org.mockito.Mockito;
+import org.skyscreamer.jsonassert.JSONAssert;
+import org.testng.Assert;
```
aklish: There should be no dependencies on testng.  Replace with junit5.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/StringValue.java
```diff
@@ -0,0 +1,33 @@
...
+import lombok.NonNull;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * A {@link StringValue} is a string value that is quoted in its serialized form.
+ *
+ * @see EnumValue
+ */
+@RequiredArgsConstructor
+public class StringValue implements ValueWithVariable {
```
aklish: Do we really need all these wrapper classes for the types?  Why can't we just use jackson and serialize plain old Java objects?

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
```diff
@@ -0,0 +1,585 @@
...
+     *
+     * @param name  The name of the selected entity/field that would appear in a GraphQL query
+     * @param selectionSet  The fields of the entity that are selected
+     *
+     * @return a top-level selection
+     *
+     * @see <a href="https://graphql.org/learn/queries/#fields">Fields</a>
+     * @see <a href="https://graphql.org/learn/schema/#object-types-and-fields">Object Types and Fields</a>
+     */
+    public static Selection entity(String name, SelectionSet selectionSet) {
```
aklish: Let's collapse entity and field into a single concept - and call it field.   Fewer concepts == better.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
```diff
@@ -0,0 +1,585 @@
...
+
+    /**
+     * Creates a entity(object field) response attribute.
+     *
+     * @param name  The name of the selected entity/field that would appear in a GraphQL query
+     * @param nodes  All instances of the entity in the response data
+     *
+     * @return an entity selection instantiation
+     */
+    public static Selection responseField(String name, SelectionSet... nodes) {
```
aklish: Is it possible to collapse responseField into field as well?  Again - fewer concepts.  If you are concerned about serialization with and without quotes, maybe that can just be an option to the DSL itself.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/ValueWithVariable.java
```diff
@@ -0,0 +1,41 @@
...
+ *         arrayValueWithVariable |
+ *         objectValueWithVariable;
+ * }
+ * </pre>
+ * Each sub-definition is an implementation of {@link ValueWithVariable}.
+ * <p>
+ * This is a {@link java.util.function functional interface} whose functional method is {@link #toGraphQLSpec()}.
+ */
+@FunctionalInterface
+public interface ValueWithVariable extends Serializable {
```
aklish: I'd like to see if it's possible remove this interface and all of its implementations and just use the java objects directly (a string, an integer, an actual enum, an object) in place of these.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/Argument.java
```diff
@@ -0,0 +1,57 @@
...
+    @NonNull
+    @Getter(AccessLevel.PRIVATE)
+    private final String name;
+
+    /**
+     * A GraphQL concept that defines argument value.
+     */
+    @NonNull
+    @Getter(AccessLevel.PRIVATE)
+    private final ValueWithVariable value;
```
aklish: This can just be Object.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/TypedOperation.java
```diff
@@ -0,0 +1,108 @@
...
+ * operation name and variables used in query.
+ */
+@RequiredArgsConstructor
+public class TypedOperation implements OperationDefinition {
+
+    /**
+     * Models the required {@code operation type}.
+     */
+    @RequiredArgsConstructor
+    public enum OperationType implements Serializable {
```
aklish: Instead of having an enum here that is a member of a class - I think it would be simpler/cleaner to just have classes for Mutation, Subscription (which we can leave out for now since we don't use it), and Query.  Also, I'm a bit confused by OperationDefinition and Definition.  I realize they are concepts referenced in GraphQL - but I'm not sure we really need them.

elide-integration-tests/src/test/java/example/Author.java
QubitPi: @aklish Strange. This was not there on my branch. Anyway it's gone now

elide-integration-tests/src/test/java/example/Book.java
QubitPi: @aklish When I was serialize a book instance Jackson keeps serializing this field so I asks Jackson to ignore it.

Let me put a hold on it and see if I can remove it after the refactoring and tests which I'll finish today

elide-graphql/pom.xml
```diff
@@ -64,18 +64,24 @@
...
             <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-core</artifactId>
+            <artifactId>logback-classic</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>org.testng</groupId>
-            <artifactId>testng</artifactId>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-core</artifactId>
```
QubitPi: I have no idea why git represents my change in this way. I simply added two new dependencies: junit5 & ant(a helper class from old junit)

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/StringValue.java
QubitPi: Done

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/ValueWithVariable.java
QubitPi: Done

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
QubitPi: Done

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/Argument.java
QubitPi: Done

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
QubitPi: Done

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/TypedOperation.java
QubitPi: I can remove `OperationDefinition`, but if we also remove `Definition`, then I think it will become a very loose type system filled with Objects that is going to make DSL entirely loose compile-time check.

elide-graphql/src/test/java/graphqlEndpointTestModels/Book.java
QubitPi: This is used in a javadoc reference.

```
/**
 * Model for book.
 * <p>
 * <b>CAUTION: DO NOT DECORATE IT WITH {@link Builder}, which hides its no-args constructor. This will result in
 * runtime error at places such as {@code entityClass.newInstance();}</b>
 */
```

elide-graphql/src/test/java/example/Book.java
QubitPi: This is used in a javadoc reference.

```
/**
 * Model for book.
 * <p>
 * <b>CAUTION: DO NOT DECORATE IT WITH {@link Builder}, which hides its no-args constructor. This will result in
 * runtime error at places such as {@code entityClass.newInstance();}</b>
 */
```

elide-graphql/src/test/java/example/Author.java
QubitPi: This is used in a javadoc:

```
/**
 * Model for authors.
 * <p>
 * <b>CAUTION: DO NOT DECORATE IT WITH {@link Builder}, which hides its no-args constructor. This will result in
 * runtime error at places such as {@code entityClass.newInstance();}</b>
 */
```

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
```diff
@@ -0,0 +1,482 @@
...
+ * {@code
+ * String graphQLQuery = document(
+ *         typedOperation(
+ *                 TypedOperation.OperationType.QUERY,
+ *                 "myQuery",
+ *                 variableDefinitions(
+ *                         variableDefinition("bookId", "[String]")
+ *                 ),
+ *                 selections(
+ *                         entity(
```
aklish: I thought the plan was to combine entity and field into a single concept here?

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
```diff
@@ -0,0 +1,482 @@
...
+ *     }
+ * }
+ * }
+ * </pre>
+ * In addition, {@link GraphQLDSL} also allows programmatically constructing GraphQL response data. For example,
+ * <pre>
+ * {@code
+ * document(
+ *         selection(
+ *                 responseField(
```
aklish: The plan was to combine responseField and field into a single concept.  THe comments need to all be updated to reflect the DSL changes.

elide-contrib/elide-test-helpers/src/test/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSLTest.java
```diff
@@ -0,0 +1,266 @@
...
+                "[{\"node\":" +
+                "{\"id\":\"1\"," +
+                "\"title\":\"My first book\"," +
+                "\"authors\":{\"edges\":[{\"node\":{" +
+                "\"name\":\"Ricky Carmichael\"}}]}}}]}}}";
+
+        String actual = document(
+                selection(
+                        field(
+                                RESPONSE,
```
aklish: I don't feel like we need RESPONSE defined here and also a separate method called toResponse.  Just use the latter and remove this extra field parameter.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/Definition.java
```diff
@@ -0,0 +1,45 @@
...
+@FunctionalInterface
+public interface Definition extends Serializable {
+
+    /**
+     * Returns the query string that corresponds to the a {@link graphql.language.Definition} part
+     * of a GraphQL query.
+     *
+     * @return a sub-string of a GraphQL query
+     */
+    String toGraphQLSpec();
```
aklish: What about having two methods here - toResponse and toQuery?  Then you can just use those everywhere instead of having the the special QUERY and RESPONSE variables on the field class.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/Document.java
```diff
@@ -0,0 +1,71 @@
...
+        return String.format("{\"data\":%s}", serialize(false));
+    }
+
+    /**
+     * Turns this {@link Document} into its string representation.
+     *
+     * @return a string representation of a GraphQL query
+     */
+    private String serialize(boolean isQuery) {
+        Definition.IS_QUERY = isQuery;
```
aklish: It would be better not to have side effects like this and instead use a separate method (toResponse and toQuery) throughout.  I would not expect a serialize function to set a variable.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
```diff
@@ -0,0 +1,535 @@
...
+import com.yahoo.elide.contrib.testhelpers.graphql.elements.Field;
+import com.yahoo.elide.contrib.testhelpers.graphql.elements.Mutation;
+import com.yahoo.elide.contrib.testhelpers.graphql.elements.Node;
+import com.yahoo.elide.contrib.testhelpers.graphql.elements.Query;
+import com.yahoo.elide.contrib.testhelpers.graphql.elements.Selection;
+import com.yahoo.elide.contrib.testhelpers.graphql.elements.SelectionSet;
+import com.yahoo.elide.contrib.testhelpers.graphql.elements.VariableDefinition;
+import com.yahoo.elide.contrib.testhelpers.graphql.elements.VariableDefinitions;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
```
aklish: Is there a reason we need to use two different serialization frameworks (GSON and Jackson)?  Can we just use one of the two?

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
```diff
@@ -0,0 +1,535 @@
...
+     * @param name  The name of the selected entity/field that would appear in a GraphQL query
+     * @param selectionSet  The fields of the entity that are selected
+     *
+     * @return a top-level selection
+     *
+     * @see <a href="https://graphql.org/learn/queries/#fields">Fields</a>
+     * @see <a href="https://graphql.org/learn/schema/#object-types-and-fields">Object Types and Fields</a>
+     */
+    public static Selection field(String name, Object... selectionSet) {
+        if (selectionSet.length == 1 && !(selectionSet[0] instanceof SelectionSet)) {
```
aklish: I"m not following this.  A couple points:
 1. instanceof is a generally a code smell in Java (unless the code is doing reflection like the EntityDictionary).
 2.  It looks like this code is trying to infer from arguments if this is a query or a response.  The only time that should happen is inside the toQuery or toResponse functions (not during DSL construction).

Construction should be dumb/simple logic and serialization can be smarter.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
```diff
@@ -0,0 +1,535 @@
...
+     * @param name  The name of the argument
+     * @param value  An object that models argument value.
+     * @param quoted  Whether or not the serialized {@code value} should be double-quote
+     *
+     * @return a complete specification of selection arguments
+     *
+     * @see <a href="https://graphql.org/learn/queries/#arguments">Arguments</a>
+     */
+    public static Argument argument(String name, Object value, boolean quoted) {
+        if (value instanceof String) {
```
aklish: Whether the argument is a string or not, it should be convertable to JSON - why not just always run it through the JSON parser?

elide-integration-tests/src/test/java/example/Author.java
aklish: This is still there.

elide-graphql/src/test/java/graphqlEndpointTestModels/Book.java
aklish: I meant Builder.  I don't see you using the import.

elide-graphql/src/test/java/example/Book.java
aklish: I meant the Builder import.

elide-graphql/src/test/java/example/Author.java
aklish: I mean can you remove the import since it is unused?

elide-graphql/src/test/java/com/yahoo/elide/graphql/PersistentResourceFetcherTest.java
```diff
@@ -159,13 +162,13 @@ protected void assertQueryEquals(String graphQLRequest, String expectedResponse,
             requestScope.saveOrCreateObjects();
         }
         requestScope.getTransaction().commit(requestScope);
-        Assert.assertEquals(result.getErrors().size(), 0, "Errors [" + errorsToString(result.getErrors()) + "]:");
+        assertEquals(result.getErrors().size(), 0, "Errors [" + errorsToString(result.getErrors()) + "]:");
```
aklish: Whenever you switch an assert from testNG to Jupiter, you need to make sure the "expected" and "actual" variables are swapped.  Otherwise the error messages will be wrong when the tests fail.

elide-graphql/pom.xml
aklish: What happens when you remove it then?  This was likely added by your IDE.

elide-graphql/pom.xml
```diff
@@ -59,18 +64,24 @@
...
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-classic</artifactId>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>5.5.1</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.ant</groupId>
+            <artifactId>ant</artifactId>
```
aklish: Why do we need ant?  Can we remove this dependency?

elide-contrib/elide-test-helpers/src/test/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSLTest.java
```diff
@@ -0,0 +1,249 @@
...
+import static com.yahoo.elide.contrib.testhelpers.graphql.GraphQLDSL.variableDefinition;
+import static com.yahoo.elide.contrib.testhelpers.graphql.GraphQLDSL.variableDefinitions;
+
+import org.testng.Assert;
+import org.testng.annotations.Test;
+
+public class GraphQLDSLTest {
+
+    @Test
+    public void verifyBasicRequest() {
```
aklish: Let's make sure the coverage of the new DSL is very high.  This should touch on every entry function you created in the DSL.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/VariableDefinitions.java
```diff
@@ -0,0 +1,54 @@
...
+ * <pre>
+ * {@code
+ * variableDefinitions : '(' variableDefinition+ ')';
+ * }
+ * </pre>
+ * {@link VariableDefinitions} is a space-separated list of {@link VariableDefinition}s surrounded by a pair of
+ * parenthesis.
+ */
+@RequiredArgsConstructor
+public class VariableDefinitions implements Serializable {
```
aklish: Why not have this import from Definition.  The toResponse method can be defaulted in Definition to throw an UnsupportedOperationException

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/VariableDefinition.java
```diff
@@ -0,0 +1,76 @@
...
+ * variableDefinition : variable ':' type defaultValue?;
+ *
+ * variable : '$' name;
+ * }
+ * </pre>
+ * A {@link VariableDefinition} is a key-value pair of string and type literal with optional default value for that
+ * variable.
+ */
+@RequiredArgsConstructor
+public class VariableDefinition implements Serializable {
```
aklish: Same comment - why not have import from Definition (see below)

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/TypedOperation.java
```diff
@@ -0,0 +1,73 @@
...
+                getVariableDefinitions() == null
+                        ? ""
+                        : getVariableDefinitions().toGraphQLSpec() + " ",
+                getSelectionSet().toGraphQLSpec()
+        );
+    }
+
+    @Override
+    String toResponse() {
+        return toGraphQLSpec();
```
aklish: I would not have toResponse call toGraphQLSpec.  Let's make these functions be recursive - so we don't want one behavior to recurse accidentally into the other.  Keep the logic separate.

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
aklish: So it sounds like this is mixing two use cases:
1) A list of selection sets (but those aren't objects, they are selection sets>
2) A field where the value is a singular object.

These ought to be two different functions.

PR #909, opened 2019-08-12
Elide-Testing Refactor BookAuthorIT & AuditIT

hellohanchen:
- Update BookAuthorIT with new integration test framework and Junit 5. https://github.com/yahoo/elide/pull/897
- Add jsonData() method to jsonApiDSL


elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/jsonapi/JsonApiDSL.java
```diff
@@ -36,6 +39,20 @@
  * }
  */
 public class JsonApiDSL {
+    /**
+     * Create a new json data object
+     * @return Map that Rest-Assured can serialize to Json API format
+     */
+    public static Map<String, Object> jsonData(Resource... resources) {
```
aklish: I believe the data/datum methods already do this?

elide-integration-tests/src/test/java/com/yahoo/elide/auditTests/AuditIT.java
```diff
@@ -6,110 +6,230 @@
...
+                .body(
+                        datum(AUDIT_1_RELATIONSHIP),
+                        ObjectMapperType.GSON
+                )
+                .patch("/auditEntity/1")
+                .then()
+                .statusCode(HttpStatus.SC_NO_CONTENT);
 
-        Assert.assertFalse(logger.logMessages.contains("Inverse entities: [Value: update id 1 through id 2 relationship: 2]"));
+        assertFalse(logger.logMessages.contains("Inverse entities: [Value: updated value relationship: 2]"));
```
hellohanchen: The log message here is changed because this test doesn't depend on `testAuditWithDuplicateLineageEntry()` anymore.

PR #933, opened 2019-08-21
Add data collection class hierarchy parameterized attributes

### LINKED ISSUE #890: Aggregation Store: Hydrate GraphQL Schema with parameterized attributes

Chandrasekar-Rajasekar:
Resolves #890 

## Description
Added `addArgumentsToAttribute` to `EntityDictionary` that accepts Map object as argument for an attribute in an Entity. `getAttributeArguments` method retrieves this object.
This method is used by EntityProjectionMaker to populate the argument set for the attribute. 
ModelBuilder uses getAttributeArguments method to expose the argument in the schema.

## How Has This Been Tested?
Unit testing or `ModelBuilder` class checks whether the GraphQL schema contains arguments to the attribute that was added by explicitly calling `addArgumentsToAttributes` method.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
@@ -52,6 +52,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Deque;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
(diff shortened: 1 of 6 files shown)


elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -116,10 +117,12 @@
     public final ConcurrentHashMap<String, Class<?>> fieldsToTypes = new ConcurrentHashMap<>();
     public final ConcurrentHashMap<String, String> aliasesToFields = new ConcurrentHashMap<>();
     public final ConcurrentHashMap<Method, Boolean> requestScopeableMethods = new ConcurrentHashMap<>();
+    public final ConcurrentHashMap<AccessibleObject, Map<String, Object>> attributeArgumets = new ConcurrentHashMap<>();
```
aklish: The entity dictionary should not store argument values directly (as those are tied to a request).  The dictionary would only know the argument name and type.

aklish: Rather than store a Map<String, Object> - we might want to store a Set<Argument> and create a new Argument type for readability.

elide-core/src/main/java/com/yahoo/elide/jsonapi/EntityProjectionMaker.java
```diff
@@ -307,6 +308,15 @@ public EntityProjection visitIncludePath(Path path) {
                 .map(attributeName -> Attribute.builder()
                     .name(attributeName)
                     .type(dictionary.getType(entityClass, attributeName))
+                    .arguments(
+                            dictionary.getAttributeArguments(entityClass, attributeName)
```
aklish: JSON-API doesn't really support attribute arguments.  We'll need to augment the GraphQL logic that builds the entity projection instead.

PR #939, opened 2019-08-29
GraphQL EntityProjection Maker

hellohanchen:


elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,410 @@
...
+    public Collection<EntityProjection> make(String query) {
+        Parser parser = new Parser();
+        Document parsedDocument = parser.parseDocument(query);
+
+        parsedDocument.getDefinitions().forEach(definition -> {
+            if (definition instanceof OperationDefinition) {
+                // Operations would be converted into EntityProjection tree
+                OperationDefinition operationDefinition = (OperationDefinition) definition;
+                if (operationDefinition.getOperation() != OperationDefinition.Operation.QUERY) {
+                    // TODO: support MUTATION and SUBSCRIPTION
```
aklish: You will need to support Mutation.  It should be identical to Query in our case.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,410 @@
...
+    /**
+     * Root projection would be an operation applied on an entity class. There should be only one selection, which
+     * is the entity, in the selection set. The EntityProjection tree would be constructed recursively to add all
+     * child projections.
+     */
+    private void addRootProjections(SelectionSet selectionSet) {
+        List<Selection> selections = selectionSet.getSelections();
+
+        if (selections.size() != 1) {
+            throw new InvalidEntityBodyException("Can't select multiple entities in graphQL QUERY operation.");
```
aklish: I don't think this is true.  YOu can create multiple roots - and there can be aliases for the same root asked for twice in the same query.  There is an IT tests for this.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,410 @@
...
+        if (edges.getSelections().size() != 1) {
+            throw new InvalidEntityBodyException("Entity selection must have one 'edges' graphQL field node.");
+        }
+
+        Selection edgesSelection = edges.getSelections().get(0);
+        if (!(edgesSelection instanceof Field)) {
+            throw new InvalidEntityBodyException("Edges selection must be a graphQL field.");
+        }
+
+        return ((Field) edgesSelection).getSelectionSet();
```
aklish: Should we verify the fieldname is edges?

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,410 @@
...
+        Value argumentValue = argument.getValue();
+
+        if (isPaginationArgument(argumentName)) {
+            addPagination(argumentName, argumentValue, entityProjection);
+        } else if (isSortingArgument(argumentName)) {
+            addSorting(argumentValue, entityProjection);
+        } else {
+            Class<?> entityType = entityProjection.getType();
+
+            if (!entityDictionary.isValidField(entityType, argumentName)) {
```
aklish: Arguments come in two flavors:
1. Arguments elide defines for filters, pagination, and sorting
2. attribute arguments which are up to the datastore to interpret.

The latter are not attribute names, but something like:
`{ playerStats { score(aggregation : sum) date(grain : day) } }`

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,410 @@
...
+
+    /**
+     * Creates a {@link Pagination} object from pagination GraphQL argument and attaches it to the
+     * {@link EntityProjection}.
+     *
+     * @param paginationArgument  A string that contains the key to a value of sorting spec
+     * @param paginationValue  A string that contains the value of pagination spec
+     * @param entityProjection projection that has the pagination argument
+     */
+    private void addPagination(
```
aklish: Your logic doesn't need to add this just yet.  Elide core logic adds pagination, sorting, and filters for you.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,410 @@
...
+        return ModelBuilder.ARGUMENT_SORT.equals(argumentName);
+    }
+
+    /**
+     * Creates a {@link Sorting} object from sorting GraphQL argument value and attaches it to the entity sorted
+     * according to the newly created {@link Sorting} object.
+     *
+     * @param argumentValue A string that contains the value of sorting spec
+     */
+    private void addSorting(Object argumentValue, EntityProjection entityProjection) {
```
aklish: YOu don't need to add sorting just yet.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,410 @@
...
+        return ((Field) nodeSelection).getSelectionSet();
+    }
+
+    /**
+     * Add a {@link Selection} that represents a field/fragment to an {@link EntityProjection}
+     *
+     * @param fieldSelection field/fragment to add
+     * @param parentProjection projection that has this field/fragment
+     */
+    private void addField(Selection fieldSelection, final EntityProjection parentProjection) {
```
aklish: addField method should only handle Field arguments.  The parent caller can determine the type and call the appropriate method.  That should keep the lines per function smaller.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,410 @@
...
+                // FragmentDefinition should be stored first and then inserted into EntityProjections that use it
+                FragmentDefinition fragmentDefinition = (FragmentDefinition) definition;
+                fragmentDefinitions.put(fragmentDefinition.getName(), fragmentDefinition);
+            } else {
+                throw new InvalidEntityBodyException(
+                        String.format("Unsupported definition type {%s}.", definition.getClass()));
+            }
+        });
+
+        // make sure there is no fragment loop and undefined fragments in fragment definitions
```
aklish: Let's break this up into more functions.  Validations can be separate functions.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,410 @@
...
+
+/**
+ * This class converts a GraphQL query string into an Elide {@link EntityProjection} using {@link #make(String)}.
+ */
+@Slf4j
+public class GraphQLEntityProjectionMaker {
+    private final EntityDictionary entityDictionary;
+    private final ElideSettings elideSettings;
+
+    private final Map<String, List<EntityProjection>> fragmentToProjections = new HashMap<>();
```
aklish: I have an idea.  What if you first translate the query to remove all fragments?  Then you don't need to keep all this state around when building the projections.  

It would also break up the code into smaller bits (that are mostly unrelated).

PR #941, opened 2019-08-30
Graphql IT test

QubitPi:


elide-integration-tests/src/test/java/com/yahoo/elide/tests/GraphQLIT.java
```diff
@@ -0,0 +1,702 @@
...
+
+        // This test demonstrates that multiple roots can be manipulated within a _single_ transaction
+        String graphQLRequest = document(
+                selections(
+                        field(
+                                "book",
+                                argument(
+                                        argument(
+                                                "ids",
+                                                Arrays.asList("1")
```
aklish: This test is different than the deleted test.  The test has to both UPSERT and object in one root document, and fetch it back in another root document (all in the same query).

elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/VariableFieldSerializer.java
```diff
@@ -0,0 +1,63 @@
...
+ * To serialize a String entity field in such a way, add the following annotation to the field, as shown above:
+ * <pre>
+ * {@code
+ * {@literal @}JsonSerialize(using = VariableFieldSerializer.class, as = String.class)
+ * }
+ * </pre>
+ *
+ * @see <a href="https://graphql.org/learn/queries/#variables">Variables</a>
+ */
+public class VariableFieldSerializer extends JsonSerializer<String> {
```
aklish: We can remove this now.

PR #949, opened 2019-09-03
Add EntityProjection plumbing

aklish:
## Overview

Adds EntityProjection (internal query concept for Elide).

## Caveats:
- GraphQL module & tests are disabled until a subsequent PR can build EntityProjections for GraphQL.
- We'll rename EntityProjection to ElideQuery later

## License


elide-core/src/main/java/com/yahoo/elide/request/EntityProjection.java
```diff
@@ -0,0 +1,171 @@
...
+
+    private Sorting sorting;
+
+    private Pagination pagination;
+
+    /**
+     * Returns the entity name.
+     * @return the entity name
+     */
+    public String getName() {
```
hellohanchen: Does entityProjection support alias?

aklish: I don't think it needs to.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -158,13 +177,86 @@ default Object loadObject(Class<?> entityClass,
...
+     * @param collection - The data collection to fetch
+     * It is optional for the data store to attempt evaluation.
+     * @param scope - contains request level metadata.
+     * @return the object in the relation
+     */
+    default Object getRelation(
+            DataStoreTransaction relationTx,
+            Object entity,
+            String relationName,
+            EntityProjection collection,
```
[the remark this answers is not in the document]
aklish: I need to clean this up.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -208,8 +227,13 @@ public boolean matchesId(String checkId) {
             Optional<FilterExpression> permissionFilter = getPermissionFilterExpression(loadClass,
                     requestScope);
             Class<?> idType = dictionary.getIdType(loadClass);
-            obj = tx.loadObject(loadClass, (Serializable) CoerceUtil.coerce(id, idType),
-                    permissionFilter, requestScope);
+
+            projection = projection
```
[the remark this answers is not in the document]
aklish: We are adding permission filters to the projection.  They are not part of the original request.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -305,9 +327,15 @@ public boolean matchesId(String checkId) {
             }
         }
 
+        EntityProjection modifiedProjection = requestScope.getEntityProjection()
```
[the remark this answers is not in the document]
aklish: Long term, we can add the permission rules to the filter expression outside the PersistentResource.  For now, it has to be added.

elide-core/src/main/java/com/yahoo/elide/core/datastore/inmemory/InMemoryStoreTransaction.java
```diff
@@ -71,6 +74,80 @@ public InMemoryStoreTransaction(DataStoreTransaction tx) {
...
+        DataFetcher fetcher = new DataFetcher() {
+            @Override
+            public Object fetch(Optional<FilterExpression> filterExpression,
+                                Optional<Sorting> sorting,
+                                Optional<Pagination> pagination,
+                                RequestScope scope) {
+
+                return tx.getRelation(relationTx, entity, relationship.copyOf()
+                        .projection(relationship.getProjection().copyOf()
+                                .filterExpression(filterExpression.orElse(null))
```
hellohanchen: If we are not gonna to change filter/sorting/pagination here, why do we need to make a copy of the projection? Same for `loadObjects(EntityProjection projection, RequestScope scope)`

elide-core/src/main/java/com/yahoo/elide/core/datastore/inmemory/InMemoryStoreTransaction.java
```diff
@@ -39,6 +42,8 @@
  */
 public class InMemoryStoreTransaction implements DataStoreTransaction {
```
hellohanchen: Can we make `InMemoryStoreTransaction` extends `TransactionWrapper`?

elide-core/src/main/java/com/yahoo/elide/jsonapi/models/Resource.java
```diff
@@ -146,13 +148,20 @@ public boolean equals(Object obj) {
...
+
         if (cls == null) {
             throw new UnknownEntityException(type);
         }
         if (id == null) {
             throw new InvalidObjectIdentifierException(id, type);
         }
-        return PersistentResource.loadRecord(cls, id, requestScope);
+
+        EntityProjection projection = EntityProjection.builder()
```
hellohanchen: Do we need to build a new projection every time we use this method?

elide-core/src/main/java/com/yahoo/elide/parsers/state/CollectionTerminalState.java
```diff
@@ -186,8 +172,8 @@ private PersistentResource createObject(RequestScope requestScope)
                     + " to type: " + entityClass);
         }
 
-        PersistentResource pResource = PersistentResource.createObject(
-                parent.orElse(null), newObjectClass, requestScope, Optional.ofNullable(id));
+        PersistentResource pResource = PersistentResource.createObject(parent.orElse(null), newObjectClass,
+                    requestScope, Optional.ofNullable(id));
```
hellohanchen: Do we need all 3 tabs here?

elide-core/src/main/java/com/yahoo/elide/core/datastore/inmemory/InMemoryStoreTransaction.java
aklish: I tried, but I'm not sure it can be done.

aklish: I originally tried this, but had trouble delegating a method call to DataStoreTransaction.  It has to call the default loadObject method in DataStoreTransaction.  Maybe there is a way to do that - but I couldn't get it to work.

elide-core/src/main/java/com/yahoo/elide/core/datastore/inmemory/InMemoryStoreTransaction.java
aklish: So the filter expression sorting and pagination can change here.  It is not obvious, but this is a function that is called after the in memory store decides whether to pass sorting, fltering, and pagination to the underlying wrapped store.  In the case of filtering, it can be a subset of the filter expression.

elide-core/src/main/java/com/yahoo/elide/jsonapi/models/Resource.java
aklish: Where would it come from?  This is not tied directly to the request path.  This typically comes from the JSON body.  There is no root projection in that case.  The object is just be loaded by its ID.

PR #950, opened 2019-09-03
Implement GraqlEntityProjectionMaker

hellohanchen:
Resolves https://github.com/yahoo/elide/issues/890 and https://github.com/yahoo/elide/issues/823

## Description
Implement `GraphQLEntityProjectionMaker` to construct an `EntityProjection` from a GraphQL `Document` object
- Implement `FragmentResolver` to parse and validate fragments in GraphQL request
- 
- Update `Entity` class in Elide/GraphQL module to contain projection in which this entity is fetched, this change match the latest API in `PersistentResource`.
- Update `GraphQLEndpoint` to use `GraphQLEntityProjectionMaker` for parsing incoming GraphQL request, this change is also applied for test classes.
- Update `PersistentResourceFetcher` to use projection
- Add enum `Keyword` for all GraphQL keywords we use in this module

## How Has This Been Tested?
These changes pass all unit test in `elide-graphql` module.

## License


elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,348 @@
...
+                .map(definition -> (FragmentDefinition) definition)
+                .collect(Collectors.toList());
+        fragmentMap.putAll(FragmentResolver.resolve(fragmentDefinitions));
+
+        // resolve operation definitions
+        parsedDocument.getDefinitions().forEach(definition -> {
+            if (definition instanceof OperationDefinition) {
+                // Operations would be converted into EntityProjection tree
+                OperationDefinition operationDefinition = (OperationDefinition) definition;
+                if (operationDefinition.getOperation() == OperationDefinition.Operation.MUTATION) {
```
aklish: I don't think we want to skip mutations.  We just want to skip subscriptions.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,348 @@
...
+        Attribute argumentAttribute = entityProjection.getAttributeByName(argumentName);
+        com.yahoo.elide.request.Argument elideArgument = com.yahoo.elide.request.Argument.builder()
+                .name(argumentName)
+                .value(argument.getValue())
+                .build();
+
+        if (argumentAttribute == null) {
+            argumentAttribute = Attribute.builder()
+                    .type(entityDictionary.getType(entityType, argumentName))
+                    .name(argumentName)
```
aklish: I think this is wrong.  The name of the attribute should be the field name - not the argument name.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,348 @@
...
+     */
+    private void addField(Field field, EntityProjection parentProjection) {
+        Class<?> parentType = parentProjection.getType();
+        String fieldName = field.getName();
+
+        // this field would either be a relationship field or an attribute field
+        if (entityDictionary.getRelationshipType(parentType, fieldName) != RelationshipType.NONE) {
+            // handle the case of a relationship field
+            final Class<?> relationshipType =
+                    entityDictionary.getParameterizedType(parentType, fieldName) == null
```
aklish: This check for null shouldn't be necessary.  getParameterized type will return the correct type regardless of whether the type is actually parameterized.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
hellohanchen: I think the argument here is something like `id`, which is one of the fields of the projected entity, isn't it?

PR #966, opened 2019-09-15
Add heroku archetype project

aklish:
## Description
First of several PRs to improve the documentation and provide an archetype setup for an elide project.

## Motivation and Context
Elide getting started docs are pretty dated - and not as helpful as they ought to be.

## How Has This Been Tested?
Tested with Heroku deployment.

## License


app.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "Elide Sample Service",
+  "description": "An example Elide service with an in-memory DB",
+  "repository": "https://github.com/yahoo/elide/tree/master/elide-example/elide-blog-example",
+  "logo": "https://elide.io//assets/images/elide-white-logo.png",
+  "keywords": ["java", "elide", "graphql", "json-api"]
```
jkusa: Can we add 
```
"addons": [ "heroku-postgresql" ]
```

aklish:
Do you think we should do Postgres instead of H2?

jkusa:
@aklish 

> If an app has a free web dyno, and that dyno receives no web traffic in a 30-minute period, it will sleep.

If we configure Postgres, user's data will persist across sessions. Also, they will be able to inspect the db data via the heroku-postgresql tools.

elide-example/elide-blog-example/pom.xml
```diff
@@ -1,148 +1,72 @@
...
-        <groupId>com.yahoo.elide</groupId>
-        <artifactId>elide-example-parent-pom</artifactId>
-        <version>4.5.3-SNAPSHOT</version>
-    </parent>
+    <groupId>com.yahoo.elide</groupId>
+    <version>4.5.3-SNAPSHOT</version>
 
     <properties>
-        <mysql.data.directory>${project.build.directory}/mysql-data</mysql.data.directory>
+        <jar.finalName>testapp</jar.finalName>
```
jkusa: what about `<jar.finalName>elide-blog-example</jar.finalName>`?

aklish:
Going to add H2 IT tests (with the changes to use Postgres).  Then will make reviewable again.

README.md
```diff
@@ -75,7 +75,13 @@ More information about Elide can be found at [elide.io](http://elide.io/).
 
 ## Install
 
-The fastest way to get started with Elide is to use [elide-standalone](https://github.com/yahoo/elide/tree/master/elide-standalone) which allows you to quickly setup a local instance
+To try out an Elide example service (with an in-memory database), you can deploy via Heroku.
```
jkusa: `(with an in-memory database)` is this true?

elide-example/elide-blog-example/pom.xml
jkusa: ?

PR #986, opened 2019-09-23
Implement GraphQLEntityProjectionMaker

hellohanchen:
Resolves https://github.com/yahoo/elide/issues/890 and https://github.com/yahoo/elide/issues/823

## Description
Implement `GraphQLEntityProjectionMaker` to construct an `EntityProjection` from a GraphQL `Document` object
- Implement `FragmentResolver` to parse and validate fragments in GraphQL request
- 
- Update `Entity` class in Elide/GraphQL module to contain projection in which this entity is fetched, this change match the latest API in `PersistentResource`.
- Update `GraphQLEndpoint` to use `GraphQLEntityProjectionMaker` for parsing incoming GraphQL request, this change is also applied for test classes.
- Update `PersistentResourceFetcher` to use projection
- Add enum `Keyword` for all GraphQL keywords we use in this module

## How Has This Been Tested?
These changes pass all unit test in `elide-graphql` module.

## License


elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
```diff
@@ -2670,6 +2670,7 @@ public void testSpecialCharacterLikeQueryHQL(FilterPredicate filterPredicate, in
         when(pagination.isGenerateTotals()).thenReturn(true);
         tx.loadObjects(EntityProjection.builder()
                 .type(Book.class)
+
```
aklish: There are some other graphQL things I commented out in the base branch.  For example, I removed the graphQL endpoint from JAX-RS.  We'll need to undo all of those so we don't have a regression.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/VariableResolver.java
```diff
@@ -0,0 +1,118 @@
...
+
+import javax.ws.rs.BadRequestException;
+
+/**
+ * Class that contains variables provided in graphql request and can resolve variables based on
+ * {@link graphql.language.OperationDefinition} scope.
+ */
+class VariableResolver {
+    private final Map<String, Object> requestVariables;  // variables defined in request is global
+    private Map<String, Object> scopeVariables;  // variables defined in each operation is operation-scoped
```
aklish: If I have multiple scopes, won't one scope overwrite the other's scopes variables?  Is that an issue?

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLProjectionInfo.java
```diff
@@ -0,0 +1,32 @@
...
+@AllArgsConstructor
+public class GraphQLProjectionInfo {
+    @Getter
+    private final SourceLocation rootLocation;
+
+    @Getter
+    private final Collection<EntityProjection> projections;
+
+    @Getter
+    private final Map<SourceLocation, Relationship> relationshipMap;
```
aklish: I'm not sure you need all this if you use recursion to resolve the EntityProjection given the current sourceLocation.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,537 @@
...
+        Class<?> entityType = entityDictionary.getEntityClass(entityName);
+        if (entityType == null) {
+            throw new InvalidEntityBodyException(String.format("Unknown entity {%s}.", entityName));
+        }
+
+        rootLocation = rootSelection.getSourceLocation();
+        Field fields = (Field) rootSelection;
+
+        // merging partial graphql selections
+        selections.stream().skip(1).forEach(selection -> {
```
aklish: Can we add a comment why we are skipping one here?

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,537 @@
...
+            if (!(selection instanceof Field)) {
+                throw new InvalidEntityBodyException("Entity selection must be a graphQL field.");
+            }
+            if (!entityName.equals(((Field) selection).getName())) {
```
aklish: I don't understand what this whole check is about.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,537 @@
...
+        }
+    }
+
+    /**
+     * Add a new graphQL {@link Field} into an {@link EntityProjection}
+     *
+     * @param field graphQL field
+     * @param parentProjection entity projection that contains this field
+     */
+    private void addField(Field field, EntityProjection parentProjection) {
```
aklish: Cyclomatic complexity and line length are quite large.  How did this pass checkstyles?  Let's break this function into more functions.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,537 @@
...
+            Relationship relationship = Relationship.builder()
+                    .name(fieldName)
+                    .alias(field.getAlias())
+                    .projection(relationshipProjection)
+                    .build();
+
+            relationshipMap.put(field.getSourceLocation(), relationship);
+
+            // add this relationship projection to its parent projection
+            parentProjection.getRelationships().add(relationship);
```
aklish: Projections are really intended to be immutable.  If you want to edit a projection, you should be dealing with a Builder - not the projection itself.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -158,6 +158,7 @@ default Object loadObject(EntityProjection entityProjection,
      * @param scope - contains request level metadata.
      * @return a collection of the loaded objects
      */
+    @Deprecated
```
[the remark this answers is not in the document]
hellohanchen: I might put these annotations at wrong spot when I was rebasing, I would remove them.

elide-graphql/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -90,24 +96,42 @@ private void setRelationships() {
...
+            this.data.entrySet().stream()
+                    .filter(entry -> dictionary.isRelation(this.entityClass, entry.getKey()))
+                    .forEach(entry -> {
+                        String relationshipName = entry.getKey();
+                        Class<?> relationshipClass =
+                                dictionary.getParameterizedType(this.entityClass, relationshipName);
+
+                        // if this data contains a relationship that is not in the projection tree, create a temporary
+                        // projection for that relationship
+                        EntityProjection relationshipProjection =
```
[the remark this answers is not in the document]
hellohanchen: The provided data can contains relationship fields like `Book.Authors`, but if we only do upsert for `Book`, there would not be projection for `Authors`.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLProjectionInfo.java
hellohanchen: This is for getting EntityProjection from the `GraphQLProjectionInfo` only. When the request goes through the `GraphQL` core code, a new `document` would be generated and parsed by `GraphQL` module itself, which would contain the same content as the one we parsed but it would be a different object. And the graphql `context` would contain the graphql object generated by `GraphQL` module. So I need this `SourceLocation` to get the projection I generated.

elide-graphql/src/main/java/com/yahoo/elide/graphql/Entity.java
```diff
@@ -32,19 +33,24 @@
     @Getter private RequestScope requestScope;
     @Getter private Set<Attribute> attributes;
     @Getter private Set<Relationship> relationships;
+    @Getter private EntityProjection projection;
```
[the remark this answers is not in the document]
hellohanchen: This would be pointer to a constructed projection. To be used in `toPersistenceResource()` method which needs a projection and `setRelationships()` method in which we can get the `Relationship` object from the projection without constructing it again.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
hellohanchen: 👍

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
hellohanchen: 👍

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
hellohanchen: This is for something querying two different classes at the same time like
```
query {
  book {
    ...
  }
  country {
    ...
  }
}
```

aklish: Let's add this example to the comment here to show why we are throwing the exception.

elide-graphql/src/test/resources/graphql/requests/fetch/fragmentInline.graphql
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+fragment AuthorFields on author {
+  edges {
+    node {
+      id
+      books {
+        ... on book {
+          ...BookFields
```
aklish: Isn't this another loop test? How is this different than the other loop failure test?

elide-graphql/src/main/java/com/yahoo/elide/graphql/containers/NodeContainer.java
```diff
@@ -46,10 +45,18 @@ public Object processFetch(Environment context, PersistentResourceFetcher fetche
...
-                    fieldName, context.ids, context.offset, context.first, context.sort, context.filters,
-                    generateTotals);
+            // get the relationship from constructed projections
+            Relationship relationship = context.requestScope
+                    .getProjectionInfo()
+                    .getRelationshipMap()
+                    .getOrDefault(context.field.getSourceLocation(), null);
+
+            if (relationship == null) {
+                throw new BadRequestException(
```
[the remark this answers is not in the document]
hellohanchen: This is to match the `BadRequestException` we thrown for **Unrecognized Object**.

elide-integration-tests/src/test/java/com/yahoo/elide/errorEncodingTests/VerboseEncodedErrorResponsesIT.java
```diff
@@ -148,19 +146,4 @@ public void graphQLMutationError() {
             .statusCode(HttpStatus.SC_OK)
             .body(equalTo(expected));
     }
-
-    @Test
-    @Disabled
-    public void graphQLFetchError() {
```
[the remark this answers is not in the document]
hellohanchen: It is now doing the same encoding as `EncodedErrorResponseIT` because the error is now thrown by `GraphQLEntityProjectionMaker`, the verbose would not make difference.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/VariableResolver.java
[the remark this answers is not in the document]
hellohanchen: Then we need to reassign new `VariableResolver` in `GraphQLEntityProjectionMaker` every time. I think that's similar to reassign `scopeVariables` in `VariableResolver`.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -0,0 +1,596 @@
...
+
+        // merging partial graphql selections, skip(1) because the first selection is already processed
+        selections.stream().skip(1).forEach(selection -> {
+            if (!(selection instanceof Field)) {
+                throw new InvalidEntityBodyException("Entity selection must be a graphQL field.");
+            }
+            if (!entityName.equals(((Field) selection).getName())) {
+                throw new InvalidEntityBodyException("Can't select multiple entities in graphQL QUERY operation.");
+            }
+            fields.getSelectionSet().getSelections().addAll(((Field) selection).getSelectionSet().getSelections());
```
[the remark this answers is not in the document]
hellohanchen:
```java
fields.getSelectionSet().getSelections().addAll(((Field) selection).getSelectionSet().getSelections());
```
Currently this is not checking field name or alias name, if a Elide field is queried in both graphQL `Field`, it would be queried twice.

elide-graphql/src/test/resources/graphql/requests/fetch/fragmentInline.graphql
hellohanchen: It would break at the inline fragment first.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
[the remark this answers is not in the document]
hellohanchen: If two fields, one from each query, have the same name and alias, they would be merged into a single field. If two fields, one from each query, have different name but same alias, they would be ambiguous.

elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -272,17 +243,16 @@ private ConnectionContainer upsertOrUpdateObjects(Environment context,
...
         }
 
         /* form entities */
         Optional<Entity> parentEntity;
         if (!context.isRoot()) {
-            parentEntity = Optional.of(new Entity(Optional.empty(),
-                    null,
-                    context.parentResource.getResourceClass(),
-                    context.requestScope));
+            parentEntity = Optional.of(new Entity(Optional.empty(), null, entityClass, context.requestScope));
```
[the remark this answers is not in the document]
hellohanchen: Fixed.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/FragmentResolver.java
```diff
@@ -0,0 +1,139 @@
+/*
+ * Copyright 2019, Oath Inc.
```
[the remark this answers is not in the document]
hellohanchen: Fixed.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -342,16 +345,18 @@ private void addRelationship(Field relationshipField, EntityProjectionBuilder pr
     private void addAttributeField(Field attributeField, EntityProjectionBuilder projectionBuilder) {
         Class<?> parentType = projectionBuilder.getType();
         String attributeName = attributeField.getName();
+        String attributeAlias = attributeField.getAlias() == null ? attributeName : attributeField.getAlias();
 
-        Class<?> attributeType = entityDictionary.getType(parentType, attributeName);
+        Class<?> attributeType = entityDictionary.getParameterizedType(parentType, attributeName);
```
aklish: For attribute - you want to use getType instead of getParameterizedType

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -567,24 +561,34 @@ private FilterExpression buildFilter(String typeName, Object filterString) {
...
 
-        Attribute argumentAttribute = projectionBuilder.getAttributeByName(argumentName);
+        Attribute argumentAttribute = projectionBuilder.getAttributeByAlias(argumentName);
         com.yahoo.elide.request.Argument elideArgument = com.yahoo.elide.request.Argument.builder()
                 .name(argumentName)
                 .value(variableResolver.resolveValue(argument.getValue()))
                 .build();
 
         if (argumentAttribute == null) {
+            Class<?> attributeType = entityDictionary.getParameterizedType(entityType, argumentName);
```
aklish: getType instead of getParameterizedType

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -568,13 +554,24 @@ private void addAttributeArgument(Argument argument, EntityProjectionBuilder pro
...
-        Attribute argumentAttribute = projectionBuilder.getAttributeByAlias(argumentName);
+        Attribute existingAttribute = projectionBuilder.getAttributeByAlias(argumentName);
+
         com.yahoo.elide.request.Argument elideArgument = com.yahoo.elide.request.Argument.builder()
                 .name(argumentName)
                 .value(variableResolver.resolveValue(argument.getValue()))
                 .build();
 
-        if (argumentAttribute == null) {
+        if (existingAttribute != null) {
```
aklish: Do we really need to differentiate these two cases.  The resulting code in each branch is nearly identical.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -568,13 +554,24 @@ private void addAttributeArgument(Argument argument, EntityProjectionBuilder pro
         String argumentName = argument.getName();
         Class<?> entityType = projectionBuilder.getType();
 
-        Attribute argumentAttribute = projectionBuilder.getAttributeByAlias(argumentName);
+        Attribute existingAttribute = projectionBuilder.getAttributeByAlias(argumentName);
```
aklish: This doesn't seem right to me.  We shouldn't be fetching the attribute by the argument name.

PR #991, opened 2019-10-02
Added AggregationDataStore Code

SushrutV:
## Description
Added AggregationDataStore class which converts input parameters into
a query type that can then be used by a QueryEngine.

## How Has This Been Tested?
Unit testing/IntegrationTests using the AggregationDataStore with the 
SQLQueryEngine.

## License


elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/elements/Field.java
```diff
@@ -88,7 +88,17 @@ public String toResponse() {
                             ? "{\"edges\":[]}"
                             : getSelectionSet().toString()
             );
-        } else {
+        }
+        else if (selectionSet instanceof Number) {
+            return String.format(
```
aklish: Can we combine this case with the one above? ```if (selectionSet instanceof String || selectionSet instanceOf Number)```

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -111,7 +127,55 @@
...
+            <type>test-jar</type>
+        </dependency>
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-graphql</artifactId>
+            <version>4.5.2-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-core</artifactId>
```
aklish: Should this be for test?

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -111,7 +127,55 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
```
aklish: We might want to organize the dependencies a bit more.  Break them into test and non-test sections.  Also, dependencies from the same group should be adjacent to one another.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -0,0 +1,78 @@
...
+
+public abstract class AggregationDataStore implements DataStore {
+
+    private QueryEngine queryEngine;
+
+    public AggregationDataStore(QueryEngine queryEngine) {
+        this.queryEngine = queryEngine;
+    }
+
+    public AggregationDataStore() {
```
aklish: Do we need this constructor?  How would this work without a QueryEngine?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -0,0 +1,78 @@
...
+
+    }
+
+    @Override
+    public abstract void populateEntityDictionary(EntityDictionary dictionary);
```
aklish: I don't think this ought to be abstract.  Maybe this could delegate to the QueryEngine?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public DataStoreTransaction beginTransaction() {
+        return new AggregationDataStoreTransaction(queryEngine);
+    }
+
+    public static Query buildQuery(EntityProjection entityProjection, RequestScope scope) {
+        SQLSchema schema = new SQLSchema(entityProjection.getType(), scope.getDictionary());
```
aklish: SQLSchema I believe is specific to the SQLQueryEngine.  This class should not have dependencies on query engine classes.

elide-datastore/elide-datastore-aggregation/pom.xml
hellohanchen: Let's have a comment like `<!-- Junit -->` here and for all other dependencies you added.

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -59,6 +59,22 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
         </dependency>
+        <dependency>
```
hellohanchen: `<scope>test</scope>` dependencies should be placed after regular compile dependencies.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -0,0 +1,78 @@
+package com.yahoo.elide.datastores.aggregation;
```
hellohanchen: Copyright here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
hellohanchen: Why do we have this constructor?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+    public AggregationDataStoreHelper(Schema schema, EntityProjection entityProjection) {
+        this.schema = schema;
+        this.entityProjection = entityProjection;
+    }
+
+    public FilterExpression getWhereFilter(FilterExpression filterExpression) {
+        FilterExpression whereFilter;
+        try {
+            whereFilter = filterExpression.accept(new SplitFilterExpressionVisitor(schema)).getWhereExpression();
+        } catch (NullPointerException npe) {
```
aklish: Why do we catch NPE?  Is this expected behavior?  Usually NPEs are not caught.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public DataStoreTransaction beginTransaction() {
+        return new AggregationDataStoreTransaction(queryEngine);
+    }
+
+    public static Query buildQuery(EntityProjection entityProjection, RequestScope scope) {
+        SQLSchema schema = new SQLSchema(entityProjection.getType(), scope.getDictionary());
+        AggregationDataStoreHelper agHelper = new AggregationDataStoreHelper(schema, entityProjection);
+        FilterExpression filterExpression = entityProjection.getFilterExpression();
```
hellohanchen: `FilterExpression`, `Sorting` and `Pagination` are all from entityProjection, can we just add them into the helper too?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
+package com.yahoo.elide.datastores.aggregation;
```
hellohanchen: Copyright

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -0,0 +1,78 @@
...
+import com.yahoo.elide.request.EntityProjection;
+import sun.awt.image.ImageWatched;
+
+import java.sql.Time;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+public abstract class AggregationDataStore implements DataStore {
```
hellohanchen: Need a javadoc for each class

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+        } catch (NullPointerException npe) {
+            whereFilter = null;
+        }
+        return whereFilter;
+    }
+
+    public FilterExpression getHavingFilter(FilterExpression filterExpression) {
+        FilterExpression havingFilter;
+        try {
+            havingFilter = filterExpression.accept(new SplitFilterExpressionVisitor(schema)).getHavingExpression();
```
hellohanchen: Do we need to construct a `SplitFilterExpressionVisitor` every time we call `getHavingFilter` or `getWhereFilter`?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+import com.yahoo.elide.request.EntityProjection;
+import com.yahoo.elide.request.Relationship;
+
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+public class AggregationDataStoreHelper {
```
hellohanchen: Java doc for class and all public methods.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+        return havingFilter;
+    }
+
+
+    public void populateDimensionList(Set<Dimension> dimensions, Set<TimeDimension> timeDimensions) {
+        Set<String> dimensionNames = getDimensionNames();
+        Set<String> metricNames = getMetricNames(); // time dimensions are under attribute in entity projection
+        for (Dimension dimension : schema.getDimensions()) {
+            if (dimension instanceof TimeDimension) {
+                if(metricNames.contains(dimension.getName()))
```
aklish: Checkstyles won't allow if statements without quotes.  See my later comment about getMetricNames.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+        this.schema = schema;
+        this.entityProjection = entityProjection;
+    }
+
+    public FilterExpression getWhereFilter(FilterExpression filterExpression) {
+        FilterExpression whereFilter;
+        try {
+            whereFilter = filterExpression.accept(new SplitFilterExpressionVisitor(schema)).getWhereExpression();
+        } catch (NullPointerException npe) {
+            whereFilter = null;
```
hellohanchen: If which case this exception is happening? can we check for those conditions before splitting the filter? I am just concerned about returning `null` here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public Map<Metric, Class<? extends Aggregation>> getMetricMap(Set<Metric> metrics) {
+        Map<Metric, Class<? extends Aggregation>> metricMap = new LinkedHashMap<>();
+        for (Metric metric : metrics) {
+            metricMap.put(metric, metric.getAggregations().get(0));
+        }
+        return metricMap;
+    }
+
+    private Set<String> getMetricNames() {
```
aklish: Instead of saying getMetricNames, let's call this getAttributes.  An attribute may be a metric, or a time dimension, or a degenerate dimension.  So getMetricNames makes the calling code look wrong.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+        } catch (NullPointerException npe) {
+            havingFilter = null;
+        }
+        return havingFilter;
+    }
+
+
+    public void populateDimensionList(Set<Dimension> dimensions, Set<TimeDimension> timeDimensions) {
+        Set<String> dimensionNames = getDimensionNames();
+        Set<String> metricNames = getMetricNames(); // time dimensions are under attribute in entity projection
```
hellohanchen: If `dimensionNames` and `metricNames` are all from the entityProjection and we should not modify them, maybe we can make them `private final` variables for this class and assign values for them in the constructor?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+        try {
+            havingFilter = filterExpression.accept(new SplitFilterExpressionVisitor(schema)).getHavingExpression();
+        } catch (NullPointerException npe) {
+            havingFilter = null;
+        }
+        return havingFilter;
+    }
+
+
+    public void populateDimensionList(Set<Dimension> dimensions, Set<TimeDimension> timeDimensions) {
```
hellohanchen: Maybe we can change the pattern from **populateSomething** to **getSomething**?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
```diff
@@ -0,0 +1,51 @@
+package com.yahoo.elide.datastores.aggregation;
```
hellohanchen: Copyright

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+        Set<String> metricNames = new LinkedHashSet<>();
+        for (Attribute attribute : entityProjection.getAttributes()) {
+            metricNames.add(attribute.getName());
+        }
+        return metricNames;
+    }
+
+    private Set<String> getDimensionNames() {
+        Set<String> dimensionNames = new LinkedHashSet<>();
+        for (Relationship relationship : entityProjection.getRelationships()) {
```
aklish: While relationships are always dimensions, attributes can be dimensions as well.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,98 @@
...
+    public void populateMetricList(Set<Metric> metrics) {
+        Set<String> metricNames = getMetricNames();
+        for (Metric metric : schema.getMetrics()) {
+            if (metricNames.contains(metric.getName())) {
+                metrics.add(metric);
+            }
+        }
+    }
+
+    public Map<Metric, Class<? extends Aggregation>> getMetricMap(Set<Metric> metrics) {
```
aklish: This can be rewritten using Java 8 streams in one line.  Checkout collectors.toMap and the identity function.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/StitchList.java
```diff
@@ -81,7 +82,8 @@ public boolean shouldStitch() {
      * @param value  The foreign key between the entity instance and the field entity.
      */
     public void todo(Object entityInstance, String fieldName, Object value) {
-        getTodoList().add(new Todo(entityInstance, fieldName, value));
+        Object coercedValue = CoerceUtil.coerce(value, getEntityDictionary().getIdType(getEntityDictionary().getParameterizedType(entityInstance, fieldName)));
```
hellohanchen: Let's add a comment for this fix, as it is actually fixing a bug

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLEntityHydrator.java
```diff
@@ -68,7 +71,8 @@ public SQLEntityHydrator(
                 .getResultList();
 
         return loaded.stream()
-                .map(obj -> new AbstractMap.SimpleImmutableEntry<>((Object) getEntityDictionary().getId(obj), obj))
+                .map(obj -> new AbstractMap.SimpleImmutableEntry<>(
+                        getEntityDictionary().getValue(obj, "id", null), obj))
```
aklish: getValue is not a safe way to get the ID of an entity.  Why did we change this?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -0,0 +1,361 @@
...
+                                )
+                        )
+                )
+        ).toResponse();
+
+        runQueryWithExpectedResult(graphQLRequest, expected);
+    }
+
+    @Test
+    @Disabled
```
hellohanchen: Add `FIXME` comment here.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -0,0 +1,361 @@
...
+
+    private JsonNode toJsonNode(String query, Map<String, Object> variables) {
+        ObjectNode graphqlNode = JsonNodeFactory.instance.objectNode();
+        graphqlNode.put("query", query);
+        if (variables != null) {
+            graphqlNode.set("variables", JSON_MAPPER.valueToTree(variables));
+        }
+        return graphqlNode;
+    }
+
```
hellohanchen: so many new lines.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/VariableFieldSerializer.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public void serialize(String value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
+        if (value.startsWith(VARIABLE_SIGN)) {
+            // this is a variable
+            gen.writeRawValue(value);
+        } else {
+            gen.writeString(value);
+        }
+    }
+}
```
hellohanchen: \n

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -0,0 +1,361 @@
...
+    private void runQueryWithExpectedResult(
+            String graphQLQuery,
+            Map<String, Object> variables,
+            String expected
+    ) throws IOException {
+        compareJsonObject(runQuery(graphQLQuery, variables), expected);
+    }
+
+    private void runQueryWithExpectedResult(String graphQLQuery, String expected) throws IOException {
```
aklish: A lot of this code looks like it was copied from GraphQLIT or something.  Maybe we should refactor this into a common class?

elide-datastore/elide-datastore-aggregation/src/test/resources/player_stats.csv
```diff
@@ -1,4 +1,5 @@
 id,highScore,lowScore,overallRating,country_id,player_id,recordedDate
-Jon Doe,1234,72,Good,840,1,2019-07-12 00:00:00
+Jon Doe,1234,35,Good,840,1,2019-07-12 00:00:00
 Jane Doe,2412,241,Great,840,2,2019-07-11 00:00:00
 Han,1000,72,Good,344,3,2019-07-13 00:00:00
+
```
hellohanchen: extra \n

elide-datastore/elide-datastore-aggregation/src/test/resources/video_games.csv
```diff
@@ -0,0 +1,5 @@
+id,rounds,timeSpent
+1,10,50
+2,20,150
+3,30,1000
+4,10,200
```
hellohanchen: \n

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/VariableFieldSerializer.java
```diff
@@ -0,0 +1,63 @@
...
+ * To serialize a String entity field in such a way, add the following annotation to the field, as shown above:
+ * <pre>
+ * {@code
+ * {@literal @}JsonSerialize(using = VariableFieldSerializer.class, as = String.class)
+ * }
+ * </pre>
+ *
+ * @see <a href="https://graphql.org/learn/queries/#variables">Variables</a>
+ */
+public class VariableFieldSerializer extends JsonSerializer<String> {
```
aklish: This looks like it was copied from the graphQL DSL test helpers.  Is that correct? If so, why didn't we use the original code?

elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLEndpointTest.java
```diff
@@ -69,11 +69,11 @@
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 public class GraphQLEndpointTest {
 
-    private GraphQLEndpoint endpoint;
-    private final SecurityContext user1 = Mockito.mock(SecurityContext.class);
-    private final SecurityContext user2 = Mockito.mock(SecurityContext.class);
-    private final SecurityContext user3 = Mockito.mock(SecurityContext.class);
-    private final AuditLogger audit = Mockito.mock(AuditLogger.class);
+    protected GraphQLEndpoint endpoint;
```
hellohanchen: Are you modifying this file? Are these changes necessary for your tests?

elide-integration-tests/src/test/java/com/yahoo/elide/initialization/IntegrationTest.java
```diff
@@ -70,10 +70,10 @@ protected IntegrationTest(final Class<? extends ResourceConfig> resourceConfig,
         }
     }
 
-    private DataStoreTestHarness createHarness() {
+    protected DataStoreTestHarness createHarness() {
         try {
             final String dataStoreSupplierName = System.getProperty("dataStoreHarness");
-
+            System.out.println("Datastore: " + dataStoreSupplierName);
```
hellohanchen: Remove this print

aklish: Remove.

elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLEndpointTest.java
```diff
@@ -174,8 +174,120 @@ public Object apply(SecurityContext securityContext) {
         tx.save(noShare, null);
 
         tx.commit(null);
+//        createBookAndAuthor();
+//        createBookAndAuthor2();
     }
 
+    public void createBookAndAuthor() throws Exception {
```
aklish: Not against more tests - just wondering why we added them?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/Author.java
```diff
@@ -0,0 +1,24 @@
...
+import com.yahoo.elide.datastores.aggregation.VariableFieldSerializer;
+import lombok.Getter;
+import lombok.Setter;
+
+import javax.persistence.Entity;
+import javax.persistence.Id;
+
+@Entity
+@Include(rootLevel = true)
+public class Author {
```
aklish: Why did we add an author and book the aggregation data store test directories?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLEntityHydrator.java
hellohanchen: Get id is always returning String.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -1,33 +1,21 @@
...
-
-import java.sql.Time;
-import java.util.LinkedHashSet;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
 
+/**
+ * Aggregation Data Store translates given information into query which can be used by a {@link QueryEngine} to return back results.
+ */
```
hellohanchen: Does AggregationDataStore just **translates** queries? I think something like "DataStore that supports aggregation functionality" would be better for this class.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -15,43 +21,75 @@
...
 import java.util.Map;
-import java.util.Optional;
 import java.util.Set;
 
+/**
+ * Helper for Aggregation Data Store which does the work associated with extracting {@link Query}.
+ */
 public class AggregationDataStoreHelper {
 
+    private static final int AGGREGATION_METHOD_INDEX = 0;
```
hellohanchen: Add a `TODO` comment to mark we are gonna to support selecting aggregation method in the future.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -15,43 +21,75 @@
...
-    public FilterExpression getHavingFilter(FilterExpression filterExpression) {
-        FilterExpression havingFilter;
-        try {
-            havingFilter = filterExpression.accept(new SplitFilterExpressionVisitor(schema)).getHavingExpression();
-        } catch (NullPointerException npe) {
+    /**
+     * Gets whereFilter and havingFilter based on provided filter expression from {@link EntityProjection}.
+     * @param filterExpression {@link FilterExpression} The filter expression to parse.
+     */
+    private void getFilters(FilterExpression filterExpression) {
```
hellohanchen: Do we need to pass in a FilterExpression or we can just use the private variable `entityProjection`. And as this method is not returning anything, we can named it `resolveFilters` or `splitFilters`, similar for other methods in this class.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -63,36 +101,55 @@ else if (dimensionNames.contains(dimension.getName()) || metricNames.contains(di
...
+     * @param metrics Set of {@link Metric} objects.
+     */
+    public void getMetricMap(Set<Metric> metrics) {
+        metricMap = new LinkedHashMap<>();
         for (Metric metric : metrics) {
-            metricMap.put(metric, metric.getAggregations().get(0));
+            metricMap.put(metric, metric.getAggregations().get(AGGREGATION_METHOD_INDEX));
         }
-        return metricMap;
+//        return metrics.stream()
```
hellohanchen: Remove this comment block.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -259,6 +239,7 @@ public void whereFilterTest() throws Exception {
 
     @Test
     @Disabled
+    //FIXME
```
hellohanchen: This comment should explain what features need to be supported to make this test pass.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
SushrutV: Fixed naming conventions

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
SushrutV: GraphQLIT wasn't in the version I was working with, but I can still refactor out these methods for future use if that would be helpful.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
SushrutV: Done

elide-datastore/elide-datastore-aggregation/pom.xml
SushrutV: Tried to fix it as best as I can, if there's still things wrong let me know and I can fix them.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,151 @@
...
+        whereFilter = constraints.getWhereExpression();
+        havingFilter = constraints.getHavingExpression();
+    }
+
+    /**
+     * Gets dimensions and timeDimensions based on relationships and attributes from {@link EntityProjection}.
+     * @param dimensions Empty set of {@link Dimension} objects.
+     * @param timeDimensions Empty set of {@link TimeDimension} objects.
+     */
+    public void resolveDimensionLists(Set<Dimension> dimensions, Set<TimeDimension> timeDimensions) {
```
hellohanchen: Could these methods be `private`?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,151 @@
...
+        Set<String> dimensionNames = getRelationships();
+        Set<String> metricNames = getAttributes(); // time dimensions are under attribute in entity projection
```
hellohanchen: We'd better rename these two to `relationshipNames` and `metricNames` based on your following logic, dimensions can come from both `dimensionNames` and `metricNames`.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,151 @@
...
+            metricMap.put(metric, metric.getAggregations().get(AGGREGATION_METHOD_INDEX));
+        }
+    }
+
+    /**
+     * Gets attribute names from {@link EntityProjection}.
+     * @return attributes list of {@link Attribute} names
+     */
+    private Set<String> getAttributes() {
+        Set<String> attributes = new LinkedHashSet<>();
```
hellohanchen: I think this and `getRelationships()` can use `stream()...`.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -0,0 +1,151 @@
...
+            }
+        }
+    }
+
+    /**
+     * Constructs map between {@link Metric} objects and the type of {@link Aggregation} we want to use.
+     * @param metrics Set of {@link Metric} objects.
+     */
+    public void resolveMetricMap(Set<Metric> metrics) {
+        metricMap = new LinkedHashMap<>();
```
hellohanchen: Maybe it would be better if we put the instantiation of `metricMap` in the constructor.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLEntityHydrator.java
aklish: The keyword "id" is not safe in this context.  This won't work for entities that have remapped the ID field to something named other than ID.  We need to use getId.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
aklish: Sure. You can do it as a separate PR if you want or in this one.

PR #1008, opened 2019-10-11
Moved entityManager creation to happen separately for each query

SushrutV:
Changed SQLQueyEngine so that EntityManager is created separately for each new query instead of being created once when the Engine was initialized.


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/engine/SQLQueryEngine.java
```diff
@@ -82,6 +83,7 @@ public Schema getSchema(Class<?> entityClass) {
 
     @Override
     public Iterable<Object> executeQuery(Query query) {
+        EntityManager entityManager = emf.createEntityManager();
```
aklish: entityManager shoudl be closed after each query.  We should put this in a try block so we don't leak it:
```java 
try ( EntityManager entityManager = emf.createEntityManager()) {
   //do stuff
}
```java

PR #1017, opened 2019-10-16
Column annotation

SushrutV:

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Column.java
```diff
@@ -41,11 +43,16 @@
...
+        this(schema, field, annotation, fieldType, field);
+    }
+
+    public Column(Schema schema, String field, Meta annotation, Class<?> fieldType, String columnName) {
         this.name = Objects.requireNonNull(field, "field");
         this.longName = annotation == null || annotation.longName().isEmpty() ? field : annotation.longName();
         this.description = annotation == null || annotation.description().isEmpty() ? field : annotation.description();
         this.dataType = Objects.requireNonNull(fieldType, "fieldType");
         this.schema = schema;
+        this.columnName = Objects.requireNonNull(columnName, "field");
```
hellohanchen: This looks wrong, should it be `this.columnName = columnName == null ? this.name : columnName`?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngineFactory.java
```diff
@@ -0,0 +1,14 @@
+/*
+ * Copyright 2019, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.datastores.aggregation;
+
+import com.yahoo.elide.core.EntityDictionary;
+
+public interface QueryEngineFactory {
```
hellohanchen: Javadoc

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/SQLQueryEngineFactory.java
```diff
@@ -0,0 +1,21 @@
...
+import com.yahoo.elide.datastores.aggregation.engine.SQLQueryEngine;
+
+import javax.persistence.EntityManagerFactory;
+import javax.persistence.Persistence;
+
+public class SQLQueryEngineFactory implements QueryEngineFactory {
+
+    @Override
+    public QueryEngine buildQueryEngine(EntityDictionary dictionary) {
+        EntityManagerFactory emf = Persistence.createEntityManagerFactory("aggregationStore");
```
hellohanchen: This is not correct, the EntityManagerFactory should be passed in.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/schema/Schema.java
```diff
@@ -182,12 +182,18 @@ protected Metric constructMetric(String metricField, Class<?> cls, EntityDiction
                         )
                 );
 
+        String columnName = metricField;
```
aklish: There is already a function in SQLSchema that does this.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -8,23 +8,34 @@
...
+        this.queryEngineFactory = queryEngineFactory;
+        queryEngine = null;
     }
 
     @Override
     public abstract void populateEntityDictionary(EntityDictionary dictionary);
 
     @Override
     public DataStoreTransaction beginTransaction() {
+        if (queryEngine == null) {
```
aklish: Instead of doing this here in the transaction - we should do this in populateEntityDictionary.

aklish: populateEntityDictionary should:

1. Create the query engine
2. call getSchemas (a new method) on the QueryEngine.
3. Use the schemas returned to bind all the entities.
4. No longer be abstract

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/Column.java
```diff
@@ -29,6 +29,8 @@
     protected final Class<?> dataType;
     @Getter
     protected final Schema schema;
+    @Getter
+    protected final String columnName;
```
aklish: This doesn't belong here.   The concept of a columnName is specific to SQL.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/SQLQueryEngineFactory.java
aklish: Agreed.  The factory will be created by the service at startup and passed in.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metric/AggregatedMetric.java
```diff
@@ -49,7 +49,18 @@ public AggregatedMetric(
...
+        this(schema, metricField, annotation, fieldType, aggregations, metricField);
+    }
+
+    public AggregatedMetric(
+            Schema schema,
+            String metricField,
+            Meta annotation,
+            Class<?> fieldType,
+            List<Class<? extends Aggregation>> aggregations,
+            String columnName
```
aklish: I don't think we should add columnName as a concept at this level.  If we need it, it should go into the SQL subclasses/extensions of these classes.

elide-datastore/elide-datastore-aggregation/src/test/resources/create_tables.sql
```diff
@@ -25,6 +25,6 @@ CREATE TABLE IF NOT EXISTS player
 CREATE TABLE IF NOT EXISTS videoGame
     (
       id BIGINT,
-      sessions BIGINT,
+      rounds BIGINT,
```
aklish: Why the rename?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTestHarness.java
```diff
@@ -15,21 +15,23 @@
 import com.yahoo.elide.datastores.aggregation.example.VideoGame;
 
 public class AggregationDataStoreTestHarness implements DataStoreTestHarness {
-    private QueryEngine qE;
+    private QueryEngineFactory queryEngineFactory;
```
aklish: We should add an explicit test for a metric column with an '_' in it.

PR #1020, opened 2019-10-16
Add SQLMetrix, rearrange packages

hellohanchen:


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/schema/SQLSchema.java
```diff
@@ -80,6 +85,38 @@ protected Dimension constructDimension(String dimensionField, Class<?> cls, Enti
...
+     */
+    @Override
+    protected Metric constructMetric(String metricField, Class<?> cls, EntityDictionary entityDictionary) {
+        Meta metaData = entityDictionary.getAttributeOrRelationAnnotation(cls, Meta.class, metricField);
+        Class<?> fieldType = entityDictionary.getType(cls, metricField);
+
+        List<Class<? extends Aggregation>> aggregations = getAggregations(metricField, cls, entityDictionary);
+
+        String columnName = metricField;
+        if (entityDictionary.attributeOrRelationAnnotationExists(cls, metricField, javax.persistence.Column.class)) {
```
aklish: Can we use the functionality it SQLSchema to do this instead?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -18,15 +18,15 @@
...
-import com.yahoo.elide.datastores.aggregation.metric.Metric;
+import com.yahoo.elide.datastores.aggregation.queryengines.sql.annotation.FromSubquery;
+import com.yahoo.elide.datastores.aggregation.queryengines.sql.annotation.FromTable;
+import com.yahoo.elide.datastores.aggregation.queryengines.sql.annotation.JoinTo;
+import com.yahoo.elide.datastores.aggregation.queryengines.sql.schema.SQLDimension;
+import com.yahoo.elide.datastores.aggregation.queryengines.sql.schema.SQLSchema;
 import com.yahoo.elide.datastores.aggregation.schema.Schema;
+import com.yahoo.elide.datastores.aggregation.schema.dimension.Dimension;
+import com.yahoo.elide.datastores.aggregation.schema.metric.Aggregation;
+import com.yahoo.elide.datastores.aggregation.schema.metric.Metric;
```
aklish: Don't we need to make changes to this class to call the new getMetricName function?

hellohanchen: No, the metric expression is constructed by the metric itself.

PR #1029, opened 2019-10-18
Fixed sorting on aggregation fields issue

### LINKED ISSUE #1023: Elide 5.0 - SQLQueryEngine: Fix Sorting on AggregationMetric
```SQL
```
```SQL
ORDER BY field_name
```
Which is not correct.

## Possible Solution
And add test

SushrutV:
Resolves #1023 

## Description
Added code in extractOrderBy to handle sorting on aggregated metrics.

## How Has This Been Tested?
mvn clean test

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
@@ -5,6 +5,7 @@
  */
 package com.yahoo.elide.datastores.aggregation;
 
+import com.yahoo.elide.core.EntityDictionary;
 import com.yahoo.elide.core.Path;
 import com.yahoo.elide.core.exceptions.InvalidOperationException;
 import com.yahoo.elide.core.filter.FilterPredicate;
@@ -12,6 +13,7 @@ import com.yahoo.elide.core.filter.expression.AndFilterExpression;
 import com.yahoo.elide.core.filter.expression.FilterExpression;
 import com.yahoo.elide.core.filter.expression.NotFilterExpression;
 import com.yahoo.elide.core.filter.expression.OrFilterExpression;
+import com.yahoo.elide.core.sort.Sorting;
 import com.yahoo.elide.datastores.aggregation.annotation.TimeGrainDefinition;
 import com.yahoo.elide.datastores.aggregation.filter.visitor.FilterConstraints;
 import com.yahoo.elide.datastores.aggregation.filter.visitor.SplitFilterExpressionVisitor;
@@ -51,8 +53,9 @@ public class AggregationDataStoreHelper {
     private Map<Metric, Class<? extends Aggregation>> metricMap;
     private FilterExpression whereFilter;
     private FilterExpression havingFilter;
+    private EntityDictionary dictionary;
 
-    public AggregationDataStoreHelper(Schema schema, EntityProjection entityProjection) {
+    public AggregationDataStoreHelper(Schema schema, EntityProjection entityProjection, EntityDictionary dictionary) {
         this.schema = schema;
         this.entityProjection = entityProjection;
         dimensionProjections = resolveNonTimeDimensions();
@@ -62,6 +65,8 @@ public class AggregationDataStoreHelper {
         metricMap = new LinkedHashMap<>();
         resolveMetricMap(metrics);
         splitFilters();
+        this.dictionary = dictionary;
+        validateSorting(entityProjection.getSorting());
     }
 
     /**
@@ -259,6 +264,37 @@ public class AggregationDataStoreHelper {
             validateHavingClause(((OrFilterExpression) havingClause).getRight());
         } else if (havingClause instanceof NotFilterExpression) {
             validateHavingClause(((NotFilterExpression) havingClause).getNegated());
+        }
+    }
+
+    private void validateSorting(Sorting sorting) {
+        if (sorting != null) {
+            Map<Path, Sorting.SortOrder> sortClauses =
+                    sorting.getValidSortingRules(schema.getEntityClass(), dictionary);
+            sortClauses.entrySet().stream()
+                    .forEach((entry) -> {
+                        Path path = entry.getKey();
+                        if (path.lastElement().get().getType() == schema.getEntityClass()) {
+                            String currentField = path.lastElement().get().getFieldName();
+                            checkFieldStatus(currentField);
+                        }
+
+            });
+        }
+    }
+
+    private void checkFieldStatus(String currentField) {
+        if (schema.getDimensions().stream().anyMatch(dim -> dim.getName().contains(currentField))) {
+            if (!dimensionProjections.stream().anyMatch(dim -> dim.getName().equals(currentField))) {
+                // Add current field to dimensionProjections
+                dimensionProjections.add(schema.getDimension(currentField).toProjectedDimension());
+            }
+        }
+        else if (schema.getMetrics().stream().anyMatch(m -> m.getName().contains(currentField))) {
+            if (!metricMap.keySet().stream().anyMatch(m -> m.getName().equals(currentField))) {
+                // Metric not specified in query
+                throw new InvalidOperationException("Can't sort on metric that is not present in query");
+            }
         }
     }
 }
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
@@ -268,6 +268,11 @@ public class AggregationDataStoreHelper {
         }
     }
 
+    /**
+     * Gets fields that are required for sorting based on {@link EntityProjection}.
+     * @param sorting {@link Sorting} that represents the sort order specified by the user
+     * @return List of names of fields that need to be sorted on
+     */
     private Set<String> getSortingFields(Sorting sorting) {
         Set<Path.PathElement> sortingFields = new LinkedHashSet<>();
         if (sorting != null) {
@@ -285,10 +290,14 @@ public class AggregationDataStoreHelper {
                         sortingFields.add(currentElement);
             });
         }
-        validateSorting(sortingFields);
+        sortingFields.stream().forEach(this::validateFieldStatus);
         return sortingFields.stream().map(sf -> sf.getFieldName()).collect(Collectors.toSet());
     }
 
+    /**
+     * Verifies that the current element in the path can be sorted on
+     * @param currentElement The field that we are inspecting and potentially sorting on
+     */
     private void validateFieldStatus(Path.PathElement currentElement) {
         String currentField = currentElement.getFieldName();
         Class<?> currentClass = currentElement.getType();
@@ -298,12 +307,8 @@ public class AggregationDataStoreHelper {
                 throw new InvalidOperationException("Can't sort on metric that is not present in query");
             }
         }
-        if (dictionary.getIdFieldName(currentClass).equals(currentField)) {
+        if (dictionary.getIdFieldName(currentClass).equals(currentField) || currentField.equals("id")) {
             throw new InvalidOperationException("Sorting on id field is not permitted");
         }
-    }
-
-    private void validateSorting(Set<Path.PathElement> sortingFields) {
-        sortingFields.stream().forEach(this::validateFieldStatus);
     }
 }
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
@@ -93,51 +93,57 @@ public class AggregationDataStoreHelper {
     }
 
     //TODO - Add tests in the next PR.
+    /**
+     * Gets time dimensions based on relationships and attributes from {@link EntityProjection}.
+     *
+     * @return projections for time dimension columns
+     * @throws InvalidOperationException Thrown if a requested time grain is not supported.
+     */
     private Set<TimeDimensionProjection> resolveTimeDimensions() {
-        Set<TimeDimensionProjection> timeDims = new LinkedHashSet<>();
-        //Only attributes can be time dimensions
-        entityProjection.getAttributes().stream().forEach((attribute -> {
-            TimeDimensionColumn timeDim = schema.getTimeDimension(attribute.getName());
-            if (timeDim == null) {
-                return;
-            }
+        return entityProjection.getAttributes().stream()
+                .filter(attribute -> schema.getTimeDimension(attribute.getName()) != null)
+                .map(attribute -> {
+                    TimeDimensionColumn timeDim = schema.getTimeDimension(attribute.getName());
 
-            Argument timeGrainArgument = attribute.getArguments().stream()
-                    .filter(attr -> attr.getName().equals("grain"))
-                    .findAny()
-                    .orElse(null);
+                    Argument timeGrainArgument = attribute.getArguments().stream()
+                            .filter(attr -> attr.getName().equals("grain"))
+                            .findAny()
+                            .orElse(null);
 
-            TimeGrainDefinition requestedGrainDefinition;
-            if (timeGrainArgument == null) {
+                    TimeGrainDefinition requestedGrainDefinition;
+                    if (timeGrainArgument == null) {
 
-                //The first grain is the default.
-                requestedGrainDefinition = timeDim.getSupportedGrains().iterator().next();
-            } else {
-                String requestedGrainName = timeGrainArgument.getValue().toString();
+                        //The first grain is the default.
+                        requestedGrainDefinition = timeDim.getSupportedGrains().stream()
+                                .findFirst()
+                                .orElseThrow(() -> new IllegalStateException(
+                                        String.format("Requested default grain, no grain defined on %s",
+                                                attribute.getName())));
+                    } else {
+                        String requestedGrainName = timeGrainArgument.getValue().toString();
 
-                TimeGrain requestedGrain;
-                try {
-                    requestedGrain = TimeGrain.valueOf(requestedGrainName);
-                } catch (IllegalArgumentException e) {
-                    throw new InvalidOperationException(String.format("Invalid grain %s", requestedGrainName));
-                }
+                        TimeGrain requestedGrain;
+                        try {
+                            requestedGrain = TimeGrain.valueOf(requestedGrainName);
+                        } catch (IllegalArgumentException e) {
+                            throw new InvalidOperationException(String.format("Invalid grain %s", requestedGrainName));
+                        }
 
-                requestedGrainDefinition = timeDim.getSupportedGrains().stream()
-                        .filter(supportedGrainDef -> supportedGrainDef.grain().equals(requestedGrain))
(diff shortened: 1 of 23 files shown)


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -356,9 +356,18 @@ private String extractOrderBy(Class<?> entityClass, Map<Path, Sorting.SortOrder>
...
                     Path.PathElement last = path.lastElement().get();
 
-                    return FilterPredicate.getTypeAlias(last.getType())
+                    String orderByType = FilterPredicate.getTypeAlias(last.getType())
                             + "."
-                            + getColumnName(entityClass, last.getFieldName())
+                            + getColumnName(entityClass, last.getFieldName());
+
+                    List<Metric> metricField = query.getMetrics().keySet().stream()
+                            .filter(m -> m.getName().equals(last.getFieldName())).collect(Collectors.toList());
```
aklish: It isn't safe to just check the field name of the last element of the path.  Field names could be duplicated in different entities.  We really need to check here that the path is of size one (a single path element) and the field name matches a metric in the query.

I also think it would be good to promote the function to search for a metric in the query as a default method in the Query itself.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -226,6 +226,114 @@ public void whereFilterTest() throws Exception {
         runQueryWithExpectedResult(graphQLRequest, expected);
     }
 
+    @Test
+    public void sortingTest() throws Exception {
```
aklish: Can we also add a test to the SQLQueryEngineTest class as well?  These tests are great - but I'd like to have full coverage for proper SQL generation at the UT level.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngineFactory.java
```diff
@@ -12,6 +12,6 @@
  */
 public interface QueryEngineFactory {
 
-    public QueryEngine buildQueryEngine(EntityDictionary dictionary);
+    QueryEngine buildQueryEngine(EntityDictionary dictionary);
```
hellohanchen: I think we can leave it public here.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -226,6 +226,114 @@ public void whereFilterTest() throws Exception {
...
+                                )
+                        )
+                )
+        ).toResponse();
+
+        runQueryWithExpectedResult(graphQLRequest, expected);
+    }
+
+    @Test
+    public void aggregationSortingTest() throws Exception {
```
hellohanchen: Can we have a test that sorting on 2 fields? 1 is metric and 1 is dimension.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -195,4 +215,86 @@ private void resolveMetricMap(Set<Metric> metrics) {
...
+                        if (path.lastElement().get().getType() == schema.getEntityClass()) {
+                            String currentField = path.lastElement().get().getFieldName();
+                            checkFieldStatus(currentField);
+                        }
+
+            });
+        }
+    }
+
+    private void checkFieldStatus(String currentField) {
```
hellohanchen: Javadoc

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -195,4 +215,86 @@ private void resolveMetricMap(Set<Metric> metrics) {
...
+    }
+
+    private void validateSorting(Sorting sorting) {
+        if (sorting != null) {
+            Map<Path, Sorting.SortOrder> sortClauses =
+                    sorting.getValidSortingRules(schema.getEntityClass(), dictionary);
+            sortClauses.entrySet().stream()
+                    .forEach((entry) -> {
+                        Path path = entry.getKey();
+                        if (path.lastElement().get().getType() == schema.getEntityClass()) {
```
hellohanchen: I think if `path.lastElement().get().getType() != schema.getEntityClass()` then we are sorting on a field of a relationship, then we need to make sure that relationship field is in group by dimension. For example
```
www.haha.com:4443/book?sorting=publisher.age
```
In this case we need to make sure we are grouping by publisher.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -195,4 +216,94 @@ private void resolveMetricMap(Set<Metric> metrics) {
...
+            validateHavingClause(((AndFilterExpression) havingClause).getRight());
+        } else if (havingClause instanceof OrFilterExpression) {
+            validateHavingClause(((OrFilterExpression) havingClause).getLeft());
+            validateHavingClause(((OrFilterExpression) havingClause).getRight());
+        } else if (havingClause instanceof NotFilterExpression) {
+            validateHavingClause(((NotFilterExpression) havingClause).getNegated());
+        }
+    }
+
+    private Set<String> getSortingFields(Sorting sorting) {
```
hellohanchen: Javadoc for new methods

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -195,4 +216,94 @@ private void resolveMetricMap(Set<Metric> metrics) {
...
+    private void validateFieldStatus(Path.PathElement currentElement) {
+        String currentField = currentElement.getFieldName();
+        Class<?> currentClass = currentElement.getType();
+        if (schema.getMetrics().stream().anyMatch(m -> m.getName().contains(currentField))) {
+            if (metricMap.keySet().stream().noneMatch(m -> m.getName().equals(currentField))) {
+                // Metric not specified in query
+                throw new InvalidOperationException("Can't sort on metric that is not present in query");
+            }
+        }
+        if (dictionary.getIdFieldName(currentClass).equals(currentField)) {
```
hellohanchen: Also reject query for field name **id** which is a reserved keyword for all id fields.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -195,4 +216,94 @@ private void resolveMetricMap(Set<Metric> metrics) {
...
+            throw new InvalidOperationException("Sorting on id field is not permitted");
+        }
+    }
+
+    private void validateSorting(Set<Path.PathElement> sortingFields) {
```
hellohanchen: Do we need this method?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -360,9 +357,18 @@ private String extractOrderBy(Class<?> entityClass, Map<Path, Sorting.SortOrder>
...
 
-                    return FilterPredicate.getTypeAlias(last.getType())
+                    String orderByType = FilterPredicate.getTypeAlias(last.getType())
                             + "."
-                            + getColumnName(entityClass, last.getFieldName())
+                            + getColumnName(entityClass, last.getFieldName());
+
+                    if (last.getType() == entityClass) {
+                        Metric currentMetric = schema.getMetric(getColumnName(entityClass, last.getFieldName()));
+                        if (currentMetric != null && query.getMetrics().keySet().contains(currentMetric)) {
```
hellohanchen: Can we use `query.getMetrics().containsKey()`?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -360,9 +357,18 @@ private String extractOrderBy(Class<?> entityClass, Map<Path, Sorting.SortOrder>
 
                     Path.PathElement last = path.lastElement().get();
 
-                    return FilterPredicate.getTypeAlias(last.getType())
+                    String orderByType = FilterPredicate.getTypeAlias(last.getType())
                             + "."
-                            + getColumnName(entityClass, last.getFieldName())
+                            + getColumnName(entityClass, last.getFieldName());
+
+                    if (last.getType() == entityClass) {
```
aklish: We should use equals instead of ==.

aklish: Scratch that.  The check is unnecessary - just remove it.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -360,9 +357,18 @@ private String extractOrderBy(Class<?> entityClass, Map<Path, Sorting.SortOrder>
...
+                        Metric currentMetric = schema.getMetric(getColumnName(entityClass, last.getFieldName()));
+                        if (currentMetric != null && query.getMetrics().containsKey(currentMetric)) {
```
aklish: The else clause for this if statement doesn't really make sense.  What if the column in question is a metric but it isn't in the query metrics - what should we do?  This would be an error and we ought to throw an exception.  However, I don't think it is a good idea to have query validation code scattered everywhere.  We should put that all in one place in a function that validates the query is valid.  All the remaining code can and should assume the query is safe.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -360,9 +357,18 @@ private String extractOrderBy(Class<?> entityClass, Map<Path, Sorting.SortOrder>
...
                     Path.PathElement last = path.lastElement().get();
 
-                    return FilterPredicate.getTypeAlias(last.getType())
+                    String orderByType = FilterPredicate.getTypeAlias(last.getType())
                             + "."
-                            + getColumnName(entityClass, last.getFieldName())
+                            + getColumnName(entityClass, last.getFieldName());
+
+                    if (last.getType() == entityClass) {
+                        Metric currentMetric = schema.getMetric(getColumnName(entityClass, last.getFieldName()));
```
aklish: We should not be using the columnName to lookup a metric in the schema.

The columnName is the physical name of the column in the database.  The schema stores things by their logical names.  This will be a bug if the column and schema names don't match.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/SubCountry.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * A root level entity for testing AggregationDataStore with @Subselect annotation
+ */
+@Data
+@Entity
+@Include(rootLevel = true)
+@Subselect(value = "select * from countries")
+@Cardinality(size = CardinalitySize.SMALL)
+public class SubCountry {
```
aklish: I'm not aware of a concept of a SubCountry.  It would be nice if we kept the test model representing something that is meaningful and fits the overall model example.  Let's find something else to model this use case.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -195,4 +216,99 @@ private void resolveMetricMap(Set<Metric> metrics) {
...
+     * the fields in having clause must be either or these two:
+     * 1. A grouped by dimension in this query
+     * 2. An aggregated metric in this query
+     *
+     * All grouped by dimensions are defined in the entity bean, so the last entity class of a filter path
+     * must match entity class of the query.
+     *
+     * @param havingClause having clause generated from this query
+     */
+    private void validateHavingClause(FilterExpression havingClause) {
```
aklish: In general, we should avoid class names like Helper.  This class can (over time) become a catch all for random functionality.  Let's move this functionality into proper classes.  

One thing we clearly need is a concept of a QueryValidator.  Functions like this one can be moved there.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -195,4 +216,99 @@ private void resolveMetricMap(Set<Metric> metrics) {
...
+        if (havingClause instanceof FilterPredicate) {
```
aklish: This functions should be broken up or we'll get cyclomatic complexity lint errors

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
aklish: Another class we need is a class which translates from the EntityProjection to the Query class.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -90,64 +99,76 @@ private void splitFilters() {
...
 
     /**
-     * Gets dimensions based on relationships and attributes from {@link EntityProjection}.
+     * Gets dimensions except time dimensions based on relationships and attributes from {@link EntityProjection}.
      */
     private Set<DimensionProjection> resolveNonTimeDimensions() {
 
         Set<String> allColumns = getAttributes();
         allColumns.addAll(getRelationships());
+        Set<String> sortingFields = getSortingFields(entityProjection.getSorting());
```
aklish: We shouldn't need to add the sortingFields.  If we are sorting by the field, we are also projecting by it (and the query validation logic can make sure that's the case).  Let's remove this redundancy.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
aklish: We should be able to remove this whole function (see comment above).

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/schema/SQLSchema.java
```diff
@@ -138,14 +141,36 @@ public static String getColumnName(EntityDictionary entityDictionary, Class<?> c
...
+        Subselect subselectAnnotation = entityDictionary.getAnnotation(cls, Subselect.class);
+
+        if (subselectAnnotation == null) {
+            Table tableAnnotation = entityDictionary.getAnnotation(cls, Table.class);
+
+            return (tableAnnotation == null)
+                    ? entityDictionary.getJsonAliasFor(cls)
+                    : tableAnnotation.name();
+        } else {
+            return "(" + subselectAnnotation.value() + ")";
```
aklish: I feel a bit uncomfortable adding SQL formatting here like parenthesis.  That's spreading the responsibility of generating the SQL code across many classes.  The schema should just be about returning metadata - not doing SQL expansion.  

Also - should this really be a static method?   @hellohanchen - I think this overlaps with your metadata work.  We need a metadata repository about AnalyticViews and Entities.  An Entity should be able to fetch where its table name comes from.  I think your PR is going to be a pretty large refactor of where this kind of logic lives.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -169,6 +169,222 @@ public void havingFilterTest() throws Exception {
         runQueryWithExpectedResult(graphQLRequest, expected);
     }
 
+    /**
+     * Test the case that a where clause is promoted into having clause.
+     * @throws Exception
+     */
+    @Test
+    public void wherePromotionTest() throws Exception {
+        String graphQLRequest = document(
```
aklish: @hellohanchen @SushrutV We had discussed about not promoting where clauses into havings because it allows things that simply don't make sense.    @hellohanchen Do you remember the context for that discussion?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -226,6 +439,273 @@ public void whereFilterTest() throws Exception {
...
+                )
+        ).toQuery();
+
+        String expected = "\"Exception while fetching data (/playerStats) : Invalid operation: 'Can't sort on metric that is not present in query'\"";
+
+        runQueryWithExpectedError(graphQLRequest, expected);
+    }
+
+    @Test
+    public void invalidQueryTest() throws Exception {
```
aklish: Let's make the function name more explicit about what is invalid here.  We'll probably have lots of invalid query tests.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -226,6 +439,273 @@ public void whereFilterTest() throws Exception {
...
+                                )
+                        )
+                )
+        ).toResponse();
+
+        runQueryWithExpectedResult(graphQLRequest, expected);
+    }
+
+    @Test
+    public void implicitNestedSortingTest() throws Exception {
```
aklish: What is this trying to test?  I can't glean it from the function name.

SushrutV:
Split this into separate PRs from a new branch

PR #1037, opened 2019-10-21
Avoid checking share permission for redundant add

suriyasundar92:
## Description
We have been calling addRelation to parent entity for updates to child
entities in graphql. This checks the share permission for the child
entity and may fail in some cases.
This PR avoids checking the share permission in case when the entity is
already a part of the relation with the parent.

## How Has This Been Tested?
The unit tests are in the file below.
elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceNoopUpdateTest.java

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -636,12 +636,37 @@ public class PersistentResource<T> implements com.yahoo.elide.security.Persisten
     }
 
     /**
+     * Checks whether the new entity is already a part of the relationship
+     * @param fieldName which relation link
+     * @param toAdd the new relation
+     * @return boolean representing the result of the check
+     *
+     * This method does not handle the case where the toAdd entity is newly created,
+     * since newly created entities have null id there is no easy way to check for identity
+     */
+    public boolean relationshipAlreadyExists(String fieldName, PersistentResource toAdd) {
+        Object relation = this.getValueUnchecked(fieldName);
+        String toAddId = toAdd.getId();
+        if (toAddId == null) {
+            return false;
+        }
+        if (relation instanceof Collection) {
+            return ((Collection) relation).stream().anyMatch((obj -> dictionary.getId(obj).equals(toAddId)));
+        } else {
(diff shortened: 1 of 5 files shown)


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -635,13 +635,38 @@ public void removeRelation(String fieldName, PersistentResource removeResource)
...
+     * since newly created entities have null id there is no easy way to check for identity
+     */
+    public boolean relationshipAlreadyExists(String fieldName, PersistentResource toAdd) {
+        Object relation = this.getValueUnchecked(fieldName);
+        String toAddId = toAdd.getId();
+        if (toAddId == null) {
+            return false;
+        }
+        if (relation instanceof Collection) {
+            return ((Collection) relation).stream().anyMatch((obj -> dictionary.getId(obj).equals(toAddId)));
```
aklish: Let's swap the operands of the equals here since we only checked for null on toAddId.

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceNoopUpdateTest.java
```diff
@@ -0,0 +1,104 @@
...
+
+        DataStoreTransaction tx = mock(DataStoreTransaction.class);
+
+        RequestScope goodScope = new RequestScope(null, null, tx, goodUser, null, elideSettings);
+        PersistentResource<FunWithPermissions> funResource = new PersistentResource<>(fun, null, "3", goodScope);
+        PersistentResource<Child> childResource = new PersistentResource<>(child, null, null, goodScope);
+        //We do not want the update to one method to be called when we add the existing entity to the relation
+        funResource.addRelation("relation1", childResource);
+        verify(tx, never()).updateToManyRelation(eq(tx), eq(child), eq("relation1"), any(), any(), eq(goodScope));
+        //verify(tx, times(1)).updateToManyRelation(eq(tx), eq(child), eq("relation1"), any(), any(), eq(goodScope));
```
aklish: Can we remove the comment?

Issue #1027, opened 2019-10-18
Elide 5.0 - SQLQueryEngine: Add join clause for attributes in having clause


hellohanchen:
When QueryEngine splits a query's original filter expression into where clause and having clause, it would do **where clause promotion** in some cases which pushes all where clause into having clause. For example
```
WHERE a OR HAVING b
```
to be 
```
HAVING a OR b
```
If `WHERE a` is a where clause that require joins, after it is pushed into having clause, the joins would not happen as `toSQL` method currently doesn't create joins for having clause.

## Solution
Fix the issue and add test cases

chczy:
This issue is closed as the PR is already merge.

PR #1039, opened 2019-10-21
ISSUE-1027 Support join for having clause

hellohanchen:
Resolve https://github.com/yahoo/elide/issues/1027
- Add join predicates for having clause.
- Add having clause validation in `AggregationDataStoreHelper` to make sure fields in having clause satisfy aggregation condition.
- Add unit test and integration tests.


elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -169,6 +169,222 @@ public void havingFilterTest() throws Exception {
...
+     * @throws Exception
+     */
+    @Test
+    public void invalidHavingClauseClassTest() throws Exception {
+        String graphQLRequest = document(
+                selection(
+                        field(
+                                "playerStats",
+                                arguments(
+                                        argument("filter", "\"country.isoCode==\\\"USA\\\",lowScore<\\\"45\\\"\"")
```
aklish: What if I wanted to filter by country.id (and also group by country).  In that case, the query could still work.  I think our logic might be too strict for that corner case.

PR #1042, opened 2019-10-22
Add time grain to GraphQL schema

aklish:
## Description
The AggregationDataStore now supports an `@Temporal` annotation that marks columns as time dimensions that support different grains.

## Motivation and Context
Time based dimensions can be annotated as follows:
```java
    @Temporal(grains = {
            @TimeGrainDefinition(grain = TimeGrain.DAY, expression = DAY_FORMAT),
            @TimeGrainDefinition(grain = TimeGrain.MONTH, expression = MONTH_FORMAT)
    }, timeZone = "UTC")
   Date myDate;
```

SQL expressions provided to the `TimeGrainDefinitions` are templates:
```java
    public static final String DAY_FORMAT = "PARSEDATETIME(FORMATDATETIME(%s, 'yyyy-MM-dd'), 'yyyy-MM-dd')";
    public static final String MONTH_FORMAT = "PARSEDATETIME(FORMATDATETIME(%s, 'yyyy-MM-01'), 'yyyy-MM-dd')";
```

In graphQL, these dimensions can be requested like:
```
{ playerStats { edges { node { highScore, myDate(grain: "month") } } } }
```
## How Has This Been Tested?
New unit and IT tests.

## License


elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1422,6 +1423,16 @@ public void addArgumentsToAttributes(Class<?> cls, String attributeName, Set<Arg
         getEntityBinding(cls).addArgumentsToAttribute(attributeName, arguments);
     }
 
+    /**
+     * Add a single argument to the attribute
+     * @param cls The entity
+     * @param attributeName attribute name to which argument has to be added
+     * @param argument A single argument
+     */
+    public void addArgumentToAttributes(Class<?> cls, String attributeName, ArgumentType argument) {
```
hellohanchen: It looks like this is `addArgumentToAttribute`, no `s` at the end.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/schema/Schema.java
```diff
@@ -53,7 +53,7 @@
 public class Schema {
 
     @Getter
-    protected final Class<?> entityClass;
+    public final Class<?> entityClass;
```
hellohanchen: Should this be a PUBLIC level getter?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -109,46 +110,47 @@ private void splitFilters() {
...
-        return entityProjection.getAttributes().stream()
-                .filter(attribute -> schema.getTimeDimension(attribute.getName()) != null)
-                .map(attribute -> {
-                    TimeDimensionColumn timeDim = schema.getTimeDimension(attribute.getName());
 
-                    Argument timeGrainArgument = attribute.getArguments().stream()
-                            .filter(attr -> attr.getName().equals("grain"))
-                            .findAny()
-                            .orElse(null);
+        Set<TimeDimensionProjection> timeDims = new LinkedHashSet<>();
```
hellohanchen: This method is reverted.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/schema/SQLSchema.java
```diff
@@ -76,7 +75,7 @@ protected DimensionColumn constructDimension(String dimensionField,
             String columnName = getColumnName(entityClass, dimensionField);
 
             if (dim instanceof TimeDimensionColumn) {
-                return new SQLTimeDimensionColumn((TimeDimensionColumn) dim, columnName, getAlias());
+               return new SQLTimeDimensionColumn((TimeDimensionColumn) dim, columnName, getAlias());
```
hellohanchen: One space is missing here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/schema/SQLTimeDimensionColumn.java
```diff
@@ -50,4 +55,21 @@ public TimeZone getTimeZone() {
...
     }
+
+    /**
+     * Returns a String that identifies this dimension in a SQL query.
+     * @param requestedGrain The requested time grain.
+     *
+     * @return Something like "table_alias.column_name"
+     */
+    public String getColumnReference(TimeGrain requestedGrain) {
+        Preconditions.checkArgument(getSupportedGrains().stream()
```
hellohanchen: Why don't we do `findFirst().orElseThrow()` here?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
```diff
@@ -477,6 +477,36 @@ public void aggregationComputedMetricTest() throws Exception {
         runQueryWithExpectedResult(graphQLRequest, expected);
     }
 
+    @Test
+    public void timeGrainAggregationTest() throws Exception {
```
hellohanchen: I think we can add more records into the test data and do both daily and monthly aggregation.

aklish: The point of the IT tests is more to test the integration of the API with the backend.  If we need more tests - they belong in the SQLQueryEngineTest.  I did both daily and monthly tests there.  Do you feel like we should add more?

hellohanchen: I personally prefer to put more tests in IntegrationTest to make sure we are generating the expected `Query` object.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
aklish: 👍

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreIntegrationTest.java
aklish: What tests do you suggest adding to more fully test the Query object?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/schema/SQLTimeDimensionColumn.java
aklish: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/schema/Schema.java
aklish: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
aklish: 👍

PR #1044, opened 2019-10-25
Support for multiple queries at root is added

### LINKED ISSUE #994: Elide 5.0: Running Multiple Queries in same transaction is broke in GraphQL
As `EntityProjection` currently requires an entityClass, we can't have multiple queries, where each query has its own entity class, in the same transaction. This wouldn't affect the JSON api endpoint but GraphQL endpoint previously support running multiple queries at the same time.

## Expected Behavior
We should support multiple queries in one request.
```GraphQL
query {
  Book {
    ...
  }
  Author {
    ...
  }
}
```

## Current Behavior
This request above would be rejected by graphql endpoint.

Chandrasekar-Rajasekar:
Resolves #994

## Description
GraphQL now handles multiple queries all at the root level. The EntityProjection for all entities that are queried at the root level is mapped to its entity name in GraphQLProjectionInfo. The root container selects the EntityProjection using the entity name which is obtained from Environment Context.

## Motivation and Context
Multiple queries in the same transaction were previously not working. 

## How Has This Been Tested?
Unit testing that hits GraphQL Endpoint querying Author and Book entity both at root level. 

## License

### DIFF
--- a/elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLRequestScope.java
+++ b/elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLRequestScope.java
@@ -38,9 +38,9 @@ public class GraphQLRequestScope extends RequestScope {
         super("/", null, transaction, user, new MultivaluedHashMap<>(), elideSettings);
         this.projectionInfo = projectionInfo;
 
-        // TODO: handle multiple projection in one graphQL query
+        // Entity Projection is retrieved from projectionInfo. This shall be removed.
         if (projectionInfo.getProjections().size() > 0) {
-            this.setEntityProjection(projectionInfo.getProjections().iterator().next());
+            this.setEntityProjection(projectionInfo.getProjections().entrySet().iterator().next().getValue());
         }
     }
 }
--- a/elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
+++ b/elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
@@ -44,7 +44,6 @@ import graphql.language.SourceLocation;
 import graphql.parser.Parser;
 import lombok.extern.slf4j.Slf4j;
 import java.math.BigInteger;
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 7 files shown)


elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLRequestScope.java
```diff
@@ -38,9 +38,9 @@ public GraphQLRequestScope(
         super("/", null, transaction, user, new MultivaluedHashMap<>(), elideSettings);
         this.projectionInfo = projectionInfo;
 
-        // TODO: handle multiple projection in one graphQL query
+        // Entity Projection is retrieved from projectionInfo. This shall be removed.
         if (projectionInfo.getProjections().size() > 0) {
-            this.setEntityProjection(projectionInfo.getProjections().iterator().next());
+            this.setEntityProjection(projectionInfo.getProjections().entrySet().iterator().next().getValue());
```
aklish: Rather than set this to something random.  I wonder if we should set this to null and mention this is unused in GraphQL.

elide-graphql/src/main/java/com/yahoo/elide/graphql/parser/GraphQLEntityProjectionMaker.java
```diff
@@ -165,55 +164,45 @@ public GraphQLProjectionInfo make(String query) {
...
-            return;
-        }
-
-        Class<?> entityType = entityDictionary.getEntityClass(entityName);
-        if (entityType == null) {
-            throw new InvalidEntityBodyException(String.format("Unknown entity {%s}.", entityName));
-        }
-
-        Field fields = (Field) rootSelection;
+        Map<String, Field> rootSelectionField = new HashMap<>();
```
aklish: Maybe rename to rootSelectionFields.

elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLEndpointTest.java
```diff
@@ -791,6 +791,89 @@ void testQueryAMapWithBadFields() throws IOException {
         assertHasErrors(response);
     }
 
+
+    @Test
+    public void testMultipleRoot() throws JSONException {
```
aklish: Add another test for two roots of the same type but with different aliases.

PR #1051, opened 2019-10-28
Support Non JPA Entity in AggregationDataStore

hellohanchen:
Update 'AggregationDataStore' to apply the changes in https://github.com/yahoo/elide/pull/1087 to support non-jpa entities.

## Description
In the use case of `AggregationDataStore`, there are entities without physical `id` field such as fact tables. Those entities are considered as **Elide Entity** which is different from **JPA Entity** who has an id field. The `AggregationDataStore` should support relationship to **Elide Entity** and expose fact table as root level.
- For a Non-JPA Elide entity that doesn't support aggregation, there should be `@Include` annotation on the class and can use `@Table` or `@Subselect` to indicate the physical table or select statement.
- For a Non-JPA Elide entity that supports aggregation, there should be `@Include` annotation on the class and can use `@FromTable` or `@FromSubquery` to indicate the physical table or select statement.

Main changes:
- Update models in `MetaDataStore` to handle tables without `id` field.
- Update `@JoinTo` annotation on a relationship field, users can provide customized join clause expression using this annotation. **There has to be a join clause expression if the relationship is to a non-JPA entity.**
- Add `UnitTest` as a framework class for all sql query engine unit tests.
- Add `AggregationTestBinder` and `AggregationResourceConfig` to use `NonEntityDictionary` in integration tests
- Test data for `PlayerStat` doesn't need `id` field in the csv file as it is now a non-JPA entity

## Motivation and Context
In the real aggregation data store, there are many cases that the dimension or fact table doesn't contain an id field, which would conflict with the JPA Entity concept.

## How Has This Been Tested?
Unit tests and integration tests for view are added.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -166,7 +168,7 @@ protected SQLQuery toSQL(Query query, SQLSchema schema) {
 
         builder.projectionClause(extractProjection(query));
 
-        Set<Path.PathElement> joinPredicates = new HashSet<>();
+        Set<Path.PathElement> joinPredicates = new LinkedHashSet<>();
```
hellohanchen: The joins need to happen following the order in the `Path`.
Change to `LinkedHashSet` to preserve the join order.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/annotation/JoinTo.java
```diff
@@ -24,5 +25,13 @@
...
-    String path();
+    String path() default "";
+
+    /**
+     * Join on clause constraints for customizing relationship joins.
+     * Use "%from" and "%join% to represent the two sides of join.
+     *
+     * @return join constraint like <code>%from.col1 = %join.col2</code>
+     */
+    JoinExpression[] constraints() default {};
```
aklish: The join clause could use OR instead of AND to join them.  Since we break out constraints rather than just the ON clause - we have to pick (likely AND) and ignore OR.  I think it would be better to just have a single ON clause expression (rather than breaking this up).  Another thing that is missing is parenthetical expressions.  The more structure we add to the ON clause - the more complicated the configuration.  We should just let them supply the ON clause.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/annotation/View.java
```diff
@@ -0,0 +1,38 @@
...
+     * Name defined for this view.
+     * @return view name
+     */
+    String name() default "";
+
+    /**
+     * Whether this view is from a table directly
+     * @return True if the view is from a table
+     */
+    boolean isTable() default true;
```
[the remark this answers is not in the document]
hellohanchen: `from()` should always be provided, it can either be a table name or a select statement.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/core/ViewDictionary.java
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Extension of {@link EntityDictionary} that manage view models.
+ */
+@Slf4j
+public class ViewDictionary extends EntityDictionary {
+    private final ConcurrentHashMap<Class<?>, ViewBinding> viewBindings = new ConcurrentHashMap<>();
+    private final static ViewBinding EMPTY_BINDING = new ViewBinding();
+
+    public ViewDictionary(Map<String, Class<? extends Check>> checks) {
```
aklish: Maybe the only constructor for ViewDictionary should just take another dictionary that it copies.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -34,7 +37,8 @@
     /**
      * These are the classes the Aggregation Store manages.
      */
-    private static final Class[] AGGREGATION_STORE_CLASSES = { FromTable.class, FromSubquery.class };
+    private static final Class[] AGGREGATION_STORE_CLASSES = {
+            Subselect.class, Table.class, FromTable.class, FromSubquery.class };
```
aklish: I don't think we want AggregationStore to manage Entities at all.  In that case - we should remove Table and Subselect.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/CountryViewNested.java
```diff
@@ -0,0 +1,54 @@
...
+import javax.persistence.Id;
+import javax.persistence.Table;
+
+/**
+ * A nested view for testing.
+ */
+@Data
+@Include
+@Table(name = "countries")
+public class CountryViewNested {
```
aklish: I would prefer if we created a model that reflects something meaningful.  It will be easier to understand the tests.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -242,12 +242,13 @@ private GraphQLObjectType buildQueryObject(Class<?> entityClass) {
                 .name("_node__" + entityName);
 
         String id = dictionary.getIdFieldName(entityClass);
-
-        /* our id types are DeferredId objects (not Scalars.GraphQLID) */
-        builder.field(newFieldDefinition()
-                .name(id)
-                .dataFetcher(dataFetcher)
-                .type(GraphQLScalars.GRAPHQL_DEFERRED_ID));
+        if (id != null) {
```
aklish: Our APIs require an ID.  Not sure this is the best approach.  Let's discuss.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
hellohanchen: Currently the **view**s are using `@Table` and `@Subselect` to distinguish themselves from **analyticView**s. Those models need to be managed by AggregationStore.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
hellohanchen: This change is made before we decide that we keep using row number as the id for analyticViews.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/CountryViewNested.java
hellohanchen: When I work on Sashruts' PR I can change this view to ContinentView

aklish: :+1:

PR #1055, opened 2019-10-31
Support multiple query of same entity with different alias

### LINKED ISSUE #994: Elide 5.0: Running Multiple Queries in same transaction is broke in GraphQL

Chandrasekar-Rajasekar:
Resolves #994 (if appropriate)

## Description
We don't want to merge two root level queries on the same entity. We support such queries if they have different alias. Else we reject it. 

Extension of https://github.com/yahoo/elide/pull/1044

## License


elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLEndpointTest.java
```diff
@@ -800,12 +786,7 @@ public void testMultipleRoot() throws JSONException {
                                 "author",
                                 selections(
                                         field("id"),
-                                        field("name")
```
aklish: This is no longer testing multiple roots.  Let's rename the method or remove the test.

elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLEndpointTest.java
```diff
@@ -337,22 +337,8 @@ void testPartialResponse() throws IOException, JSONException {
                 )
         ).toQuery();
 
-        String expectedData = document(
```
aklish: What about testing the sort parameter for two aliased root level collections?

elide-graphql/src/main/java/com/yahoo/elide/graphql/containers/RootContainer.java
```diff
@@ -14,12 +14,16 @@
 public class RootContainer implements GraphQLContainer {
     @Override
     public Object processFetch(Environment context, PersistentResourceFetcher fetcher) {
+        String entityName = context.field.getName();
+        String aliasName = context.field.getAlias();
+        String keyName = (aliasName == null ? "" : aliasName) + ":" + entityName;
```
aklish: Let's put this into a static function so the logic won't be duplicated.

elide-graphql/src/test/java/com/yahoo/elide/graphql/FetcherFetchTest.java
```diff
@@ -200,11 +200,6 @@ public void testAliasSameRelationship() throws Exception {
         runComparisonTest("aliasSameRelationship");
     }
 
-    @Test
```
aklish: Why did we remove this test?

Chandrasekar-Rajasekar: This test checks if there is two root level query of same entity, (without alias) then the response should merge these two queries. We decided to not support these conditions. 
Request- 
`
{
  book(ids: ["1"]) {
    edges {
      node {
        alias1: id
      }
    }
  }
  book(ids: ["1"]) {
    edges {
      node {
        alias2: id
      }
    }
  }
}
`

Response-
`
{
  "book": {
    "edges": [
      {
        "node": {
          "alias1": "1",
          "alias2": "1"
        }
      }
    ]
  }
}
`

elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLEndpointTest.java
Chandrasekar-Rajasekar: This one still tests two root level query, one author and one book without alias.

Issue #1024, opened 2019-10-18
Elide 5.0 - SQLQueryEngine: Sorting by id should be prohibited


hellohanchen:
When SQLQueryEngine generates an aggregation query, it would not fetch the **id** field because that's meaningless and the ids are generated in-memory after aggregated results return. So we don't actually query **id** in the aggregation query and that prevent user from sorting on **id** field.

## Expected Behavior
Check sorting clause on id and reject or ignore it.

## Current Behavior
No checks for this situation, the query would fail as **id** is not defined in the query.

## Solution
Fix the issue and add test cases

chczy:
Merged.

PR #1067, opened 2019-11-08
Added sorting on aggregated metrics revision 2

### LINKED ISSUE #1023: Elide 5.0 - SQLQueryEngine: Fix Sorting on AggregationMetric

SushrutV:
Resolves #1023, #1024

## Description
Fixes sorting on AggregatedMetrics and disables sorting on id field.
Summary of different sorting queries and how they are currently handled here: https://docs.google.com/document/d/1-2ZEk6emqP838DG3HOS6me1bW0JLNfYwTzzO0S1UHp8/edit?usp=sharing

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -261,4 +268,43 @@ private void validateHavingClause(FilterExpression havingClause) {
...
+        Map<Path, Sorting.SortOrder> sortClauses = sorting.getValidSortingRules(entityProjection.getType(), dictionary);
+        sortClauses.keySet().forEach((path) -> validatePath(path, allFields));
+    }
+
+    /**
+     * Verifies that the current path can be sorted on
+     * @param path The path that we are validating
+     * @param allFields Set of all field names included in initial query
+     */
+    private void validatePath(Path path, Set<String> allFields) {
```
hellohanchen: Let this be `validateSortingPath`

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -261,4 +268,43 @@ private void validateHavingClause(FilterExpression havingClause) {
...
+        //TODO add support for double nested sorting
+        if (pathElemenets.size() > 2) {
+            throw new UnsupportedOperationException(
+                    "Currently sorting on double nested fields is not supported");
+        }
+        Path.PathElement currentElement = pathElemenets.get(0);
+        String currentField = currentElement.getFieldName();
+        Class<?> currentClass = currentElement.getType();
+        if (!allFields.stream().anyMatch(field -> field.equals(currentField))) {
+            throw new InvalidOperationException("Can't sort on field that is not present in query");
```
hellohanchen: Let's print out name of the invalid field.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -261,4 +268,43 @@ private void validateHavingClause(FilterExpression havingClause) {
...
+        }
+        if (dictionary.getIdFieldName(currentClass).equals(currentField) || currentField.equals("id")) {
```
hellohanchen: I think there is a `REGULAR_ID_NAME` or something like this that represents the reserved `id` string. Let's replace `id` with that.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreHelper.java
```diff
@@ -261,4 +268,44 @@ private void validateHavingClause(FilterExpression havingClause) {
...
+     * Method to verify that all the sorting options provided
+     * by the user are valid and supported.
+     */
+    public void validateSorting() {
+        Sorting sorting = entityProjection.getSorting();
+        if (sorting == null) {
+            return;
+        }
+        Set<String> allFields = getRelationships();
+        allFields.addAll(getAttributes());
```
hellohanchen: This is the second time we have this logic (check line 166), I think we can make it into a method like `getAllFields()`.

PR #1074, opened 2019-11-14
Expose opaque user in audits

patelh:

## Description
Expose opaque user through audits so it can be logged through log statements.

## Motivation and Context
Right now there is no way to get to opaque user information through audit logging interface.  It might be fine if we inject user information through slf4j mdc but if we want to implement our own logger and want to use the log statements through elide to generate the final messages, we need to expose the opaque user.

## How Has This Been Tested?
Test was added.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/audit/LogMessage.java
+++ b/elide-core/src/main/java/com/yahoo/elide/audit/LogMessage.java
@@ -122,17 +122,14 @@ public class LogMessage {
                 singleElementContext.setVariable(name, singleElementExpression);
             }
 
-            final User user = getUser();
+            final Object user = getUser();
             if (user != null) {
-                final Object opaqueUser = user.getOpaqueUser();
-                if (opaqueUser != null) {
-                    final ValueExpression opaqueUserValueExpression = EXPRESSION_FACTORY
-                        .createValueExpression(
-                            opaqueUser, Object.class
-                        );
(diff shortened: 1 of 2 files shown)


elide-core/src/main/java/com/yahoo/elide/audit/LogMessage.java
```diff
@@ -120,6 +121,19 @@ public String getMessage() {
                 ctx.setVariable(name, expression);
                 singleElementContext.setVariable(name, singleElementExpression);
             }
+
+            final User user = record.getRequestScope().getUser();
```
aklish: Changes look good.  Can we move this snippet to a helper function?  This function is already pretty large with high cyclomatic complexity.

patelh: Done.

elide-core/src/main/java/com/yahoo/elide/audit/LogMessage.java
```diff
@@ -120,6 +121,19 @@ public String getMessage() {
                 ctx.setVariable(name, expression);
                 singleElementContext.setVariable(name, singleElementExpression);
             }
+
+            final User user = getUser();
+            if (user != null) {
+                final Object opaqueUser = user.getOpaqueUser();
```
aklish: Let's move the this line and the null check into your getUser function.  Then I think we are good to go.

PR #1075, opened 2019-11-15
Expose audit logger in standalone

patelh:

## Description
Change the elide standalone settings interface to be able to provide an audit logger.

## Motivation and Context
Right now elide standalone has no way of configuring the audit logger, this exposes it as part of the interface so it can be customized.

## How Has This Been Tested?
Updated existing test.

## License


patelh:
@aklish I have no way of seeing what the error is for Better Code Hub, it says I have no permissions... Any way I can find out what it's complaining about?

aklish:
Don't worry about better code hub.  The linters are not that accurate at finding real issues.

PR #1085, opened 2019-11-22
Elide sorting on aggregated metrics

### LINKED ISSUE #1023: Elide 5.0 - SQLQueryEngine: Fix Sorting on AggregationMetric

SushrutV:
Resolves #1023, #1024

## Description
Fixes sorting on AggregatedMetrics and disables sorting on id field.
Summary of different sorting queries and how they are currently handled here: https://docs.google.com/document/d/1-2ZEk6emqP838DG3HOS6me1bW0JLNfYwTzzO0S1UHp8/edit?usp=sharing

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryValidator.java
```diff
@@ -0,0 +1,156 @@
...
+            throw new UnsupportedOperationException(
+                    "Currently sorting on double nested fields is not supported");
+        }
+        Path.PathElement currentElement = pathElemenets.get(0);
+        String currentField = currentElement.getFieldName();
+        Class<?> currentClass = currentElement.getType();
+        if (!allFields.stream().anyMatch(field -> field.equals(currentField))) {
+            throw new InvalidOperationException("Can't sort on " + currentField + " as it is not present in query");
+        }
+        if (dictionary.getIdFieldName(currentClass).equals(currentField)
```
hellohanchen: flip this expression `dictionary.getIdFieldName(currentClass).equals(currentField)` to avoid the case that `getIdFieldName()` returns `null`

chczy:
This PR is updated in another branch and merged.

PR #1087, opened 2019-11-25
Add auto configuration for aggregation store

aklish:
## Description
Fixes:
1. AggregationDataStore only populates the entityDictionary with Elide models that are `@FromTable` or `@FromSubquery`.
2. Spring Boot Autoconfigure supports Aggregation Data Store and Query Engine.
3. Entity Dictionary supports non `@Entity` classes.
4. Consolidated logic which does class scans.

## Motivation and Context
This adds the ability to configure the AggregationDataStore and its query engine via spring boot.

## How Has This Been Tested?
Added Spring Boot Autoconfigure Test.  Verified all other tests still pass.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/MetaDataStore.java
```diff
@@ -32,28 +39,52 @@
...
+    @Getter
+    private final String scanPackageName;
+
+    public MetaDataStore(Package scanPackage) {
+        this (scanPackage.getName());
+    }
+
+    public MetaDataStore(String scanPackageName) {
         super(META_DATA_PACKAGE);
+        this.scanPackageName = scanPackageName;
```
hellohanchen: How does this handle the case that one metadatastore needs to manage more than one package?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -35,8 +49,13 @@ public AggregationDataStore(QueryEngineFactory queryEngineFactory, MetaDataStore
...
+        for (Class<? extends Annotation> cls : AGGREGATION_STORE_CLASSES) {
+            ClassScanner.getAnnotatedClasses(modelPackageName, cls).stream().forEach(modelClass -> {
+                dictionary.bindEntity(modelClass);
+            });
+        }
+
+        queryEngine = queryEngineFactory.buildQueryEngine(metaDataStore);
 
         /* Add 'grain' argument to each TimeDimensionColumn */
         for (AnalyticView table : metaDataStore.getMetaData(AnalyticView.class)) {
```
hellohanchen: We need to ensure that the metadatastore is populated before this datastore. In current logic, the `metadatastore.loadMetadata()` can happen after aggregationDataStore populating the dictionary.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTestHarness.java
```diff
@@ -6,42 +6,33 @@
...
+        AggregationDataStore aggregationDataStore = new AggregationDataStore(queryEngineFactory, metaDataStore);
+
+        DataStore jpaStore = new JpaDataStore(
+                () -> { return queryEngineFactory.getEmf().createEntityManager(); },
+                (entityManager) -> { return new NonJtaTransaction(entityManager); }
+        );
 
         // meta data store needs to be put at first to populate meta data models
-        return new MultiplexManager(metaDataStore, aggregationDataStore);
+        return new MultiplexManager(jpaStore, metaDataStore, aggregationDataStore);
```
hellohanchen: So the JPA models are not managed by metaDataStore?

aklish: Correct.  JPA models are managed by JPA Store.  MetaDataStore only manages a small subset of entities (Table, Column, etc).

hellohanchen: I mean the data models in JPA store. Would those models be added into the metadatastore? For example, there is a JPA model call `People`, would be schema of `People` be stored in metadatastore?

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1037,18 +1061,71 @@ public String getId(Object value) {
...
+    }
+
+    /**
+     * Follow for this class or super-class for Include annotation.
+     *
+     * @param objClass provided class
+     * @return class with Include annotation or
+     */
+    public Class<?> lookupIncludeClass(Class<?> objClass) {
+        Annotation first = getFirstAnnotation(objClass, Arrays.asList(Exclude.class, Include.class));
```
hellohanchen: I think we should update `getFirstAnnotation()` as well, for this scenario 
```java
@Exclude 
class Foo { }

@Include
class Bar extends Foo {}

class Baz extends Bar {}
```
I think `getFirstAnnotation(Baz.class, Exclude.class, Include.class)` would return `Exclude.class`?

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1415,9 +1490,11 @@ private boolean isValidParameterizedMap(Map<?, ?> values, Class<?> keyType, Clas
      * @param entityClass the class to bind.
      */
     private void bindIfUnbound(Class<?> entityClass) {
-        if (! entityBindings.containsKey(lookupEntityClass(entityClass))) {
-            bindEntity(entityClass);
+        if (lookupBoundClass(entityClass) != null) {
+            return;
         }
+
+        bindEntity(entityClass);
```
hellohanchen:
```Java
if (lookupBoundClass(entityClass) == null) {
  bindEntity(entityClass);
}
```
Might be simpler or more straight forward?

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
aklish: Added a test.  It does the right thing.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
aklish: :+1:

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTestHarness.java
aklish: MetadataStore has two dictionaries - the one it exposes to Elide core (which just has Table, Column, etc), and other dictionary which contains the entities it adds to the metadata.  That is populated by scanning the class path.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
aklish: Fixed

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/MetaDataStore.java
aklish: I'm going to punt on this one for now.  I had a similar thought, but likely a single package is sufficient for nearly every case.

elide-core/src/test/java/com/yahoo/elide/core/EntityDictionaryTest.java
```diff
@@ -486,14 +487,156 @@ public void testChildBinding() {
...
+        @Include
+        class Bar extends Foo {
+
+        }
+
+        class Baz extends Bar {
+
+        }
+
+        Annotation first = getFirstAnnotation(Baz.class, Arrays.asList(Exclude.class, Include.class));
```
hellohanchen: This case should not happen, but do we need to check the case that `@Include` and `@Exclude` are declared on the same class together?

elide-core/src/test/java/com/yahoo/elide/core/PersistenceResourceTestSetup.java
```diff
@@ -63,6 +71,15 @@ void init() {
         dictionary.bindEntity(ContainerWithPackageShare.class);
         dictionary.bindEntity(ShareableWithPackageShare.class);
         dictionary.bindEntity(UnshareableWithEntityUnshare.class);
+        return dictionary;
+    }
+    protected static ElideSettings initSettings() {
```
hellohanchen: We can add a new line here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/MetaDataStore.java
```diff
@@ -32,28 +39,43 @@
     public static final Package META_DATA_PACKAGE =
             Package.getPackage("com.yahoo.elide.datastores.aggregation.metadata.models");
 
-    public MetaDataStore() {
-        super(META_DATA_PACKAGE);
+    private static final Class[] METADATA_STORE_CLASSES = {
+            FromTable.class, FromSubquery.class, Subselect.class, javax.persistence.Table.class};
```
hellohanchen: Can we change `METADATA_STORE_CLASSES` to  `METADATA_STORE_ANNOTATIONS`?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/MetaDataStore.java
```diff
@@ -32,28 +39,43 @@
...
-
-        classes.stream()
-                .filter(cls -> !META_DATA_PACKAGE.equals(cls.getPackage()))
-                .forEach(cls -> addTable(
-                        isAnalyticView(cls)
-                                ? new AnalyticView(cls, dictionary)
-                                : new Table(cls, dictionary)));
+    @Override
+    public void populateEntityDictionary(EntityDictionary dictionary) {
+        ClassScanner.getAllClasses(Table.class.getPackage().getName()).stream().forEach(cls -> {
```
hellohanchen: Can we use `ClassScanner.getAllClasses(META_DATA_PACKAGE.getName())`?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -55,9 +55,9 @@
     @Getter
     private Map<Class<?>, Table> tables;
 
-    public SQLQueryEngine(EntityManagerFactory emf, EntityDictionary dictionary, MetaDataStore metaDataStore) {
+    public SQLQueryEngine(EntityManagerFactory emf, MetaDataStore metaDataStore) {
         this.emf = emf;
-        this.dictionary = dictionary;
+        this.dictionary = metaDataStore.getDictionary();
```
hellohanchen: Can we change `dictionary` to `metaDataDictionary` as this dictionary is not the exposed Elide core dictionary?

PR #1088, opened 2019-11-26
Consolidate class scanning logic

aklish:
## Description
1. Consolidates class scanning logic across Elide.
2. Removes unnecessary elide.modelPackages setting in Spring Boot.
3. Allows a non-entity to be registered in the EntityDictionary.

## Motivation and Context
Replicates some changes happening in Elide 5.0.

## How Has This Been Tested?
Verified existing tests pass.

## License


elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1449,7 +1461,7 @@ private boolean isValidParameterizedMap(Map<?, ?> values, Class<?> keyType, Clas
      * @param entityClass the class to bind.
      */
     private void bindIfUnbound(Class<?> entityClass) {
-        if (lookupBoundClass(entityClass) == null) {
+        if (! isClassBound(entityClass)) {
```
chczy: extra space after !

PR #1106, opened 2019-12-10
Migrate legacy getting started to elide standalone readme

aklish:
## Description
Main Elide _Getting Started_ docs are going to change to describe spring boot setup.  This PR migrates the current getting starting docs to this README since they reference Elide standalone.

## License

### DIFF
--- a/elide-standalone/README.md
+++ b/elide-standalone/README.md
@@ -1,3 +1,11 @@
+Table of Contents
+=================
+
+  * [Overview](#overview)
+  * [Who is this for?](#whofor)
+  * [Getting Started](#gettingstarted)
+  * [More Detailed Examples](#moredetail)
+
 ## <a name="overview"></a>Overview
 
 The Elide standalone application is a configurable web server using Elide. While Elide is typically a pluggable **middleware** framework, we have constructed a flexible and complete service to allow you to get started quickly.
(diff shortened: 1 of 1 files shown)


elide-standalone/README.md
```diff
@@ -36,23 +37,298 @@ To include `elide-standalone` into your project, add the single dependency:
...
+      @ManyToOne
+      public ArtifactProduct artifact;
+  }
+  ```
+
+### Spin up the API
+
+So now we have some models, but without an API it is not very useful. Before we add the API component, we need to create the schema in the database that our models will use.   Our example uses liquibase to manage the schema.  When Heroku releases the application, our example will execute the [database migrations][https://github.com/aklish/elide-heroku-example/blob/master/src/main/resources/db/changelog/changelog.xml] to configure the database with some test data automatically.  This demo uses Postgres.  Feel free to modify the migration script if you are using a different database provider.
+
+There may be more tables in your database than models in your project.  Similarly, there may be more columns in a table than in a particular model.  Not only will our models work just fine, but we expect that models will normally expose only a subset of the fields present in the database. Elide is an ideal tool for building micro-services - each service in your system can expose only the slice of the database that it requires.
```
hellohanchen: We can actually also have more models than tables I think.

elide-standalone/README.md
```diff
@@ -36,23 +37,298 @@ To include `elide-standalone` into your project, add the single dependency:
...
+      @Override
+      public Properties getDatabaseProperties() {
+          Properties options = new Properties();
+          
+          //Here we use H2 in memory instead of Postgres 
+          options.put("hibernate.dialect", "org.hibernate.dialect.H2Dialect");
+          options.put("javax.persistence.jdbc.driver", "org.h2.Driver");
+          options.put("javax.persistence.jdbc.url", "jdbc:h2:mem:db1;DB_CLOSE_DELAY=-1");
+          options.put("javax.persistence.jdbc.user", "sa");
+          options.put("javax.persistence.jdbc.password", "");
```
hellohanchen: Should we explain these configurations too? Or put a link to all references.

elide-standalone/README.md
```diff
@@ -36,23 +37,298 @@ To include `elide-standalone` into your project, add the single dependency:
...
+  ```curl
+  curl http://localhost:8080/api/v1/group
+  ```
+
+#### GraphQL
+  ```curl
+  curl -g -X POST -H"Content-Type: application/json" -H"Accept: application/json" \
+      "http://localhost:8080/graphql/api/v1" \
+      -d'{   
+             "query" : "{ group { edges { node { name commonName description } } } }"
```
hellohanchen: The graphql request example in elide.io only contains this part `{ group { edges { node { name commonName description } } } }`, we should also update that one too. And I think we can also explain what this query means.

elide-standalone/README.md
aklish: 👍

elide-standalone/README.md
aklish: 👍

PR #1131, opened 2020-01-10
Adds a registry based mechanism for registering Custom GraphQL scalars 

### LINKED ISSUE #1128: Question: How to handle OffsetDateTime with Elide Spring Boot
I have a property in my entity with `OffsetDateTime` type, but Eldie seems to throw stackOverfow error whenever i try to start my Spring Boot Application. I assume there is no in build support for the same in Eldie  

Please suggest a way to register new GraphQL scalar type. I know about Serde from documentation, but it would be great if you could provide a way to register new ones with Eldie.

1. Add `OffsetDateTime` in your Entity

## Your Environment
* Elide version used: 4.5.9
* Environment name and version (Java 1.8.0_152): 1.8
* Operating System and version: Windows 10

murtuza-ranapur:
Resolves #1128 

## Description
1. Adds Support for java.time.OffsetDateTime
2. Adds new annotation, using which users can add support for new GraphQL Scalar types

## Motivation and Context
Currently there is no mechanism in place using which user can register new GraphQL scalars with Elide Engine. This PR suggests a Annotation and Auto Discovery based approach to solve this issue. To auto discover use of ClassGraph Library has been made.  

## How Has This Been Tested?
Junit test cases for the same has been written

## License


aklish:
Thanks for the PR.  I'll take a look at the travis build.  

I like the approach overall.  I have a few suggestions though:

1. It would be nice if the auto-scan would work for both JSON-API and GraphQL.  To that end, I'm wondering if the scan could discover and register Serdes instead of Coercing classes.  GraphConversionUtils could then extract the Serdes from CoerceUtil (CoerceUtil would need a new function to return the list of Serdes).
2. Elide uses ClassGraph for scans, but it hides the details behind a core class called ClassScanner (we have changed the class scanner library in the past).  
3. I think it might be safe to generalize a Coercing class tied to a Serde.   We can use that Coercing class for all types that are registered in CoerceUtil.  Something like:

```java
@Data
public class SerdeCoercing<I, O> implements Coercing<I, O> {
    private String errorMessage;
    Serde<O, I> serde;

    @Override
    public O serialize(Object dataFetcherResult) {
        return serde.serialize((I) dataFetcherResult);
    }

    @Override
    public I parseValue(Object input) {
        return serde.deserialize((O) input);
    }

    public I parseLiteral(Object o) {
        Object input;
        if (o instanceof IntValue) {
            input = ((IntValue) o).getValue().longValue();
        } else if (o instanceof StringValue) {
            input = ((StringValue) o).getValue();
        } else if (o instanceof FloatValue) {
            input = ((FloatValue) o).getValue().floatValue();
        } else {
            throw new CoercingParseValueException(errorMessage);
        }
        return parseValue(input);
    }
}
```

murtuza-ranapur:
1. Sure will try to make auto scan work for Serde, will think in that direction
2. I had noticed `ClassScanner`, will maybe add a function which returns `ClassInfo` directly there only 
3. Just want to add one more point here. So now with`SerdeCoercing` there will be 1-1 mapping between a Serde and Coercing which is also the most common case. But let's say I want to use 1 Serde for a number of other types too, mostly child types, then should the annotation have an attribute that allows users to specify the child/extending types. Let me know your thoughts on this one.

aklish:
Comments inline:

`I had noticed ClassScanner, will maybe add a function which returns ClassInfo directly there only`

Since ClassInfo is part of class graph, it would break encapsulation to leak that.  Maybe just return the list of Serde classes with the annotation - and then use standard reflection libraries to get your annotation and metadata from each class returned.

`Just want to add one more point here. So now withSerdeCoercing there will be 1-1 mapping between a Serde and Coercing which is also the most common case. But let's say I want to use 1 Serde for a number of other types too, mostly child types, then should the annotation have an attribute that allows users to specify the child/extending types. Let me know your thoughts on this one`

What if instead of coercing for a single class, the new annotation can have an array of classes that the Serde coerces for.   We could change the annotation to take an array of classes to coerce for.  Thoughts?

murtuza-ranapur:
Yes, that's what I was getting at with annotation having an extra attribute and child/sub types. So with this, annotation will look like
```
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.TYPE)
public @interface ElideTypeConvertor {
    Class<?> type();
    String name();
    String description() default "Custom Elide type";
    Class<?> [] subtype() default {};
}
```

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -77,6 +92,69 @@ public Elide(ElideSettings elideSettings) {
         elideSettings.getSerdes().forEach((targetType, serde) -> {
             CoerceUtil.register(targetType, serde);
         });
+
+        registerCustomSerde();
+        registerCustomSerdeInObjectMapper();
+    }
+
+    private void registerCustomSerde() {
```
aklish: I refactored this a bit by:
  1. Using a different member of ClassScanner (see other comment in this PR).
  2. Linking the methods together to reduce some duplication between them.

```java
    private void registerCustomSerde() {
        Set<Class<?>> classes = ClassScanner.getAnnotatedClasses(ElideTypeConverter.class);

        for (Class<?> clazz : classes) {
            if (!Serde.class.isAssignableFrom(clazz)) {
                log.warn("Skipping Serde registration (not a Serde!): {}", clazz);
                continue;
            }
            Serde serde;
            try {
                serde = (Serde) clazz.newInstance();
            } catch (InstantiationException | IllegalAccessException e) {
                String errorMsg = String.format("Error while registering custom Serde: %s", e.getLocalizedMessage());
                log.error(errorMsg);
                throw new UnableToAddSerdeException(errorMsg);
            }
            ElideTypeConverter converter = clazz.getAnnotation(ElideTypeConverter.class);
            Class baseType = converter.type();
            List<Class<?>> allTypes = Arrays.asList(converter.subTypes());
            allTypes.add(baseType);

            for (Class type : allTypes) {
                if (!baseType.isAssignableFrom(type)) {
                    throw new IllegalArgumentException("Mention type " + type
                            + " not subtype of " + baseType);
                }
                log.info("Registering serde for type : {}", type);
                CoerceUtil.register(type, serde);
                registerCustomSerdeInObjectMapper(type, serde, converter.name());
            }
        }
    }
    private void registerCustomSerdeInObjectMapper(Class<?> type, Serde serde, String name) {
        ObjectMapper objectMapper = mapper.getObjectMapper();
        objectMapper.registerModule(new SimpleModule(name)
            .addSerializer(type, new JsonSerializer<Object>() {
                @Override
                public void serialize(Object obj, JsonGenerator jsonGenerator, SerializerProvider serializerProvider)
                        throws IOException, JsonProcessingException {
                    jsonGenerator.writeObject(serde.serialize(obj));
                }
            }));
    }
```

elide-core/src/main/java/com/yahoo/elide/utils/ClassScanner.java
```diff
@@ -57,6 +58,18 @@
         }
     }
 
+    /**
+     * Scans all classes accessible from the context class loader which belong to the current class loader.
+     *
+     * @param annotation  Annotation to search
+     * @return The classeInfos
+     */
+    static public List<ClassInfo> getAnnotatedClassInfos(Class<? extends Annotation> annotation) {
```
aklish: This method leaks ClassInfo.  We don't want ClassGraph concepts to leak into core Elide.  I would just use getAnnotatedClasses instead.

elide-core/src/main/java/com/yahoo/elide/utils/coerce/converters/ElideTypeConvertor.java
```diff
@@ -0,0 +1,20 @@
...
+package com.yahoo.elide.utils.coerce.converters;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface ElideTypeConvertor {
```
aklish: I think the correct spelling is converter

elide-core/src/main/java/com/yahoo/elide/utils/coerce/CoerceUtil.java
```diff
@@ -78,6 +78,10 @@
         return (Serde<S, T>) SERDES.getOrDefault(targetType, null);
     }
 
+    public static Map<Class<?>, Serde<?, ?>> getSerdes() {
+        return SERDES;
```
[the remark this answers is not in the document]
murtuza-ranapur: Sure

elide-core/src/main/java/com/yahoo/elide/utils/ClassScanner.java
murtuza-ranapur: It slipped through. Its not used anywhere. Will simply remove it.

elide-core/src/main/java/com/yahoo/elide/Elide.java
murtuza-ranapur: Replaced my code with yours, made some minor changes here and there. Thanks! for the lesson on short circuit though :)

elide-core/src/main/java/com/yahoo/elide/utils/coerce/converters/ElideTypeConvertor.java
murtuza-ranapur: Noted, will correct the typo.

murtuza-ranapur:
Applied all the requested changes. I think with this user will be able to register new serde. But i found one problem when i wrote my own OffsetDateTime serde in the project that uses Elide, i was hopping it would replace the in-build implementation, but it didn't. My custom serde was added to registry before in-build implementation hence the reverse process happened. 

I have used ClassGraph in past but i am not really sure in what order does it perform class scanning. If there is something already in-build in ClassGraph to tackle it let me know. Or else we may need some attribute to provide Precedence in the annotation(The way Spring does), in-Build implementations may have lowest priority. Let me know your thoughts.

aklish:
> Applied all the requested changes. I think with this user will be able to register new serde. But i found one problem when i wrote my own OffsetDateTime serde in the project that uses Elide, i was hopping it would replace the in-build implementation, but it didn't. My custom serde was added to registry before in-build implementation hence the reverse process happened.
> 
> I have used ClassGraph in past but i am not really sure in what order does it perform class scanning. If there is something already in-build in ClassGraph to tackle it let me know. Or else we may need some attribute to provide Precedence in the annotation(The way Spring does), in-Build implementations may have lowest priority. Let me know your thoughts.

I think the solution to this is for built-in types - we don't use the annotation, but rather just register the Serde directly.  The scanning can happen after that registration and replace builtin types with user defined types.

murtuza-ranapur:
Fair enough. Really liked the project though! Hope to contribute more.

aklish:
Thanks for the contribution!  If you'd like to help with tickets, you can check out the project kanban and look for unassigned tickets in the "To Do" column:

https://github.com/yahoo/elide/projects/3

We're always looking to get new committers.

Issue #1043, opened 2019-10-22
Elide 5.0 - SQLQueryEngine should support SQL expressions for each metric & dimension


aklish:
## Expected Behavior

We need a new annotation that takes a given column and applies an arbitrary SQL expression over it to extract the value.  This annotation should apply for both metrics and dimension columns.  For dimensions it should apply for both degenerate dimensions and `JoinTo` dimensions.

```java
@SQLExpression(expression = "blahblah(%s)")
```

## Current Behavior
Current behavior is limited to using the column name directly in the SQL.  For actual data marts, the columns may require extraction from a denormalized data structure, or have to be cast to the correct type.

PR #1133, opened 2020-01-10
@MetricFormula & @DimensionFormula

hellohanchen:
Solved https://github.com/yahoo/elide/issues/1043.
Add `@MetricFormula` and `@DimensionFormula` annotations to support arbitrary post-aggregation and pre-aggregation sql expression.

## Description
- Add `@MetricFormula`
  - `@MetricFormula` would be used on metric fields to perform post-aggregation expression, it can only be used on metric fields and can only refer to metric fields.
  - Update `Metric` class to resolve `@MetricFormula`.
- Add `@DimensionFormula`.
  - `@DimensionFormula` would be used on dimension fields to perform pre-aggregation expression, it can only be used on dimension fields.
  - It can use `@JoinTo` to refer to a field defined in other class.
  - Update `SQLQueryEngine.generateColumnReference()` method to resolve. `@DimensionFormula`. This method would use `Path` to navigate through data models.
  - Update `SQLColumn` to use new method.
- These two annotations are similar but they are used in different stages of query construction.

## How Has This Been Tested?
2 new unit tests classes are added.
Totally 8 unit tests and 2 integration tests are added.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/annotation/SQLExpression.java
```diff
@@ -0,0 +1,30 @@
...
+ * Sql expression can only be applied on degenerated dimension field or physical column reference.
+ * Using this annotation on metric field is invalid as metric function expression already has the same functionality.
+ * Using this annotation on relationship field is invalid as relationship field is not physical table column.
+ * Sql expression can be carried when used with @JoinTo, the outer layer expression would be applied after the initial
+ * joined to sql expression.
+ */
+@Documented
+@Target({ElementType.FIELD, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+public @interface SQLExpression {
```
aklish: I think we should:
1. Unify this concept with MetricComputation.  They are closely related.
2. Instead of arbitrary names like 'reference', we should use model field names like %overallRating and %highScore.   This will allow metric definitions as well.

We still need to differentiate whether something is a metric or dimension though.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -254,19 +253,35 @@ private SQLQuery toPageTotalSQL(SQLQuery sql) {
      *
      * @param path The predicate path to convert
      * @param dictionary dictionary to expand joinTo path
+     * @param expressions outer layer sql expression to be applied on this field
      * @return A SQL fragment that references a database column
      */
-    public static String generateColumnReference(Path path, EntityDictionary dictionary) {
+    public static String generateColumnReference(Path path, EntityDictionary dictionary, List<String> expressions) {
```
aklish: Instead of List<String>, I think we want more of a symbol table concept here with concrete names that map to the model.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -279,4 +294,24 @@ public static String generateColumnReference(Path path, EntityDictionary diction
...
+    /**
+     * Generate alias for represent the join path.
+     * The path would start with the class alias of the first element, and then each field would append "_fieldName" to
+     * the result.
+     * The last element would not be included as that's not a part of the join path.
+     *
+     * @param path path that represents a relationship chain
+     * @return join path alias, i.e. <code>foo.bar.baz</code> would be <code>foo_bar</code>
+     */
+    public static String getJoinPathAlias(Path path) {
```
aklish: Can this be unified with FilterPredicate logic?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryConstructor.java
```diff
@@ -320,22 +324,25 @@ private String extractOrderBy(Map<Path, Sorting.SortOrder> sortClauses, SQLQuery
...
                     MetricFunctionInvocation metric = template.getMetrics().stream()
                             // TODO: filter predicate should support alias
                             .filter(invocation -> invocation.getAlias().equals(last.getFieldName()))
                             .findFirst()
                             .orElse(null);
 
-                    String orderByClause = metric == null
-                            ? generateColumnReference(expandedPath, dictionary)
-                            : metric.getFunctionExpression();
+                    // if the sorted field is a metric field on this queried table, it should be present in the request
```
aklish: What is metric != null but the last type doesn't match the queried table class?

hellohanchen: The metric is searched using alias. There can be a dimension class field that have the same as the projected metric.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
hellohanchen: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/annotation/SQLExpression.java
hellohanchen: My initial thought is that MetricComputation is resolved as MetricFunction, SQLExpression is for dimension fields. We would not use dimension fields for metric computation, right?

hellohanchen: We can support `@SQLExpression` on metric field, like
```java
@SQLExpression("%reference / 1000")
@MetricAggregation(function = SqlMax.class)
@Column(name = "col1")
private float highScore;
```
Would be resolved as `SELECT MAX(col / 1000) AS highScore`.
I idea is to keep `SQLExpression` separated from metrics, as metrics would be loaded into metadata store to provide information like arguments, functions and so on.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -269,22 +290,130 @@ private SQLQuery toPageTotalSQL(SQLQuery sql) {
...
+     * select statement, group by clause, where clause, having clause or order by clause).
+     *
+     * @param path the full join path to be resolved.
+     * @param toResolve join paths and join paths fragments that are not resolved yet.
+     * @param resolved join paths and join fragments that are already resolved
+     * @param dictionary dictionary instance
+     * @return A SQL fragment that references a database column
+     */
+    public static String generateColumnReference(Path path,
+                                                 LinkedHashSet<Path> toResolve,
```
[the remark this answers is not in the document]
aklish: I think we should break this up.  JoinTo should be expanded independently from the expression logic.  Each column that has a JoinTo clause - can create and store a singular path that is fully expanded.  The logic which expands this can check for cycles.

Separately from the JoinTo expansion, we can handle expression expansion.  This logic should be consolidated for metrics and dimensions.   This might require some refactoring of those respective classes.  Let's get together and map out a design.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/MetaDataStore.java
```diff
@@ -180,4 +180,16 @@ public static boolean isMetricField(EntityDictionary dictionary, Class<?> cls, S
...
+
+    /**
+     * Construct a column name as meta data
+     *
+     * @param tableClass table class
+     * @param fieldName field name
+     * @param dictionary entity dictionary to use
+     * @return <code>tableAlias.fieldName</code>
+     */
+    public static String constructColumnName(Class<?> tableClass, String fieldName, EntityDictionary dictionary) {
```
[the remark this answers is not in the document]
hellohanchen: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Metric.java
```diff
@@ -32,19 +41,150 @@
...
+     *
+     * @param tableClass talbe class
+     * @param fieldName metric field name
+     * @param toResolve references that are not resolved yet, to detect cycle-reference
+     * @param resolved references that are already resolved
+     * @param meta meta annotation on the field
+     * @param dictionary dictionary with entity information
+     * @return resolved metric function instance
+     */
+    private static MetricFunction resolveFormula(Class<?> tableClass,
```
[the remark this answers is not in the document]
hellohanchen: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Metric.java
```diff
@@ -32,19 +41,150 @@
...
+    private static MetricFunction resolveAggregation(Class<?> tableClass,
+                                                     String fieldName,
+                                                     MetricAggregation aggregation,
+                                                     Meta meta,
+                                                     EntityDictionary dictionary) {
+        String columnName = constructColumnName(tableClass, fieldName, dictionary);
         try {
-            this.metricFunction = metric.function().newInstance();
-            metricFunction.setName(getId() + "[" + metricFunction.getName() + "]");
+            MetricFunction metricFunction = aggregation.function().newInstance();
```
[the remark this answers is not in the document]
hellohanchen: The reason why I do this is that the `@Metric` annotation only takes in the class name of function class. And in the metadata package, `Metric` should not know the difference between `SQLMetricFunction` and `MetricFunction`.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLColumn.java
```diff
@@ -23,20 +28,20 @@
     private final String reference;
 
     @Getter
-    private final Path joinPath;
+    private final List<Path> joinPaths = new ArrayList<>();
```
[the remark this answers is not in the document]
hellohanchen: A `@DimensionFormula` can contains more that one references that need joins.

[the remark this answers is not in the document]
aklish: But those would reference other columns which would be annotated with their own JoinTo.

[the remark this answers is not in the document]
aklish: Or `@DimensionFormula` can reference paths (books.authors.name) directly as a symbol.

PR #1140, opened 2020-01-16
Fix travis log length

aklish:
## Description

Travis is killing builds because Elide 5.0 is exceeding its max log length.  This is a cheap fix.

## License


wcekan:
Remove the `.log().all()` and replace with logging only on failed tests.

```java
RestAssured.enableLoggingOfRequestAndResponseIfValidationFails();
```

PR #1162, opened 2020-01-25
Core tests

wcekan:
## Description
Adding some missing tests.
- dictionary field lookups
- field coercion
- field path building
- check lifecycle handling during exceptions

## Motivation and Context
Better elide-core code coverage

## How Has This Been Tested?
Unit testing

## License

### DIFF
--- a/elide-core/src/test/java/com/yahoo/elide/core/LifeCycleTest.java
+++ b/elide-core/src/test/java/com/yahoo/elide/core/LifeCycleTest.java
@@ -11,10 +11,12 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyString;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.ArgumentMatchers.isA;
 import static org.mockito.ArgumentMatchers.isNull;
 import static org.mockito.Mockito.clearInvocations;
+import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.times;
@@ -48,11 +50,13 @@ import com.yahoo.elide.security.User;
 import com.yahoo.elide.security.checks.Check;
 
 import com.google.common.collect.Sets;
+
 import example.Author;
 import example.Book;
 import example.Editor;
 import example.Publisher;
 import example.TestCheckMappings;
+
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
@@ -61,9 +65,11 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
+
 import javax.persistence.Entity;
 import javax.persistence.Id;
 import javax.persistence.Transient;
+import javax.validation.ConstraintViolationException;
 import javax.ws.rs.core.MultivaluedHashMap;
 import javax.ws.rs.core.MultivaluedMap;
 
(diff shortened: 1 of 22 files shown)


elide-core/src/test/java/com/yahoo/elide/jsonapi/JsonApiTest.java
```diff
@@ -326,5 +339,21 @@ public void readListWithMeta() throws IOException {
...
+
+    private void checkEquality(JsonApiDocument doc1) {
+        JsonApiDocument doc2;
+        try {
+            String json = mapper.writeJsonApiDocument(doc1);
+            doc2 = mapper.readJsonApiDocument(json);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        if (doc2.getData() == null) {
```
aklish: What's this for?

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
```diff
@@ -110,6 +117,7 @@ public PersistentResourceTest() {
         badUserScope = new RequestScope(null, null, mock(DataStoreTransaction.class),
                 new User(-1), null, elideSettings);
         reset(goodUserScope.getTransaction());
+        assertFalse(goodUserScope.isUseFilterExpressions());
```
aklish: This is not a test.  Why use assert here?

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
```diff
@@ -2167,6 +2176,57 @@ public void testRelationChangeSpecType() {
             assertTrue(model.updateRelation("child", null));
     }
 
+    @Test
+    public void testPatchRequestScope() {
+        DataStoreTransaction tx = mock(DataStoreTransaction.class);
+        PatchRequestScope parentScope =
```
aklish: Does the code actually nest patch scopes like this?  I thought patch was the parent scope - but child scopes were always normal request scopes.

elide-core/src/test/java/com/yahoo/elide/core/LifeCycleTest.java
```diff
@@ -115,22 +127,21 @@ public TestEntityDictionary(Map<String, Class<? extends Check>> checks) {
...
+        ImmutableList.of(
+                OnUpdatePostCommit.class, OnUpdatePreCommit.class, OnUpdatePreSecurity.class)
+                .stream().forEach(cls -> dictionary.bindTrigger(Book.class, cls, "title", callback));
         dictionary.bindTrigger(Book.class, OnUpdatePreCommit.class, onUpdateDeferredCallback, true);
         dictionary.bindTrigger(Book.class, OnUpdatePreSecurity.class, onUpdateImmediateCallback, true);
         dictionary.bindTrigger(Book.class, OnUpdatePostCommit.class, onUpdatePostCommitCallback, true);
         dictionary.bindTrigger(Author.class, OnUpdatePostCommit.class, onUpdatePostCommitAuthor, true);
+        // enable trace
+        Logger rootLogger = (Logger) LoggerFactory.getILoggerFactory().getLogger(Logger.ROOT_LOGGER_NAME);
+        rootLogger.setLevel(Level.TRACE);
```
aklish: Do we want to leave this log level on?

elide-core/src/test/java/com/yahoo/elide/jsonapi/JsonApiTest.java
wcekan: See #1162

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
wcekan: Why is this not a test?

elide-core/src/test/java/com/yahoo/elide/core/LifeCycleTest.java
wcekan: Yes, otherwise we are not testing logging works without throwing exceptions.

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
wcekan: Yes, they are nested.  That is what this tests.  See

https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/extensions/JsonApiPatch.java#L65

PR #1163, opened 2020-01-26
JsonApiDocument hashCode and equals were inconsistent

wcekan:
## Description
Pulled out only the code changes from #1162 for easier review.
- JsonApiDocument hashCode and equals were inconsistent.
- Correctly process empty ConstraintViolations
- Remove unnecessary initializations in code

## Motivation and Context
Stability

## How Has This Been Tested?
Unit testing

## License

### DIFF
--- a/elide-core/src/test/java/com/yahoo/elide/jsonapi/JsonApiTest.java
+++ b/elide-core/src/test/java/com/yahoo/elide/jsonapi/JsonApiTest.java
@@ -35,6 +35,7 @@ import org.junit.jupiter.api.Test;
 import org.mockito.Answers;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashSet;
@@ -90,6 +91,7 @@ public class JsonApiTest {
 
         String doc = mapper.writeJsonApiDocument(jsonApiDocument);
         assertEquals(expected, doc);
+        checkEquality(jsonApiDocument);
     }
 
     @Test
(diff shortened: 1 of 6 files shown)


elide-core/src/test/java/com/yahoo/elide/jsonapi/JsonApiTest.java
```diff
@@ -326,5 +339,21 @@ public void readListWithMeta() throws IOException {
...
+
+    private void checkEquality(JsonApiDocument doc1) {
+        JsonApiDocument doc2;
+        try {
+            String json = mapper.writeJsonApiDocument(doc1);
+            doc2 = mapper.readJsonApiDocument(json);
+        } catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        if (doc2.getData() == null) {
```
aklish: Why do we do this?

wcekan: Good question.  Seems the null is not handled well.

I found the problem. Somehow serializing changes the value of doc.data!

wcekan: I removed this hack and added a verification that data is unchanged after serialization, which fails :(

wcekan: It seems "Empty" object and Null object are written the same, but read differently.

wcekan: expected: 
<JsonApiDocument(data=Data(values=[], relationshipType=MANY_TO_ONE), meta=null, links=null, includedRecs=[], included=null)> 
but was: 
<JsonApiDocument(data=null, meta=null, links=null, includedRecs=[], included=null)>

wcekan: Will now treat "Empty" object and Null data object as equal.

PR #1172, opened 2020-01-27
null access suggestions

wcekan:
## Description
Respond to null pointer alerts:
https://lgtm.com/projects/g/yahoo/elide/alerts?mode=list

## Motivation and Context
Stability

## How Has This Been Tested?
Unit tests and product tests

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/jsonapi/models/Relationship.java
+++ b/elide-core/src/main/java/com/yahoo/elide/jsonapi/models/Relationship.java
@@ -77,6 +77,7 @@ public class Relationship {
                     if (data.isToOne() && resource == null) {
                         continue;
                     }
+                    // TODO: what if resource==null here?
                     res.add(resource.toPersistentResource(requestScope));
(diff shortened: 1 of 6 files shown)


elide-core/src/main/java/com/yahoo/elide/jsonapi/models/Relationship.java
```diff
@@ -77,6 +77,7 @@ public Relationship(@JsonProperty("links") Map<String, String> links,
                     if (data.isToOne() && resource == null) {
                         continue;
                     }
+                    // TODO: what if resource==null here?
```
wcekan: @aklish What is the intent here when resource is null?

aklish: We need a test to find out.

wcekan: Added test and fixed TODO.

PR #1176, opened 2020-01-30
Added IsEmpty operation for filter predicate

### LINKED ISSUE #1160: Add IsEmpty Filter Operator
Elide should provide a filter operator to test if a collection is empty.

## Expected Behavior

Support filters of the variety.

`filter=book.authors=isempty=true`
`filter=book.authors=isempty=1`

Both attribute and relationship collections should be supported.

## Current Behavior
Filters that span relationships is partially supported in JSON-API and not supported in GraphQL.  Issue #1141 will remedy that.

Also, filters today are only supported on model attributes.  This needs to be expanded to include relationships in limited operator contexts (like this one).

## Possible Solution
We must first address #1141 

This operator will likely require different behavior for JPQL stores and for in memory filtering.  

For JPQL stores, we'll need to use the collection empty operator:
https://docs.oracle.com/html/E13946_04/ejb3_langref.html#ejb3_langref_empty_comp

For in memory stores, we'll need to check if the field is null OR if the collection is empty.

Chandrasekar-Rajasekar:
Resolves #1160 

## Description
Added IsEmpty Filter operation with syntax `=isempty=`. 

## Motivation and Context
isempty filter operation are useful in most cases and the community needed support for this operation. 

## How Has This Been Tested?
Added New Test Cases

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/IsEmptyPredicate.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/IsEmptyPredicate.java
@@ -10,7 +10,7 @@ import com.yahoo.elide.core.Path;
 import java.util.Collections;
 
 /**
- * Is EMopty Predicate calss
+ * Is EMopty Predicate Class.
  */
 public class IsEmptyPredicate extends FilterPredicate {
 
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/IsEmptyPredicate.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/IsEmptyPredicate.java
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2019, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core.filter;
+
+import com.yahoo.elide.core.Path;
+
+import java.util.Collections;
+
+/**
+ * Is EMopty Predicate calss
+ */
+public class IsEmptyPredicate extends FilterPredicate {
+
+    public IsEmptyPredicate(Path path) {
+        super(path, Operator.ISEMPTY, Collections.emptyList());
+    }
+
+    public IsEmptyPredicate(Path.PathElement pathElement) {
+        super(pathElement, Operator.ISEMPTY, Collections.emptyList());
+    }
+}
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/IsEmptyPredicate.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/IsEmptyPredicate.java
@@ -10,7 +10,7 @@ import com.yahoo.elide.core.Path;
 import java.util.Collections;
 
 /**
- * Is EMopty Predicate Class.
+ * Is Empty Predicate Class.
  */
 public class IsEmptyPredicate extends FilterPredicate {
 
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
@@ -336,6 +336,7 @@ public enum Operator {
 
     private static <T> Predicate<T> isEmpty(String field, RequestScope requestScope) {
         return (T entity) -> {
+
             Object val = getFieldValue(entity, field, requestScope);
             if (val == null) { return true; }
             if (val instanceof String) {
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
@@ -14,8 +14,10 @@ import com.yahoo.elide.utils.coerce.CoerceUtil;
 import lombok.Getter;
 import lombok.RequiredArgsConstructor;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
+import java.util.Map;
 import java.util.function.Function;
 import java.util.function.Predicate;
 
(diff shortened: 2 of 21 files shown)


elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -368,5 +377,25 @@ private FilterExpression buildIsNullOperator(Path path, List<String> arguments)
...
+        private FilterExpression buildIsEmptyOperator(Path path, List<String> arguments) {
+            String arg = arguments.get(0);
+            try {
+                boolean wantsEmpty = CoerceUtil.coerce(arg, boolean.class);
+                if (wantsEmpty) {
+                    return new IsEmptyPredicate(path);
+                }
+                return new NotEmptyPredicate(path);
+            } catch (InvalidValueException ignored) {
+                throw new RSQLParseException(String.format("Invalid value for operator =isnull= '%s'", arg));
```
PatRolland: copy pasted =isnull=. here should be =isempty=

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -316,6 +334,24 @@ public static Operator fromString(final String string) {
         return (T entity) -> false;
     }
 
+    private static <T> Predicate<T> isEmpty(String field, RequestScope requestScope) {
+        return (T entity) -> {
+            Object val = getFieldValue(entity, field, requestScope);
+            if (val == null) { return true; }
+            if (val instanceof String) {
```
PatRolland: i feel like only collection and map should be handled here

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -316,6 +334,24 @@ public static Operator fromString(final String string) {
         return (T entity) -> false;
     }
 
+    private static <T> Predicate<T> isEmpty(String field, RequestScope requestScope) {
+        return (T entity) -> {
+            Object val = getFieldValue(entity, field, requestScope);
+            if (val == null) { return true; }
```
PatRolland: null and empty do not represent the same state imo

Chandrasekar-Rajasekar: `For in memory stores, we'll need to check if the field is null OR if the collection is empty.` This is the requirement given in the issue.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
Chandrasekar-Rajasekar: So the string `""` is not considered empty?

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
aklish: I agree - they are not the same thing.  But I think you could interpret this either way.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
aklish: So we generate JPQL queries based on IS EMPTY and IS NOT EMPTY.  I'm not sure those will work on Strings when we push this to the datastore.  

I'm inclined to think the JPQL queries will blow up if we try this on Strings in the DB.  Either remove the String case - or we need to add explicit tests to see that they work as expected.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
PatRolland: if you look at it in a db table way, a column referring to children might be null or have a value with no associations .
consider 3 tables
TableA(a_id), TableB(b_id) and TableAB(ab_id, a_id, b_id).
You might end up with a row in TableAB such as ab_id = 1, a_id = 1 and b_id = null thus representing an empty collection.
You migth also end up having no row in TableAB thus representing a null collection.

I'm not sure if this is a concept worth representing in the =isempty= filter but that is how I can represent the distinction here.

aklish: @jkusa Any thoughts on this one?

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
aklish: @Chandrasekar-Rajasekar Let's remove the string case.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
aklish: We'll run a test and see what JPQL does.  The in-memory result should match the JPQL result.

elide-datastore/elide-datastore-hibernate5/pom.xml
```diff
@@ -37,6 +37,10 @@
         <url>https://github.com/yahoo/elide.git</url>
         <tag>HEAD</tag>
     </scm>
+    <properties>
+        <excludeTags>emptyOnAttributeCollection</excludeTags>
```
Chandrasekar-Rajasekar: Looks like the fix for querying attribute collection on EMPTY operator is fixed in version `5.4.11`. We are using `5.4.2.Final`. 
One option is we use `where size(book.awards) = 0` as suggested in below link. 
Or we could continue using Empty operator and change the version to 5.4.11 when it gets released.

https://hibernate.atlassian.net/browse/HHH-6686?attachmentOrder=asc&oldIssueView=true
http://hibernate.org/orm/releases/5.4/

aklish: We can update Hibernate version on this branch as well.

Chandrasekar-Rajasekar: It is not released yet http://hibernate.org/orm/releases/5.4/

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
PatRolland: Makes sense. Is there a way where I could make the distinction between an absence of association between and an empty one then? I feel like is isNull and isEmpty operators would clash in some way.

aklish: @PatRolland - Not quite following

PatRolland: Doesn't matter I'm being picky 😄

elide-core/src/main/java/com/yahoo/elide/core/filter/IsEmptyPredicate.java
```diff
@@ -0,0 +1,24 @@
...
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core.filter;
+
+import com.yahoo.elide.core.Path;
+
+import java.util.Collections;
+
+/**
+ * Is EMopty Predicate Class.
```
aklish: typo

Issue #1005, opened 2019-10-10
RFC: Create Metadata For Aggregation Store


aklish:
## Overview

Elide will export metadata (as elide models) for every table it manages via the Aggregation Data Store.   

### Supporting Types

Elide & Navi will start with the following set of primitive types:
 - Time
 - Integer
 - Decimal
 - Money
 - Text
 - Coordinate
 - Boolean

```java
public enum ValueType {
    TIME,
    INTEGER,
    DECIMAL,
    MONEY,
    TEXT,
    COORDINATE,
    BOOLEAN
}
```

### Tables

Tables are not physical tables but represent what can logically be projected by Navi.

```java
@Include(rootLevel = true, type = "table")
@Data
@ToString
@Builder
public class Table {
    @Id
    private String id;

    private String name;

    private String description;

    private String category;

    @Exclude
    private boolean hidden = false;

    private CardinalitySize cardinality;

    @OneToMany
    @ToString.Exclude
    private Set<Metric> metrics;

    @OneToMany
    @ToString.Exclude
    private Set<Dimension> dimensions;

    @OneToMany
    @ToString.Exclude
    private Set<TimeDimension> timeDimensions;

    @ToString.Exclude
    private Set<String> tags;
}
```
Tables support a set of tags that can either be used:
1.   by elide for tooling purposes.  
2.  defined by the client

Tags include a namespace in the form `namespace.tag`.  The namespace  _elide_ is reserved.

### Columns

Columns are the base type for metrics and dimensions inside tables.  Columns can be projected, filtered, and sorted on.

```java
@Data
@ToString
@Builder
@Include
public abstract class Column {
    @Id
    private String id;

    private String name;

    private String description;

    @ToOne
    private Table table;

    //Defaults to self. Point to the terminal source column
    @ToOne
    private Column sourceColumn;

    @Exclude
    private boolean hidden = false;

    private String category;

    private ValueType valueType;

    @ToString.Exclude
    private Set<String> tags;
}
```

Columns support a set of tags that can either be used:
1.   by elide for tooling purposes.  
2.  defined by the client

Tags include a namespace in the form `namespace.tag`.  The namespace  _elide_ is reserved.

### Metrics

Metrics are functions that are computed over a group of related rows.

```java
@EqualsAndHashCode(callSuper = true)
@Include(type = "metric")
@Data
public class Metric extends Column {
    private Format defaultFormat;

    @ManyToOne
    @ToString.Exclude
    private MetricFunction metricFunction;
[...]

PR #1185, opened 2020-02-11
@Join and JoinPath

hellohanchen:
Resolves https://github.com/yahoo/elide/issues/1005

## Description
- Add `@Join` annotation for `Table` to use to join to other tables or entities.
- Add `JoinPath` which is a path but can navigate through models using `@Join` annotation.
- Add `TableTag` and `ColumnTag`.
- Add API in `EntityDictionary` to allow hiding a field from API.

## How Has This Been Tested?
unit tests and integration tests.

## License


elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/PlayerStatsWithView.java
```diff
@@ -200,8 +200,7 @@ public void setSubCountryIsoCode(String isoCode) {
         this.subCountryIsoCode = isoCode;
     }
 
-    @ToOne
-    @JoinTo(joinClause = "%from.country_id = %join.id")
+    @Join("%from.country_id = %join.id")
```
aklish: I'm worried countryView is going to show up in the GraphQL schema as an attribute.  While we can always have the developer add `@Exclude`, it would be better if JoinTo automatically excluded the property.

One way to do this is to promote AggregationStore annotations to elide-annotations as  first class citizens.  I think Elide 5.0 should collapse elide-annotations and elide-core because there are natural dependencies between them.
I don't want to do that collapse now though because it will make rebasing harder.

Another option is to allow the entity dictionary to accept hints when the data store binds the entities.  We could make another bindEntity method that takes additional annotations to exclude while binding.  I think for now, I favor this option.

hellohanchen: The field is hidden in both JSON API and GraphQL. I have this latest commit.
https://github.com/yahoo/elide/pull/1185/commits/3053cdf49c4b019a7204f1418e77678efe8eac40

aklish: I don't want to add a new annotation for this.  We already have Exclude.   I think Join should be implicitly hidden (without having to add another declaration).

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -872,7 +889,29 @@ public void bindEntity(Class<?> cls) {
...
+        }
+    }
+
+    /**
+     * Add an EntityBinding instance to dictionary.
+     *
+     * @param entityBinding EntityBinding instance
+     */
+    public void bindEntity(EntityBinding entityBinding) {
+        Class<?> declaredClass = entityBinding.entityClass;
```
aklish: Why not have this method call the new bindEntity method with an empty set (or a set that includes Exclude)?

elide-core/src/main/java/com/yahoo/elide/core/Path.java
```diff
@@ -70,7 +87,20 @@ public Path(Class<?> entityClass, EntityDictionary dictionary, String dotSeparat
...
+
+    /**
+     * Check whether a field need navigation to another entity.
+     *
+     * @param entityClass entity class
+     * @param fieldName field name
+     * @param dictionary dictionary
+     * @return True if the field requires navigation.
+     */
+    protected boolean needNavigation(Class<?> entityClass, String fieldName, EntityDictionary dictionary) {
```
aklish: Nice

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -178,6 +178,13 @@
             <version>2.5.0</version>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.ant</groupId>
```
aklish: Why do we need ant?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Table.java
```diff
@@ -55,17 +57,21 @@
...
+    @ToString.Exclude
+    private Set<String> tableTags;
+
     @Exclude
     @ToString.Exclude
     private final Map<String, Column> columnMap;
 
     public Table(Class<?> cls, EntityDictionary dictionary) {
-        if (!dictionary.getBindings().contains(cls)) {
+        if (!dictionary.getBoundClasses().contains(cls)) {
```
aklish: If Table is going to be a simple POJO - we probably don't need to pass dictionary to it.  This will also simplify the Fili integration.

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
hellohanchen: This `bindEntity` method needs to keep all information from the existing binding, as the `hiddenAnnotations` are not tracked in `EntityBinding`, if we bind the class again in a different dictionary, previous `hiddenAnnotations` would not be applied.

elide-datastore/elide-datastore-aggregation/pom.xml
hellohanchen: Used for read in graphql expected response from json file.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Table.java
hellohanchen: If using builder, there is not need to provide a dictionary.

elide-datastore/elide-datastore-aggregation/pom.xml
aklish: I'd rather not add a dependency on ant in Elide.   There are other places in Elide where we slurp in files.  Let's just do the same approach.

PR #1200, opened 2020-02-24
Cleanup Error Reporting

### LINKED ISSUE #650: Errors should be objects
In the JSON API [spec](http://jsonapi.org/format/#error-objects), the error is described as “Error Objects” and has some members. The GraphQL [spec](http://facebook.github.io/graphql/October2016/#sec-Errors) also require this.

**Actual Result**
```json
{
    "errors": [
        "error description"
    ]
}
```

**Expected Result**
```json
{
    "errors": [
        {
            "detail": "error description"
        }
    ]
}
```

aklish:
Resolves #650 and many other improvements around error handling/reporting.

## Description

1. JSON-API requires error objects _status_ field to be text (not an integer).  This is fixed to match the specification.
2. JSON-API patch extension is an array of error objects.  Each object in the array needs a top-level errors array.  This is fixed to match the specification.
3. All error text is HTML encoded.
4. Simplified Elide settings to turn on verbose error reporting (This is a flag now).
5. Remove unused settings in Elide settings.
6. Added fixes where errors were not being reported.
7. Improved error text for ForbiddenAccessExceptions.
8. Removed exception names from errors. 

## Motivation and Context
Errors were not matching the specification for JSON-API.   Errors were missing useful text in some cases. 

## How Has This Been Tested?
All existing tests pass.

## License


aklish:
@rmihalko Correct - Error encoding is always on and there is no easy way to disable it.

PR #1203, opened 2020-02-26
Elide 5.x async

Abhino:
Elide Async Feature

Co-authored-by: moizarafat

Related pull requests:

elide-spring changes for Elide-Async: https://github.com/yahoo/elide/pull/1204
elide-standalone changes for Elide-Async: https://github.com/yahoo/elide/pull/1205

## Description
These changes add the Elide Async Feature to Elide-5.x versions. The Elide Async API will support two new entities for asynchronous scheduled queries. Each Elide instance will run a background ExecutorService process which is responsible for running the query (AsyncQuery) in the background and posting the results to the database (AsyncQueryResult).

The ExecutorService executes queries programmatically using the Elide and QueryRunner interfaces. A post-commit hook (Elide lifecycle hook) on the AsyncRequest model queues up a task in the executor service to run the query. 

Additionally there are two additional services to support this feature:
1) The AsyncQueryInterruptor service handles any long running query beyond the max runtime. This service terminates the async query if max run time is reached and sets the status in the result to TIMEDOUT.

2) Should the service be stopped or crash with tasks running, the tasks can hang in the PROCESSING state. Therefore, the AsyncQueryCleaner service looks for expired AsnycQueries in the PROCESSING state and mark them as TIMEDOUT after a timeout.

Both the elide Standalone and Elide-Spring have been updated to include this feature and are configurable - the updates are part of separate pull requests as mentioned above.
    
## Motivation and Context
The Async module would help users who have potentially long running queries/requests submit asynchronous requests and view the results later. 

## How Has This Been Tested?
The feature has been tested with elide-standalone and elide-spring with sample application implementations.
Additional unit tests and integration tests will be added for the Async features.

## License


elide-async/pom.xml
```diff
@@ -0,0 +1,261 @@
...
+            <version>5.0.0-pr7-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.10</version>
+        </dependency>
+        <dependency>
+            <groupId>org.antlr</groupId>
+            <artifactId>antlr4-runtime</artifactId>
```
aklish: Are all of these dependencies really needed or was this copied from another pom?  Why do we need antlr?  Can we slim this down to just the direct dependencies?

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -0,0 +1,106 @@
...
+    @PreUpdate
+    public void preUpdate() {
+        this.updatedOn = new Date();
+    }
+
+    public void setResult(AsyncQueryResult result) {
+        this.result = result;
+    }
+
+    public Date getUpdatedOn() {
```
aklish: Can we use lombok and get rid of all the boilerplate getters & setters?

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -0,0 +1,106 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Model for Async Query
+ */
+@Entity
+@Include(type = "query", rootLevel = true)
+@ReadPermission(expression = "Principal is Owner")
+@UpdatePermission(expression = "Prefab.Role.None")
```
aklish: What about DeletePermission?

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQueryResult.java
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Model for Async Query Result
+ */
+@Entity
+@Include(type="queryResult")
+@ReadPermission(expression = "Principal is Owner")
+@UpdatePermission(expression = "Prefab.Role.None")
+@CreatePermission(expression = "Prefab.Role.None")
+public class AsyncQueryResult implements PrincipalOwned {
```
aklish: What about DeletePermission

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQueryResult.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private Date createdOn;
+
+    private Date updatedOn;
+
+    @OneToOne
+    private AsyncQuery query;
+
+    @Exclude
+    public String getPrincipalName() {
```
aklish: Let's use lombok to remove getters & setters.

elide-async/src/main/java/com/yahoo/elide/async/models/security/AsyncQueryOperationChecks.java
```diff
@@ -0,0 +1,50 @@
...
+    @SecurityCheck(AsyncQueryOwner.PRINCIPAL_IS_OWNER)
+    public static class AsyncQueryOwner extends OperationCheck<Object> {
+
+        public static final String PRINCIPAL_IS_OWNER = "Principal is Owner";
+
+        @Override
+        public boolean ok(Object object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
+            Principal principal = ((Principal) requestScope.getUser().getOpaqueUser());
+
+            if(object.getClass().equals(AsyncQuery.class)) {
```
aklish: Cast to PrincipalOwned.  Then you don't need to handle the two cases differently.  

Also - let's file a ticket to move this to a FilterExpressionCheck

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,95 @@
...
+            EntityProjection asyncQueryCollection = EntityProjection.builder()
+                    .type(AsyncQuery.class)
+                    .filterExpression(filter)
+                    .build();
+
+            Iterable<Object> loaded = tx.loadObjects(asyncQueryCollection, scope);
+            Iterator<Object> itr = loaded.iterator();
+            while(itr.hasNext()) {
+                AsyncQuery query = (AsyncQuery) itr.next();
+                long differenceInMillies = Math.abs((new Date()).getTime() - query.getCreatedOn().getTime());
```
aklish: Move all this math to a separate function:

```java
boolean isTimedOut(AsyncQuery query) { .. }
```

aklish: Do the date.getTime outside the loop once.  There could be a lot of records.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,95 @@
...
+                long difference = TimeUnit.MINUTES.convert(differenceInMillies, TimeUnit.MILLISECONDS);
+
+                // Check if its twice as long as max run time. It means the host/app crashed or restarted.
+                if(difference > maxRunTime * 2) {
+                    log.info("Updating Async Query Status to TIMEDOUT");
+                    query.setQueryStatus(QueryStatus.TIMEDOUT);
```
aklish: This thread should delete the query rather than timing it out.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryInterruptThread.java
```diff
@@ -0,0 +1,104 @@
...
+        interruptQuery();
+    }
+
+    /**
+     * This is the main method which interrupts the Async Query request, if it has executed beyond
+     * the maximum run time.
+     */
+    protected void interruptQuery() {
+        try {
+            long interruptTimeInMillies = interruptTime * 60 * 1000;
```
aklish: Let's make these configurable.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryInterruptThread.java
```diff
@@ -0,0 +1,104 @@
...
+            long differenceInMillies = interruptTimeInMillies - ((new Date()).getTime() - submittedOn.getTime());
+            
+            if(differenceInMillies > 0) {
+               log.debug("Sleeping for {}", differenceInMillies);
+               Thread.sleep(differenceInMillies);
```
aklish: Instead of sleeping here, just wait on the future with the given timeout:
```java
future.get(long timeout, TimeUnit unit)
```
If the task returns early, this thread can exit early too.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryInterruptThread.java
```diff
@@ -0,0 +1,104 @@
...
+        DataStoreTransaction tx = elide.getDataStore().beginTransaction();
+
+        // Creating new RequestScope for Datastore transaction
+        RequestScope scope = new RequestScope(null, null, tx, null, null, elide.getElideSettings());
+
+        try {
+            EntityProjection asyncQueryCollection = EntityProjection.builder()
+                    .type(AsyncQuery.class)
+                    .build();
+            AsyncQuery query = (AsyncQuery) tx.loadObject(asyncQueryCollection, asyncQueryId, scope);
```
aklish: Let's commonize all of this logic to load an AsyncQuery and do something (pass in a function that manipulates the query).

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -0,0 +1,106 @@
...
+@Slf4j
+public class AsyncQuery implements PrincipalOwned {
```
aklish: All of the Elide models should implement hashcode & equals.  There are some best practices around how to make this work with ORMs.  You can look at BaseId in Elide to get one of the approaches.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,95 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Runnable thread for updating AsyncQueryThread status
+ * beyond the max run time and if not terminated by interrupt process
+ * due to app/host crash or restart.
+ */
+@Slf4j
+public class AsyncQueryCleanerThread implements Runnable {
```
aklish: To be on the safe side, I would add lombok hashcode and equals to all threads that are being managed by a thread pool.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
moizarafat: If we delete the Query entry on the restart, it will confuse the user where the query went because an application crash or restart will not be known to him.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
moizarafat: Will make the DeletePermission as None .

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQueryResult.java
Abhino: Should we set the DeletePermission as None for the AsyncQueryResult ?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryInterruptThread.java
moizarafat: interruptTime is set when the elide instance comes up based on the properties/settings set by the user in application.yaml/settings.java and it gets applied to all the queries. If the user does not set it, its defaulted to 60 minutes.  Do you mean it should be an attribute in AsyncQuery and every query should have its own timeout provided by the user? if not provided, we default it.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * This method updates the model for AsyncQuery with passed status value.
+     * @param status new status based on the enum QueryStatus
+     * @param asyncQueryId queryId from asyncQuery request
+     * @throws IOException IOException from DataStoreTransaction
+     * @return AsyncQuery Object
+     */
+    protected AsyncQuery updateAsyncQueryStatus(QueryStatus status, UUID asyncQueryId) throws IOException {
+        log.debug("Updating AsyncQuery status to {}", status);
+        DataStoreTransaction tx = elide.getDataStore().beginTransaction();
```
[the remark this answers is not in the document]
Abhino: Right now this method throws a exception and is caught by the surrounding try - catch block in the processQuery method. That try catch block should catch any of the leaked leak DB connections cause by the database transactions. Should we put a try catch block in each of the DB transactions methods separately?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    /**
+     * This method updates the model for AsyncQuery with result object,
+     * @param asyncQueryResult AsyncQueryResult object to be associated with the AsyncQuery object
+     * @param asyncQueryId UUID of the AsyncQuery to be associated with the AsyncQueryResult object
+     * @throws IOException IOException from DataStoreTransaction
+     */
+    protected void updateAsyncQueryStatus(AsyncQueryResult asyncQueryResult, UUID asyncQueryId) throws IOException {
+        log.debug("Updating AsyncQueryResult to {}", asyncQueryResult);
```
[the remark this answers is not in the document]
Abhino: Right now we have a overloaded method for updateAsyncQueryStatus - one of them is used to update Query Status and the other one is used to add the AsyncQueryResult Object. We do it this way because at any instance we only need to update either the query status or add the result object.  Could you please elaborate on how we can create this generic updateAsyncQuery function ?

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
Abhino: We are adding @Data for AsyncQuery.java which internally includes @EqualsAndHashCode. Do we need to add these separately for the models?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
Abhino: We are already adding `@Data` for AsyncQueryCleanerThread.java which internally includes `@EqualsAndHashCode`. Do we need to add these separately for the threads?

elide-async/src/main/java/com/yahoo/elide/async/models/security/AsyncQueryOperationChecks.java
moizarafat: I can not remember the reason, but can you explain what will be the advantage of moving this to FilterExpressionCheck vs the OperationCheck?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,219 @@
...
+                log.debug("JSONAPI_V1_0 getBody: {}", response.getBody());
+            }
+            else if (queryType.equals(QueryType.GRAPHQL_V1_0)) {
+                response = runner.run(query, user);
+                log.debug("GRAPHQL_V1_0 getResponseCode: {}", response.getResponseCode());
+                log.debug("GRAPHQL_V1_0 getBody: {}", response.getBody());
+            }
+            // if 200 - response code then Change async query to complete else change to Failure
+            if (response.getResponseCode() == 200) {
+                asyncQuery = updateAsyncQueryStatus(QueryStatus.COMPLETE, id);
```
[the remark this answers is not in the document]
aklish: @jkusa

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
aklish: `@Data` should be sufficient.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
aklish: `@Data` should be sufficient.

aklish: @Data is not sufficient for use with ORMs.

pom.xml
```diff
@@ -25,6 +25,7 @@
         <module>elide-contrib</module>
         <module>elide-standalone</module>
         <module>elide-spring</module>
+        <module>elide-async</module>
```
aklish: The list of modules specifies the order of the build.  elide-async should be moved higher up so we can use it in other places (like spring and standalone).

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncDbUtil.java
```diff
@@ -0,0 +1,100 @@
...
+            .build();
+        RequestScope scope = new RequestScope(null, null, tx, null, null, elide.getElideSettings());
+        AsyncQuery query = (AsyncQuery) tx.loadObject(asyncQueryCollection, asyncQueryId, scope);
+        if (updatedAttribute.getClass().getSimpleName().equals("QueryStatus")) {
+            query.setStatus((QueryStatus) updatedAttribute);
+        }
+        else if (updatedAttribute.getClass().getSimpleName().equals("AsyncQueryResult")) {
+            query.setResult((AsyncQueryResult)updatedAttribute);
+        }
+        tx.save(query, scope);
```
aklish: My comments about using functional style here still applies.

elide-async/pom.xml
```diff
@@ -0,0 +1,97 @@
...
+            <version>5.0.0-pr7-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>javax.persistence</groupId>
+            <artifactId>javax.persistence-api</artifactId>
+            <version>2.2</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
```
aklish: What dependency do we have on http client?

moizarafat: For URI Builder used at  https://github.com/yahoo/elide/blob/7058821b3ed64054ebb7c7206d52cd5f97d020b1/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java#L110

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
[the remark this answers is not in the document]
jkusa: If we can fetch the query status (whether the query failed or succeeded) we should return a `200`.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -0,0 +1,102 @@
...
+@UpdatePermission(expression = "Prefab.Role.None")
+@DeletePermission(expression = "Prefab.Role.None")
+@Slf4j
+@Data
+public class AsyncQuery implements PrincipalOwned {
+    @Id
+    private UUID id; //Can be generated or provided.
+
+    //Extracted from the Principal object
+    private String principalName;
```
aklish: The principalName should be set by a lifecycle (OnCreatePreCommit) hook automatically rather than having the user provide it in the API request.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private QueryType queryType; //GRAPHQL, JSONAPI
+
+    @UpdatePermission(expression = "Principal is Owner AND value is Cancelled")
+    private QueryStatus status;
+
+    @OneToOne(mappedBy = "query")
+    private AsyncQueryResult result;
+
+    private Date createdOn;
```
aklish: An alternative way to generate dates is:

```java
    @Generated(GenerationTime.INSERT)
    @Column(updatable = false, insertable = false, columnDefinition = "timestamp default current_timestamp")
    @Temporal(TemporalType.TIMESTAMP)
    private Date createdOn;

    @Generated(GenerationTime.ALWAYS)
    @Column(updatable = false, insertable = false, columnDefinition = "timestamp default current_timestamp")
    @Temporal(TemporalType.TIMESTAMP)
    private Date updatedOn;
```

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -0,0 +1,102 @@
...
+
+    private Date updatedOn;
+
+    @Inject
+    @Transient
+    private AsyncExecutorService asyncExecutorService;
+
+    @Transient
```
aklish: The natural key needs to be persisted to get hashCode & equals to work correctly.  hashCode & equals only uses the natural key or comparisons.  Remove Transient and add Exclude.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @PreUpdate
+    public void preUpdate() {
+        this.updatedOn = new Date();
+    }
+
+    @OnCreatePostCommit
+    public void executeQueryFromExecutor(RequestScope scope) {
+        log.info("AsyncExecutorService executor object: {}", asyncExecutorService);
+        asyncExecutorService.executeQuery(query, queryType, scope.getUser(), id);
```
aklish: Why not just pass the Query object in here instead of breaking out params for it.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQueryResult.java
```diff
@@ -0,0 +1,85 @@
...
+
+    private String responseBody; //success or errors
+
+    private Integer status; // HTTP Status
+
+    private Date createdOn;
+
+    private Date updatedOn;
+
+    @Transient
```
aklish: Change to `@Exclude`

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleaner.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+
+import javax.inject.Singleton;
+
+/**
+ * Class for initializing the Query Cleaner.
+ */
+@Singleton
+class AsyncQueryCleaner {
```
aklish: Do we need both a a Cleaner and a CleanerService?  Why not just combine these into a single concept?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryExecutor.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import javax.inject.Singleton;
+
+/**
+ * Class for initializing the Query Executor.
+ */
+@Singleton
+class AsyncQueryExecutor {
```
aklish: Do we need both an Executor and an ExecutorService?  Can we combine them?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,140 @@
...
+    /**
+     * This method updates the status of long running async query which
+     * were not interrupted due to host crash/app shutdown to TIMEDOUT.
+     * */
+    @SuppressWarnings("unchecked")
+    private void deleteAsyncQuery() {
+        AsyncDbUtil asyncDbUtil = AsyncDbUtil.getInstance(elide);
+        Iterable<Object> loaded = (Iterable<Object>) asyncDbUtil.executeInTransaction(elide.getDataStore(), (tx, scope) -> {
+            try {
+                EntityDictionary dictionary = elide.getElideSettings().getDictionary();
```
aklish: You can move a lot of this setup code outside the closure (and hence outside the transaction).

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,140 @@
...
+                log.error("Exception: {}", e.getMessage());
+            }
+            return null;
+        });
+        Iterator<Object> itr = loaded.iterator();
+        while(itr.hasNext()) {
+            AsyncQuery query = (AsyncQuery) itr.next();
+
+            log.info("Found a query to DELETE");
+            asyncDbUtil.deleteAsyncQueryAndResult(query.getId());
```
aklish: This is where using the DataStoreTransaction interface in Elide is less than ideal.  There is no bulk edit interface.  This will generate lots of round trips to the database to delete objects one at a time (rather than delete them in bulk).  We should at least change the DbUtil.deleteAsync function take a list of objects to delete - and to remove them in a single transaction.  DdUtil should be an interface that can be created and overridden (as a Spring bean) if there are performance problems with this general approach.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,140 @@
...
+            }
+        }
+    }
+
+    private boolean isTimedOut(long currentTime, AsyncQuery query) {
+        long differenceMillies = Math.abs(currentTime - query.getCreatedOn().getTime());
+        long differenceMinutes = TimeUnit.MINUTES.convert(differenceMillies, TimeUnit.MILLISECONDS);
+
+        // Check if its twice as long as max run time. It means the host/app crashed or restarted.
+        return (differenceMinutes > maxRunTimeMinutes * 2);
```
aklish: Multiplying maxRunTimeMinutes by 2 should be done in a single place across all the code.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,140 @@
...
+                log.info("Updating Async Query Status to TIMEDOUT");
+                asyncDbUtil.updateAsyncQuery(query.getId(), (asyncQueryObj) -> {
+                    asyncQueryObj.setStatus(QueryStatus.TIMEDOUT);
+                    });
+            }
+        }
+    }
+
+    private boolean isTimedOut(long currentTime, AsyncQuery query) {
+        long differenceMillies = Math.abs(currentTime - query.getCreatedOn().getTime());
```
aklish: We can push this logic into the RSQL filter expression.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,140 @@
...
+        timeoutAsyncQuery();
+    }
+
+    /**
+     * This method updates the status of long running async query which
+     * were not interrupted due to host crash/app shutdown to TIMEDOUT.
+     * */
+    @SuppressWarnings("unchecked")
+    private void deleteAsyncQuery() {
+        AsyncDbUtil asyncDbUtil = AsyncDbUtil.getInstance(elide);
```
aklish: The two functions can generalized with a common function that does a load and takes an FilterExpression to filter the results.

The two functions can call the generalized function to do a filtered load and then either update or delete the queries respectively.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryInterruptor.java
```diff
@@ -0,0 +1,38 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+
+import javax.inject.Singleton;
+
+/**
+ * Class for initializing the Query Interruptor.
+ */
+@Singleton
+class AsyncQueryInterruptor {
```
aklish: Can this be combined with the Interruptor service?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,140 @@
...
+     * values for AsyncQuery and AsyncQueryResult models accordingly.
+     */
+    protected void processQuery() {
+        AsyncDbUtil asyncDbUtil = AsyncDbUtil.getInstance(elide);
+        try {
+            // Change async query to processing
+            asyncDbUtil.updateAsyncQuery(id, (asyncQuery) -> {
+                asyncQuery.setStatus(QueryStatus.PROCESSING);
+                });
+            //Thread.sleep(180000);
```
aklish: Remove.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,140 @@
...
+            ElideResponse response = null;
+            log.debug("query: {}", query);
```
aklish: It would be good to log the entire query object here.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,140 @@
...
+            log.debug("queryType: {}", queryType);
+            AsyncQuery asyncQuery;
+            AsyncQueryResult asyncQueryResult;
+            if (queryType.equals(QueryType.JSONAPI_V1_0)) {
+                MultivaluedMap<String, String> queryParams = getQueryParams(query);
+                response = elide.get(getPath(query), queryParams, user);
+                log.debug("JSONAPI_V1_0 getResponseCode: {}", response.getResponseCode());
```
aklish: Can we collapse these log lines into one?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,140 @@
...
+                log.debug("JSONAPI_V1_0 getBody: {}", response.getBody());
+            }
+            else if (queryType.equals(QueryType.GRAPHQL_V1_0)) {
+                response = runner.run(query, user);
+                log.debug("GRAPHQL_V1_0 getResponseCode: {}", response.getResponseCode());
```
aklish: Cab we collapse these log lines into one?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,140 @@
...
+                log.debug("GRAPHQL_V1_0 getBody: {}", response.getBody());
+            }
+            if (response != null){
+                asyncQuery = asyncDbUtil.updateAsyncQuery(id, (asyncQueryObj) -> {
+                    asyncQueryObj.setStatus(QueryStatus.COMPLETE);
+                });
+
+                // Create AsyncQueryResult entry for AsyncQuery
+                asyncQueryResult = asyncDbUtil.createAsyncQueryResult(response.getResponseCode(), response.getBody(), asyncQuery, id);
```
aklish: Creating the response and updating the status should be done inside a single transaction to the database.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,140 @@
...
+     * @return MultivaluedMap with query parameters
+     */
+    protected MultivaluedMap<String, String> getQueryParams(String query) throws URISyntaxException {
+        URIBuilder uri;
+        uri = new URIBuilder(query);
+        MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
+        for (NameValuePair queryParam : uri.getQueryParams()) {
+            queryParams.add(queryParam.getName(), queryParam.getValue());
+        }
+        log.debug("QueryParams: {}", queryParams);
```
aklish: If we log the entire asyncQuery object, we won't need all of these separate, sprinkled log messages.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -0,0 +1,140 @@
...
+     * This method parses the url and gets the query params and retrieves path
+     * to be used by underlying Elide.get method
+     * @param query query from the Async request
+     * @throws URISyntaxException URISyntaxException from malformed or incorrect URI
+     * @return Path extracted from URI
+     */
+    protected String getPath(String query) throws URISyntaxException {
+        URIBuilder uri;
+        uri = new URIBuilder(query);
+        log.debug("Retrieving path from query");
```
aklish: Remove log statement.  Let's consolidate logging into fewer (more complete log statements).

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
moizarafat: As discussed, we will keep as is since `@Generated` is hibernate annotations.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
moizarafat: Had to use OnCreatePreSecurity as using OnCreatePreCommit threw Null Pointer Exception.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQueryResult.java
```diff
@@ -0,0 +1,84 @@
...
+        createdOn = updatedOn = new Date();
+    }
+
+    @PreUpdate
+    public void preUpdate() {
+        updatedOn = new Date();
+    }
+
+    @Override
+    public int hashCode() {
```
aklish: I would move all of this common code to a base class like BaseId

aklish: You can even move the createdOn, updatedOn, and owner fields to a common super class.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
```diff
@@ -0,0 +1,73 @@
...
+    private final int MAX_CLEANUP_INTIAL_DELAY_MINUTES = 100;
+
+    private static AsyncCleanerService asyncCleanerService;
+    private ScheduledExecutorService cleanerService;
+
+    @Inject
+    public AsyncCleanerService(Elide elide, Integer maxRunTimeMinutes, Integer queryCleanupDays, AsyncQueryDAO asyncQueryDao) {
+
+    	//If query is still running for twice than maxRunTime, then interrupt did not work due to host/app crash.
+    	int queryRunTimeThresholdMinutes = maxRunTimeMinutes * 2;
```
aklish: Spacing is off in this function.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryDAO.java
```diff
@@ -0,0 +1,76 @@
...
+ * Utility interface which uses the elide datastore to modify, update and create
+ * AsyncQuery and AsyncQueryResult Objects
+ */
+public interface AsyncQueryDAO {
+
+    /**
+     * Set elide object
+     * @param elide Elide Object.
+     */
+    public void setElide(Elide elide);
```
aklish: This seems like a constructor argument rather than an interface contract.  I would remove.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryDAO.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * Set data store object
+     * @param dataStore Datastore Object from Elide.
+     */
+    public void setDataStore(DataStore dataStore);
```
aklish: Same comment.  We don't need this in the interface.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryDAO.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     * This method updates the model for AsyncQuery with passed value.
+     * @param asyncQueryId Unique UUID for the AsyncQuery Object
+     * @param updateFunction Functional interface for updating AsyncQuery Object
+     * @return AsyncQuery Object
+     */
+    public AsyncQuery updateAsyncQuery(UUID asyncQueryId, UpdateQuery updateFunction);
```
aklish: I don't think we should expose implementation details like UpdateQuery in a DAO interface.  

We should have methods like ```updateStatus(UUID asyncQueryId, QueryStatus status)```

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryDAO.java
```diff
@@ -0,0 +1,76 @@
...
+    public AsyncQueryResult setAsyncQueryAndResult(Integer status, String responseBody, AsyncQuery asyncQuery, UUID asyncQueryId);
+
+    /**
+     * This method creates a transaction from the datastore, performs the DB action using
+     * a generic functional interface and closes the transaction.
+     * @param dataStore Elide datastore retrieved from Elide object
+     * @param action Functional interface to perform DB action
+     * @return Object Returns Entity Object (AsyncQueryResult or AsyncResult)
+     */
+    public Object executeInTransaction(DataStore dataStore, Transactional action);
```
aklish: Remove - see earlier comment.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryDAO.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public Iterable<Object> updateAsyncQueryCollection(Iterable<Object> asyncQueryList, UpdateQuery updateFunction);
+
+    /**
+     * This method deletes a collection of AsyncQuery and AsyncQueryResult objects from database and
+     * returns the objects deleted.
+     * @param asyncQueryList Iterable list of AsyncQuery objects to be deleted
+     * @return query object list deleted
+     */
+    public Iterable<Object> deleteAsyncQueryAndResultCollection(Iterable<Object> asyncQueryList);
```
aklish: Let's not use Object here.  This should delete a collection of AsyncQuery objects.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryDAO.java
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public AsyncQuery updateAsyncQuery(UUID asyncQueryId, UpdateQuery updateFunction);
+
+    /**
+     * This method updates a collection of AsyncQuery objects from database and
+     * returns the objects updated.
+     * @param asyncQueryList Iterable list of AsyncQuery objects to be updated
+     * @return query object list updated
+     */
+    public Iterable<Object> updateAsyncQueryCollection(Iterable<Object> asyncQueryList, UpdateQuery updateFunction);
```
aklish: Remove or replace with more specific functions.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -0,0 +1,149 @@
...
+        });
+        return queryResultObj;
+    }
+
+    @Override
+    public Object executeInTransaction(DataStore dataStore, Transactional action) {
+        log.debug("executeInTransaction");
+        Object result = null;
+        try (DataStoreTransaction tx = dataStore.beginTransaction()) {
+	        JsonApiDocument jsonApiDoc = new JsonApiDocument();
```
aklish: Spacing is off.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -0,0 +1,149 @@
...
+                updateFunction.update(asyncQuery);
+                tx.save(asyncQuery, scope);
+            }
+            return asyncQueryList;
+        });
+        return asyncQueryList;
+    }
+
+    @Override
+    public Iterable<Object> deleteAsyncQueryAndResultCollection(Iterable<Object> asyncQueryList) {
```
aklish: Separate the load and delete.  This function should only delete (and not load anything).

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -0,0 +1,149 @@
...
+        log.debug("updateAsyncQueryCollection");
+        executeInTransaction(dataStore, (tx, scope) -> {
+            EntityProjection asyncQueryCollection = EntityProjection.builder()
+                    .type(AsyncQuery.class)
+                    .build();
+
+            Iterator<Object> itr = asyncQueryList.iterator();
+            while(itr.hasNext()) {
+                AsyncQuery query = (AsyncQuery) itr.next();
+                AsyncQuery asyncQuery = (AsyncQuery) tx.loadObject(asyncQueryCollection, query.getId(), scope);
```
aklish: I don't think this function should load anything.  Load can be a separate function.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryDAO.java
```diff
@@ -0,0 +1,76 @@
...
+     * @param elide Elide Object.
+     */
+    public void setElide(Elide elide);
+
+    /**
+     * Set data store object
+     * @param dataStore Datastore Object from Elide.
+     */
+    public void setDataStore(DataStore dataStore);
+
```
aklish: Add a function (see later comments):
```Collection<AsyncQuery> loadQueries(FilterExpression filter)```

File: elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java [hunk 52]
```diff
@@ -0,0 +1,124 @@
...
+                MultivaluedMap<String, String> queryParams = getQueryParams(queryObj.getQuery());
+                log.debug("Extracted QueryParams from AsyncQuery Object: {}", queryParams);
+                response = elide.get(getPath(queryObj.getQuery()), queryParams, user);
+                log.debug("JSONAPI_V1_0 getResponseCode: {}, JSONAPI_V1_0 getBody: {}", response.getResponseCode(), response.getBody());
+            }
+            else if (queryObj.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
+                response = runner.run(queryObj.getQuery(), user);
+                log.debug("GRAPHQL_V1_0 getResponseCode: {}, GRAPHQL_V1_0 getBody: {}", response.getResponseCode(), response.getBody());
+            }
+            if (response != null){
```
aklish: We are logging the response above before we check for null here.

aklish: Technically, I don't think this ever returns null.  I think it either returns non-null or throws an exception.

aklish: Instead of checking for null, we can just add an assert here that will throw an exception if response is null.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,118 @@
...
+    
+    /**
+     * This method updates the status of long running async query which
+     * were not interrupted due to host crash/app shutdown to TIMEDOUT.
+     * */
+	@SuppressWarnings("unchecked")
+    private void timeoutAsyncQuery() {
+
+        //Calculate date to filter for clean up
+        Calendar cal = Calendar.getInstance();
```
aklish: There is some duplicate code here with the prior function.   We should write a utility function to consolidate it.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -0,0 +1,118 @@
...
+        Format dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'");
+        String filterDateFormatted = dateFormat.format(filterDate);
+        log.debug("FilterDateFormatted = {}", filterDateFormatted);
+        String filterExpression = "status=in=(" + QueryStatus.PROCESSING.toString() + ","
+                + QueryStatus.QUEUED.toString() + ");createdOn=le='" + filterDateFormatted + "'";
+
+        Iterable<Object> loaded = getFilteredResults(filterExpression);
+
+        asyncQueryDao.updateAsyncQueryCollection(loaded, (asyncQuery) -> {
+            asyncQuery.setStatus(QueryStatus.PROCESSING);
```
aklish: Shouldn't this be TIMEDOUT instead of PROCESSING?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
```diff
@@ -0,0 +1,73 @@
...
+    	
+        // Setting up query cleaner that marks long running query as TIMEDOUT.
+        ScheduledExecutorService cleaner = AsyncCleanerService.getInstance().getExecutorService();
```
aklish: This doesn't look right to me.  I don't think a Constructor should be calling the a factory method for itself (the singleton).

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
```diff
@@ -0,0 +1,73 @@
...
+    private static AsyncCleanerService getInstance() {
+        if (asyncCleanerService == null) {
+          synchronized (AsyncCleanerService.class) {
+        	  asyncCleanerService = new AsyncCleanerService();
+          }
+        }
+        return asyncCleanerService;
+    }
+
+    private AsyncCleanerService() {
```
aklish: Why do we have the two constructors?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
```diff
@@ -0,0 +1,82 @@
...
+    private static AsyncExecutorService getInstance(int threadPoolSize) {
+        if (asyncExecutorService == null) {
+          synchronized (AsyncExecutorService.class) {
+        	  asyncExecutorService = new AsyncExecutorService(threadPoolSize);
+            }
+          }
+        return asyncExecutorService;
+    }
+
+    private AsyncExecutorService(int threadPoolSize) {
```
aklish: Why do we need these two constructors?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
```diff
@@ -0,0 +1,82 @@
...
+    private static AsyncExecutorService asyncExecutorService;
+    private AsyncQueryDAO asyncQueryDao;
+    
+
+    @Inject
+    public AsyncExecutorService(Elide elide, Integer threadPoolSize, Integer maxRunTime, AsyncQueryDAO asyncQueryDao) {
+        this.elide = elide;
+        this.runner = new QueryRunner(elide);
+        this.maxRunTime = maxRunTime;
+        executor = AsyncExecutorService.getInstance(threadPoolSize == null ? DEFAULT_THREADPOOL_SIZE : threadPoolSize).getExecutorService();
```
aklish: This looks wrong to me.  A constructor should not call a factory method on itself.  Maybe we should just eliminate all the singletons?

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -127,8 +149,38 @@ public AsyncQueryResult setAsyncQueryAndResult(Integer status, String responseBo
...
-    @Override
-    public Object executeInTransaction(DataStore dataStore, Transactional action) {
+    @SuppressWarnings("unchecked")
+    public Collection<AsyncQuery> loadQueries(String filterExpression) {
+        EntityDictionary dictionary = elide.getElideSettings().getDictionary();
+        RSQLFilterDialect filterParser = new RSQLFilterDialect(dictionary);
+
+        Collection<AsyncQuery> loaded = (Collection<AsyncQuery>) executeInTransaction(dataStore, (tx, scope) -> {
+            try {
+                FilterExpression filter = filterParser.parseFilterExpression(filterExpression, AsyncQuery.class, false);
```
aklish: This one line can be moved outside the transaction.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -127,8 +149,38 @@ public AsyncQueryResult setAsyncQueryAndResult(Integer status, String responseBo
...
+
+                EntityProjection asyncQueryCollection = EntityProjection.builder()
+                        .type(AsyncQuery.class)
+                        .filterExpression(filter)
+                        .build();
+
+                Iterable<Object> loadedObj = tx.loadObjects(asyncQueryCollection, scope);
+                return loadedObj;
+            } catch (Exception e) {
```
aklish: I wonder if this will swallow exceptions that would have been caught and handled by the outer transaction function.  Would it be better to let that function handle the exceptions?

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
aklish: Same comment still applies.  All of these objects have already been loaded.  This will result in a lot of unnecessary calls to the DB.  

Instead of loading them, just call the updateFunction and save.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -87,14 +109,14 @@ public AsyncQuery updateAsyncQuery(UUID asyncQueryId, UpdateQuery updateFunction
...
-    public Iterable<Object> deleteAsyncQueryAndResultCollection(Iterable<Object> asyncQueryList) {
+    public Collection<AsyncQuery> deleteAsyncQueryAndResultCollection(Collection<AsyncQuery> asyncQueryList) {
         log.debug("deleteAsyncQueryAndResultCollection");
         executeInTransaction(dataStore, (tx, scope) -> {
             EntityProjection asyncQueryCollection = EntityProjection.builder()
                     .type(AsyncQuery.class)
                     .build();
 
-            Iterator<Object> itr = asyncQueryList.iterator();
+            Iterator<AsyncQuery> itr = asyncQueryList.iterator();
```
aklish: Same comment about the load below.  The objects can be assumed to have already been loaded.  This function can just call tx.delete.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
Abhino: If we move this line outside the anonymous function - we would have to put it in a try catch block and move the filter variable as a class level variable else we get: Local variable filter defined in an enclosing scope must be final or effectively final. Do you want this variable to be a class level variable?

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
moizarafat: We are able to remove the load object logic for the update, but does not work with delete. Even calling save before delete does not work. It throws TransactionException trying to remove a detached entity.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncBase.java
```diff
@@ -0,0 +1,47 @@
+package com.yahoo.elide.async.models;
```
aklish, line 1: Copyright.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -167,6 +168,60 @@ default boolean enableJSONAPI() {
...
+        return 7;
+    }
+
+    /**
+     * Implementation of AsyncQueryDAO to use.
+     *
+     * @return AsyncQueryDAO type object.
+     */
+    default AsyncQueryDAO getAsyncQueryDAO() {
+        return null;
```
aklish, line 223: Maybe return the default one here?

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -0,0 +1,202 @@
...
+    @Override
+    @SuppressWarnings("unchecked")
+    public Collection<AsyncQuery> deleteAsyncQueryAndResultCollection(String filterExpression) {
+        log.debug("deleteAsyncQueryAndResultCollection");
+        EntityDictionary dictionary = elide.getElideSettings().getDictionary();
+        RSQLFilterDialect filterParser = new RSQLFilterDialect(dictionary);
+
+        Collection<AsyncQuery> asyncQueryList = (Collection<AsyncQuery>) executeInTransaction(dataStore, (tx, scope) -> {
+            try {
+                FilterExpression filter = filterParser.parseFilterExpression(filterExpression, AsyncQuery.class, false);
```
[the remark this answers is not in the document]
moizarafat: We are unable to move this outside the try block because the ParseException is thrown by the filterParser.parseFilterExpression method itself.

PR #1204, opened 2020-02-26
Elide 5.x elide-async integration in elide-spring

moizarafat:
Co-authored-by: Abhino <abhino@gmail.com>

Depends On: https://github.com/yahoo/elide/pull/1203

## Description
- Integrating with elide-async. 
- New properties class for Async related properties like enable/disable, number of threads for executing the async queries in parallel, the maximum run time for individual threads before they are automatically killed, number of instances hosting the elide service in parallel (to avoid all hosts trying to update status of orphan queries in parallel).
- New Async Configuration for binding AsyncExecutor Service when Async is enabled. It also adds Async Models to the dictionary.
- Adds the Async models to Swagger if Async and Swagger are enabled. The Async models are available on /swagger/doc/async path
- Added EntityScan to Main class with model packages path since earlier it will bind all models with `@Entity` annotation, but we don't want Async models to be bound if async is not enabled. 

## Motivation and Context
To allow submission of queries that will be executed asynchronously.

## How Has This Been Tested?
- Used elide-spring-boot-example with the updated version of the elide-spring artifact and verify query submissions and executions. 

NA

## License


elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Default thread pool size.
+     */
+    private int threadPoolSize = 6;
+
+    /**
+     * Default max query run time.
+     */
+    private int maxRunTime = 60;
```
aklish: Time variables should always include Units in their name so they are unambiguous.

aklish: I think we'll want two times here - one for the timeout and one for cleanup (deletion).  Both should be configurable.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -0,0 +1,31 @@
...
+
+    /**
+     * Default number of hosts running Elide.
+     */
+    private int numberOfHosts = 1;
```
aklish: Do we need this?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+public class ElideAsyncConfiguration {
+
+    /**
+     * Updates the entity dictionary for Elide by binding the Async Models.
+     * @param dictionary Contains the static metadata about Elide models.
+     * @return updated instance of dictionary.
+     */
+    public EntityDictionary bindAsyncDictionary(EntityDictionary dictionary) {
+
+        dictionary.bindEntity(AsyncQuery.class);
```
aklish: If we are using the JpaDataStore, these classes are automatically discovered.  There is no need to explicitly bind them.

aklish: We should be able to remove this function

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/spring/App.java
```diff
@@ -15,6 +16,7 @@
  */
 @Slf4j
 @SpringBootApplication
+@EntityScan(basePackages = {"${elide.modelPackage}"})
```
aklish: I don't think we need this.  Why did we add?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
moizarafat: We are calculating the cleanup time in elide-async based on provided maxRunTime. It's set to the maximum of 6 Hours or Twice of the maxRunTime provided by the user.  If we pass both params then we will have to do checks to make sure cleanup time is not less than the timeout.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
moizarafat: so we are using this param to randomize the start-up time for cleanup executor. If we have more than 1 hosts running and then if all of them have the cleanup service executing at the same time they could possibly try to update/delete the record at the same time and have lock error etc.  so this is how we are using it here. https://github.com/yahoo/elide/blob/c88d08a94c994f16ee51c6a51c39694411fb8c59/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java#L59-L63

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
moizarafat: https://github.com/yahoo/elide/blob/c88d08a94c994f16ee51c6a51c39694411fb8c59/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java#L63

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
moizarafat: Did this because we did not want the Async Models to be discovered/binded if the enableAsync is set to false.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/spring/App.java
moizarafat: If we don't this, the Async Models are bound by default to API and swagger. We wanted to avoid that and are explicitly telling it to bind/discover the entities in the provided package.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
moizarafat: @aklish By default Spring scans for entities in the package where App.java (with annotation `@SpringBootApplication`) will be placed.  so it will not be scanning and adding these entities automatically.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/spring/App.java
moizarafat: @aklish Since we are putting `@EntityScan(basePackageClasses = AsyncQuery.class)` in the `ElideAsyncConfiguration.java`, it overrides the default behavior of spring to scan all the entities in the package where App.java (with annotation @SpringBootApplication) will be placed. In my testing, I found that we can change `@EntityScan(basePackages = {"${elide.modelPackage}"})` to `@EntityScan` in this class and it should work fine. This will be a simple change for users that can be included in the documentation. 
The other option would be as suggested by you that we include the logic from the dynamic config design to use our own implementation of `LocalContainerEntityManagerFactoryBean` and add the Async models ourselves but it could potentially take time to implement and test out fullt. Let me know your thoughts.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncCleanupConfiguration.java
```diff
@@ -0,0 +1,40 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Async Configuration For Elide Services.  Override any of the beans (by defining your own) to change
+ * the default behavior.
+ */
+@Configuration
+@EnableConfigurationProperties(ElideConfigProperties.class)
+@ConditionalOnExpression("${elide.async.cleanupEnabled:false}")
```
aklish: This should also be conditional on async being enabled (no point to have the cleaner in that case).

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -0,0 +1,42 @@
...
+    /**
+     * Default retention of async query and results.
+     */
+    private int queryCleanupDays = 7;
+
+    /**
+     * Whether or not to use the default implementation of AsyncQueryDAO.
+     * If false, the user will provide custom implementation of AsyncQueryDAO.
+     */
+    private boolean defaultAsyncQueryDAO = true;
```
aklish: In auto-configuration, I thought this was done with ConditionalOnMissingBean annotations rather than config.

elide-spring/elide-spring-boot-autoconfigure/src/test/resources/application.yaml
```diff
@@ -13,6 +13,13 @@ elide:
     path: /doc
     enabled: true
     version: "1.0"
+  async:
+    enabled: true
```
aklish: We need a URL path for this as well.   Async should probably inherit from the other endpoint configs.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncQueryDAOConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Async Configuration For Elide Services.  Override any of the beans (by defining your own) to change
+ * the default behavior.
+ */
+@Configuration
+@EnableConfigurationProperties(ElideConfigProperties.class)
+@ConditionalOnExpression("${elide.async.defaultAsyncQueryDAO:true}")
```
aklish: ConditionalOnMissingBean should suffice.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Async Configuration For Elide Services.  Override any of the beans (by defining your own) to change
+ * the default behavior.
+ */
+@Configuration
+@EntityScan(basePackageClasses = AsyncQuery.class)
+@EnableConfigurationProperties(ElideConfigProperties.class)
+@ConditionalOnExpression("${elide.async.enabled:false}")
+public class ElideAsyncConfiguration {
```
aklish: Instead of breaking out all of these beans into separate files - why not just put them in a single ElideAsyncConfiguration class?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
moizarafat: Has moved all logic to single config.

moizarafat: If the user provides his bean we will not be able to control the injection order so we should disable our injection using this param and also be consistent with standalone where the user overrides the getAsyncQueryDAO method.

elide-spring/elide-spring-boot-autoconfigure/src/test/resources/application.yaml
moizarafat: As discussed we will leave this param as is 

> The url param does exist but I have not set it specifically in application.yaml because the swagger documentation for async models is available at /doc, and the models are also accessible via jsonapi and graphgl so i did not set it specifically to something there.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncQueryDAOConfiguration.java
moizarafat: see previous comment on AsyncProperties.java about adding `defaultAsyncQueryDAO`

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
moizarafat: Merging them into one config now

elide-spring/elide-spring-boot-autoconfigure/src/main/resources/META-INF/spring.factories
```diff
@@ -1,5 +1,7 @@
 org.springframework.boot.autoconfigure.EnableAutoConfiguration=\
  com.yahoo.elide.spring.config.ElideAutoConfiguration, \
+ com.yahoo.elide.spring.config.ElideAsyncQueryDAOConfiguration, \
```
aklish: We can remove DAOConfiguration now

PR #1205, opened 2020-02-26
Elide 5.x elide-async integration in elide-standalone

moizarafat:
Co-authored-by: Abhino <abhino@gmail.com>

Depends on: https://github.com/yahoo/elide/pull/1203

## Description
- Integrating with elide-async. 
- Changes to the methods in Util class to combine the Async models with package models.  
- Binds AsyncExecutor Service when Async is enabled.
- Adds the Async models to Swagger if Async and Swagger are enabled. The Async models are available on /swagger/doc/async path
- Additional methods added to ElideStandaloneSettings.java for configuring the Async Feature like enable/disable, number of threads for executing the async queries in parallel, the maximum run time for individual threads before they are automatically killed, number of instances hosting the elide service in parallel (to avoid all hosts trying to update status of orphan queries in parallel). 

## Motivation and Context
To allow submission of queries that will be executed asynchronously.

## How Has This Been Tested?
- Used elide-standalone-example with the updated version of the elide-standalone artifact and verify query submissions and executions. 

NA

## License

### DIFF
--- a/elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
+++ b/elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
@@ -75,7 +75,6 @@ public class Util {
         if (includeAsyncModel) {
             modelEntities.addAll(getAllEntities(AsyncQuery.class.getPackage().getName()));
         }
-
         return modelEntities;
     }
 
--- a/elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
+++ b/elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
@@ -5,6 +5,7 @@
  */
 package com.yahoo.elide.standalone;
 
+import com.yahoo.elide.async.models.AsyncQuery;
 import com.yahoo.elide.utils.ClassScanner;
 import org.hibernate.jpa.boot.internal.EntityManagerFactoryBuilderImpl;
 import org.hibernate.jpa.boot.internal.PersistenceUnitInfoDescriptor;
@@ -23,7 +24,7 @@ import javax.persistence.spi.PersistenceUnitInfo;
  */
 public class Util {
 
-    public static EntityManagerFactory getEntityManagerFactory(String modelPackageName, String asyncModelPackageName, 
+    public static EntityManagerFactory getEntityManagerFactory(String modelPackageName, boolean includeAsyncModel, 
             Properties options) {
 
         // Configure default options for example service
@@ -53,7 +54,7 @@ public class Util {
         }
 
         PersistenceUnitInfo persistenceUnitInfo = new PersistenceUnitInfoImpl("elide-stand-alone",
-                combineModelEntities(modelPackageName, asyncModelPackageName), options);
+                combineModelEntities(modelPackageName, includeAsyncModel), options);
 
         return new EntityManagerFactoryBuilderImpl(
                 new PersistenceUnitInfoDescriptor(persistenceUnitInfo), new HashMap<>())
@@ -64,15 +65,15 @@ public class Util {
      * Combine the model entities with Async model.
      *
      * @param modelPackageName Package name
-     * @param asyncModelPackageName Async model package Name
+     * @param includeAsyncModel Include Async model package Name
      * @return All entities combined from both package.
      */
-    public static List<String> combineModelEntities(String modelPackageName, String asyncModelPackageName) {
+    public static List<String> combineModelEntities(String modelPackageName, boolean includeAsyncModel) {
 
         List<String> modelEntities = getAllEntities(modelPackageName);
 
-        if (asyncModelPackageName != null) {
-            modelEntities.addAll(getAllEntities(asyncModelPackageName));
+        if (includeAsyncModel) {
+            modelEntities.addAll(getAllEntities(AsyncQuery.class.getPackage().getName()));
         }
 
         return modelEntities;
--- a/elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
+++ b/elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
@@ -23,7 +23,8 @@ import javax.persistence.spi.PersistenceUnitInfo;
  */
 public class Util {
 
-    public static EntityManagerFactory getEntityManagerFactory(String modelPackageName, Properties options) {
+    public static EntityManagerFactory getEntityManagerFactory(String modelPackageName, String asyncModelPackageName, 
+            Properties options) {
 
         // Configure default options for example service
         if (options.isEmpty()) {
@@ -52,11 +53,29 @@ public class Util {
         }
 
         PersistenceUnitInfo persistenceUnitInfo = new PersistenceUnitInfoImpl("elide-stand-alone",
-                getAllEntities(modelPackageName), options);
+                combineModelEntities(modelPackageName, asyncModelPackageName), options);
 
         return new EntityManagerFactoryBuilderImpl(
                 new PersistenceUnitInfoDescriptor(persistenceUnitInfo), new HashMap<>())
                 .build();
+    }
+
+    /**
+     * Combine the model entities with Async model.
+     *
+     * @param modelPackageName Package name
+     * @param asyncModelPackageName Async model package Name
+     * @return All entities combined from both package.
+     */
+    public static List<String> combineModelEntities(String modelPackageName, String asyncModelPackageName) {
+
+        List<String> modelEntities = getAllEntities(modelPackageName);
+
+        if (asyncModelPackageName != null) {
+            modelEntities.addAll(getAllEntities(asyncModelPackageName));
+        }
+
+        return modelEntities;
     }
 
     /**
--- a/elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
+++ b/elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
@@ -62,7 +62,7 @@ public class ElideResourceConfig extends ResourceConfig {
         register(new AbstractBinder() {
             @Override
             protected void configure() {
-                bind(Util.combineModelEntities(settings.getModelPackageName(), settings.getAsyncModelPackageName())).to(Set.class)
+                bind(Util.combineModelEntities(settings.getModelPackageName(), settings.enableAsync())).to(Set.class)
                         .named("elideAllModels");
             }
         });
--- a/elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
+++ b/elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
@@ -13,6 +13,7 @@ import com.yahoo.elide.standalone.Util;
 import com.yahoo.elide.async.models.AsyncQuery;
 import com.yahoo.elide.async.models.AsyncQueryResult;
 import com.yahoo.elide.async.service.AsyncExecutorService;
+import com.yahoo.elide.async.service.AsyncCleanerService;
 import com.yahoo.elide.contrib.swagger.SwaggerBuilder;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.health.HealthCheckRegistry;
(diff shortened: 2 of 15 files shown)


elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -179,6 +195,42 @@ default boolean enableJSONAPI() {
...
+    default Integer getMaxRunTime() {
+        return 60;
+    }
+
+    /**
+     * Number of hosts running Elide Service.
+     *
+     * @return Default: 1
+     */
+    default Integer getNumberOfHosts() {
```
aklish: I don't think we need this.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -179,6 +195,42 @@ default boolean enableJSONAPI() {
...
+    default Integer getAsyncThreadSize() {
+        return 5;
+    }
+
+    /**
+     * Maximum Query Run time for Async Queries to mark as TIMEDOUT.
+     *
+     * @return Default: 60
+     */
+    default Integer getMaxRunTime() {
```
aklish: See comments I Had on Spring in general.  Settings that involve time should have units in their name.  We need two time settings (timeout and delete).

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -131,6 +132,21 @@ default int getPort() {
...
+     * recursively scanned for @Entity's and registered with Elide.
+     *
+     * NOTE: This will scan for all entities in that package and bind this data to a set named "elideAllModels".
+     *       If providing a custom ElideSettings object, you can inject this data into your class by using:
+     *
+     *       <strong>@Inject @Named("elideAllModels") Set&lt;Class&gt; entities;</strong>
+     *
+     * @return
+     */
+    default String getAsyncModelPackageName() {
```
aklish: We don't need a setting for this.  The package name is fixed (by us).

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -23,7 +23,8 @@
  */
 public class Util {
 
-    public static EntityManagerFactory getEntityManagerFactory(String modelPackageName, Properties options) {
+    public static EntityManagerFactory getEntityManagerFactory(String modelPackageName, String asyncModelPackageName,
```
aklish: Instead of passing the package name for async here (which Elide already knows), you could just pass a flag about whether or not to include async models.

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -52,13 +53,31 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
+     * @param modelPackageName Package name
+     * @param asyncModelPackageName Async model package Name
+     * @return All entities combined from both package.
+     */
+    public static List<String> combineModelEntities(String modelPackageName, String asyncModelPackageName) {
+
+        List<String> modelEntities = getAllEntities(modelPackageName);
+
+        if (asyncModelPackageName != null) {
+            modelEntities.addAll(getAllEntities(asyncModelPackageName));
```
aklish: asyncModelPackageName can just be AsyncQuery.class.getPackage().getName()

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -58,7 +63,8 @@ public ElideResourceConfig(ServiceLocator injector, @Context ServletContext serv
         register(new AbstractBinder() {
             @Override
             protected void configure() {
-                bind(Util.getAllEntities(settings.getModelPackageName())).to(Set.class).named("elideAllModels");
+                bind(Util.combineModelEntities(settings.getModelPackageName(), settings.getAsyncModelPackageName())).to(Set.class)
```
aklish: See my earlier comment here.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -64,7 +65,7 @@
      */
     default ElideSettings getElideSettings(ServiceLocator injector) {
         EntityManagerFactory entityManagerFactory = Util.getEntityManagerFactory(getModelPackageName(),
-                getDatabaseProperties());
+                getAsyncModelPackageName(), getDatabaseProperties());
```
aklish: See my earlier comment here.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
moizarafat: We are using this param to randomize the start-up time for the cleanup executor. If we have more than 1 host running and then if all of them have the cleanup service executing at the same time they could possibly try to update/delete the record at the same time and have lock error etc.  so this is how we are using it here. https://github.com/yahoo/elide/blob/c88d08a94c994f16ee51c6a51c39694411fb8c59/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java#L59-L63

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
moizarafat: Done

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
moizarafat: Done

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
moizarafat: Done

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
moizarafat: Done

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
moizarafat: Done

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
moizarafat: Done

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
moizarafat: done

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -167,6 +169,60 @@ default boolean enableJSONAPI() {
...
+    default Integer getAsyncThreadSize() {
+        return 5;
+    }
+
+    /**
+     * Maximum Query Run time for Async Queries to mark as TIMEDOUT.
+     *
+     * @return Default: 60
+     */
+    default Integer getMaxRunTimeMinutes() {
```
aklish: This should be qualified with Async

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -167,6 +169,60 @@ default boolean enableJSONAPI() {
...
+        return 60;
+    }
+
+    /**
+     * Number of days history to retain for async query executions and results.
+     *
+     * @return Default: 7
+     */
+    default Integer getQueryCleanupDays() {
```
aklish: Same comment.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -167,6 +169,60 @@ default boolean enableJSONAPI() {
...
+        return 7;
+    }
+
+    /**
+     * Implementation of AsyncQueryDAO to use.
+     *
+     * @return AsyncQueryDAO type object.
+     */
+    default AsyncQueryDAO getAsyncQueryDAO() {
```
aklish: Instead of resolving this as a setting - this can be injected into the async code (either via lifecycle hook or bean).

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -85,6 +112,21 @@ protected void configure() {
...
+                    if(settings.enableAsync()) {
+                        EntityDictionary dictionary = new EntityDictionary(new HashMap());
+                        dictionary.bindEntity(AsyncQuery.class);
+                        dictionary.bindEntity(AsyncQueryResult.class);
+                         
+                        Info info = new Info().title("Async Service").version("1.0");
+
+                        SwaggerBuilder builder = new SwaggerBuilder(dictionary, info);
+
+                        Swagger swagger = builder.build().basePath("/api/v1");
```
aklish: Most of the other APIs have a config setting for the base path.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -76,6 +84,25 @@ protected void configure() {
                 bind(elideSettings).to(ElideSettings.class);
                 bind(elideSettings.getDictionary()).to(EntityDictionary.class);
                 bind(elideSettings.getDataStore()).to(DataStore.class).named("elideDataStore");
+
+                // Binding async service
+                if(settings.enableAsync()) {
+                    AsyncQueryDAO asyncQueryDao = settings.getAsyncQueryDAO();
```
aklish: Just create the DAO here.  We don't need a setting.  Users can override this function if they want something different.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
moizarafat: As discussed no config setting exists will add a new one.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
moizarafat: As discussed will leave it as is.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
moizarafat: As discussed will leave this as is.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -211,17 +211,26 @@ default Integer getMaxRunTimeMinutes() {
...
+    default Integer getAsyncQueryCleanupDays() {
         return 7;
     }
 
+    /**
+     *  Root path specification for the Async Models Swagger endpoint.
+     *
+     * @return Default: /api/v1
+     */
+    default String getAsyncSwaggerPath() {
```
aklish: Instead of making this the async swagger path - how about a config setting for swagger in general (including non-asyc models)?

aklish: This should just come from getJsonApiPathSpec()

PR #1210, opened 2020-03-04
Elide 5.x- Dynamic Config

### LINKED ISSUE #1178: RFC : Elide Dynamic Model Configuration (Tables Only)
## Overview

## Location of Models

```
models
└───entities
    │   entity1.hjson  
└───tables
    │   table1.hjson
    │   table2.hjson
└───variables.hjson
└───security.hjson
└───models.asc
```

## Model Format

### Example

```
{
  name: PlayerStats
  table: player_stats
  schema: gamedb
  description:
  '''
  A long description
  '''
  cardinality : large

  joins: [
      {  
         name: playerCountry
         to: country
         type: toOne
      }
  ]

  measures : [
     {  
        name : highScore
        type : number
        definition: 'MAX(score)'
     }
  ]

  dimensions : [
     { 
       name : countryCode
       type : text
     },
     { 
       name : createdOn
       type : time 
       definition : create_on
       grains: [{{hour}}, {{day}}]
     }
  ]
}
```

### Table Properties

### Joins

`${to|from}.columnName`

### Measures

`definition: 'SUM(column)'`

1. by name: `column`

### Dimensions

`definition: createdOn`

1. by name: `column`

### Time Dimensions

```
{
   name : createdOn
   definition : create_on
   type: time
}
```

```
{

   grain: month
}
```

## Metric and dimension types

| Type | Applies To |
| ---- | ----------- |
| time | Dimension  |
| text | Dimension  |
| coordinate | Dimension |
| boolean | Dimension |

## Variable Substitution

```
{
   foo: [1, 2, 3]
   bar: blah
}
```

### Permission Rules

```
{
    rules: [
        {
            name: User belongs to company
        },
        {
            filter: id==${principal.id}
            name: Principal is owner
        },
    ]
}
```

1. Minimal restrictions:
  - Quotations will be escaped
2. Names only

## Validation

## Caveats 

moizarafat:
Authored-by: Ayeswarya
Authored-by: AvaniMakwana

## Description
JSON schema representing the config to be provided by users representing their Models.

## Motivation and Context
Please refer #1178 

## How Has This Been Tested?

## License


elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideSecuritySchema.json
```diff
@@ -0,0 +1,52 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "$id": "http://elide.io/schemas/security_schema_v1#",
+    "description": "Elide Security config json/hjson schema",
+    "type": "object",
+    "properties": {
+       "roles": {
+               "title": "Security Roles",
```
aklish: Spacing is off here.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,296 @@
...
+          "title": "Metric definition",
+          "description": "The definition of the metric",
+          "type": "string"
+        },
+        "type": {
+          "title": "Field type",
+          "description": "The data type of the field",
+          "type": "string",
+          "enum": [
+            "time",
```
aklish: This needs to stay in sync with:
https://github.com/yahoo/elide/issues/1005

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,296 @@
...
+          "default": "number"
+        }
+      },
+      "required": [
+        "name",
+        "definition"
+      ]
+    },
+    "dimension": {
+      "title": "Dimension",
```
aklish: Dimension needs to be a union of different types (dimension and timedimension are two different types).  Time dimension has all of the fields of dimension though.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,296 @@
...
+            "type": "string"
+          },
+          "default": []
+        },
+        "grains": {
+          "title": "Dimension grains",
+          "description": "Time definition granularity",
+          "type": "array",
+          "items": {
+            "type": "string"
```
aklish: Grains are objects

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,296 @@
...
+            "pattern": "^[A-Z][A-Za-z]*$"
+          },
+          "schema": {
+            "title": "Table Schema",
+            "description": "The database or schema where the model lives.",
+            "type": "string",
+            "pattern": "^([A-Za-z]*_?[A-Za-z]*)*$",
+            "default": ""
+          },
+          "table": {
```
aklish: The table should have one of (sql, table) but not both.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,296 @@
...
+            "uniqueItems": true,
+            "items": {
+              "type": "string"
+            },
+            "default": []
+          }
+        },
+        "required": [
+          "name",
+          "dimensions"
```
aklish: one of (sql | table) is also required.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,296 @@
...
+            "title": "Table SQL",
+            "description": "SQL query which is used to populate the table.",
+            "type": "string",
+            "default": ""
+          },
+          "description": {
+            "title": "Table Model description",
+            "description": "A long description of the model.",
+            "type": "string",
+            "default": "[$name]"
```
aklish: I haven't seen this before.  Does this actually reference the name?  Can you point me to the docs for this? Thanks.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
ayeswarya: The values in #1005 do not have number type. Number was specified as default in https://github.com/yahoo/elide/issues/1178. What should be the default for the new types. thanks

moizarafat: Integer should be used as default

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
ayeswarya: I read about it in a custom Json implementation and tried to use it. However turns out there is no way to do that as of now. So it is going to be handled in the code.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
ayeswarya: Added this change using oneOf and also added extends

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
ayeswarya: Grains are objects now with grain as enum

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,435 @@
...
+          "description": "Read permission for the metric.",
+          "type": "string",
+          "default": "Allow All"
+        },
+        "definition": {
+          "title": "Metric definition",
+          "description": "The definition of the metric",
+          "type": "string"
+        },
+        "type": {
```
aklish: We should probably break this out as a separate type and then use JSON scheme references to reference it.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,435 @@
...
+          "description": "The definition of the dimension",
+          "type": "string",
+          "default": ""
+        },
+        "type": {
+          "title": "Dimension field type",
+          "description": "The data type of the dimension field",
+          "type": "string",
+          "enum": [
+            "text",
```
aklish: This list of types is smaller than the measure list of types.  Dimension should have all types available to them. If we are going to restrict types, we'd restrict the measure types.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,435 @@
...
+            "coordinate",
+            "number",
```
aklish: Number is no longer a type - we now have integer and decimal or something along those lines.  

Check out ValueType in Elide.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,435 @@
...
+        "definition"
+      ],
+      "additionalProperties": false
+    },
+    "timeDimension": {
+      "title": "Time Dimension",
+      "description": "Time Dimensions represent labels for measures. Dimensions are used to filter and group measures.",
+      "type": "object",
+      "properties": {
+        "name": {
```
aklish: Rather than redefining the same fields in multiple places, you can define them once and then use references and _allOf_  to create something that looks like inheritance.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,435 @@
...
+              "$ref": "#/definitions/measure"
+            },
+            "default": []
+          },
+          "dimensions": {
+            "title": "Table dimensions",
+            "description": "One or more dimension definitions.",
+            "type": "array",
+            "items": {
+              "oneOf": [
```
aklish: Should this be anyOf?

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideVariableSchema.json
```diff
@@ -0,0 +1,17 @@
+{
+    "$schema": "http://json-schema.org/schema#",
+    "$id": "http://elide.io/schemas/variable_schema_v1#",
```
aklish: Should the ids be https or http?

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideVariableSchema.json
```diff
@@ -0,0 +1,17 @@
...
+    "description": "Elide Variable config json/hjson schema",
+    "type": "object",
+    "patternProperties": {
+        "^([A-Za-z]*_?[A-Za-z]*)*$": {
+          "anyOf": [
+            {"type": "string"},
+            {"type": "array"},
```
aklish: What about an object?

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
ayeswarya: anyOf is for atleast one, since the json/hjson can only match one or the other only one (one of ) works for this case.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/ElideHjsonUtil.java
```diff
@@ -0,0 +1,177 @@
...
+            return false;
+        }
+    }
+
+    /**
+     * Null check for the string.
+     * @param input string to validate
+     * @return whether string is null or not
+     */
+    public static boolean isNull(String input) {
```
moizarafat: Lets rename this to isNullOrEmpty. Also we could just do 
`return (input == null || input.trim().length() == 0)` instead `return (input == null || input.trim().length() == 0) ? true : false`

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideSecurityToPojo.java
```diff
@@ -0,0 +1,72 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * The ElideSecurityToPojo class allows conversion of Security config Hjson 
+ * format to a Elide Security POJO.
+ */
+@Slf4j
+public class ElideSecurityToPojo {
```
moizarafat: There is a lot of duplicate code between `ElideSecurityToPojo`, `ElideTableToPojo`, `ElideVariableToPojo`. I think we can have a single class like `ElideConfigToPojo` with all 3 methods taking 1 additional argument `String schemaType`. The parse method can then can do a `switch` statement  to call `ObjectMapper().readValue` with appropriate `.class` to transform to.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/ElideHjsonUtil.java
```diff
@@ -0,0 +1,177 @@
...
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+/**
+ * Util class for HJSON parsing
+ */
+@Slf4j
+public class ElideHjsonUtil {
```
moizarafat: Thoughts on changing this to just `util.java` so that we can expand it for including any other methods to be used in Spring and Standalone. I think the HTTP reader and File Reader will be needed in Spring and Standalone as well.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/Util.java
```diff
@@ -0,0 +1,144 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URL;
+import java.nio.charset.Charset;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Map;
+
```
moizarafat: Please include JavaDoc for class and methods.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,65 @@
...
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.contrib.dynamicconfighelpers.parser;
+
+import com.yahoo.elide.contrib.dynamicconfighelpers.Util;
+
+import org.json.JSONObject;
+
+import lombok.extern.slf4j.Slf4j;
+
```
moizarafat: Please add Javadoc for the class.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,65 @@
...
+@Slf4j
+public class ElideConfigParser {
+
```
moizarafat: Please add Java Doc.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,65 @@
...
+            return Util.getModelPojo(inputConfigType, jsonConfig);
+        }
+        else {
+            log.error(Util.INVALID_ERROR_MSG);
+            return null;
+        }
+    }
+
+    /**
+     * Pass File path containing hjson config; http or local.
```
moizarafat: Should `Pass` be `Parse`?

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-checkstyle-plugin</artifactId>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>1.8</source>
```
moizarafat, line 148: Please change to $maven.compiler.source

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+                    <target>1.8</target>
```
moizarafat, line 149: Please change to maven.compiler.target

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/model/Measure.java
```diff
@@ -0,0 +1,65 @@
...
+    private String description;
+
+    @JsonProperty("category")
+    private String category;
+
+    @JsonProperty("hidden")
+    private Boolean hidden = false;
+
+    @JsonProperty("readAccess")
+    private String readAccess = "Allow All";
```
moizarafat, line 49: Please change to `Prefab.Role.All`

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/model/Table.java
```diff
@@ -0,0 +1,137 @@
...
+    private Boolean hidden = false;
+
+    @JsonProperty("description")
+    private String description;
+
+    @JsonProperty("cardinality")
+    private Table.Cardinality cardinality = Table.Cardinality.fromValue("tiny");
+
+    @JsonProperty("readAccess")
+    private String readAccess = "Allow All";
```
moizarafat, line 68: Please change to `Prefab.Role.All`

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/model/Dimension.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @JsonProperty("category")
+    private String category;
+
+
+    @JsonProperty("hidden")
+    private Boolean hidden = false;
+
+    @JsonProperty("readAccess")
+    private String readAccess = "Allow All";
```
moizarafat, line 59: Please change to `Prefab.Role.All`

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,409 @@
...
+          "title": "Hide/Show measure",
+          "description": "Whether this metric is exposed via API metadata",
+          "type": "boolean",
+          "default": false
+        },
+        "readAccess": {
+          "title": "Measure read access",
+          "description": "Read permission for the metric.",
+          "type": "string",
+          "default": "Allow All"
```
moizarafat, line 118: Please change to `Prefab.Role.All`

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,409 @@
...
+          "title": "Hide/Show dimension",
+          "description": "Whether this dimension is exposed via API metadata",
+          "type": "boolean",
+          "default": false
+        },
+        "readAccess": {
+          "title": "Dimension read access",
+          "description": "Read permission for the dimension.",
+          "type": "string",
+          "default": "Allow All"
```
moizarafat, line 173: Please change to `Prefab.Role.All`

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,409 @@
...
+              "large",
+              "huge"
+            ],
+            "default": "tiny"
+          },
+          "readAccess": {
+            "title": "Table read access",
+            "description": "Read permission for the table.",
+            "type": "string",
+            "default": "Allow All"
```
moizarafat, line 310: Please change to `Prefab.Role.All`

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <source>${maven.compiler.source}</source>
+                    <target>${maven.compiler.target}</target>
+                </configuration>
+            </plugin>
+            <plugin>
```
aklish, line 152: I feel like this pom could/should inherit most of the plugins from the parent pom.   My guess is whatever this was copied from also needs to be fixed.

aklish: In general, I would gut this to only what is needed and start adding back dependencies and plugins until we get to minimal form.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+@Slf4j
+/**
+ * Util class for Dynamic config helper module.
+ */
+public class DynamicConfigHelpersUtil {
```
aklish, line 38: HelpersUtil seems redundant.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,103 @@
...
+    private DynamicConfigHelpersUtil util = new DynamicConfigHelpersUtil();
+
+    /**
+     * Parse File path containing hjson config; http or local.
+     * @param localFilePath : File Path to hjson config
+     * @throws NullPointerException
+     */
+    public void parseConfigPath(String localFilePath) {
+        try {
+      if (util.isNullOrEmpty(localFilePath)) {
```
aklish, line 42: Spacing is off.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,103 @@
...
+
+        //table
+        populateTablesPojo(util.getJsonConfig(localConfigPath, DynamicConfigHelpersUtil.SCHEMA_TYPE_TABLE));
+    }
+
+    @SuppressWarnings("unchecked")
+    private void populatePojo(String jsonConfig, String configType) throws Exception {
+        switch (configType) {
+        case DynamicConfigHelpersUtil.SCHEMA_TYPE_VARIABLE:
+      this.variables = (Map<String, Object>) parseJsonConfig(jsonConfig,
```
aklish, line 69: Spacing is off.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/TableSchemaValidationTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+    private InputStream loadStreamFromClasspath(String resource) throws Exception {
+        return TableSchemaValidationTest.class.getResourceAsStream(resource);
+    }
+
+    private Reader loadReaderFromClasspath(String resource) throws Exception {
+        return new InputStreamReader(loadStreamFromClasspath(resource));
+    }
+
+    private JsonNode loadJsonFromClasspath(String resource, boolean translate) throws Exception {
```
aklish, line 71: These functions are duplicated in several places.  Let's consolidate them - maybe into a base class that these test classes can inherit from.   Another alternative is to put them in a utility class.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+
+    private static final String CHAR_SET = "UTF-8";
+    private static final String NEW_LINE = "\n";
+
+    /**
+     * Converts hjson input string to valid json string.
+     * @param hjson : input config in hjson format
+     * @return valid json string
+     */
+    public String hjsonToJson(String hjson) {
```
aklish, line 61: Add static.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+     */
+    public boolean isConfigValid(JSONObject schemaObj, String jsonConfig) {
+
+        try {
+            JSONObject data = new JSONObject(new JSONTokener(jsonConfig));
+            Schema schema = SchemaLoader.load(schemaObj);
+            schema.validate(data);
+            return true;
+        } catch (Exception e) {
+            e.printStackTrace();
```
aklish, line 79: Log the exception (using `@Slf4j`) - don't print.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+     * @return true or false
+     */
+    public boolean isConfigValid(JSONObject schemaObj, String jsonConfig) {
+
+        try {
+            JSONObject data = new JSONObject(new JSONTokener(jsonConfig));
+            Schema schema = SchemaLoader.load(schemaObj);
+            schema.validate(data);
+            return true;
+        } catch (Exception e) {
```
aklish, line 78: Generally not a good practice to catch Exception (which can swallow runtime exceptions) except in uncaught handlers.  Let's be specific for what we expect to blow up.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+            return false;
+        }
+    }
+
+    /**
+     * Checks whether input is null or empty.
+     * @param input : input string
+     * @return true or false
+     */
+    public boolean isNullOrEmpty(String input) {
```
aklish, line 90: static

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+        return JsonValue.readHjson(hjson).toString();
+    }
+
+    /**
+     * Validates json config against json schema.
+     * @param schemaObj : schema object
+     * @param jsonConfig : config to be validated against schema
+     * @return true or false
+     */
+    public boolean isConfigValid(JSONObject schemaObj, String jsonConfig) {
```
aklish, line 71: I believe all of the methods in this class can be static.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+        case SCHEMA_TYPE_TABLE:
+            return loadSchema(ELIDE_TABLE_VALIDATION_SCHEMA);
+
+        case SCHEMA_TYPE_SECURITY:
+            return loadSchema(ELIDE_SECURITY_SCHEMA);
+
+        case SCHEMA_TYPE_VARIABLE:
+            return loadSchema(ELIDE_VARIABLE_SCHEMA);
+        default:
+            return null;
```
aklish, line 112: Throw illegalArgumentExctpion.  This shouldn't happen.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /**
+     * Converts json string to appropriate POJO.
+     * @param inputConfigType ex. table, security or variable
+     * @param jsonConfig input json config
+     * @return config pojo
+     * @throws Exception
+     */
+    public Object getModelPojo(String inputConfigType, String jsonConfig) throws Exception {
```
aklish, line 123: You can use Java Generics here so this class will return the correct type (instead of Object).  Then the calling code won't need to cast.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+import java.util.List;
+import java.util.Map;
+
+@Slf4j
+/**
+ * Util class for Dynamic config helper module.
+ */
+public class DynamicConfigHelpersUtil {
+
+    public static final String SCHEMA_TYPE_TABLE = "table";
```
aklish, line 40: Instead of making the SCHEMA_TYPE strings - make an enum that takes a constructor argument which is the POJO class you expect to deserialize.  Then you have a nice mapping between the enum name and the POJO class and you don't need so many switch statements below.

aklish: In the same enum, you can tie together three bits of information - a friendly name, the POJO class to deserialize, and the JSON file where the schema lives.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
```diff
@@ -0,0 +1,217 @@
...
+            configs.add(hjsonToJson(readConfigFile(basePath  + SECURITY_CONFIG_PATH)));
+            return configs;
+
+        case SCHEMA_TYPE_TABLE:
+            return getTablesConfig(basePath + TABLE_CONFIG_PATH);
+        }
+        return null;
+    }
+
+    private static String readFileContent(BufferedReader reader) throws IOException {
```
aklish, line 169: There are libraries that can do this already.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/model/Rule.java
```diff
@@ -0,0 +1,41 @@
...
+    "name"
+})
+@Data
+@EqualsAndHashCode()
+@AllArgsConstructor
+@NoArgsConstructor
+public class Rule {
+
+    @JsonProperty("type")
+    private String type;
```
aklish, line 34: This could be an enum.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Parse File path containing hjson config; http or local.
+     * @param localFilePath : File Path to hjson config
+     * @throws NullPointerException
+     */
+    public void parseConfigPath(String localFilePath) {
+        try {
+      if (util.isNullOrEmpty(localFilePath)) {
+    throw new NullPointerException("Config path is null");
```
aklish, line 43: IllegalArgumentException

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,103 @@
...
+      }
+      parse(localFilePath);
+        } catch (Exception e) {
+      e.printStackTrace();
```
aklish, line 47: Don't print stack trace.  Don't catch Exception (too broad).

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,103 @@
...
+      break;
+
+        case DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY:
+      this.elideSecurityConfig = (ElideSecurityConfig) parseJsonConfig(jsonConfig,
+        DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY);
+      break;
+        }
+    }
+
+    private void populateTablesPojo(List<String> tablesJson) throws Exception {
```
aklish, line 80: Instead of setting a private member here (ideally all private members are initialized in the constructor), it might make more sense to return a List or Set of ElideTableConfig from this function (they are never used outside the function).

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,103 @@
...
+      ElideTableConfig table = (ElideTableConfig) parseJsonConfig(tableJson,
+              DynamicConfigHelpersUtil.SCHEMA_TYPE_TABLE);
+      tables.addAll(table.getTables());
+        }
+        this.elideTableConfig = new ElideTableConfig();
+        this.elideTableConfig.setTables(tables);
+    }
+
+
+    private Object parseJsonConfig(String jsonConfig, String inputConfigType) throws Exception {
```
aklish, line 92: Try to use generics instead of Object.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,103 @@
...
+        JSONObject configType = util.schemaToJsonObject(inputConfigType);
+
+        if (configType != null && util.isConfigValid(configType, jsonConfig)) {
+      return util.getModelPojo(inputConfigType, jsonConfig);
+        }
+        else {
+      log.error(DynamicConfigHelpersUtil.INVALID_ERROR_MSG);
+      return null;
```
aklish, line 100: If this is really an error throw an exception.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -0,0 +1,103 @@
...
+        //security
+        String securityJson = util.getJsonConfig(localConfigPath, DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY).get(0);
+        populatePojo(securityJson, DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY);
+
+        //table
+        populateTablesPojo(util.getJsonConfig(localConfigPath, DynamicConfigHelpersUtil.SCHEMA_TYPE_TABLE));
+    }
+
+    @SuppressWarnings("unchecked")
+    private void populatePojo(String jsonConfig, String configType) throws Exception {
```
aklish, line 66: Similar comments - I don't think we should ideally be setting state outside a constructor.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideSecuritySchema.json
```diff
@@ -0,0 +1,52 @@
...
+                "properties": {
+                    "type": {
+                        "title": "Rule Type",
+                        "description": "Type of security rule",
+                        "type": "string"
+                    },
+                    "filter": {
+                        "title": "Rule Filter",
+                        "description": "Rule filter expression",
+                        "type": "string"
```
aklish, line 31: Let's make this an enum (for now with just one value - filter).

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideSecuritySchema.json
```diff
@@ -0,0 +1,52 @@
...
+                "required": [
+                    "filter",
+                    "name"
+                ],
+                "additionalProperties": false
+            }
+        }
+    },
+    "required": [
+        "roles",
```
aklish, line 48: Security sections are optional overall (the entire file can also be missing).

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,409 @@
...
+        "name": {
+          "title": "Join name",
+          "description": "The name of the join relationship.",
+          "type": "string"
+        },
+        "to": {
+          "title": "Join table name",
+          "description": "The name of the table that is being joined to",
+          "type": "string",
+          "pattern": "^([A-Za-z]*_?[A-Za-z]*)*$"
```
aklish, line 48: I believe table names can have numbers too

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,409 @@
...
+    "measure": {
+      "title": "Measure",
+      "description": "Metric definitions are extensible objects that contain a type field and one or more additional attributes. Each type is tied to logic in Elide that generates a metric function.",
+      "type": "object",
+      "properties": {
+        "name": {
+          "title": "Metric name",
+          "description": "The name of the metric. This will be the same as the POJO field name.",
+          "type": "string",
+          "pattern": "^([A-Za-z]*_?[A-Za-z]*)*$"
```
aklish, line 95: Metric names can include numbers.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
aklish: :+1:

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,409 @@
...
+    "dimensionRef": {
+      "title": "Dimension",
+      "description": "Dimensions represent labels for measures. Dimensions are used to filter and group measures.",
+      "type": "object",
+      "properties": {
+        "name": {
+          "title": "Dimension name",
+          "description": "The name of the dimension. This will be the same as the POJO field name.",
+          "type": "string",
+          "pattern": "^([A-Za-z]*_?[A-Za-z]*)*$"
```
aklish, line 150: Include numbers.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,409 @@
...
+        },
+        "description": {
+          "title": "Dimension description",
+          "description": "A long description of the dimension.",
+          "type": "string",
+          "default": "The dimension name"
```
aklish, line 156: Remove the default - it's not the actual default.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
aklish: :+1:

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,409 @@
...
+      "items": {
+        "title": "Elide Table Models",
+        "description": "Array of elide Table Models",
+        "type": "object",
+        "properties": {
+          "name": {
+            "title": "Table Model Name",
+            "description": "The name of the model. This will be the same as the POJO class name.",
+            "type": "string",
+            "pattern": "^[A-Z][A-Za-z]*$"
```
aklish, line 272: name and schema I believe can also include numbers in all but the first character.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
```diff
@@ -0,0 +1,409 @@
...
+            "title": "Hide/Show Table",
+            "description": "Whether this table is exposed via API metadata",
+            "type": "boolean",
+            "default": false
+          },
+          "description": {
+            "title": "Table Model description",
+            "description": "A long description of the model.",
+            "type": "string",
+            "default": "The model name"
```
aklish, line 291: Remove default here.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/elideTableSchema.json
aklish: :+1:

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.everit.json</groupId>
+            <artifactId>org.everit.json.schema</artifactId>
```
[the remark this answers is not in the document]
AvaniMakwana: for validating json config against schema

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java
moizarafat: wont be needed due to refactoring of the class.

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -42,20 +42,26 @@
...
-        <everit.version>1.0.0</everit.version>
         <jackson.version>2.10.2</jackson.version>
         <commons-lang3.version>3.0</commons-lang3.version>
+        <commons-io.version>2.6</commons-io.version>
         <slf4j-api.version>1.7.26</slf4j-api.version>
-        <jsonschema2pojo-core.version>1.0.2</jsonschema2pojo-core.version>
         <lombok.version>1.18.10</lombok.version>
+        <handlebars.version>4.2.0</handlebars.version>
         <json-schema-validator.version>2.2.12</json-schema-validator.version>
+        <junit.jupiter.version>5.6.0</junit.jupiter.version>
```
aklish, line 52: We should inherit the JUnit versions from the parent pom (Junit should be the same across all maven modules).

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -42,20 +42,26 @@
         <maven.compiler.source>1.8</maven.compiler.source>
         <maven.compiler.target>1.8</maven.compiler.target>
         <hjson.version>3.0.0</hjson.version>
-        <everit.version>1.0.0</everit.version>
         <jackson.version>2.10.2</jackson.version>
         <commons-lang3.version>3.0</commons-lang3.version>
+        <commons-io.version>2.6</commons-io.version>
```
aklish, line 47: Any version that is in the parent pom shouldn't be redefined here - but rather inherited from the parent.    Take a look and remove all that fit that description.  That should include lombok, slf4j, and others.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -0,0 +1,124 @@
...
+        return (input == null || input.trim().length() == 0);
+    }
+
+    /**
+     * format config file path.
+     * @param basePath : path to hjson config.
+     * @return formatted file path.
+     */
+    public static String formatFilePath(String basePath) {
+        if (!basePath.endsWith("/")) {
```
aklish, line 55: Using "/" directly is not portable.  Java has a system method to get the directory/path separator character.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -5,99 +5,48 @@
...
 
-        if (configType != null && util.isConfigValid(configType, jsonConfig)) {
-      return util.getModelPojo(inputConfigType, jsonConfig);
-        }
-        else {
-      log.error(DynamicConfigHelpersUtil.INVALID_ERROR_MSG);
-      return null;
+        } catch (JsonProcessingException e) {
+            log.error("Error while parsing dynamic config at location " + localFilePath);
+            log.error(e.getMessage());
```
aklish, line 49: If the config format is busted, I wonder if we should throw an IllegalStateException here.

elide-contrib/elide-dynamic-config-helpers/src/test/resources/elideSecuritySchema.json
```diff
@@ -44,9 +44,5 @@
             }
         }
     },
-    "required": [
```
aklish: Why are we duplicating this file in the test resources?  Why not just use the one in the main resources?

moizarafat:
Please remove elide-contrib/.DS_Store

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * converts security.hjson to ElideSecurityConfig Pojo.
+     * @param basePath : root path to model dir.
+     * @return ElideSecurityConfig Pojo
+     * @throws JsonProcessingException
+     */
+    public static ElideSecurityConfig getElideSecurityPojo(String basePath) throws JsonProcessingException {
+        String jsonConfig = hjsonToJson(readConfigFile(new File(basePath + SECURITY_CONFIG_PATH)));
```
moizarafat, line 99: we can do a check here instead if the file exists or does not exists and log message with results. Then make a call to readConfig method. Right now readConfigFile will return with a exception if file does not exist and print an error message which could be misleading. Same is applicable for Variables.

elide-contrib/elide-dynamic-config-helpers/pom.xml
moizarafat: Junit Platform is only added in dynamic config so that version is included here.

aklish: We can move platform to the parent pom.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -31,11 +31,14 @@
  */
 public class DynamicConfigHelpers {
 
-    private static final String TABLE_CONFIG_PATH = "tables/";
+    private static String TABLE_CONFIG_PATH = "tables/";
     private static final String SECURITY_CONFIG_PATH = "security.hjson";
     private static final String VARIABLE_CONFIG_PATH = "variables.hjson";
     private static final String NEW_LINE = "\n";
 
+    public static void setTableConfigPath(String input) {
```
aklish, line 39: Instead of using statics (which are hard to test), the config path should be passed in as an argument.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersTest.java
```diff
@@ -0,0 +1,73 @@
...
+        DynamicConfigHelpers.setTableConfigPath("valid/");
+        String path = "src/test/resources/tables";
+        File file = new File(path);
+        String absolutePath = file.getAbsolutePath();
+        try {
+            ElideTableConfig config =  DynamicConfigHelpers.getElideTablePojo(
+                    DynamicConfigHelpers.formatFilePath(absolutePath));
+            assertNotNull(config);
+        } catch (JsonProcessingException e) {
+            log.error(e.getMessage());
```
aklish, line 70: By catching the exception, we are not failing the test.  Tests like this should just propagate all unexpected exceptions.

PR #1220, opened 2020-03-18
Elide 5.x dynamic config handlebars hydration

### LINKED ISSUE #1178: RFC : Elide Dynamic Model Configuration (Tables Only)

moizarafat:
Authored-by: rishi-aga

## Description
Hydrate the HandleBar Template for Table/Security with hjson parameters.

## Motivation and Context
Please refer #1178 

## How Has This Been Tested?

## License


elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,83 @@
...
+@ToString
+@Entity
+@Data
+{{#if table}}@FromTable(name = "{{table}}"){{/if}}
+public class {{#capitalizeFirstLetter name}}{{/capitalizeFirstLetter}} {
+
+    @Id
+    private String name;
+
+{{#each dimensions}}
```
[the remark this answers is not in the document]
aklish: Each dimension is missing the `@Meta` annotation.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,83 @@
...
+    private {{#getDimensionType this}}{{/getDimensionType}} {{name}};
+{{/each}}
+
+{{#each joins}}
```
aklish: We generally shouldn't need getter & setter methods anywhere in the generated class files.  We should put all the annotations on the fields.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,83 @@
...
+    private {{#capitalizeFirstLetter to}}{{/capitalizeFirstLetter}} {{ name }};
```
aklish: So the join type is either toOne or toMany.  What is generated above is correct for toOne.  We are not correctly handling the toMany case.

For toMany - this will be a collection (Set) of the type.  For example, (book.authors) would have a `Set<Author>` objects instead of just Author.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHelper.java
```diff
@@ -0,0 +1,246 @@
...
+     * @param dim Elide dimension object
+     * @param options options object with type/string to match
+     * @return template if matched
+     * @throws IOException IOException
+     */
+    public CharSequence ifDimTypeMatches(Dimension dim, Options options) throws IOException {
+
+        String dimDataType = dim.getType().toString();
+        String typeToMatch = options.param(0, null);
+        boolean result = new EqualsBuilder().append(dimDataType, typeToMatch).isEquals();
```
aklish: Why not just dimDataType.equals(typeToMatch)?

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHelper.java
```diff
@@ -0,0 +1,246 @@
...
+            case INTEGER:
+                return LONG;
+            case TEXT:
+                return STRING;
+            case TIME:
+                return DATE;
+            case DECIMAL:
+                return DOUBLE;
+            case MONEY:
+                return DOUBLE;
```
aklish: BigDecimal.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHelper.java
```diff
@@ -0,0 +1,246 @@
...
+            case COORDINATE:
+                return STRING;
+            case INTEGER:
+                return LONG;
+            case TEXT:
+                return STRING;
+            case TIME:
+                return DATE;
+            case DECIMAL:
+                return DOUBLE;
```
aklish: BigDecimal.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHelper.java
```diff
@@ -0,0 +1,246 @@
...
+        return String.join(UNDERSCORE,
+                Arrays.asList(nameSplit).stream().map(str -> toUpperCase(str)).collect(Collectors.toList()));
+    }
+
+    /**
+     * Get Name of the security role from security rule.
+     * @param name Name of the security rule
+     * @return Last word in lower case
+     */
+    public String getRoleName(String name) {
```
aklish: The role name should not be extracted from the Rules - but rather the Roles.  As such, we don't need to parse text like "Principal is ADMIN".

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,83 @@
...
+    }, timeZone = "UTC")
+
+    {{#generateDimGetterMethod this}}{{/generateDimGetterMethod}}
+
+    {{#generateDimSetterMethod this}}{{/generateDimSetterMethod}}
+
+{{!--  --}}
+{{else}}
+    {{#ifContains definition "."}}
+        @JoinTo(path = "{{definition}}")
```
aklish: I don't think you need to worry about JoinTo annotation.  You can just use the new DimensionFormula annotation.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,83 @@
...
+    {{else}}
+        @DimensionFormula("{{definition}}")
+        {{#generateDimGetterMethod this}}{{/generateDimGetterMethod}}
+    {{!--  --}}
+    {{/ifContains}}
+{{/ifDimTypeMatches}}
+{{/each}}
+
+{{#each joins}}
+    @Join("{{ definition }}")
```
aklish: This annotation should be placed on the field.   We can remove getter/setter.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,83 @@
...
+{{else}}
+    {{#ifContains definition "."}}
+        @JoinTo(path = "{{definition}}")
+        @Column(updatable = false, insertable = false)
+        {{#generateDimGetterMethod this}}{{/generateDimGetterMethod}}
+
+        {{#generateDimSetterMethod this}}{{/generateDimSetterMethod}}
+    {{!--  --}}
+    {{else}}
+        @DimensionFormula("{{definition}}")
```
aklish: This annotation can be placed on the field.  We can remove getter/setter.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHelper.java
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+     * If string contains passed value.
+     * @param str String to be searched
+     * @param search pattern to search
+     * @return true/false if matched
+     */
+    public boolean contains(String str, String search) {
+
+        return (str == null || str.indexOf(search) < 0);
```
aklish: This logic feels inverted to me.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHelper.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    /**
+     * Generate setter method.
+     * @param type Type of the field
+     * @param fieldName Field Name
+     * @return setter method
+     */
+    public String generateSetterMethod(String type, String fieldName) {
```
aklish: We can remove all the generate getter and setter methods.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHelper.java
```diff
@@ -0,0 +1,246 @@
...
+                        + "return " + fieldName + ";" + NEWLINE + FOUR_SPACES + "}";
+    }
+
+    // Security.java helper methods
+    /**
+     * Get the Class Name for Security.
+     * @param name Name of the security rule
+     * @return Capitalize First Letter of Each Word and remove spaces
+     */
+    public String getClassSuffix(String name) {
```
aklish: All the security helpers can probably be removed since they are parsing the rules and not the roles.  Rules is phase 2.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHydrator.java
```diff
@@ -0,0 +1,112 @@
...
+        {"&", "&amp;" }
+    });
+
+    /**
+     * Method to hydrate the Table template.
+     * @param table ElideTable object
+     * @return table java class list
+     * @throws IOException IOException
+     */
+    public List<String> hydrateTableTemplate(ElideTable table) throws IOException {
```
aklish: ElideTable needs a better class name since it is really a configuration object and not a singular table.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHydrator.java
```diff
@@ -0,0 +1,112 @@
...
+        return tableClassStringNameList;
+    }
+
+    /**
+     * Method to hydrate the Security template.
+     * @param security ElideSecurity Object
+     * @return security java class string
+     * @throws IOException IOException
+     */
+    public String hydrateSecurityTemplate(ElideSecurity security) throws IOException {
```
aklish: Rename ElideSecurity to ElideSecurityConfig

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHydrator.java
```diff
@@ -0,0 +1,112 @@
...
+        Handlebars handlebars = new Handlebars(loader).with(MY_ESCAPING_STRATEGY);
+        handlebars.registerHelpers(ConditionalHelpers.class);
+        handlebars.registerHelpers(new HandlebarsHelper());
+        Template template = handlebars.compile("table");
+
+        for (Table t : table.getTables()) {
+            tableClassStringList.add(template.apply(t));
+        }
+
+        return tableClassStringList;
```
aklish: Its hard to say if these method signatures are the right ones without seeing the calling code.  I'm confused why this is returning a list of Strings for example and not a Map.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHydratorTest.java
```diff
@@ -0,0 +1,344 @@
...
+            + "            filter: id==${principal.id}\n"
+            + "            name: Principal is member\n"
+            + "        },\n"
+            + "    ]\n"
+            + "}";
+
+    private static final String VALID_SECURITY_JAVA = "package com.yahoo.elide.contrib.dynamicconfig.model;\n"
+            + "\n"
+            + "import java.security.Principal;\n"
+            + "import java.util.Optional;\n"
```
moizarafat, line 283: We

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/security.hbs
```diff
@@ -0,0 +1,19 @@
+package com.yahoo.elide.contrib.dynamicconfig.model;
+
+import java.security.Principal;
+import java.util.Optional;
```
moizarafat, line 4: Optional is not used. Please remove

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/security.hbs
```diff
@@ -0,0 +1,19 @@
...
+import com.yahoo.elide.annotation.SecurityCheck;
+import com.yahoo.elide.security.checks.prefab.Role.RoleMemberCheck;
+
+public class DynamicConfigOperationChecks {
```
moizarafat, line 8: We need to remove the top level DynamicConfigOperationChecks and instead move PrincipalIs* Classes as the top level class. For ex. the class should be generated as
`package example;

import com.yahoo.elide.annotation.SecurityCheck;
import com.yahoo.elide.security.checks.prefab.Role.RoleMemberCheck;

@SecurityCheck(DynamicConfigOperationChecksPrincipalIsAdmin.PRINCIPAL_IS_ADMIN)
public class DynamicConfigOperationChecksPrincipalIsAdmin extends RoleMemberCheck {

	public static final String PRINCIPAL_IS_ADMIN = "Principal is Admin";

	public DynamicConfigOperationChecksPrincipalIsAdmin() {
		super("admin"); 

	}
}`

moizarafat: We should return each roles as separate java class back to the caller program. This is to assist during the inmemory compile as we will not be able to parse the top level class to get the individual security class.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHydrator.java
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public String hydrateSecurityTemplate(ElideSecurity security) throws IOException {
+
+        TemplateLoader loader = new ClassPathTemplateLoader("/templates");
+        Handlebars handlebars = new Handlebars(loader).with(MY_ESCAPING_STRATEGY);
+        handlebars.registerHelpers(ConditionalHelpers.class);
+        handlebars.registerHelpers(new HandlebarsHelper());
+        Template template = handlebars.compile("security");
+
+        return template.apply(security);
```
moizarafat, line 91: As mentioned in the other comment, this should return multiple classes by looping the individual entries in roles. So this class should also return a map like hydrateTableTemplate.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -72,12 +72,14 @@ private void parse(String localConfigPath) throws Exception {
...
         String securityJson = util.getJsonConfig(localConfigPath, DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY).get(0);
-        populatePojo(securityJson, DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY);
+        populatePojo(util.resolveVariables(securityJson, this.variables),
+                DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY);
 
         //table
         Set<Table> tables = new HashSet<>();
         for (String tableJson : util.getJsonConfig(localConfigPath, DynamicConfigHelpersUtil.SCHEMA_TYPE_TABLE)) {
-            ElideTableConfig table = (ElideTableConfig) parseJsonConfig(tableJson,
+            ElideTableConfig table = (ElideTableConfig) parseJsonConfig(
```
moizarafat, line 81: We should store the output of `util.resolveVariables(tableJson, this.variables)` in a variable and then call populatePojo method.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParserTest.java
```diff
@@ -41,13 +42,15 @@ public void testValidateVariablePath() throws Exception {
...
+            // test hydration, variable substitution
+            assertEquals(Type.INTEGER, t.getMeasures().get(0).getType());
         }
     }
 
     @Test
     public void testNullConfig() {
         try {
-            testClass.parseConfigPath(null);
+//            testClass.parseConfigPath(null);
```
moizarafat, line 53: Is this intended?

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,81 @@
...
+@Entity
+@Data
+{{#if table}}@FromTable(name = "{{#if schema}}{{schema}}.{{/if}}{{table}}"){{/if}}
+{{#if sql}}@FromSubquery(sql = "{{sql}}"){{/if}}
+{{#if readAccess}}@ReadPermission(expression = "{{readAccess}}"){{/if}}
+{{#if description}}@Meta(description = "{{description}}"){{/if}}
+public class {{#capitalizeFirstLetter name}}{{/capitalizeFirstLetter}} {{#if extend}}extends {{#capitalizeFirstLetter extend}}{{/capitalizeFirstLetter}}{{/if}}{
+
+    @Id
+    private String name;
```
moizarafat, line 42: change name to Id

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,81 @@
...
+import javax.persistence.Entity;
+
+/**
+ * A root level entity for testing AggregationDataStore.
+ */
+@Include(rootLevel = true)
+@Cardinality(size = CardinalitySize.{{#toUpperCase cardinality}}{{/toUpperCase}})
+@EqualsAndHashCode
+@ToString
+@Entity
```
moizarafat, line 33: remove @Entity annotation here.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,81 @@
...
+
+import java.util.Date;
+import javax.persistence.Column;
+import javax.persistence.Id;
+import javax.persistence.Entity;
+
+/**
+ * A root level entity for testing AggregationDataStore.
+ */
+@Include(rootLevel = true)
```
moizarafat, line 29: When Hidden is true we need to use Exclude Annotation. https://elide.io/pages/guide/15-annotations.html#annotation-Exclude

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,81 @@
...
+    @Temporal(grains = {
+    {{#each grains}}
+            @TimeGrainDefinition(grain = TimeGrain.{{grain}}, expression = "{{sql}}"){{#if @last}}{{else}}, {{/if}}
+    {{/each}}
+    }, timeZone = "UTC")
+{{/ifTypeMatches}}
+
+    {{#if readAccess}}@ReadPermission(expression = "{{readAccess}}"){{/if}}
+    {{#if description}}@Meta(description = "{{description}}"){{/if}}
+    @DimensionFormula("{{definition}}")
```
moizarafat, line 56: Include logic for exclude.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,81 @@
...
+    private Set<{{#capitalizeFirstLetter to}}{{/capitalizeFirstLetter}}> {{name}};
+{{else}}
+    private {{#capitalizeFirstLetter to}}{{/capitalizeFirstLetter}} {{name}};
+{{/ifTypeMatches}}
+
+{{/each}}
+
+{{#each measures}}
+
+    @MetricFormula("{{definition}}")
```
moizarafat, line 75: Include logic for exclude.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParserTest.java
AvaniMakwana: resolved in https://github.com/yahoo/elide/pull/1210

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
AvaniMakwana: resolved in https://github.com/yahoo/elide/pull/1210

suppressions.xml
```diff
@@ -3,4 +3,25 @@
...
       <cve>CVE-2018-1258</cve>
    </suppress>
+   <suppress until="2019-12-01Z">
+       <notes><![CDATA[file name: jackson-databind-2.9.9.3.jar]]></notes>
+       <gav regex="true">^com\.fasterxml\.jackson\.core:jackson-databind.*$</gav>
+       <cve>CVE-2019-16335</cve>
+       <cve>CVE-2019-14540</cve>
+   </suppress>
+   <suppress>
+       <notes><![CDATA[file name: handlebars-4.1.2.jar: handlebars-v4.0.4.js]]></notes>
```
aklish, line 13: There's no fix for any of these?  Paranoids don't like unbounded security supressions.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHelper.java
```diff
@@ -0,0 +1,113 @@
...
+    /**
+     * Transform string to capitalize first letter of each word and remove spaces.
+     * @param str String to be transformed
+     * @return Capitalize First Letter of Each Word and remove spaces
+     */
+    public String firstCharOnlyToUpper(String str) {
+
+        return (str == null || str.length() == 0) ? str
+                : String.join(EMPTY_STRING, Arrays.asList(str.trim().split(WHITESPACE_REGEX)).stream().map(
+                        s -> s.substring(0, 1).toUpperCase(Locale.ENGLISH) + s.substring(1).toLowerCase(Locale.ENGLISH))
```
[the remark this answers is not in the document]
rishi-aga: capitalizeFirstLetter is to change first character to uppercase and keep others as is.
firstCharOnlyToUpper is to change first character of each word to uppercase, change other
characters to lower case and remove spaces. Renaming method name  to "titleCaseRemoveSpaces"
Also changing definition to "s -> toUpperCase(s.substring(0, 1)) + toLowerCase(s.substring(1)))" ,

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/security.hbs
```diff
@@ -0,0 +1,13 @@
+package com.yahoo.elide.contrib.dynamicconfig.model;
```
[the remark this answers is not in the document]
rishi-aga: Added.

suppressions.xml
moizarafat: there has been no version released since Oct 2018. https://mvnrepository.com/artifact/com.github.jknack/handlebars . Its a open issue on their github page recently.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHelper.java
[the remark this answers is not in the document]
moizarafat: Calling capitalizeFirstLetter will run substring again even for 1 character passed to it, so we decided to skip that call.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
```diff
@@ -56,7 +56,8 @@ private void parse(String localConfigPath) throws Exception {
 
         //security
         String securityJson = util.getJsonConfig(localConfigPath, DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY).get(0);
-        populatePojo(securityJson, DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY);
+        populatePojo(util.resolveVariables(securityJson, this.variables),
+                DynamicConfigHelpersUtil.SCHEMA_TYPE_SECURITY);
 
         //table
         populateTablesPojo(util.getJsonConfig(localConfigPath, DynamicConfigHelpersUtil.SCHEMA_TYPE_TABLE));
```
moizarafat, line 63: util.resolveVariables should also be called in populateTablesPojo.

suppressions.xml
moizarafat: There is a new release of handlebars 4.2.0 for this issue. we will switch to that and verify

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/ElideConfigParser.java
rishi-aga: populateTablesPojo method is calling util.resolveVariables

suppressions.xml
aklish: Any update on this?

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -131,6 +131,29 @@
...
         </dependency>
+        <dependency>
+            <groupId>com.github.jknack</groupId>
+            <artifactId>handlebars</artifactId>
+            <version>${handlebars.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <version>${junit.jupiter.version}</version>
```
aklish, line 142: Any dependency defined in the parent pom dependency management should not set explicit versions.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * A root level entity for testing AggregationDataStore.
+ */
+@Cardinality(size = CardinalitySize.{{#toUpperCase cardinality}}{{/toUpperCase}})
+@EqualsAndHashCode
+@ToString
+@Data
+{{#if table}}@FromTable(name = "{{#if schema}}{{schema}}.{{/if}}{{table}}"){{/if}}
+{{#if sql}}@FromSubquery(sql = "{{sql}}"){{/if}}
```
aklish, line 39: Can this be part of an else statement with respect to `{{#if table}}`

rishi-aga: As per schema it should be one of {table, sql, extend} so both table and sql can be undefined

rishi-aga: changed to use if-else for table and sql

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -100,16 +100,6 @@
             <artifactId>hjson</artifactId>
             <version>${hjson.version}</version>
         </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
```
aklish, line 104: Can you remove the version variables above as well?

rishi-aga: Removed.

Issue #1113, opened 2019-12-17
Elide 5.0 - Simplify Life Cycle Hooks


aklish:
## Expected Behavior
Deprecate/Remove the OnCreate..., OnRead..., OnDelete..., OnUpdate annotations. 
We'll create a singular, new annotation for all of these concepts:

```java
@Target({TYPE, METHOD, FIELD})
public @interface LifeCycleHookBinding {
    Class<? extends LifeCycleHook> hook();
    Operation operation(); //CREATE, READ, UPDATE, DELETE
    TransactionPhase phase(); //PRESECURITY, PRECOMMIT, POSTCOMMIT
    boolean oncePerRequest(); 
}
```

Rather than binding the hook to a function in the Elide model, the annotation will provide the class of the LifeCycleHook function to invoke.  

The three behaviors for hooks should be preserved:
1) A hook is invoked once per class per request (when bound to the model).
2) A hook is invoked once per field per request (when bound to a model field or method).
3) A hook is invoked one or more times per class per request (when bound to a model and oncePerRequest is false).

## Current Behavior
Documented here:
https://elide.io/pages/guide/02-data-model.html

## Possible Solution
The entity dictionary should bind any hooks it finds on a class when that class is bound to the dictionary.

We should preserve the EntityDictionary bindTrigger functions to the extent possible in scenarios where a developer would rather manually bind hooks to the dictionary.

## Context
We want life cycles hooks to be declarative in the model but to decouple the logic of the hook from the model.

PR #1224, opened 2020-03-22
Simplify life cycle hooks

aklish:
Resolves Issue #1113

## Description
See Issue #1113.

New tests found two legacy issues with life cycle hooks (both are not fixed on this branch):
1. Delete will trigger READ hooks on relationships.  READ should not be triggered.
2. When adding or removing multiple members of a relationship, each element added or removed triggers an UPDATE (rather than a singular update for the entire collection).

There will be separate issues filed to fix these issues.

## Motivation and Context
Decouples entity models from lifecycle hooks.

## How Has This Been Tested?
Rewrote existing unit and IT tests.

## License


README.md
```diff
@@ -136,6 +136,7 @@ Add Lifecycle hooks to your models to embed custom business logic that execute i
 @CreatePermission("Admin OR Publisher")
 @DeletePermission("Noone")
 @UpdatePermission("Noone")
```
jkusa, line 138:
```suggestion
@DeletePermission("None")
@UpdatePermission("None")
```
??

PR #1236, opened 2020-03-25
Enforce ReadPermission for filter joins

wcekan:
## Description
Enforce ReadPermission for filter joins.
Clean up duplicated filter expression builder in RequestScope.getLoadFilterExpression

## Motivation and Context
Restricts filter join access to fields that pass ReadPermission.

## How Has This Been Tested?
Build tests.  Added additional Integration test.  
Tested locally with end product.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
@@ -31,15 +31,14 @@ public class EnforceJoinFilterExpressionVisitor implements FilterExpressionVisit
         this.resource = resource;
     }
 
+    /**
+     * Enforce ReadPermission on provided query filter
+     *
+     * @return true if allowed, false if rejected
+     */
     @Override
     public Boolean visitPredicate(FilterPredicate filterPredicate) {
         RequestScope requestScope = resource.getRequestScope();
-
-        // enforce filter checking only when user filter is present
-        if (!requestScope.getQueryParams().isPresent()
-                || !requestScope.getQueryParams().get().keySet().stream().anyMatch(k -> k.startsWith("filter["))) {
-            return true;
-        }
         Set<PersistentResource> val = Collections.singleton(resource);
         for (Path.PathElement pathElement : filterPredicate.getPath().getPathElements()) {
             Class<?> entityClass = pathElement.getType();
--- a/elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2020, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core;
+
+import com.yahoo.elide.annotation.ReadPermission;
+import com.yahoo.elide.core.exceptions.ForbiddenAccessException;
+import com.yahoo.elide.core.filter.FilterPredicate;
+import com.yahoo.elide.core.filter.expression.AndFilterExpression;
+import com.yahoo.elide.core.filter.expression.FilterExpressionVisitor;
+import com.yahoo.elide.core.filter.expression.NotFilterExpression;
+import com.yahoo.elide.core.filter.expression.OrFilterExpression;
+
+import java.util.Collections;
+import java.util.Objects;
+import java.util.Set;
+import java.util.Spliterators;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * Enforce read permission on filter join
+ */
+public class EnforceJoinFilterExpressionVisitor implements FilterExpressionVisitor<Boolean> {
+    private PersistentResource<?> resource;
+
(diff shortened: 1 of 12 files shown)


elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
```diff
@@ -0,0 +1,106 @@
...
+        // neither rejected
+        return left && right;
+    }
+
+    @Override
+    public Boolean visitOrExpression(OrFilterExpression expression) {
+        Boolean left = expression.getLeft().accept(this);
+        Boolean right = expression.getRight().accept(this);
+        // neither rejected
+        return left && right;
```
aklish: should be left || right.

elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    @Override
+    public Boolean visitNotExpression(NotFilterExpression expression) {
+        // check rejected
+        return expression.getNegated().accept(this);
```
aklish: We need to negate the value of the accept.

elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
```diff
@@ -0,0 +1,106 @@
...
+import java.util.Set;
+import java.util.Spliterators;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
+
+/**
+ * Enforce read permission on filter join
+ */
+public class EnforceJoinFilterExpressionVisitor implements FilterExpressionVisitor<Boolean> {
```
aklish: Maybe a better name would be `VerifyFieldAccessFilterExpressionVisitor`

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1557,7 +1558,8 @@ protected void addInverseRelation(String relationName, Object inverseObj) {
      * @return Filtered set of resources
      */
     protected static Set<PersistentResource> filter(Class<? extends Annotation> permission,
-                                                    Set<PersistentResource> resources) {
+            Optional<FilterExpression> filter,
+            Set<PersistentResource> resources) {
```
aklish: Update the javadoc above.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/FilterIT.java
```diff
@@ -1362,6 +1362,21 @@ void testFailFilterAuthorBookByChapter() throws IOException {
...
+    void testFailFilterBookByAuthorAddress() throws IOException {
+        /* Test default */
+        JsonNode result = getAsNode("book?filter[book.authors.homeAddress]=main&include=authors");
+        JsonNode data = result.get("data");
+        assertEquals(0, data.size(), result.toString());
+
+        /* Test RSQL */
+        result = getAsNode(
+                String.format("book?filter[book]=authors.homeAddress=='main'", hemingwayId),
+                HttpStatus.SC_BAD_REQUEST);
```
aklish: Why does this give a bad request but the above tests returns an empty data?  I suspect the RSQL filter is formatted incorrectly.

elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
aklish: Can we add a Unit test for this class to test the different visitor methods?

elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
wcekan: No. true is valid, false is rejected.  If either is rejected, we want false.

elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
wcekan: No.  We are not negating allowed or rejected. 
`book.authors.homeAddress[NOT]='main'` should still be rejected.

elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
wcekan: sure.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/FilterIT.java
wcekan: Bad merge.  Will fix that.

elide-core/src/main/java/com/yahoo/elide/core/EnforceJoinFilterExpressionVisitor.java
wcekan: Can try.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
wcekan: done.

PR #1240, opened 2020-03-27
call PersistentResource get relationship

wcekan:
## Description
Updated VerifyFieldAccessFilterExpressionVisitor to call PersistentResorce get relationship to cross datastore transition.

## Motivation and Context
Will properly use proper transaction to request the relationship results.

## How Has This Been Tested?
Build testing.  Locally testing with our product.

## License


wcekan:
@aklish I built this on your suggestion, but I am not sure this is required at all.  I do not think we need to go back to the transaction to get the relationship again.  We need to follow those PersistentResources that are already collected from the query.

PR #1243, opened 2020-03-29
Optimize field annotation lookup

wcekan:
## Description
Profiling showed a hotspot for `entityClass.getMethod(method)` in the FilterExpressionCheck.

Cleanup some `get/if null put` logic with more readable `computeIfAbsent`
 
## Motivation and Context
Reduces overhead for `FilterExpressionPath` lookup.

## How Has This Been Tested?
Local testing with profiler.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
@@ -580,7 +580,7 @@ public class EntityBinding {
      * @param <A> annotation type
      * @return the annotation
      */
-    public <A extends Annotation> A getAnnotation(Class<A> annotationClass, String method) {
+    public <A extends Annotation> A getMethodAnnotation(Class<A> annotationClass, String method) {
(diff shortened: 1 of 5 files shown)


elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -572,6 +572,26 @@ public void bindTrigger(Class<? extends Annotation> annotationClass,
...
 
+    /**
+     * Return annotation for provided method.
+     *
+     * @param annotationClass the annotation class
+     * @param method the method
+     * @param <A> annotation type
+     * @return the annotation
+     */
+    public <A extends Annotation> A getAnnotation(Class<A> annotationClass, String method) {
```
aklish: We have a lot of getAnnotation methods.  This should be renamed to getMethodAnnotation.

wcekan: Sure

PR #1254, opened 2020-03-30
Fixes #1244: Incorrect pagination with ToMany filtering

### LINKED ISSUE #1244: Incorrect pagination with ToMany filtering
## Expected Behavior
When issuing query with filter on ToMany relationship and pagination should return correct number of results vs page size

## Current Behavior
When issuing query with filter on ToMany relationship and pagination,
then resulting product of join vs distinct rows differs resulting in broken pagination

## Possible Solution
distinct should be applied in this case

## Your Environment

* Elide version used: 4.5.14
* Environment name and version (Java 11.0.2):
* Operating System and version:
* Link to your project: https://github.com/FAForever/faf-java-api/issues/338

bukajsytlos:
Resolves #1244

## Description
When creating query, check if pagination and ToMany relationship is present, if so add DISTINCT to query.

## Motivation and Context
See the issue

## How Has This Been Tested?
Unit tests

## License

### DIFF
--- a/elide-datastore/elide-datastore-hibernate/src/test/java/com/yahoo/elide/datastores/hibernate/hql/RootCollectionFetchQueryBuilderTest.java
+++ b/elide-datastore/elide-datastore-hibernate/src/test/java/com/yahoo/elide/datastores/hibernate/hql/RootCollectionFetchQueryBuilderTest.java
@@ -6,13 +6,20 @@
 package com.yahoo.elide.datastores.hibernate.hql;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 
 import com.yahoo.elide.core.EntityDictionary;
 import com.yahoo.elide.core.Path;
+import com.yahoo.elide.core.exceptions.InvalidValueException;
 import com.yahoo.elide.core.filter.FilterPredicate;
 import com.yahoo.elide.core.filter.InPredicate;
+import com.yahoo.elide.core.filter.dialect.CaseSensitivityStrategy;
+import com.yahoo.elide.core.filter.dialect.ParseException;
+import com.yahoo.elide.core.filter.dialect.RSQLFilterDialect;
+import com.yahoo.elide.core.filter.expression.FilterExpression;
 import com.yahoo.elide.core.filter.expression.OrFilterExpression;
 import com.yahoo.elide.core.hibernate.hql.RootCollectionFetchQueryBuilder;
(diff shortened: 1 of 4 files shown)


elide-datastore/elide-datastore-hibernate/src/test/java/com/yahoo/elide/datastores/hibernate/hql/RootCollectionFetchQueryBuilderTest.java
```diff
@@ -128,6 +129,56 @@ public void testRootFetchWithJoinFilter() {
         assertEquals(expected, actual);
     }
 
+    @Test
+    public void testDistinctRootFetchWithToManyJoinFilterAndPagination() {
```
aklish, line 133: Can we test the exception behavior as well?

bukajsytlos: yes, there is/are failing tests right now because of this, but I am struggling little bit with running and debugging them locally. when I resolve that, I'll try to write test for this case

elide-datastore/elide-datastore-hibernate/src/test/java/com/yahoo/elide/datastores/hibernate/hql/RootCollectionFetchQueryBuilderTest.java
```diff
@@ -128,6 +129,56 @@ public void testRootFetchWithJoinFilter() {
...
+        final Pagination pagination = Pagination.fromOffsetAndLimit(10, 0, false);
+        List<Path.PathElement> chapterTitlePath = Arrays.asList(
```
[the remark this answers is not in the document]
bukajsytlos: ok, I just copied this from other test :)

bukajsytlos:
@aklish I am slightly confused. There was a failing test in SortingIT which I have changed here. Looks like elide is creating default pagination, but I don't understand under which conditions. I guess users won't be happy if my change would throw exceptions on them, when they are not specifying pagination, but my exception states otherwise

aklish:
Elide puts a cap on the number of records returned if the client doesn't provide pagination params.  That's called _default pagination_.  There is a global setting that sets how many records this is - typically something large like 10k.

aklish:
Overall - the change is correct.  I'm not sure if this will break anyone's use case - but I'm hoping it won't.  We'll do a a minor version release.

PR #1255, opened 2020-04-03
in memory compilation integrated with dynamic config helpers

### LINKED ISSUE #1178: RFC : Elide Dynamic Model Configuration (Tables Only)

ayeswarya:
Authored-by: Ayeswarya
One of the multiple PRs that will resolve #1178 
## Description
In-memory compiler, custom class loader for spring/hibernate to know where to look for class files.
Integrate the utilities developed above to first convert Variables.hjson etc to POJO and then resolve Tables Handlebar HJSON and then resolve it to POJO.
Resolve the Model Handlebar java classes using the POJOs generated in step above. 
Resolve the Security Handlebar java classes using the POJOs generated from security.hjson. 

## Motivation and Context
Please refer #1178

## How Has This Been Tested?
Unit tests included.
Tested the new feature with elide-spring-boot-example

## License


elide-contrib/elide-dynamic-config-helpers/src/test/resources/models/tables/table2.hjson
```diff
@@ -20,15 +20,15 @@
         measures : [
             {
             name : highScore
-            type : "{{measure_type}}"
+            type : "INTEGER"
```
moizarafat, line 23: you can still have it as <%measure_type%>. This way the original test case where this value is replaced from the value in variables.hjson is still valid.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
```diff
@@ -0,0 +1,137 @@
...
+            //Bind FromTable/FromSubSelect classes from classpath to Persistence Unit
+            bindClasses.addAll(ClassScanner.getAnnotatedClasses(FromTable.class));
+            bindClasses.addAll(ClassScanner.getAnnotatedClasses(FromSubquery.class));
+
+            //Map of JPA Properties to be be passed to EntityManager
+            Map<String, String> jpaPropMap = jpaProperties.getProperties();
+            String hibernateGetDDLAuto = hibernateProperties.getDdlAuto();
+
+            //Set the relevant property in JPA corresponding to Hibernate Property Value
+            if (jpaPropMap.get("hibernate.hbm2ddl.auto") == null && hibernateGetDDLAuto != null) {
```
moizarafat, line 74: Lets move the value `hibernate.hbm2ddl.auto` to a static final var

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/spring/tests/DynamicConfigTest.java
```diff
@@ -0,0 +1,129 @@
...
+    @Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD,
+            statements = "INSERT INTO PlayerStats (name,countryId,createdOn) VALUES\n"
+                    + "\t\t('SaniaMirza','2','2000-10-01');"
+                    + "INSERT INTO PlayerStats (name,countryId,createdOn) VALUES\n"
+                    + "\t\t('SerenaWilliams','1','2000-10-01');"
+                    + "INSERT INTO PlayerCountry (id,isoCode) VALUES\n"
+                    + "\t\t('2','IND');"
+                    + "INSERT INTO PlayerCountry (id,isoCode) VALUES\n"
+                    + "\t\t('1','USA');")
+    @Sql(executionPhase = Sql.ExecutionPhase.AFTER_TEST_METHOD,
```
moizarafat, line 117: Since this statement is common for all test cases, we can move it up at class level.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/spring/tests/DynamicConfigTest.java
```diff
@@ -0,0 +1,129 @@
...
+                                )
+                        )
+                ).toJSON()))
+                .statusCode(HttpStatus.SC_CREATED);
+    }
+    @SqlMergeMode(SqlMergeMode.MergeMode.MERGE)
+    @Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD,
+            statements = "INSERT INTO PlayerStats (name,countryId,createdOn) VALUES\n"
+                    + "\t\t('SaniaMirza','2','2000-10-01');"
+                    + "INSERT INTO PlayerStats (name,countryId,createdOn) VALUES\n"
```
moizarafat, line 111: Since this statement is common for all test cases, we can move it up at class level.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/spring/tests/DynamicConfigTest.java
```diff
@@ -0,0 +1,129 @@
...
+                    + "\t\t('SerenaWilliams','1','2000-10-01');"
+                    + "INSERT INTO PlayerCountry (id,isoCode) VALUES\n"
+                    + "\t\t('2','IND');"
+                    + "INSERT INTO PlayerCountry (id,isoCode) VALUES\n"
```
moizarafat, line 115: Since this statement is common for all test cases, we can move it up at class level.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/spring/tests/IntegrationTest.java
```diff
@@ -10,6 +10,7 @@
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.web.server.LocalServerPort;
 
+
```
moizarafat, line 13: Please remove extra line.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/dynamic/compile/ElideDynamicInMemoryClassLoader.java
```diff
@@ -0,0 +1,56 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Set;
+
+@Slf4j
+@Data
+@AllArgsConstructor
+public class ElideDynamicInMemoryClassLoader extends ClassLoader {
```
moizarafat, line 20: Please add Javadoc for the class.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/dynamic/compile/ElideDynamicPersistenceUnit.java
```diff
@@ -0,0 +1,63 @@
...
+
+import javax.persistence.SharedCacheMode;
+import javax.persistence.ValidationMode;
+import javax.persistence.spi.ClassTransformer;
+import javax.persistence.spi.PersistenceUnitInfo;
+import javax.persistence.spi.PersistenceUnitTransactionType;
+import javax.sql.DataSource;
+
+@Data
+public class ElideDynamicPersistenceUnit implements PersistenceUnitInfo {
```
moizarafat, line 22: Please add Javadoc for the class.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/dynamic/compile/ElideDynamicEntityCompiler.java
```diff
@@ -0,0 +1,99 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+
+@Slf4j
+public class ElideDynamicEntityCompiler {
```
moizarafat, line 25: Please add Javadoc for the class.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
```diff
@@ -0,0 +1,137 @@
...
+ */
+
+@Slf4j
+@Configuration
+@EnableConfigurationProperties(ElideConfigProperties.class)
+@ConditionalOnExpression("${elide.dynamic-config.enabled:false}")
+public class ElideDynamicConfiguration {
+
+    @Bean
+    public LocalContainerEntityManagerFactoryBean entityManagerFactory (EntityManagerFactoryBuilder builder,
```
moizarafat, line 52: Please add Javadoc for the method.

moizarafat:
Lets do the same updates as table.hbs to security.hbs so that it's consistent.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
```diff
@@ -0,0 +1,148 @@
...
+            bindClasses.addAll(ClassScanner.getAnnotatedClasses(FromTable.class));
+            bindClasses.addAll(ClassScanner.getAnnotatedClasses(FromSubquery.class));
+
+            //Map of JPA Properties to be be passed to EntityManager
+            Map<String, String> jpaPropMap = jpaProperties.getProperties();
+            String hibernateGetDDLAuto = hibernateProperties.getDdlAuto();
+
+            //Set the relevant property in JPA corresponding to Hibernate Property Value
+            if (jpaPropMap.get(HIBERNATE_DDL_AUTO) == null && hibernateGetDDLAuto != null) {
+               jpaPropMap.put(HIBERNATE_DDL_AUTO, hibernateGetDDLAuto);
```
aklish, line 86: There are a lot of hibernate properties we will likely want to preserve beyond just the DDL_AUTO - batch settings, fetch modes, connection pool settings, etc.  Can we just take the entire map of hibernate properties and use them all?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
```diff
@@ -0,0 +1,148 @@
...
+            }
+
+            LocalContainerEntityManagerFactoryBean bean = new LocalContainerEntityManagerFactoryBean();
+            bean.setJpaVendorAdapter(vendorAdapter);
+
+            //Add JPA Properties from Application.yaml
+            bean.setJpaPropertyMap(jpaPropMap);
+
+            //Add Classes
+            bean.setJpaPropertyMap(puiPropertyMap);
```
aklish, line 127: Does this overwrite the previous or extend it?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
```diff
@@ -0,0 +1,148 @@
...
+
+                @Override
+                public PersistenceUnitInfo obtainPersistenceUnitInfo(String persistenceUnitName)
+                        throws IllegalArgumentException, IllegalStateException {
+                    return elideDynamicPersistenceUnit;
+                }
+            });
+
+            return bean;
+        } catch (Exception e) {
```
aklish, line 143: Two things - let's not catch Exception here (it is too broad).  Also, if there is an error, we'll want to bring down the service.  Let the exceptions propagate.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/MetaDataStore.java
```diff
@@ -79,6 +79,19 @@ public void populateEntityDictionary(EntityDictionary dictionary) {
                 .forEach(cls -> dictionary.bindEntity(cls, Collections.singleton(Join.class)));
     }
 
+    /**
+     * Bind dynamic compiled classes to dictionary.
+     * @param dynamicCompiledClasses Dynamically compiled classes
+     */
+    public void populateEntityDictionary(Set<Class<?>> dynamicCompiledClasses) {
```
aklish, line 86: Would be nice if this was done in the constructor somehow (along with the other entities that are being populated).

elide-spring/elide-spring-boot-autoconfigure/pom.xml
```diff
@@ -149,11 +149,28 @@
             <version>${spring.boot.version}</version>
             <optional>true</optional>
         </dependency>
+        <dependency>
+            <groupId>org.mdkt.compiler</groupId>
+            <artifactId>InMemoryJavaCompiler</artifactId>
+            <version>1.3.0</version>
```
aklish, line 155: optional to true

elide-spring/elide-spring-boot-autoconfigure/pom.xml
```diff
@@ -149,11 +149,28 @@
...
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-dynamic-config-helpers</artifactId>
+            <version>5.0.0-pr7-SNAPSHOT</version>
+            <optional>true</optional>
+        </dependency>
 
         <!-- Test Dependencies -->
 
         <!-- YAML for spring config -->
+                <dependency>
```
aklish, line 168: Spacing.

elide-spring/elide-spring-boot-autoconfigure/pom.xml
```diff
@@ -149,11 +149,28 @@
...
+            <groupId>javax.persistence</groupId>
+            <artifactId>javax.persistence-api</artifactId>
+            <version>2.2</version>
+            <scope>provided</scope>
```
aklish, line 172: optional true

elide-spring/elide-spring-boot-autoconfigure/pom.xml
```diff
@@ -173,6 +190,12 @@
             <artifactId>spring-boot-starter-test</artifactId>
             <version>${spring.boot.version}</version>
             <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.vaadin.external.google</groupId>
```
aklish, line 195: How does this conflict?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAutoConfiguration.java
```diff
@@ -40,8 +49,32 @@
...
+    public ElideDynamicEntityCompiler buildElideDynamicEntityCompiler(ElideConfigProperties settings) {
+
+        ElideDynamicEntityCompiler compiler = null;
+
+        if (settings.getDynamicConfig().isEnabled()) {
+            compiler = new ElideDynamicEntityCompiler(settings.getDynamicConfig().getPath());
+            try {
+                compiler.compile(settings.getDynamicConfig().getPath());
+            } catch (Exception e) {
+                e.printStackTrace();
```
aklish, line 71: Don't print stack traces - log the exception.  Also - don't catch Exception.  Just let it propagate.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAutoConfiguration.java
```diff
@@ -90,32 +127,66 @@ public void inject(Object entity) {
...
+            ObjectProvider<ElideDynamicEntityCompiler> dynamicCompiler, ElideConfigProperties settings)
+            throws ClassNotFoundException {
+
         MetaDataStore metaDataStore = new MetaDataStore();
 
+        if (settings.getDynamicConfig().isEnabled()) {
+            ElideDynamicEntityCompiler compiler = dynamicCompiler.getIfAvailable();
+            Set<Class<?>> annotatedClass = findAnnotatedClasses(compiler, FromTable.class);
+            annotatedClass.addAll(findAnnotatedClasses(compiler, FromSubquery.class));
+            metaDataStore.populateEntityDictionary(annotatedClass);
```
aklish, line 160: Pass the compiler to the MetaDataStore constructor and initialize the internal dictionary in the constructor instead of doing this initialization here.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAutoConfiguration.java
```diff
@@ -125,6 +196,29 @@ public DataStore buildDataStore(EntityManagerFactory entityManagerFactory, Query
...
+     * Find classes with a particular annotation from dynamic compiler.
+     * @param compiler An instance of ElideDynamicEntityCompiler.
+     * @param annotationClass Annotation to search for.
+     * @return Set of Classes matching the annotation.
+     * @throws ClassNotFoundException
+     */
+    private Set<Class<?>> findAnnotatedClasses(ElideDynamicEntityCompiler compiler, Class annotationClass)
+            throws ClassNotFoundException {
+
+        Set<Class<?>> annotatedClass = new HashSet<Class<?>>();
```
aklish, line 209: make plural

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/dynamic/compile/ElideDynamicPersistenceUnit.java
```diff
@@ -0,0 +1,72 @@
...
+import javax.persistence.spi.ClassTransformer;
+import javax.persistence.spi.PersistenceUnitInfo;
+import javax.persistence.spi.PersistenceUnitTransactionType;
+import javax.sql.DataSource;
+
+/**
+ * Persistent Unit implementation for Dynamic Configuration.
+ */
+@Data
+public class ElideDynamicPersistenceUnit implements PersistenceUnitInfo {
```
aklish, line 25: This is a copy of another class in Elide.  Let's move it to a common place (probably JPA data store) - and have both pull from it.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/dynamic/compile/ElideDynamicEntityCompiler.java
```diff
@@ -0,0 +1,108 @@
...
+    public ElideDynamicEntityCompiler(String path) {
+
+        ElideTableConfig tableConfig = new ElideTableConfig();
+        ElideSecurityConfig securityConfig = new ElideSecurityConfig();
+        ElideConfigParser elideConfigParser = new ElideConfigParser();
+        HandlebarsHydrator hydrator = new HandlebarsHydrator();
+
+        try {
+
+            elideConfigParser.parseConfigPath(path);
```
aklish, line 54: What handles the variable substitution for the variables.hjson?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/dynamic/compile/ElideDynamicEntityCompiler.java
```diff
@@ -0,0 +1,108 @@
...
+
+            for (Entry<String, String> entry : securityClasses.entrySet()) {
+                classNames.add(PACKAGE_NAME + entry.getKey());
+            }
+
+            compiler.useParentClassLoader(
+                    new ElideDynamicInMemoryClassLoader(ClassLoader.getSystemClassLoader(),
+                            Sets.newHashSet(classNames)));
+
+        } catch (NullPointerException | IOException e) {
```
aklish, line 72: Let NPE propagate.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/dynamic/compile/ElideDynamicEntityCompiler.java
```diff
@@ -0,0 +1,108 @@
...
+            log.error("Unable to read Dynamic Configuration " + e.getMessage());
+        }
+    }
+
+    /**
+     * Compile table and security model pojos.
+     * @param path: Dynamic config hjsons root location
+     */
+    public void compile(String path) throws Exception {
```
aklish, line 81: Path is unused.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/dynamic/compile/ElideDynamicInMemoryClassLoader.java
```diff
@@ -0,0 +1,59 @@
...
+
+    private Set<String> classNames = Sets.newHashSet();
+
+    public ElideDynamicInMemoryClassLoader(ClassLoader parent, Set<String> classNames) {
+        super(parent);
+        setClassNames(classNames);
+    }
+
+    public void setClassNames(Set<String> classNames) {
```
aklish, line 32: What is the purpose of this public setter method that overrides the constructor initialization?  Can we remove it?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
ayeswarya: It extends the jpaPropertyMap to include the classloader and annotated classes

elide-spring/elide-spring-boot-autoconfigure/pom.xml
ayeswarya: org.json.JSONObject was available from two locations from android-json and everit json schema validator. Everit is used in dynamic config helpers for validation.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/dynamic/compile/ElideDynamicEntityCompiler.java
ayeswarya: Variable substitution happens in  elide contrib dynamic config helpers https://github.com/yahoo/elide/blob/e36cc01d6976ce0f511e47babd3b5b2e313b2138/elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpersUtil.java#L176

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
moizarafat: most of the hibernate properties are passed as `spring.jpa.properties.hibernate.*`. But there are 4 properties which are passed as `spring.jpa.hibernate.*` . Here we are taking those 4 properties and setting the corresponding parameters in `spring.jpa.properties.hibernate.*` if not already set so that it gets passed along to the EntityManagerFactoryBean. https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-application-properties.html

moizarafat: Ayes has expanded the logic to remaining 3 params too.

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -115,6 +112,18 @@
             <artifactId>handlebars</artifactId>
             <version>${handlebars.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.mdkt.compiler</groupId>
+            <artifactId>InMemoryJavaCompiler</artifactId>
+            <version>${mdkt.compiler.version}</version>
+            <optional>true</optional>
```
aklish, line 119: Optional=true is really meant for autoconfigure packages - not a core elide package.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/compile/ElideDynamicEntityCompiler.java
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Compiles dynamic model pojos generated from hjson files.
+ *
+ */
+@Slf4j
+public class ElideDynamicEntityCompiler {
+
+    public static ArrayList<String> classNames = new ArrayList<String>();
+
+    public static final String PACKAGE_NAME = "com.yahoo.elide.contrib.dynamicconfig.model.";
```
aklish, line 36: Prefixing the models with "com.yahoo.elide" will prohibit API versioning in Elide 5 (all elide prefixed models are not versioned).  We'll need to find another package name for the dynamic models.

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -65,6 +65,14 @@
             <artifactId>elide-datastore-multiplex</artifactId>
             <version>5.0.0-pr7-SNAPSHOT</version>
         </dependency>
+        
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-dynamic-config-helpers</artifactId>
+            <version>5.0.0-pr7-SNAPSHOT</version>
+            <optional>true</optional>
+        </dependency>
```
aklish, line 74: Remove optional = true.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
```diff
@@ -0,0 +1,164 @@
...
+            Map<String, String> jpaPropMap = jpaProperties.getProperties();
+
+            String hibernateGetDDLAuto = hibernateProperties.getDdlAuto();
+            Naming hibernateGetNaming =  hibernateProperties.getNaming();
+            String hibernateImplicitStrategy = hibernateGetNaming.getImplicitStrategy();
+            String hibernatePhysicalStrategy = hibernateGetNaming.getPhysicalStrategy();
+            Boolean hibernateGetIdenGen = hibernateProperties.isUseNewIdGeneratorMappings();
+
+            //Set the relevant property in JPA corresponding to Hibernate Property Value
+            hibernateJPAPropertyOverride(jpaPropMap, HIBERNATE_DDL_AUTO, hibernateGetDDLAuto);
```
aklish, line 92: Why can't we copy all of the properties from hibernateProperties instead of setting explicit ones.  THere are dozens of hibernate properties that might be set.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
```diff
@@ -0,0 +1,164 @@
...
+
+            //Map for Persistent Unit properties
+            Map<String, Object> puiPropertyMap = new HashMap<>();
+
+            //Bind entity classes from classpath to Persistence Unit
+            ArrayList<Class> bindClasses = new ArrayList<>();
+            bindClasses.addAll(ClassScanner.getAnnotatedClasses(Entity.class));
+
+            //Bind FromTable/FromSubSelect classes from classpath to Persistence Unit
+            bindClasses.addAll(ClassScanner.getAnnotatedClasses(FromTable.class));
```
aklish, line 79: FromTable and FromSubquery are not JPA Entities - why are we binding them to the Persistent Unit?

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/contrib/dynamicconfig/model/PlayerStats.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * A root level entity for testing AggregationDataStore.
+ */
+@Include(rootLevel = true , type = "playerStats")
+@Cardinality(size = CardinalitySize.LARGE)
+@EqualsAndHashCode
+@ToString
+@Data
+@Entity
```
aklish, line 34: Fact tables in the AggregationDataStore are never entities (because Elide manages the SQL).  JPA shouldn't know about them.  Look at the similar class in AggregationDataStore.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/contrib/dynamicconfig/model/PlayerStats.java
```diff
@@ -0,0 +1,52 @@
...
+@ReadPermission(expression = "Prefab.Role.All")
+@UpdatePermission(expression = "Prefab.Role.None")
+@DeletePermission(expression = "Prefab.Role.None")
+public class PlayerStats {
```
aklish, line 38: elide-example-models was a place to put common classes that are referenced in multiple maven sub-modules.  Rather than redefine these entities here - can we move Player stats and related models to elide-example-models?  That said, I don't want this to cascade into a ton of work - so let's discuss if this looks complicated.

aklish, line 38: Now that I think about it, why do we need this class at all?  Since we are testing dynamic classes - not static ones.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/contrib/dynamicconfig/model/PlayerCountry.java
```diff
@@ -0,0 +1,43 @@
...
+@Cardinality(size = CardinalitySize.SMALL)
+@EqualsAndHashCode
+@ToString
+@Data
+@Entity
+@ReadPermission(expression = "Prefab.Role.All")
+@UpdatePermission(expression = "Prefab.Role.None")
+@DeletePermission(expression = "Prefab.Role.None")
+
+public class PlayerCountry {
```
aklish, line 35: Why do we need a static class when we are testing dynamic class loading?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
moizarafat: Discussed with Aaron, this can be ignored.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/compile/ElideDynamicEntityCompiler.java
moizarafat: we can use dynamicconfig.models

elide-spring/elide-spring-boot-autoconfigure/src/test/java/com/yahoo/elide/contrib/dynamicconfig/model/PlayerStats.java
moizarafat: We will take care of this once all these PRs are merged into a single PR.

elide-contrib/elide-dynamic-config-helpers/src/main/resources/templates/table.hbs
```diff
@@ -31,57 +33,57 @@ import javax.persistence.Id;
...
-{{#if sql}}@FromSubquery(sql = "{{sql}}"){{/if}}
-{{#if readAccess}}@ReadPermission(expression = "{{readAccess}}"){{/if}}
-{{#if description}}@Meta(description = "{{description}}"){{/if}}
-{{#if hidden}}@Exclude{{else}}@Include(rootLevel = true){{/if}}
-public class {{#capitalizeFirstLetter name}}{{/capitalizeFirstLetter}} {{#if extend}}extends {{#capitalizeFirstLetter extend}}{{/capitalizeFirstLetter}}{{/if}}{
+<%#if table%>@FromTable(name = "<%#if schema%><%schema%>.<%/if%><%table%>")<%/if%>
+<%#if sql%>@FromSubquery(sql = "<%sql%>")<%/if%>
+<%#if readAccess%>@ReadPermission(expression = "<%readAccess%>")<%/if%>
+<%#if description%>@Meta(description = "<%description%>")<%/if%>
+<%#if hidden%>@Exclude<%else%>@Include(rootLevel = true, type = "<%#capitalizeFirstLetter name%><%/capitalizeFirstLetter%>")<%/if%>
```
moizarafat, line 44: for `type=`, instead of capitalizeFirstLetter let's use a new method lowerCaseFirstLetter so that its consistent with how we do when it's not dynamic config i.e. camelCase

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1256,7 +1256,21 @@ public void scanForSecurityChecks() {
...
         // /spring/boot/autoconfigure/ElideAutoConfiguration.java
 
-        for (Class<?> cls : ClassScanner.getAnnotatedClasses(SecurityCheck.class)) {
+        Set<Class<?>> classes = ClassScanner.getAnnotatedClasses(SecurityCheck.class);
+
+        addSecurityChecks(classes);
+    }
+
+    /**
+     * Add security checks and bind them to the dictionary.
```
aklish, line 1265: Javadoc param is missing.  This will cause a warning in the build.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideConfigProperties.java
```diff
@@ -36,6 +36,11 @@
      */
     private AsyncProperties async;
 
+    /**
+     * Settings for the Dynamic Configuration.
+     */
+    private ControllerProperties dynamicConfig;
```
aklish, line 42: dynamicConfig is not a controller - so why are we using ControllerProperties to configure it?

elide-spring/elide-spring-boot-autoconfigure/src/test/resources/models/tables/playerStats.hjson
```diff
@@ -0,0 +1,56 @@
...
+            type : INTEGER
+            definition: 'MAX(highScore)'
+            readAccess : Prefab.Role.All
+            }
+        ]
+        dimensions : [
+            {
+                name : name
+                type : TEXT
+                readAccess : Prefab.Role.All
```
aklish, line 31: SInce this is defined at the table level, you don't need to define it for each field.

elide-spring/elide-spring-boot-autoconfigure/src/test/resources/models/tables/playerCountry.hjson
```diff
@@ -0,0 +1,20 @@
...
+        '''
+        A long description
+        '''
+        cardinality : small
+        readAccess : Prefab.Role.All
+        dimensions : [
+            {
+                name : isoCode
+                type : TEXT
+                readAccess : Prefab.Role.All
```
aklish, line 15: This is redundant and can be removed.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Dynamic Configuration For Elide Services. Override any of the beans (by
+ * defining your own) and setting flags to disable in properties to change the
+ * default behavior.
+ */
+
+@Slf4j
+@Configuration
+@EnableConfigurationProperties(ElideConfigProperties.class)
+@ConditionalOnExpression("${elide.dynamic-config.enabled:false}")
```
aklish, line 46: If we turn off dynamic-config - will Elide autoconfigure work as expected with its normal EntityManagerFactory configuration?

moizarafat: yes it does. All test cases except dynamic config one succeeds.

PR #1259, opened 2020-04-10
Elide 5.x dynamic config standalone

### LINKED ISSUE #1178: RFC : Elide Dynamic Model Configuration (Tables Only)

AvaniMakwana:
Authored-by: AvaniMakwana

## Description
Enable Dynamic Model Configuration for Elide-standalone

## Motivation and Context
Please refer #1178

## How Has This Been Tested?

## License


elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -53,28 +64,61 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
             options.put("javax.persistence.jdbc.password", "elide123");
         }
 
-        PersistenceUnitInfo persistenceUnitInfo = new PersistenceUnitInfoImpl("elide-stand-alone",
-                combineModelEntities(modelPackageName, includeAsyncModel), options);
+        dynamicEntityCompiler = new ElideDynamicEntityCompiler(dynamicConfigPath);
```
moizarafat, line 67: This line should be inside the if block at L-68, otherwise even when the dynamic config is disabled it will try to generate the Config POJOs.

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -53,28 +64,61 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
+        try {
+
+            ElideDynamicEntityCompiler.bindClasses = new HashSet<>();
+            // add dynamic generated classes
+            ElideDynamicEntityCompiler.bindClasses.addAll(populateBindClasses(dynamicEntityCompiler, Entity.class));
+            ElideDynamicEntityCompiler.bindClasses.addAll(populateBindClasses(dynamicEntityCompiler, FromTable.class));
+            ElideDynamicEntityCompiler.bindClasses.addAll(populateBindClasses(dynamicEntityCompiler,
+                    FromSubquery.class));
+            // add classes
+            ElideDynamicEntityCompiler.bindClasses.addAll(ClassScanner.getAnnotatedClasses(Entity.class));
```
moizarafat, line 84: These 3 sets of classes should not be binded to `ElideDynamicEntityCompiler.bindClasses` as they are not compiled through dynamic compiler. These 3 set of classes should be put in `options` as `AvailableSettings.LOADED_CLASSES, <ClassScanner Scanned classes> ` . Refer https://github.com/yahoo/elide/blob/2d6436d07b3818aa897f401a7f8cd9a9b6e97440/elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideDynamicConfiguration.java#L97

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -53,28 +64,61 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
+
+            persistenceUnitInfo = new PersistenceUnitInfoImpl("elide-stand-alone",
+                    combineModelEntities(dynamicEntityCompiler, modelPackageName,
+                            includeAsyncModel, includeDynamicModel),
+                    options,
+                    dynamicEntityCompiler.getClassLoader());
 
+        } catch (ClassNotFoundException e) {
+            // TODO Auto-generated catch block
+            e.printStackTrace();
```
moizarafat, line 96: We should probably not catch the exception and let the error propagate.

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -89,4 +133,39 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
+        return annotatedClass;
+    }
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public static Set<Class> populateBindClasses(ElideDynamicEntityCompiler compiler,
+            Class annotationClass) throws ClassNotFoundException {
+        Set<Class> bindClasses = new HashSet<>();
+        List<String> dynamicClasses = ElideDynamicEntityCompiler.classNames;
+        for (String dynamicClass : dynamicClasses) {
+            Class<?> bindClass = compiler.getClassLoader().loadClass(dynamicClass).getClass();
```
moizarafat, line 164: Do we need to use getClass() here, wont doing just loadClass work?

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -89,4 +133,39 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
+     * Find classes with a particular annotation from dynamic compiler.
+     * @param compiler An instance of ElideDynamicEntityCompiler.
+     * @param annotationClass Annotation to search for.
+     * @return Set of Classes matching the annotation.
+     * @throws ClassNotFoundException
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public static List<String> findAnnotatedClasses(ElideDynamicEntityCompiler compiler, Class annotationClass)
+            throws ClassNotFoundException {
+        List<String> annotatedClass = new ArrayList<String>();
```
moizarafat, line 147: annotatedClasses instead of annotatedClass

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
moizarafat: if only loadClass works, there is common code between populateBindClasses and findAnnotatedClasses. Maybe we call populateBindClasses in findAnnotatedClasses and then traverse the list to get the names?

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -49,24 +51,36 @@
 
     private static MetricRegistry metricRegistry = null;
     private static HealthCheckRegistry healthCheckRegistry = null;
+    private ElideDynamicEntityCompiler dynamicEntityCompiler = null;
```
moizarafat, line 54: This is already defined as static in Util. We should reuse from there.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -49,24 +51,36 @@
...
-     * @param injector Injection instance for application
+     * @param injector Injection instance for application.
      */
     @Inject
     public ElideResourceConfig(ServiceLocator injector, @Context ServletContext servletContext) {
         this.injector = injector;
 
         settings = (ElideStandaloneSettings) servletContext.getAttribute(ELIDE_STANDALONE_SETTINGS_ATTR);
 
+        dynamicEntityCompiler = new ElideDynamicEntityCompiler(settings.getDynamicConfigPath());
```
moizarafat, line 67: This piece of code is duplicated between Util.getEntityManager and here. Instead this could be a separate method in Util like initialize dynamic compiler which takes a boolean argument derived from enableDynamicModelConfig . And based on the boolean value and if dynamicEntityCompiler is null, dynamicEntityCompiler will be initialized and classes compiled.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -110,27 +124,45 @@ protected void configure() {
...
+
+                        Info info = new Info().title("Dynamic models Service").version("1.0");
+                        SwaggerBuilder builder = new SwaggerBuilder(dictionary, info);
+
+                        //Default value of getJsonApiPathSpec() ends with /* at the end. need to remove.
+                        String dynamicBasePath = settings.getJsonApiPathSpec().replaceAll("/\\*", "");
+
+                        Swagger swagger = builder.build().basePath(dynamicBasePath);
+
+                        swaggerDocs.put("dynamic model", swagger);
```
moizarafat, line 164: lets have a singular name instead of with spaces. Just "dynamic" instead of "dynamic model";

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -110,27 +124,45 @@ protected void configure() {
...
+                        EntityDictionary dictionary = new EntityDictionary(new HashMap<>());
+                        for (Class entity : dynamicEntityCompiler.getBindClasses()) {
+                            dictionary.bindEntity(entity);
+                        }
+
+                        Info info = new Info().title("Dynamic models Service").version("1.0");
+                        SwaggerBuilder builder = new SwaggerBuilder(dictionary, info);
+
+                        //Default value of getJsonApiPathSpec() ends with /* at the end. need to remove.
+                        String dynamicBasePath = settings.getJsonApiPathSpec().replaceAll("/\\*", "");
```
moizarafat, line 160: Since this is logic is same between both blocks, we can get it out in a separate common variable.

elide-standalone/src/main/java/com/yahoo/elide/standalone/dynamic/config/ElideDynamicEntityCompiler.java
```diff
@@ -0,0 +1,122 @@
...
+    public ClassLoader getClassLoader() {
+        return inMemoryJavaCompiler.getClassloader();
+    }
+
+    /**
+     * getter for compiled dynamic class.
+     * @param name - class name
+     * @return compiled class
+     */
+    public Class<?> getCompiled(String name) {
```
moizarafat, line 110: This method is not used anywhere. we should remove it.

elide-standalone/src/test/java/com/yahoo/elide/standalone/ElideStandaloneTest.java
```diff
@@ -89,79 +102,122 @@ public void shutdown() throws Exception {
...
+                datum(
+                        resource(
+                                type("player"),
+                                id("ready-player-1"),
+                                attributes(
+                                        attr("name", "player1"),
+                                        attr("countryCode", "USA"),
+                                        attr("playerCountry", "USA"),
+                                        attr("highScore", 100)
+                                        //                        attr("createdOn", "2020-01-01"); still need to be tested with date
```
moizarafat, line 120: Please check the spring PR. I think we were able to successfully add it there. I think it should be like this 2020-01-01T00:00Z

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -89,4 +133,39 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
+            Class<?> classz = compiler.getClassLoader().loadClass(dynamicClass);
+            if (classz.getAnnotation(annotationClass) != null) {
+                annotatedClass.add(classz.getName());
+            }
+        }
+        return annotatedClass;
+    }
+
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public static Set<Class> populateBindClasses(ElideDynamicEntityCompiler compiler,
```
moizarafat: We can rename this to findAnnotatedClasses.

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -53,28 +64,61 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
-
+    public static List<String> combineModelEntities(ElideDynamicEntityCompiler compiler, String modelPackageName,
+            boolean includeAsyncModel, boolean includeDynamicModel) throws ClassNotFoundException {
         List<String> modelEntities = getAllEntities(modelPackageName);
-
         if (includeAsyncModel) {
             modelEntities.addAll(getAllEntities(AsyncQuery.class.getPackage().getName()));
         }
+        if (includeDynamicModel) {
+            modelEntities.addAll(findAnnotatedClasses(compiler, Entity.class));
```
moizarafat, line 120: We need to add classes of all 3 types Entity, FromTable, FromSubQuery

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -89,4 +133,39 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
+
+    /**
+     * Find classes with a particular annotation from dynamic compiler.
+     * @param compiler An instance of ElideDynamicEntityCompiler.
+     * @param annotationClass Annotation to search for.
+     * @return Set of Classes matching the annotation.
+     * @throws ClassNotFoundException
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public static List<String> findAnnotatedClasses(ElideDynamicEntityCompiler compiler, Class annotationClass)
```
moizarafat, line 145: we can call this findAnnotatedClassNames.

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -89,4 +135,39 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
+
+    /**
+     * Find classes with a particular annotation from dynamic compiler.
+     * @param compiler An instance of ElideDynamicEntityCompiler.
+     * @param annotationClass Annotation to search for.
+     * @return Set of Classes matching the annotation.
+     * @throws ClassNotFoundException ClassNameNotFound in compiler
+     */
+    @SuppressWarnings({ "unchecked", "rawtypes" })
+    public static List<String> findAnnotatedClassNames(ElideDynamicEntityCompiler compiler, Class annotationClass)
```
aklish, line 147: Maybe this can be a non-static method on the compiler (to return the list of classes that it manages).

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -24,8 +32,10 @@
  */
 public class Util {
 
-    public static EntityManagerFactory getEntityManagerFactory(String modelPackageName, boolean includeAsyncModel, 
-            Properties options) {
+    public static ElideDynamicEntityCompiler dynamicEntityCompiler;
+
+    public static EntityManagerFactory getEntityManagerFactory(String modelPackageName, boolean includeAsyncModel,
+            boolean includeDynamicModel, String dynamicConfigPath, Properties options) throws Exception {
```
aklish, line 38: Instead of includingDynamicModel - maybe have an` Optional<DynamicEntityCompiler>`

elide-standalone/src/main/java/com/yahoo/elide/standalone/Util.java
```diff
@@ -51,30 +61,66 @@ public static EntityManagerFactory getEntityManagerFactory(String modelPackageNa
...
+            Collection<ClassLoader> classLoaders = new ArrayList<>();
+            classLoaders.add(dynamicEntityCompiler.getClassLoader());
+            options.put(AvailableSettings.CLASSLOADERS, classLoaders);
 
-        return new EntityManagerFactoryBuilderImpl(
-                new PersistenceUnitInfoDescriptor(persistenceUnitInfo), new HashMap<>())
-                .build();
+            persistenceUnitInfo = new PersistenceUnitInfoImpl("elide-stand-alone",
+                    combineModelEntities(dynamicEntityCompiler, modelPackageName,
+                            includeAsyncModel, includeDynamicModel),
```
aklish, line 82: Instead of creating the PersistentUnitInfoImpl in the if/else - this can be simplified by just setting the classloader and list of packages to include in the if/else - and then using those to create the PersistentUnitInfo outside the control block.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -173,4 +204,15 @@ public static HealthCheckRegistry getHealthCheckRegistry() {
 
         return healthCheckRegistry;
     }
+
+    private Swagger setupSwagger(EntityDictionary dictionary, String path, String title) {
+        Info info = new Info().title(title).version("1.0");
```
aklish, line 209: Elide 5 now interprets the version set in the Swagger document.  If the non-elide models don't have this version - they vanish.  

Don't set a version.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -110,27 +134,34 @@ protected void configure() {
...
                     }
+                    // bind dynamic models
+                    if (settings.enableDynamicModelConfig()) {
+                        EntityDictionary dictionary = new EntityDictionary(new HashMap<>());
+                        try {
+                            for (Class entity : Util.dynamicEntityCompiler.findAnnotatedClasses(Entity.class)) {
+                                dictionary.bindEntity(entity);
+                            }
+                        } catch (ClassNotFoundException e) {
+                            log.error("error while binding class");
```
aklish, line 158: We can at least log the class name since we are swallowing the exception.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -110,27 +134,34 @@ protected void configure() {
...
-                        String asyncBasePath = settings.getJsonApiPathSpec().replaceAll("/\\*", "");
-
-                        Swagger swagger = builder.build().basePath(asyncBasePath);
-
+                        Swagger swagger = setupSwagger(dictionary, settings.getJsonApiPathSpec(), "Async Service");
                         swaggerDocs.put("async", swagger);
                     }
+                    // bind dynamic models
+                    if (settings.enableDynamicModelConfig()) {
+                        EntityDictionary dictionary = new EntityDictionary(new HashMap<>());
```
aklish, line 152: The swagger builder doesn't need a separate dictionary to only build a subset of models.  YOu can provide it the list of classes to load.  That's the recommended approach.  Let's follow that.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -51,30 +53,52 @@
...
 
         // Bind to injector
         register(new AbstractBinder() {
             @Override
             protected void configure() {
-                ElideSettings elideSettings = settings.getElideSettings(injector);
+                ElideSettings elideSettings = null;
+                try {
+                    elideSettings = settings.getElideSettings(injector);
+                } catch (Exception e) {
```
aklish, line 98: What exception are we catching here an why?

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -51,30 +53,52 @@
...
     @Inject
     public ElideResourceConfig(ServiceLocator injector, @Context ServletContext servletContext) {
         this.injector = injector;
 
         settings = (ElideStandaloneSettings) servletContext.getAttribute(ELIDE_STANDALONE_SETTINGS_ATTR);
 
+        if (settings.enableDynamicModelConfig()) {
+            try {
+                Util.initDynamicConfig(settings.getDynamicConfigPath());
+            } catch (Exception e) {
```
aklish, line 70: What exception are we catching here and why.  We need to be very careful about catching exception - in general - not advised.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
moizarafat: The DynamicEntityompiler compile method throws an Exception so this was catching and throwing that. Have updated DynamicCompiler to catch it and throw a run time exception now instead of compile time.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
moizarafat: The DynamicEntityompiler compile method throws an Exception so this was catching and throwing that. Have updated DynamicCompiler to catch it and throw a run time exception now instead of compile time.

elide-standalone/src/test/resources/models/tables/PostView.hjson
```diff
@@ -0,0 +1,20 @@
...
+        '''
+        A long description
+        '''
+        cardinality : large
+        readAccess : Prefab.Role.All
+        dimensions : [
+            {
+                name : content
+                type : TEXT
+                readAccess : Prefab.Role.All
```
aklish, line 15: Redundant readAccess.

elide-standalone/src/test/java/example/ElideStandaloneTest.java
```diff
@@ -159,6 +162,20 @@ public void testJsonAPIPost() {
             .then()
             .statusCode(HttpStatus.SC_CREATED)
             .extract().body().asString();
+
+        // Test the Dynamic Generated Analytical Model is accessible
+        given()
+            .when()
+            .get("/api/v1/postView")
+            .then()
```
aklish, line 170: You can combine this test with the one below in a single request (test the body and the status code).

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -286,11 +339,150 @@ default void updateServletContextHandler(ServletContextHandler servletContextHan
...
+    }
+
+    /**
+     * Gets the DataStore for elide.
+     * @param queryEngine query engine object.
+     * @param aggregationDataStore AggregationDataStore object.
+     * @param entityManagerFactory EntityManagerFactory object.
+     * @return EntityDictionary object initialized.
+     */
+    default DataStore getDataStore(SQLQueryEngine queryEngine, AggregationDataStore aggregationDataStore,
```
aklish, line 376: This should take the MetadataStore instead of the SQLQueryEngine as an argument.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -286,11 +339,150 @@ default void updateServletContextHandler(ServletContextHandler servletContextHan
...
+        return dataStore;
+    }
+
+    /**
+     * Gets the AggregationDataStore for elide.
+     * @param queryEngine query engine object.
+     * @param optionalCompiler optional dynamic compiler object.
+     * @return AggregationDataStore object initialized.
+     */
+    default AggregationDataStore getAggregationDataStore(SQLQueryEngine queryEngine,
```
aklish, line 394: INstead of binding this to the SQLQueryEngine - can we just bind it to QueryEngine?

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -58,47 +59,62 @@
...
+                        settings.enableAsync(), optionalCompiler, settings.getDatabaseProperties());
+
+                EntityDictionary dictionary = settings.getEntityDictionary(injector, optionalCompiler);
+
+                MetaDataStore metaDataStore = settings.getMetaDataStore(optionalCompiler);
+
+                SQLQueryEngine queryEngine = settings.getSQLQueryEngine(metaDataStore, entityManagerFactory);
+
+                DataStore dataStore = settings.getDataStore(
+                        settings.getSQLQueryEngine(metaDataStore, entityManagerFactory),
```
aklish, line 98: You shouldn't have to call settings.getSQLQueryEngine twice here.

PR #1270, opened 2020-04-15
Refactored metric SQL expansion to occur dynamically at query time ra…

aklish:
…ther than statically during service initialization

## Description
Metric SQL generation should be done at query time rather than resolving the references statically during service startup.  This allows us to to support parameterized metrics.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/query/SQLMetricProjection.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2019, Yahoo Inc.
```
jkusa, line 2:
```suggestion
 * Copyright 2020, Yahoo Inc.
```

PR #1275, opened 2020-04-21
Short Circuit Filter Expression Security If User Checks would pass/fail.

aklish:
## Description
Evaluating security for filter expressions currently fetches the relationships involved in the filter expression.  This can result in unnecessary database queries if the only security rules are User checks.

## Motivation and Context
Remove unnecessary database calls when we can evaluate security statically.

## How Has This Been Tested?
Unit Tests - TBD

## License


wcekan:
I like the optimization, and will look again after the new unit tests.

elide-core/src/main/java/com/yahoo/elide/core/VerifyFieldAccessFilterExpressionVisitor.java
```diff
@@ -74,6 +85,28 @@ public Boolean visitPredicate(FilterPredicate filterPredicate) {
...
+            try {
+                result = executor.checkUserPermissions(
+                        element.getType(),
+                        ReadPermission.class,
+                        element.getFieldName());
+            } catch (ForbiddenAccessException e) {
+                return ExpressionResult.FAIL;
+            }
+
+            if (result == ExpressionResult.DEFERRED) {
```
wcekan, line 103: If this is FAIL should we return FAIL here too?

PR #1281, opened 2020-04-25
Better errors for missing IDs in Patch Extension Request. (#1278)

wcekan:
## Description
Merge #1278 into Elide-5.x

## Motivation and Context
Test coverage

## How Has This Been Tested?
Unit testing

## License

### DIFF
--- a/elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
+++ b/elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
@@ -1394,6 +1394,23 @@ public class ResourceIT extends IntegrationTest {
     }
 
     @Test
+    public void invalidPatchMissingId() {
+        String request = jsonParser.getJson("/ResourceIT/invalidPatchMissingId.req.json");
+
+        String detail = "Bad Request Body'Patch extension requires all objects to have an assigned "
+                + "ID (temporary or permanent) when assigning relationships.'";
+
+        given()
+                .contentType(JSONAPI_CONTENT_TYPE_WITH_JSON_PATCH_EXTENSION)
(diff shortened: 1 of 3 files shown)


wcekan:
Current failure not expected.
```json
[{
  "errors": [{
    "detail": "Bad Request Body&#39;Patch extension requires all objects to have an assigned ID (temporary or permanent) when assigning relationships.&#39;",
    "status": "400"
  }]
},
{
  "errors": [{
    "detail": "Operation not executed. Terminated by earlier failure."
  } ]
}]
```

elide-integration-tests/src/test/java/com/yahoo/elide/tests/ResourceIT.java
```diff
@@ -1393,6 +1394,23 @@ public void createChildRelateExisting() {
...
+                + "ID (temporary or permanent) when assigning relationships.'";
+
+        given()
+                .contentType(JSONAPI_CONTENT_TYPE_WITH_JSON_PATCH_EXTENSION)
+                .accept(JSONAPI_CONTENT_TYPE_WITH_JSON_PATCH_EXTENSION)
+                .body(request)
+                .patch("/")
+                .then()
+                .statusCode(HttpStatus.SC_BAD_REQUEST)
+                .body("errors[0].detail[0]", equalTo(Encode.forHtml(detail)));
```
wcekan, line 1411: @aklish any idea why detail is now treated as an array here?

aklish: That might be a bug.  I checked JSON-API spec - and I think that should just be a string.

wcekan: Yes, I suspect it might be processed as
`(errors[0].detail)[0]` due to the outer and inner arrays.

PR #1291, opened 2020-04-27
Issue 1197 add memberof operator

### LINKED ISSUE #1197: Add MEMBER OF Operation
Elide should support MEMBER OF operation to check if a value is a member of the specified collection.

## Expected Behavior
Member of operation should work only on Collection attributes. We do not support the relationship.
-->`filter=book.awards=hasmember='Booker Prize'`
-->`filter=book.awards=hasnomember='Hugo Award'`

## Possible Solution
Should replicate hibernate behavior.
https://docs.oracle.com/html/E13946_04/ejb3_langref.html#ejb3_langref_collection_member

Chandrasekar-Rajasekar:
Resolves #1197

## Description
Add the `member of` operation for only Element Collection attributes for root entity or toOne relationship entity.
`filter[book]=awards=hasmember='Booker Prize'`
`filter[book]=publisher.phoneNumbers=hasmember='999-999-9999'`

Enables to filter on collection attribute

## License


elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -350,6 +366,25 @@ public static Operator fromString(final String string) {
         };
     }
 
+    private static <T> Predicate<T> hasMember(String field, List<Object> values, RequestScope requestScope) {
+        return (T entity) -> {
+            if (values.size() != 1) {
+                throw new InvalidPredicateException("HasMember can only take one argument");
+            }
+            Object val = getFieldValue(entity, field, requestScope);
+            String filterStr = CoerceUtil.coerce(values.get(0), String.class);
```
aklish, line 375: coerce into val.getClass().  We probably want to support lists of numbers too.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/DefaultFilterDialect.java
```diff
@@ -111,12 +112,19 @@ public FilterExpression parseGlobalExpression(String path, MultivaluedMap<String
                 throw new ParseException(String.format("Invalid predicate: %s", filterPredicate));
             }
 
+            if ((filterPredicate.getOperator().equals(Operator.HASMEMBER)
```
aklish, line 115: This check seems redundant with the validation logic below.  Do we also need this check?

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/filter/FilterTranslator.java
```diff
@@ -170,6 +172,21 @@
             return String.format("%s IS NOT EMPTY", columnAlias);
         });
 
+        operatorGenerators.put(HASMEMBER, (columnAlias, params) -> {
+            Preconditions.checkArgument(params.size() == 1);
+            String x =  String.format("%s MEMBER OF %s",
```
aklish, line 177: Can we remove the local variable here?

elide-datastore/elide-datastore-hibernate3/pom.xml
```diff
@@ -40,7 +40,7 @@
 
     <properties>
         <dataStoreHarness>com.yahoo.elide.datastores.hibernate3.HibernateDataStoreHarness</dataStoreHarness>
-        <excludeTags>emptyOnAttributeCollection</excludeTags>
+        <excludeTags>emptyOnAttributeCollection,memberOfOperation</excludeTags>
```
aklish, line 43: In Elide 5 - we make tags for stores (rather than features).

elide-integration-tests/src/test/java/com/yahoo/elide/tests/FilterIT.java
```diff
@@ -1654,6 +1655,146 @@ void testExceptionOnEmptyOperator() throws IOException {
...
+                    awardBook.add(book.get("id"));
+                    break;
+                }
+            }
+        }
+        // Test Default filter type on Root Entity
+        result = getAsNode(String.format("/book?filter[book.awards][hasmember]=%s", filterString));
+        assertEquals(awardBook.size(), result.get("data").size());
+        for (JsonNode book : result.get("data")) {
+            assertTrue(awardBook.contains(book.get("id")));
```
aklish, line 1681: There is a much cleaner way to do these comparisons with hamcrest Matchers.  There are some good examples in PaginationIT.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/DefaultFilterDialect.java
Chandrasekar-Rajasekar: This check is for Global Filter. The validation logic below is for the typed filter. Previously we did not have `toMany` relationship check for global filter. So I decided to include only Iscollection check for global filter.

Without this check in global filter, we allow filter expr like `book?filter=title=hasmember='1'` to pass through and jpql rejects these statement.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
Chandrasekar-Rajasekar: Object `val` will be an instance of Collection. To get the parameterized type of collection, I think we need to iterate over the path and get paremeterizedType from entityDict. I am not sure how to get parameterized type from Collection object

aklish: You are right, we'll need to iterate over it.  
Another alternative is to convert field into a Path object in Elide.  That will do all the type conversion for you.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -350,6 +367,25 @@ public static Operator fromString(final String string) {
         };
     }
 
+    private static <T> Predicate<T> hasMember(Path fieldPath, List<Object> values, RequestScope requestScope) {
+        return (T entity) -> {
+            if (values.size() != 1) {
+                throw new InvalidPredicateException("HasMember can only take one argument");
+            }
+            Object val = getFieldValue(entity, fieldPath, requestScope);
+            String filterStr = CoerceUtil.coerce(values.get(0), String.class);
```
aklish, line 376: We still need to fix the coercion.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/FilterIT.java
```diff
@@ -1668,6 +1674,156 @@ void testExceptionOnEmptyOperator() throws IOException {
...
+                .then()
+                .body("data", hasSize(publisherBook.size()),
+                        "data.id", contains(publisherBook.toArray())
+                );
+
+    }
+
+    @Test
+    @Tag("excludeOnHibernate3")
+    void testExceptionOnMemberOfOperator() throws IOException {
```
aklish, line 1777: Can we also add error tests for book.authors and book.publisher?

PR #1294, opened 2020-04-29
fixes for Async Models lifecycle hooks failing

moizarafat:
## Description
This is a fix for Async Lifecycle hooks failing due to AsyncExecutorService not initialized. ExecuteQueryHook is now manually binded to dictionary after the AsyncExecutorService is initialized.

## Motivation and Context
With recent change in LifeCycleHooks, all hooks are moved to a separate class and injected separately. The injection of hooks happen when elide comes up. Async Hooks inject AsyncExecutorService which is not initialized at that moment as it needs Elide to be initialized. 

## How Has This Been Tested?
Elide Standalone Example has been used for testing.
Elide SpringBoot Example has been used for testing.

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
@@ -23,7 +23,7 @@ import javax.persistence.PrePersist;
 
 /**
  * Model for Async Query.
- * ExecuteQueryHook & UpdatePrincipalNameHook is binded manually during the elide startup,
+ * ExecuteQueryHook and UpdatePrincipalNameHook is binded manually during the elide startup,
  * after asyncexecutorservice is initialized.
  */
 @Entity
--- a/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
@@ -6,7 +6,6 @@
 package com.yahoo.elide.async.models;
 
 import static com.yahoo.elide.annotation.LifeCycleHookBinding.Operation.CREATE;
-import static com.yahoo.elide.annotation.LifeCycleHookBinding.TransactionPhase.POSTCOMMIT;
 import static com.yahoo.elide.annotation.LifeCycleHookBinding.TransactionPhase.PRESECURITY;
 
 import com.yahoo.elide.annotation.DeletePermission;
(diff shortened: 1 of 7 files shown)


elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -30,14 +28,15 @@
...
+ * ExecuteQueryHook is binded manually during the elide startup,
+ * after asyncexecutorservice is initialized.
  */
 @Entity
 @Include(type = "asyncQuery", rootLevel = true)
 @ReadPermission(expression = "Principal is Owner")
 @UpdatePermission(expression = "Prefab.Role.None")
 @DeletePermission(expression = "Prefab.Role.None")
 @LifeCycleHookBinding(hook = UpdatePrincipalNameHook.class, operation = CREATE, phase = PRESECURITY)
-@LifeCycleHookBinding(hook = ExecuteQueryHook.class, operation = CREATE, phase = POSTCOMMIT)
```
aklish, line 40: Might as well make this consistent - let's remove all life cycle hook annotations from this class.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -72,4 +70,10 @@ public AsyncCleanerService buildAsyncCleanerService(Elide elide, ElideConfigProp
     public AsyncQueryDAO buildAsyncQueryDAO(Elide elide) {
         return new DefaultAsyncQueryDAO(elide, elide.getDataStore());
     }
+
+    @Override
+    public void afterPropertiesSet() {
+        ExecuteQueryHook executeQueryHook = new ExecuteQueryHook(asyncExecutorService);
```
aklish, line 76: I thought we were going to do all of this logic in the AsyncExecutorService bean?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAutoConfiguration.java
```diff
@@ -144,4 +147,20 @@ public Swagger buildSwagger(EntityDictionary dictionary, ElideConfigProperties s
...
+    /**
+     * Configure the AsyncExecutorService used for submitting async query requests.
+     * @param elide elideObject.
+     * @param settings Elide settings.
+     * @return a AsyncExecutorService.
+     */
+    @Bean
+    @ConditionalOnMissingBean
+    @ConditionalOnProperty(prefix = "elide.async", name = "enabled", matchIfMissing = false)
+    public AsyncExecutorService buildAsyncExecutorService(Elide elide, ElideConfigProperties settings,
```
aklish, line 160: Why not leave this in the ElideAsyncConfiguration file?

PR #1297, opened 2020-05-01
Implement equals/hashCode and immutability where needed

john-karp:
## Description
Make sure classes used in queries have correct equals/hashCode implementations, using Lombok if possible.

## Motivation and Context
`Query` has `Set<ColumnProjection>` and `Set<TimeDimensionProjection>` fields, but for those sets to work as expected, `ColumnProjection` and `TimeDimensionProjection` need to implement equals/hashCode correctly, as well as classes held by those classes and their superclasses.

## How Has This Been Tested?
UTs are passing.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/request/Argument.java
+++ b/elide-core/src/main/java/com/yahoo/elide/request/Argument.java
@@ -7,13 +7,13 @@
 package com.yahoo.elide.request;
 
 import lombok.Builder;
-import lombok.Data;
 import lombok.NonNull;
+import lombok.Value;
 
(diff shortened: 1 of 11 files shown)


john-karp:
Seems like codacy doesn't recognize that the `@Value` annotation makes fields `private final`, so it complains "Use explicit scoping instead of the default package private level". Not sure if I should ignore the codacy warning or add the redundant qualifiers.

aklish:
Codacy complains about package level permissions.  It prefers private or public.  Any reason not to have both Value and private?

PR #1304, opened 2020-05-06
Fixing GraphQL 500 errors due to client API errors

aklish:
## Description
There is a bug in Elide's GraphQL implementation where it returns jaxrs exceptions instead of HttpStatusExceptions.

These exceptions weren't being handle correctly and resulted in the server returning 500 error.

## Motivation and Context
Some client errors result in the server throwing a 500 error.

## How Has This Been Tested?
Added an IT test for one example error.

## License


elide-core/src/main/java/com/yahoo/elide/core/exceptions/InvalidPredicateException.java
```diff
@@ -10,6 +10,7 @@
 /**
  * Invalid predicate exception.
  */
+@Deprecated
 public class InvalidPredicateException extends HttpStatusException {
```
wcekan, line 14: Should we just have it extend BadRequestException?
```suggestion
public class InvalidPredicateException extends BadRequestException {
```

PR #1305, opened 2020-05-06
GraphQL Many-to-many filter Bug Fix

codingwhatever:
Resolves - Could not find any related issues

## Description
The GraphQL interface has a bug related to filtering values through a many-to-many relationship. An example of a broken filter would be `books.title=="1984"` on a request for the author object from the UT and integration test model. The books set in the author entity is a many-to-many relationship to the book entity. The filter fails to parse due to type errors since the `MultipleFilterDialect. parseTypedExpression` method does not properly handle parsing the nested types. 

The solution is to attempt to parse with the `MultipleFilterDialect.parseGlobalExpression` similar to the existing solution in the JsonAPI interface in `RequestScope`.

## Motivation and Context
This bug unnecessarily limits how our UI can query the GraphQL endpoint.

## How Has This Been Tested?
I have added multiple integration tests in this PR to ensure that different filters work with this change. All existing test continue to pass unchanged.

## License

### DIFF
--- a/elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -570,7 +570,6 @@ public class PersistentResourceFetcher implements DataFetcher<Object> {
                     filterKey += "[" + typeName + "]";
                 }
                 put(filterKey, Arrays.asList(filterStr));
-
             }
         };
     }
@@ -581,11 +580,12 @@ public class PersistentResourceFetcher implements DataFetcher<Object> {
         // TODO: Refactor FilterDialect interfaces to accept string or List<String> instead of (or in addition to?)
         // query params.
         return filter.map(filterStr -> {
+            String errorMessage = "";
             try {
                 return requestScope.getFilterDialect()
                         .parseGlobalExpression(typeName, getQueryParams(Optional.empty(), filterStr));
             } catch (ParseException e) {
-                log.debug(e.getMessage());
+                errorMessage = e.getMessage();
             }
 
             try {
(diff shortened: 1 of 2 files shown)


elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -563,22 +562,38 @@ private ConnectionContainer replaceObjects(Environment context) {
...
                                                    RequestScope requestScope) {
         // TODO: Refactor FilterDialect interfaces to accept string or List<String> instead of (or in addition to?)
         // query params.
         return filter.map(filterStr -> {
-            MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>() {
-                {
-                    put("filter[" + typeName + "]", Arrays.asList(filterStr));
-                }
-            };
+            String errorMessage = "";
```
wcekan, line 583: Unnecessary initialization.
```suggestion
            String errorMessage;
```

elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -563,22 +562,38 @@ private ConnectionContainer replaceObjects(Environment context) {
...
+
             try {
-                return requestScope.getFilterDialect().parseTypedExpression(typeName, queryParams).get(typeName);
+                return requestScope.getFilterDialect()
+                        .parseTypedExpression(typeName, getQueryParams(Optional.of(typeName), filterStr))
+                        .get(typeName);
             } catch (ParseException e) {
-                log.debug("Filter parse exception caught", e);
-                throw new InvalidPredicateException("Could not parse filter for type: " + typeName);
+                throw new InvalidPredicateException(errorMessage + "\n" + e.getMessage());
```
wcekan, line 596: @aklish will this conflict with #1304?

aklish: It will - but should be simple to merge.

PR #1245, opened 2020-03-30
Elide 5.x async tests

Abhino:
Elide Async Feature Tests

Co-authored-by: moizarafat

Related pull requests:
elide-async changes for Elide-Async: https://github.com/yahoo/elide/pull/1203

## Description
These changes add the Integration and unit tests for the Async Module

## How Has This Been Tested?
Tests performed using integration and unit tests.

## License
I confirm that this contribution is made under an Apache 2.0 license and that I have the authority necessary to make this contribution on behalf of its copyright owner.

PR #1311, opened 2020-05-08
Elide Async Feature Unit and Integration Tests

Abhino:
Elide Async Feature Unit and Integration Tests

Co-authored-by: moizarafat

Related pull requests:
Old PR (Closed since we did a merge instead of rebase) : https://github.com/yahoo/elide/pull/1245
Main code elide-async changes for Elide-Async: https://github.com/yahoo/elide/pull/1203

## Description
These changes add the Integration and unit tests for the Async Module. Here is a summary of the tests
1) Tests and validations for a JSONAPI query to the Async API
2) Tests and validations  for a GRAPHQL query to the Async API
3) Tests and validations for bad requests to the Async API
4) Tests for making a Async request to a model to which the user does not have permissions

Further additional basic tests have been added to test for Async Module for Elide-Standalone and Elide-Spring.

Also this includes the change Fixing legacy Hibernate entity manager store (so it doesn't recycle the entity manager): https://github.com/yahoo/elide/pull/1307
## How Has This Been Tested?
Tests performed using integration and unit tests against all datastores.

## License


elide-async/src/test/java/com/yahoo/elide/async/service/AsyncExecutorServiceTest.java
```diff
@@ -0,0 +1,86 @@
...
+        ObjectMapper objectMapper = mock(ObjectMapper.class);
+        Set<Class<?>> boundclasses = new HashSet<Class<?>>();
+        boundclasses.add(AsyncQuery.class);
+        boundclasses.add(AsyncQueryResult.class);
+
+        Set<String> apiCollection = new HashSet<String>();
+        apiCollection.add("v1");
+
+        when(elide.getElideSettings()).thenReturn(elideSettings);
+        when(elideSettings.getDictionary()).thenReturn(dictionary);
```
aklish, line 56: Normally, we don't mock the dictionary or the elide settings.  Is there a reason you can't just create a dictionary here and use it?

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryInterruptThreadTest.java
```diff
@@ -0,0 +1,83 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.util.Date;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+public class AsyncQueryInterruptThreadTest {
```
aklish, line 29: Given that the interrupt thread will be removed/rewritten, you can just delete the tests for now.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryThreadTest.java
```diff
@@ -0,0 +1,105 @@
...
+        verify(asyncQueryDao, times(1)).createAsyncQueryResult(anyInt(), anyString(), any(), any());
+    }
+
+    @Test
+    public void testProcessQueryNoResponse() {
+        String query = "{\"query\":\"{ group { edges { node { name commonName description } } } }\",\"variables\":null}";
+
+        when(queryObj.getQuery()).thenReturn(query);
+        when(queryObj.getQueryType()).thenReturn(QueryType.GRAPHQL_V1_0);
+        when(runner.run(query, user)).thenReturn(null);
```
aklish, line 98: In practice, this doesn't happen.  What does happen is that the queryRunner might throw an exception.  We should mock and test that scenario.

elide-datastore/pom.xml
```diff
@@ -100,7 +100,11 @@
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-engine</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-async</artifactId>
```
aklish, line 105: Why does the datastore depend on async?

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/AsyncIT.java
```diff
@@ -0,0 +1,541 @@
...
+                .then()
+                .statusCode(HttpStatus.SC_CREATED);
+
+    }
+
+    /**
+     * Various tests for a JSONAPI query as a Async Request.
+     */
+    @Test
+    @Tag("skipInMemory")
```
aklish, line 117: Why do we skip in memory?  As a best practice, I always put the reason so that we can remove these over time.  What doesn't work in memory?

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/AsyncIT.java
```diff
@@ -0,0 +1,541 @@
...
+                                        )
+                                )
+                        ).toJSON())
+                .when()
+                .post("/asyncQuery")
+                .then()
+                .statusCode(org.apache.http.HttpStatus.SC_CREATED);
+
+        int i = 0;
+        while (i < 1000) {
```
aklish, line 141: How does the test fail if this loop finishes without going to COMPLETE?

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/AsyncIT.java
```diff
@@ -0,0 +1,541 @@
...
+                break;
+            }
+        }
+    }
+
+    /**
+     * Various tests for a GRAPHQL query as a Async Request.
+     */
+    @Test
+    @Tag("skipInMemory")
```
aklish, line 221: Same comment - why do we do this?

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -38,10 +39,13 @@
  */
 @Singleton
 @Slf4j
+@Getter
 public class DefaultAsyncQueryDAO implements AsyncQueryDAO {
 
     @Setter private Elide elide;
     @Setter private DataStore dataStore;
+    @Setter private EntityDictionary dictionary;
```
aklish, line 47: I don't think we need setters here for testing.  Let's remove.

PR #1319, opened 2020-05-12
Add caching support to SQLQueryEngine

john-karp:
## Description
Adds support for a query result cache to `AggregationDataStore` (moved from `QueryEngine`). A new table annotation `@VersionQuery` is added to allow the user to specify a SQL query that will return a table-data-version number for that table, presumably something along the lines of a row-count query or last-modification-time query. This version is then included in the cache key, so that when the data changes, `QueryEngine` will not return any stale data. If a table does not have that annotation, cached data is always returned, no matter the age.

## Motivation and Context
API support for QueryEngine caching was added in #1297. However no implementations were provided. This PR adds an implementation for SQLQueryEngine.

## How Has This Been Tested?
New unit tests added.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
```diff
@@ -53,7 +53,7 @@ public void createObject(Object entity, RequestScope scope) {
     @Override
     public Iterable<Object> loadObjects(EntityProjection entityProjection, RequestScope scope) {
         Query query = buildQuery(entityProjection, scope);
-        return queryEngine.executeQuery(query, true);
+        return queryEngine.executeQuery(query);
```
aklish, line 56: Ideally, I think this is where the cache ought to be queried for entries.  This would allow the same cache to be used for multiple engines (Fili, SQL, etc).  I think the cache key generation logic itself could be common as well.  The key could be derived from the Query object itself (rather than the generated SQL) plus data freshness key (which the query engine could surface).

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -177,6 +184,56 @@ public MetricProjection constructMetricProjection(Metric metric,
...
+                "Running Query: " + paginationSQL
+        ).get();
+
+        if (paginationCacheKey != null) {
+            cache.put(paginationCacheKey, total);
+        }
+        return total;
+    }
+
+    private int getTableVersion(EntityManager entityManager, Query query) {
```
aklish, line 219: TableVersion as a string would be more flexible.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/annotation/VersionQuery.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2019, Yahoo Inc.
```
aklish, line 2: 2020

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -156,14 +151,32 @@ private void populateMetaData(MetaDataStore metaDataStore) {
                 .forEach(metaDataStore::addTable);
     }
 
+    public interface Transaction extends AutoCloseable {
```
aklish, line 154: I think we'll need some Javadoc here so the linter won't warn.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -156,14 +151,32 @@ private void populateMetaData(MetaDataStore metaDataStore) {
...
+        @Override
+        void close();
```
aklish, line 156: Same.  Also we are adding a cancel method in another PR.  Do see these as the same thing?  For cancel, the expectation is the method is thread safe.  If we consolidate them - we'll need the same comment here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -156,14 +151,32 @@ private void populateMetaData(MetaDataStore metaDataStore) {
...
+     * @param transaction
      * @return query results
      */
-    public abstract QueryResult executeQuery(Query query);
+    public abstract QueryResult executeQuery(Query query, Transaction transaction);
+
+    /**
+     * Get a serial number or other token indicating the version of the data in the table.
+     * No particular semantics are required, though it must change if the data changes.
+     * If one is not available, returns null.
```
aklish, line 174: You might want to mention that if one is not available, returns null which disables caching.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransactionTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+        String cacheKey = "foo;" + queryKey;
+        Mockito.verify(cache).get(cacheKey);
+        Mockito.verify(cache).put(cacheKey, queryResult);
+        Mockito.verifyNoMoreInteractions(cache);
+    }
+
+    @Test
+    void loadObjectsUsesCache() {
+        QueryResult queryResult = QueryResult.builder().data(DATA).build();
```
aklish, line 91: Does this test execution depend on the prior test executing first?  I think we should make each test stand on its own.

john-karp: No, the fields annotated with `@Mock`, such as the cache, get reset for each test, and the rest of the fields used in the tests are immutable. So `loadObjectsUsesCache` is inserting a cache entry before this test with `when(cache.get(anyString())).thenReturn(queryResult);`.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
john-karp: I assume if cancel() is called on a AggregationDataStoreTransaction that it ends the whole transaction, not just any pending calls such as loadObjects? (There won't be a use case for aborting loadObjects but still being able to use the same AggregationDataStoreTransaction after.)

PR #1322, opened 2020-05-15
TANDS-19093-adding-transactionRegistry-interface

### LINKED ISSUE #1292: RFC Elide 5.0 - TransactionRegistry

```java
    @Data
        RequestScope request;
    }

}
```

```java
public void cancel();
```

ramshakr:

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/TransactionRegistry.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/TransactionRegistry.java
@@ -18,7 +18,6 @@ public interface TransactionRegistry {
          * @see RequestScope
          * @see DataStoreTransaction 
          */
-
         public RequestScope request;
         public DataStoreTransaction transaction;
     }
--- a/elide-core/src/main/java/com/yahoo/elide/core/TransactionRegistry.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/TransactionRegistry.java
@@ -5,9 +5,9 @@
  */
 package com.yahoo.elide.core;
 
-import java.util.Set;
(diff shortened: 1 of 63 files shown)


elide-core/src/main/java/com/yahoo/elide/core/TransactionRegistry.java
```diff
@@ -0,0 +1,26 @@
+/*
+ *  * Copyright 2015, Yahoo Inc.
```
aklish, line 2: 2020

elide-core/src/main/java/com/yahoo/elide/core/TransactionRegistry.java
```diff
@@ -0,0 +1,26 @@
...
+ *   * Licensed under the Apache License, Version 2.0
+ *    * See LICENSE file in project root for terms.
+ *     */
+package com.yahoo.elide.core;
+
+/**
+ *  * Database interface library.
```
aklish, line 9: Can you remove the extra asterisks?

elide-core/src/main/java/com/yahoo/elide/core/TransactionRegistry.java
```diff
@@ -0,0 +1,25 @@
...
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core;
+import java.util.Set;
+/**
+ * Database interface library.
+ */
+public interface TransactionRegistry {
+
+    public static class TransactionEntry {
```
aklish, line 13: I think we might get build warnings for the lack of javadoc for the inner static class - and also the public methods below.  Can we add a bit of documentation?

PR #1329, opened 2020-05-18
CARBON-161  add requestId to requestScope

### LINKED ISSUE #1292: RFC Elide 5.0 - TransactionRegistry

AshishAg24:

## License


elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java
```diff
@@ -77,6 +78,7 @@
     @Getter private final int updateStatusCode;
 
     @Getter private final MultipleFilterDialect filterDialect;
+    private String requestId;
```
aklish, line 81: Can we make it final?

PR #1336, opened 2020-05-21
Fixes - Honor Doc Version in SwaggerController and Use ISO8601Dates Serde Format in Date Formatting 

moizarafat:
## Description
i) SwaggerController now honor's the model version provided while creating the document.
ii) Use a new Elide Object in DefaulyAsyncQueryDAO which has the date serde set.
iii) Reduces some warnings during build process.
iv) Remove assembly plugin from elide-dynamic-config-helpers pom.

A separate PR will be done for supporting epoch dates.

## Motivation and Context
Bug Fixes.

## How Has This Been Tested?
Existing Unit Test and Integrated Test passes.

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
@@ -12,7 +12,6 @@ import com.yahoo.elide.annotation.ReadPermission;
 import com.yahoo.elide.annotation.UpdatePermission;
 
 import lombok.Data;
-import lombok.EqualsAndHashCode;
 
 import javax.persistence.CascadeType;
 import javax.persistence.Column;
@@ -33,7 +32,6 @@ import javax.validation.constraints.Pattern;
 @UpdatePermission(expression = "Prefab.Role.None")
 @DeletePermission(expression = "Prefab.Role.None")
 @Data
-@EqualsAndHashCode(callSuper = false)
 public class AsyncQuery extends AsyncBase implements PrincipalOwned {
     @Id
     @Column(columnDefinition = "varchar(36)")
--- a/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
@@ -12,6 +12,7 @@ import com.yahoo.elide.annotation.ReadPermission;
 import com.yahoo.elide.annotation.UpdatePermission;
 
 import lombok.Data;
+import lombok.EqualsAndHashCode;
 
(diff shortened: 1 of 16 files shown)


elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQueryResult.java
```diff
@@ -29,6 +30,7 @@
 @CreatePermission(expression = "Prefab.Role.None")
 @DeletePermission(expression = "Prefab.Role.None")
 @Data
+@EqualsAndHashCode(callSuper = false)
```
aklish, line 33: This breaks correctness.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -32,6 +33,7 @@
 @UpdatePermission(expression = "Prefab.Role.None")
 @DeletePermission(expression = "Prefab.Role.None")
 @Data
+@EqualsAndHashCode(callSuper = false)
```
aklish, line 36: We have to use the natural key for equals/hashcode for Hibernate.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/SwaggerControllerProperties.java
```diff
@@ -6,11 +6,13 @@
 package com.yahoo.elide.spring.config;
 
 import lombok.Data;
+import lombok.EqualsAndHashCode;
 
 /**
  * Extra controller properties for the Swagger document endpoint.
  */
 @Data
+@EqualsAndHashCode(callSuper = false)
```
aklish, line 15: Same - you can remove callSuper = false.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -77,7 +78,8 @@ private String evaluateFormattedFilterDate(int calendarUnit, int amount) {
         cal.setTime(new Date());
         cal.add(calendarUnit, -(amount));
         Date filterDate = cal.getTime();
-        Format dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm'Z'");
+        ISO8601DateSerde serde = (ISO8601DateSerde) elide.getElideSettings().getSerdes().get(Date.class);
```
aklish, line 81: This may not be a safe cast.  There are other Serde's for Date besides ISO8601.  Also, the client has control over this so let's figure out what can work in all circumstances.

PR #1338, opened 2020-05-21
addTransaction-removeTransaction

### LINKED ISSUE #1292: RFC Elide 5.0 - TransactionRegistry

Rkr1992:

## License


elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -278,7 +283,10 @@ protected ElideResponse handleRequest(boolean isReadOnly, User user,
                                           Supplier<DataStoreTransaction> transaction,
                                           Handler<DataStoreTransaction, User, HandlerResult> handler) {
         boolean isVerbose = false;
+        UUID requestId = null;
```
AshishAg24, line 286: We are storing requestId as a string in requestScope. Can we do the same here? or is there an advantage to storing it like this?
https://github.com/yahoo/elide/pull/1331/files

Rkr1992: We are using it as a hash value in a Map so I  guess we can store it either way

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -412,4 +421,31 @@ public RequestScope getRequestScope() {
             return requestScope;
         }
     }
+
+     /**
+      * Transaction Registry class.
+      */
+    @Getter
+    public static class TransactionRegistry {
```
aklish, line 429: Can we move this outside Elide into its own file?

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -412,4 +421,31 @@ public RequestScope getRequestScope() {
...
+        private Map<UUID, DataStoreTransaction> transactionMap = new HashMap<>();
+        public Set<DataStoreTransaction> getRunningTransactions() {
+            Set<DataStoreTransaction> transactions = new HashSet<DataStoreTransaction>();
```
aklish, line 432: You can write this more directly as 

```java
transactions = transactionMap.values().stream().collect(Collectors.toSet())
```
or something similar to that.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -31,7 +32,7 @@
         PARTIAL,
         NONE
     }
-
+    UUID TRANSACTION_ID = UUID.randomUUID();
```
aklish, line 35: Won't this be static?  This doesn't look right to me.

Rkr1992: Don't we want different **TRANSACTION_ID** for each transaction? And making it static would mean 1 id for all the instances of the class?

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -263,4 +263,12 @@ default boolean supportsSorting(Class<?> entityClass, Sorting sorting) {
     default boolean supportsPagination(Class<?> entityClass, FilterExpression expression) {
         return true;
     }
+
+    /**
+     * @return UUID of transaction
+     *
+     */
+    default UUID getId() {
```
aklish, line 271: Given that this is an interface, I'm not sure a default implementation makes sense here. You can leave the method, but remove default and the implementation.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -263,4 +262,6 @@ default boolean supportsSorting(Class<?> entityClass, Sorting sorting) {
     default boolean supportsPagination(Class<?> entityClass, FilterExpression expression) {
         return true;
     }
+
+    UUID getRequestId();
```
aklish, line 266: Needs javadoc.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransactionImplementation.java
```diff
@@ -0,0 +1,19 @@
...
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core;
+
+import lombok.Getter;
+
+import java.util.UUID;
+
+public abstract class DataStoreTransactionImplementation implements DataStoreTransaction {
```
aklish, line 12: Class needs javadoc.

elide-core/src/main/java/com/yahoo/elide/core/datastore/inmemory/InMemoryStoreTransaction.java
```diff
@@ -119,7 +119,7 @@ public Object loadObject(EntityProjection projection,
                 projection.getFilterExpression()) == FeatureSupport.FULL) {
             return tx.loadObject(projection, id, scope);
         } else {
-            return DataStoreTransaction.super.loadObject(projection, id, scope);
+            return super.loadObject(projection, id, scope);
```
aklish, line 122: I think this is wrong.  This was calling the default method defined in the interface. You changed this to call the super class - which is something else - and will cause problems.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
```diff
@@ -14,14 +14,16 @@
...
 
+import java.io.IOException;
+import java.util.UUID;
 /**
  * Transaction handler for {@link AggregationDataStore}.
  */
 public class AggregationDataStoreTransaction implements DataStoreTransaction {
     private QueryEngine queryEngine;
-
+    @Getter private final UUID requestId = UUID.randomUUID();
```
aklish, line 26: Is there a reason this can't extend DataStoreTransactionImpl?

PR #1339, opened 2020-05-22
Supported embedded entities in graphql

aklish:
## Description
Adds support for embedded complex attributes (non-scalar and non collection types) in GraphQL.

## Motivation and Context
JSON-API supports but GraphQL does not.

## How Has This Been Tested?
Added new GraphQL unit tests.

## License


elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -69,12 +69,14 @@
 
     /**
      * Class constructor, constructs the custom arguments to handle mutations
-     * @param dictionary elide entity dictionary
+     * @param entityDictionary elide entity dictionary
      * @param dataFetcher graphQL data fetcher
```
moizarafat, line 73: we can add nonEntityDictionary in the param list to reduce warnings. :)

aklish: done

PR #1354, opened 2020-05-29
Elide dynamic config model verification

### LINKED ISSUE #1178: RFC : Elide Dynamic Model Configuration (Tables Only)

AvaniMakwana:
Authored-by: AvaniMakwana

Description
Verify Dynamic Model Configuration tar.gz file with available public key and signature file.

Motivation and Context
Please refer #1178

How Has This Been Tested?

License

### DIFF
--- a/elide-contrib/elide-dynamic-config-helpers/pom.xml
+++ b/elide-contrib/elide-dynamic-config-helpers/pom.xml
@@ -70,6 +70,11 @@
             <version>${commons-io.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
+            <version>1.20</version>
+        </dependency>
+        <dependency>
         <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
--- a/elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
+++ b/elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2020, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.contrib.dynamicconfighelpers.verify;
+
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveInputStream;
+import org.apache.commons.compress.compressors.gzip.GzipCompressorInputStream;
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.BufferedInputStream;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.security.InvalidKeyException;
+import java.security.KeyStore;
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.PublicKey;
+import java.security.Signature;
+import java.security.SignatureException;
+import java.security.cert.Certificate;
+import java.util.Base64;
+
+@Slf4j
+public class DynamicConfigVerifier {
+
+    /**
+     * Main Methode to Verify Signature of Model Tar file.
+     * @param args : expects 3 arguments.
+     * @throws IllegalStateException
+     */
+    public static void main(String[] args) {
+        if (args == null || args.length != 3) {
+            usage();
+            throw new IllegalStateException("No Arguments provided!");
+        }
+        String modelTarFile = args[0];
+        String signatureFile = args[1];
+        String publicKeyName = args[2];
+
+        if (verify(tarContents(modelTarFile), signatureFile, getPublicKey(publicKeyName))) {
+            log.info("Successfully Validated " + modelTarFile);
+        }
+        else {
+            log.error("Could not verify " + modelTarFile + " with details provided");
+        }
+    }
+
+    private static PublicKey getPublicKey(String keyName) {
+        PublicKey publicKey = null;
+        try {
+            KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
+            Certificate cert = keyStore.getCertificate(keyName);
+            publicKey = cert.getPublicKey();
+        } catch (KeyStoreException e) {
+            log.error(e.getMessage());
+            throw new IllegalArgumentException("Key " + keyName + " is not availabe in keystore");
+        }
+        return publicKey;
+    }
+
+    /**
+     * Verify signature of tar.gz.
+     * @param fileList
+     * @param signature
+     * @param publicKey
+     * @return whether the file can be verified by given key & signature.
+     * @throws Exception
+     */
+    public static boolean verify(String fileList, String signature, PublicKey publicKey) {
+        Signature publicSignature;
+        try {
+            publicSignature = Signature.getInstance("SHA256withRSA");
+            publicSignature.initVerify(publicKey);
+            publicSignature.update(fileList.getBytes(StandardCharsets.UTF_8));
+            byte[] signatureBytes = Base64.getDecoder().decode(signature);
+            return publicSignature.verify(signatureBytes);
+        } catch (NoSuchAlgorithmException e) {
+            log.error(e.getMessage());
+        } catch (InvalidKeyException e) {
+            log.error(e.getMessage());
+        } catch (SignatureException e) {
+            log.error(e.getMessage());
+        }
+        return false;
+    }
+
+    private static String tarContents(String archiveFile) {
+        StringBuffer sb = new StringBuffer();
+        try (TarArchiveInputStream archive = new TarArchiveInputStream(
+                new GzipCompressorInputStream(new BufferedInputStream(new FileInputStream(archiveFile))))) {
+            TarArchiveEntry entry;
+            while ((entry = archive.getNextTarEntry()) != null) {
+                sb.append(entry.getName());
+            }
+        } catch (IOException e) {
+            log.error(e.getMessage());
+            throw new IllegalArgumentException("archiveFile " + archiveFile + " is not available");
+        }
+        return sb.toString();
+    }
+
+    private static void usage() {
+        log.info("\n Usage: java -cp <Jar File Name>"
+                + " com.yahoo.elide.contrib.dynamicconfighelpers.validator.DynamicConfigVerifier\n"
+                + " <The name of the tar.gz file\n"
+                + " <The name of the file containing the signature \n"
+                + " <The name public key \n"
+                );
+    }
+}
--- a/elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifiesTest.java
+++ b/elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifiesTest.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2020, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.contrib.dynamicconfighelpers.verify;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import org.junit.jupiter.api.Test;
+
+import java.nio.charset.StandardCharsets;
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.PrivateKey;
+import java.security.SecureRandom;
+import java.security.Signature;
+import java.util.Base64;
+
+public class DynamicConfigVerifiesTest {
+
+    private KeyPair kp;
+    private String signature;
+
+    private static KeyPair generateKeyPair() throws Exception {
+        KeyPairGenerator generator = KeyPairGenerator.getInstance("RSA");
+        generator.initialize(2048, new SecureRandom());
+        KeyPair pair = generator.generateKeyPair();
+        return pair;
+    }
+
+    private static String sign(String data, PrivateKey privateKey) throws Exception {
+        Signature privateSignature = Signature.getInstance("SHA256withRSA");
+        privateSignature.initSign(privateKey);
+        privateSignature.update(data.getBytes(StandardCharsets.UTF_8));
+
+        byte[] signature = privateSignature.sign();
+
+        return Base64.getEncoder().encodeToString(signature);
+    }
+
+    private void setup() throws Exception {
+        kp = generateKeyPair();
+        signature = sign("testing-signature", kp.getPrivate());
+    }
+
+    @Test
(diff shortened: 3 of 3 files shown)


elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+@Slf4j
+public class DynamicConfigVerifier {
+
+    /**
+     * Main Methode to Verify Signature of Model Tar file.
+     * @param args : expects 3 arguments.
+     * @throws IllegalStateException
+     */
+    public static void main(String[] args) {
+        if (args == null || args.length != 3) {
```
moizarafat, line 37: Please refer https://github.com/yahoo/elide/blob/ece07420045910324dd91206ee999cd82c8ca478/elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidator.java#L64. We want to use Options class.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+                sb.append(entry.getName());
+            }
+        } catch (IOException e) {
+            log.error(e.getMessage());
+            throw new IllegalArgumentException("archiveFile " + archiveFile + " is not available");
+        }
+        return sb.toString();
+    }
+
+    private static void usage() {
```
moizarafat, line 107: Please refer https://github.com/yahoo/elide/blob/ece07420045910324dd91206ee999cd82c8ca478/elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidator.java#L270 to use HelpFormatter.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    private static PublicKey getPublicKey(String keyName) {
+        PublicKey publicKey = null;
+        try {
+            KeyStore keyStore = KeyStore.getInstance(KeyStore.getDefaultType());
+            Certificate cert = keyStore.getCertificate(keyName);
+            publicKey = cert.getPublicKey();
+        } catch (KeyStoreException e) {
```
moizarafat, line 59: Lets not catch the exception and just let the original exception propagate.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+     */
+    public static boolean verify(String fileList, String signature, PublicKey publicKey) {
+        Signature publicSignature;
+        try {
+            publicSignature = Signature.getInstance("SHA256withRSA");
+            publicSignature.initVerify(publicKey);
+            publicSignature.update(fileList.getBytes(StandardCharsets.UTF_8));
+            byte[] signatureBytes = Base64.getDecoder().decode(signature);
+            return publicSignature.verify(signatureBytes);
+        } catch (NoSuchAlgorithmException e) {
```
moizarafat, line 82: same, let the exceptions propagate back as is.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+    private static String tarContents(String archiveFile) {
+        StringBuffer sb = new StringBuffer();
+        try (TarArchiveInputStream archive = new TarArchiveInputStream(
+                new GzipCompressorInputStream(new BufferedInputStream(new FileInputStream(archiveFile))))) {
+            TarArchiveEntry entry;
+            while ((entry = archive.getNextTarEntry()) != null) {
+                sb.append(entry.getName());
+            }
+        } catch (IOException e) {
+            log.error(e.getMessage());
```
moizarafat, line 101: Let exception propagate as is.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifiesTest.java
```diff
@@ -0,0 +1,67 @@
...
+        Signature privateSignature = Signature.getInstance("SHA256withRSA");
+        privateSignature.initSign(privateKey);
+        privateSignature.update(data.getBytes(StandardCharsets.UTF_8));
+
+        byte[] signature = privateSignature.sign();
+
+        return Base64.getEncoder().encodeToString(signature);
+    }
+
+    private void setup() throws Exception {
```
moizarafat, line 43: Since this method is used by all test cases, we can annotate with @BeforeEach instead of calling setup from the individual tests.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifiesTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Test
+    public void testInvalidSignature() throws Exception {
+        setup();
+        assertFalse(DynamicConfigVerifier.verify("invalid-signature", signature, kp.getPublic()));
+    }
+
+    @Test
+    public void testFileSignature() throws Exception {
+        kp = generateKeyPair();
```
moizarafat, line 62: this is same as setup method.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+        } catch (InvalidKeyException e) {
+            log.error(e.getMessage());
+        } catch (SignatureException e) {
+            log.error(e.getMessage());
+        }
+        return false;
+    }
+
+    private static String tarContents(String archiveFile) {
```
moizarafat, line 92: readTarFile will be a better name here. tarContents sounds like we are trying to generate a tar.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,115 @@
...
+        return publicKey;
+    }
+
+    /**
+     * Verify signature of tar.gz.
+     * @param fileList
+     * @param signature
+     * @param publicKey
+     * @return whether the file can be verified by given key & signature.
+     * @throws Exception
```
moizarafat, line 72: list the exceptions instead of generic exception.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifiesTest.java
```diff
@@ -40,28 +45,47 @@ private static String sign(String data, PrivateKey privateKey) throws Exception
...
-        setup();
-        assertFalse(DynamicConfigVerifier.verify("invalid-signature", signature, kp.getPublic()));
+        assertFalse(DynamicConfigVerifier.verify("invalid-signature", 5, signature, kp.getPublic()));
     }
 
     @Test
-    public void testFileSignature() throws Exception {
-        kp = generateKeyPair();
-        signature = sign("testing-signature", kp.getPrivate());
+    public void testHelpArgumnents() {
```
moizarafat, line 65: Arguments is spelled incorrectly.

elide-contrib/elide-dynamic-config-helpers/pom.xml
```diff
@@ -76,7 +77,12 @@
             <version>${commons-io.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.slf4j</groupId>
+            <groupId>org.apache.commons</groupId>
+            <artifactId>commons-compress</artifactId>
+            <version>${commons-compress.version}</version>
+        </dependency>
+        <dependency>
+        <groupId>org.slf4j</groupId>
```
moizarafat, line 85: indentation is off here.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,159 @@
...
+import java.security.KeyStoreException;
+import java.security.NoSuchAlgorithmException;
+import java.security.PublicKey;
+import java.security.Signature;
+import java.security.SignatureException;
+import java.security.cert.Certificate;
+import java.util.Base64;
+
+@Slf4j
+public class DynamicConfigVerifier {
```
aklish, line 39: Java doc comment explaining what this is for

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,159 @@
...
+        return publicKey;
+    }
+
+    private static String readTarContents(String archiveFile) throws FileNotFoundException, IOException {
+        StringBuffer sb = new StringBuffer();
+        TarArchiveInputStream archive = new TarArchiveInputStream(
+                new GzipCompressorInputStream(new BufferedInputStream(new FileInputStream(archiveFile))));
+        TarArchiveEntry entry;
+        while ((entry = archive.getNextTarEntry()) != null) {
+            sb.append(entry.getName());
```
aklish, line 133: This seems to be only verifying the names of the files but not the file contents.  Am I reading this correct?   If so - we'll need to verify the contents as well.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,159 @@
...
+     * @throws ParseException
+     * @throws IOException
+     * @throws KeyStoreException
+     * @throws FileNotFoundException
+     * @throws SignatureException
+     * @throws NoSuchAlgorithmException
+     * @throws InvalidKeyException
+     *
+     */
+    public static void main(String[] args) throws ParseException, InvalidKeyException, NoSuchAlgorithmException,
```
aklish, line 53: We'll want a way to programmatically do this as well as run this from the command line.

AvaniMakwana: ```public static boolean verify(String fileList, long sizeOfFile, String signature, PublicKey publicKey) throws NoSuchAlgorithmException, InvalidKeyException, SignatureException {```

verify function can be called programmatically to verify the tar signature

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifier.java
```diff
@@ -0,0 +1,157 @@
...
+
+        String modelTarFile = cli.getOptionValue("tarFile");
+        String signatureFile = cli.getOptionValue("signatureFile");
+        String publicKeyName = cli.getOptionValue("publicKeyName");
+
+        if (verify(readTarContents(modelTarFile), signatureFile, getPublicKey(publicKeyName))) {
+            log.info("Successfully Validated " + modelTarFile);
+        }
+        else {
+            log.error("Could not verify " + modelTarFile + " with details provided");
```
aklish, line 79: You should set the exit status on failure to something that will cause this to abort shell scripts.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/verify/DynamicConfigVerifiesTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.security.KeyPair;
+import java.security.KeyPairGenerator;
+import java.security.PrivateKey;
+import java.security.SecureRandom;
+import java.security.Signature;
+import java.util.Base64;
+
+public class DynamicConfigVerifiesTest {
```
aklish, line 26: We need to add a test that creates a gzipped tarball, signs it, and then verifies your functions can do the reverse and verify the signature.

PR #1363, opened 2020-06-01
Disabling Legacy Filter Dialect in Swagger Documentation

### LINKED ISSUE #1362: JSONAPI Swagger filtering using the wrong dialect.
I'm new to using the **elide-spring-boot-starter** and noticed it provides swagger documentation.
However that the documentation seems to use an older dialect and not the RSQL dialect.

## Expected Behavior
That the swagger documentation should display the RSQL dialect, but appears to be displaying the old dialect.

## Current Behavior
image
As you can see from the screen it is displaying the old dialect and when you use the dialect displayed on swagger, it reports an error.

## Steps to Reproduce
1. Check out the example and build
2. Run the application
3. Go to the swagger documentation and pick a filter example
4. Then attempt to apply the filter, which files.
5. Try RSQL filter and it works. 

## Your Environment
* Elide version used: **4.6.4**
* Java version:  **1.8.0_232**
* Operating System and version: **Windows**

aklish:
Resolves #1362 

## License

### DIFF
--- a/elide-contrib/elide-swagger/src/main/java/com/yahoo/elide/contrib/swagger/SwaggerBuilder.java
+++ b/elide-contrib/elide-swagger/src/main/java/com/yahoo/elide/contrib/swagger/SwaggerBuilder.java
@@ -56,6 +56,7 @@ public class SwaggerBuilder {
     protected Map<Integer, Response> globalResponses;
     protected Set<Parameter> globalParams;
     protected Set<Operator> filterOperators;
+    protected boolean supportLegacyDialect;
 
     public static final Response UNAUTHORIZED_RESPONSE = new Response().description("Unauthorized");
     public static final Response FORBIDDEN_RESPONSE = new Response().description("Forbidden");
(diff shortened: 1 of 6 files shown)


elide-contrib/elide-swagger/src/main/java/com/yahoo/elide/contrib/swagger/SwaggerBuilder.java
```diff
@@ -604,7 +607,17 @@ private boolean lineageContainsType(PathMetaData other) {
...
     public SwaggerBuilder(EntityDictionary dictionary, Info info) {
+        this(dictionary, info, true);
+    }
+
+    /**
+     * @param dictionary The entity dictionary
+     * @param info Basic service information that cannot be generated
+     * @param supportLegacyDialect Whether or not to support the legacy dialect
+     */
+    public SwaggerBuilder(EntityDictionary dictionary, Info info, boolean supportLegacyDialect) {
```
wcekan, line 618: Since this is a builder, why is there no
```
    . supportLegacyDialect(boolean)
```

whgibbo: So does this also need to be exposed as an configuration option for **elide-spring-boot-starter** or change the autoconfiguration to disable the legacy dialect?

elide-contrib/elide-swagger/src/main/java/com/yahoo/elide/contrib/swagger/SwaggerBuilder.java
```diff
@@ -509,19 +510,21 @@ private Parameter getSortParameter() {
...
-                    /* Only filter attributes that can be assigned to strings or primitives */
-                    if (attributeClass.isPrimitive() || String.class.isAssignableFrom(attributeClass)) {
-                        params.add(new QueryParameter()
-                                .type("string")
-                                .name("filter[" + typeName + "." + name + "][" + op.getNotation() + "]")
-                                .description("Filters the collection of " + typeName + " by the attribute "
-                                        + name + " " + "using the operator " + op.getNotation()));
-                    }
-                });
+            if (supportLegacyDialect) {
```
whgibbo, line 513: I'm new to Elide, but would I be right in thinking that it only supports one dialect..  So for users that want to use the legacy dialect, then the RSQL examples should not be displayed.

elide-contrib/elide-swagger/src/main/java/com/yahoo/elide/contrib/swagger/SwaggerBuilder.java
aklish: The auto-configuration has the legacy dialect disabled by default.  Also, you can auto-configure the Swagger object construction to turn on/off the dialects you want.

elide-contrib/elide-swagger/src/main/java/com/yahoo/elide/contrib/swagger/SwaggerBuilder.java
aklish: Elide supports multiple dialects enabled at the same time.

aklish:
Updated from comments.

PR #1364, opened 2020-06-03
add explicit join

### LINKED ISSUE #920: JPA/Hibernate stores should perform left outer joins when sorting.
## Expected Behavior
If sorting on `book.publisher.name`, Elide currently uses JPQL's implicit join syntax (which performs inner joins):

```
HQL supports two forms of association joining: implicit and explicit.

The queries shown in the previous section all use the explicit form, that is, where the join keyword is explicitly used in the from clause. This is the recommended form.

The implicit form does not use the join keyword. Instead, the associations are "dereferenced" using dot-notation. implicit joins can appear in any of the HQL clauses. implicit join result in inner joins in the resulting SQL statement.

from Cat as cat where cat.mate.name like '%s%'
```

If the publisher is null, Elide will not return those records.    Instead, Elide should return all records including those where one of the relationships is null.

## Possible Solution
A possible solution is to modify the JPQL query to build explicit join clauses similar to how filter joins currently work.

Chandrasekar-Rajasekar:
Resolves #920  (if appropriate)

## Description
Query with Sorting clause should have an explicit join clause and use its alias in order by clause. 

## Motivation and Context
The implicit JPA join uses an inner join to fetch the relationship entity. We want JPA to perform left join to preserve all response records.

## How Has This Been Tested?
Unit test
IT test

## License

### DIFF
--- a/elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/SubCollectionFetchQueryBuilder.java
+++ b/elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/SubCollectionFetchQueryBuilder.java
@@ -74,6 +74,7 @@ public class SubCollectionFetchQueryBuilder extends AbstractHQLQueryBuilder {
             String filterClause = new FilterTranslator().apply(fe, USE_ALIAS);
 
             String joinClause =  getJoinClauseFromFilters(filterExpression.get())
+                    + getJoinClauseFromSort(sorting)
                     + extractToOneMergeJoins(relationship.getChildType(), childAlias);
 
             //SELECT parent_children from Parent parent JOIN parent.children parent_children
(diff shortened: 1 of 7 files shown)


elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/SubCollectionFetchQueryBuilder.java
```diff
@@ -74,6 +74,7 @@ public Query build() {
             String filterClause = new FilterTranslator().apply(fe, USE_ALIAS);
 
             String joinClause =  getJoinClauseFromFilters(filterExpression.get())
+                    + getJoinClauseFromSort(sorting)
```
aklish, line 77: Logic looks correct - but you are missing two files that need to have their sorting changed.  SubCollectionPageTotalsQueryBuilder and RootCollectionPageTotalsQueryBuilder

PR #1367, opened 2020-06-03
implement cancel function

### LINKED ISSUE #1292: RFC Elide 5.0 - TransactionRegistry

Rkr1992:

## License


elide-core/src/main/java/com/yahoo/elide/core/CancelSession.java
```diff
@@ -0,0 +1,23 @@
...
+ * See LICENSE file in project root for terms.
+ */
+
+import org.hibernate.Session;
+
+/**
+ * Cancel Session implementation.
+ */
+
+public abstract class CancelSession {
```
aklish, line 13: Instead of creating an abstract class - create a functional interface (with no dependencies on things like hibernate).

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -376,10 +378,10 @@ default AuditLogger getAuditLogger() {
      */
     default DataStore getDataStore(MetaDataStore metaDataStore, AggregationDataStore aggregationDataStore,
             EntityManagerFactory entityManagerFactory) {
-
+	CancelSession cancelSession = new CancelSession(entityManagerFactory.get().unwrap(Session.class));
```
aklish, line 381: YOu can make this a local variable and use it multiple places.

aklish: Another issue - you are unwrapping the factory - not the entity manager.  I don't think that is correct.

elide-core/src/main/java/com/yahoo/elide/core/CancelSession.java
aklish: I think the functionalInterface will require an entityManager as an argument - and as such - it should probably live in the Aggregation Store and JpaStore respectively.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -393,18 +393,19 @@ default DataStore getDataStore(MetaDataStore metaDataStore, AggregationDataStore
      * @return AggregationDataStore object initialized.
      */
     default AggregationDataStore getAggregationDataStore(QueryEngine queryEngine,
-            Optional<ElideDynamicEntityCompiler> optionalCompiler) {
+            Optional<ElideDynamicEntityCompiler> optionalCompiler, EntityManagerFactory entityManagerFactory) {
```
aklish, line 396: Remove entityManagerFactory as an argument.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -393,18 +393,19 @@ default DataStore getDataStore(MetaDataStore metaDataStore, AggregationDataStore
...
             annotatedClasses.addAll(getDynamicClassesIfAvailable(optionalCompiler, FromSubquery.class));
-            aggregationDataStore = new AggregationDataStore(queryEngine, annotatedClasses);
+	    aggregationDataStore = new AggregationDataStore(
+            	() -> { return entityManagerFactory.createEntityManager(); },
+                queryEngine, annotatedClass, (entityManager -> { entityManager.unwrap(Session.class).cancelQuery();}));
         } else {
-            aggregationDataStore = new AggregationDataStore(queryEngine);
+            aggregationDataStore = new AggregationDataStore(
+            	() -> { return entityManagerFactory.createEntityManager(); },
+                queryEngine, (entityManager -> { entityManager.unwrap(Session.class).cancelQuery();}));
```
aklish, line 407: We define the transaction cancel method four times.  Let's define this once with a static, final variable and just use that everywhere.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -163,7 +163,7 @@ private void populateMetaData(MetaDataStore metaDataStore) {
      * @param query The query customized for a particular persistent storage or storage client
      * @return query results
      */
-    public abstract QueryResult executeQuery(Query query);
+    public abstract FutureImplementation<QueryResult> executeQuery(Query query);
```
aklish, line 166: Just return a Future here instead of the Implementation.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/FutureImplementation.java
```diff
@@ -0,0 +1,21 @@
...
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+
+import com.yahoo.elide.datastores.aggregation.query.QueryResult;
+import com.yahoo.elide.datastores.aggregation.queryengines.sql.query.SQLQueryEngine;
+
+import java.util.concurrent.Future;
+
+public abstract class FutureImplementation implements Future<QueryResult> {
```
aklish, line 12: You could make this an inner class of the SQLQueryEngine - not abstract - and it can just call the TransactionCancel member.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -289,4 +291,28 @@ private SQLQuery toPageTotalSQL(SQLQuery sql) {
...
+    /**
+     * Functional interface for describing a method to supply JpaTransaction.
+     */
+    @FunctionalInterface
+    public interface TransactionCancel {
+        public void cancel(EntityManager entityManager);
+    }
+  
+    public class FutureImplementation implements Future<QueryResult> {
+        private final TransactionCancel transactionCancel;
```
aklish, line 304: An inner class can reference the members of the outer class if it is not static.  There is no need to:
1. Define private members
2. Have a constructor with arguments.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -289,4 +291,28 @@ private SQLQuery toPageTotalSQL(SQLQuery sql) {
...
+	private final EntityManagerFactory entityManagerFactory;
+        /**
+	 * Cancels transaction
+         */
+    	@Override
+	protected FutureImplementation(EntityManagerFactory entityManagerFactory, TransactionCancel transactionCancel) {
```
aklish, line 310: Let's find a better name than FutureImplementation.  How about QueryResultFuture?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -289,4 +291,28 @@ private SQLQuery toPageTotalSQL(SQLQuery sql) {
...
+	    this.transactionCancel = transactionCancel;
+	    this.entityManagerFactory = entityManagerFactory;
+	}
+    	public boolean cancel(boolean mayInterruptIfRunning) {
```
aklish, line 314: Add ```@Override```.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -60,13 +61,14 @@
 @Slf4j
 public class SQLQueryEngine extends QueryEngine {
     private final EntityManagerFactory entityManagerFactory;
-
+    private final TransactionCancel transactionCancel;
```
aklish, line 64: I don't see the implementation of executeQuery that returns a Future<QueryResult> here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
aklish, line 314: You are missing a lot of the other Future functions.  There is more than just cancel.

elide-core/src/main/java/com/yahoo/elide/core/DataStoreTransaction.java
```diff
@@ -267,4 +267,9 @@ default boolean supportsPagination(Class<?> entityClass, FilterExpression expres
      * @return UUID id
      */
     UUID getRequestId();
+
+    /**
+     * cancel running transaction
```
aklish, line 272: Can you mention in the comments that the implementation must be thread-safe?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
```diff
@@ -52,11 +56,21 @@ public void createObject(Object entity, RequestScope scope) {
     @Override
     public Iterable<Object> loadObjects(EntityProjection entityProjection, RequestScope scope) {
         Query query = buildQuery(entityProjection, scope);
-        QueryResult result = queryEngine.executeQuery(query);
-        if (entityProjection.getPagination() != null && entityProjection.getPagination().returnPageTotals()) {
-            entityProjection.getPagination().setPageTotals(result.getPageTotals());
+        queryResult = queryEngine.executeQuery(query);
+        queryResult.run();
```
aklish, line 60: You don't need to call run() here.  Calling queryResult.get() should invoke queryResult.run()

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
```diff
@@ -52,11 +56,21 @@ public void createObject(Object entity, RequestScope scope) {
...
+        try {
+            QueryResult result = queryResult.get();
+            if (entityProjection.getPagination() != null && entityProjection.getPagination().returnPageTotals()) {
+                entityProjection.getPagination().setPageTotals(result.getPageTotals());
+            }
+            return result.getData();
+        } catch (TransactionException e) {
+            throw new TransactionException(null);
```
aklish, line 68: Don't set null here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -60,13 +62,15 @@
...
     private final SQLReferenceTable referenceTable;
+    private EntityManager entityManager;
 
-    public SQLQueryEngine(MetaDataStore metaDataStore, EntityManagerFactory entityManagerFactory, Cache cache) {
-        super(metaDataStore, cache);
-        this.entityManagerFactory = entityManagerFactory;
-        this.referenceTable = new SQLReferenceTable(metaDataStore);
+    public SQLQueryEngine(MetaDataStore mDStore, EntityManagerFactory emFactory, Cache cx, TransactionCancel txCancel) {
```
aklish, line 69: cache is much more readable than cx.  Can you revert back to the more readable variable names?

aklish: Same for mDStore, qEngine, emFactory, etc.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -289,4 +302,27 @@ private SQLQuery toPageTotalSQL(SQLQuery sql) {
...
+
+    /**
+     * Functional interface for describing a method to supply JpaTransaction.
+     */
+    @FunctionalInterface
+    public interface TransactionCancel {
+        public void cancel(EntityManager entityManager);
+    }
+
+    public class QueryResultFuture<V> extends FutureTask<V> {
```
aklish, line 314: This inner class is missing javadoc.  The linter will complain.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/QueryEngineTest.java
```diff
@@ -55,8 +58,20 @@ public void testFullTableLoad() {
                 .timeDimension(toProjection(playerStatsTable.getTimeDimension("recordedDate"), TimeGrain.DAY))
                 .build();
 
-        List<Object> results = StreamSupport.stream(engine.executeQuery(query).getData().spliterator(), false)
+        List<Object> results;
+        FutureTask<QueryResult> queryResult = engine.executeQuery(query);
+        queryResult.run();
```
aklish, line 63: This code is violating DRY (do not repeat yourself).  Make a function in this class to avoid repeating the same block over and over.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SubselectTest.java
```diff
@@ -87,7 +92,10 @@ public void testJoinToFilter() throws Exception {
                         PlayerStats.class, false))
                 .build();
 
-        List<Object> results = StreamSupport.stream(engine.executeQuery(query).getData().spliterator(), false)
+        FutureTask<QueryResult> queryResult = engine.executeQuery(query);
+        queryResult.run();
```
aklish, line 96: Remove all references to queryResult.run() in this PR.

aklish: The delta between the original line and your modifications can be just 4 characters:

```java
List<Object> results = StreamSupport.stream(engine.executeQuery(query).get().getData().spliterator(), false)
```

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/ViewTest.java
```diff
@@ -46,8 +50,20 @@ public void testViewAttribute() {
...
 
-        List<Object> results = StreamSupport.stream(engine.executeQuery(query).getData().spliterator(), false)
+        List<Object> results;
+        FutureTask<QueryResult> queryResult = engine.executeQuery(query);
+        queryResult.run();
+        try {
+            QueryResult result = queryResult.get();
+            results = StreamSupport.stream(result.getData().spliterator(), false)
                 .collect(Collectors.toList());
+        } catch (TransactionException e) {
```
aklish, line 60: For UnitTests, you can simply have the test throw Exception and avoid all these try catch blocks.

Look at my prior comments - all of this can be changed by just adding 4 characters - without these large blocks of duplicate code.

elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/JpaDataStore.java
```diff
@@ -73,15 +78,15 @@ public void populateEntityDictionary(EntityDictionary dictionary) {
...
+        JpaTransaction transaction = readTransactionSupplier.get(entityManager, jpaTransactionCancel);
         transaction.begin();
         return transaction;
     }
 
     @Override
     public DataStoreTransaction beginTransaction() {
         EntityManager entityManager = entityManagerSupplier.get();
-        JpaTransaction transaction = writeTransactionSupplier.get(entityManager);
+        JpaTransaction transaction = writeTransactionSupplier.get(entityManager, jpaTransactionCancel);
```
aklish, line 89: You shouldn't need to pass transactionCancel as an argument to the supplier.  When the supplier is created, it can have a reference to transactionSupplier through closure.  Remove this extra argument.

elide-datastore/elide-datastore-jpa/src/test/java/com/yahoo/elide/datastores/jpa/JpaDataStoreHarness.java
```diff
@@ -96,7 +99,8 @@ public JpaDataStoreHarness() {
 
         store = new JpaDataStore(
                 () -> { return emf.createEntityManager(); },
-                (entityManager) -> { return new NonJtaTransaction(entityManager); }
+                (entityManager, txCancel) -> { return new NonJtaTransaction(entityManager, txCancel); },
```
aklish, line 102: So for example, there is no need to include txCancel as a parameter to the supplier.  Rewrite like:
```java
(entityManager) -> { return new NonJtaTransaction(entityManager, jpaTransactionCancel); }
```

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAutoConfiguration.java
```diff
@@ -56,6 +58,10 @@
      * @return An instance of ElideDynamicEntityCompiler.
      * @throws Exception Exception thrown.
      */
+
+     private final AbstractJpaTransaction.JpaTransactionCancel jTC = (e) -> { e.unwrap(Session.class).cancelQuery(); };
```
aklish, line 62: Because the signatures are the same, you may be able to define this once as a Function rather than the concrete types above - and just use the single reference below.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -376,10 +382,10 @@ default AuditLogger getAuditLogger() {
      */
     default DataStore getDataStore(MetaDataStore metaDataStore, AggregationDataStore aggregationDataStore,
             EntityManagerFactory entityManagerFactory) {
-
         DataStore jpaDataStore = new JpaDataStore(
                 () -> { return entityManagerFactory.createEntityManager(); },
-                (em -> { return new NonJtaTransaction(em); }));
+                ((em, txCancel) -> { return new NonJtaTransaction(em, txCancel); }),
```
aklish, line 387: Remove txCancel from the supplier.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -56,6 +58,10 @@
  */
 public interface ElideStandaloneSettings {
     /* Elide settings */
+
+     public final AbstractJpaTransaction.JpaTransactionCancel JTC = (m) -> { m.unwrap(Session.class).cancelQuery(); };
```
aklish, line 62: See if these can be consolidated into a single variable using one of Java's built in Function classes.

elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/JpaDataStore.java
```diff
@@ -8,7 +8,7 @@
 import com.yahoo.elide.core.DataStoreTransaction;
 import com.yahoo.elide.core.EntityDictionary;
 import com.yahoo.elide.core.datastore.JPQLDataStore;
-import com.yahoo.elide.datastores.jpa.transaction.AbstractJpaTransaction;
+//import com.yahoo.elide.datastores.jpa.transaction.AbstractJpaTransaction;
```
aklish, line 11: Remove this comment.

elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/transaction/JtaTransaction.java
```diff
@@ -22,11 +23,11 @@
 @Slf4j
 public class JtaTransaction extends AbstractJpaTransaction {
     private final UserTransaction transaction;
-    public JtaTransaction(EntityManager entityManager, JpaTransactionCancel jpaTransactionCancel) {
+    public JtaTransaction(EntityManager entityManager, Consumer<EntityManager> jpaTransactionCancel) {
         this(entityManager, lookupUserTransaction(), jpaTransactionCancel);
     }
 
-    public JtaTransaction(EntityManager entityManager, UserTransaction transaction, JpaTransactionCancel txCancel) {
+    public JtaTransaction(EntityManager entityManager, UserTransaction transaction, Consumer<EntityManager> txCancel) {
```
aklish, line 30: Unify the variable names txCancel and jpaTransactionCancel.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAutoConfiguration.java
```diff
@@ -59,8 +60,7 @@
      * @throws Exception Exception thrown.
      */
 
-     private final AbstractJpaTransaction.JpaTransactionCancel jTC = (e) -> { e.unwrap(Session.class).cancelQuery(); };
-     private final SQLQueryEngine.TransactionCancel txCancel = (em) -> { em.unwrap(Session.class).cancelQuery(); };
+     private final Consumer<EntityManager> func = (em) -> { em.unwrap(Session.class).cancelQuery(); };
```
aklish, line 63: txCancel was a better name than func.  I would change everywhere.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -59,8 +59,7 @@
 public interface ElideStandaloneSettings {
     /* Elide settings */
 
-     public final AbstractJpaTransaction.JpaTransactionCancel JTC = (m) -> { m.unwrap(Session.class).cancelQuery(); };
-     public final SQLQueryEngine.TransactionCancel TXCANCEL = (m) -> { m.unwrap(Session.class).cancelQuery(); };
+     public final Consumer<EntityManager> FUNC = (em) -> { em.unwrap(Session.class).cancelQuery(); };
```
aklish, line 62: Rename to TXCANCEL

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -157,6 +156,7 @@ private void populateMetaData(MetaDataStore metaDataStore) {
     public interface Transaction extends AutoCloseable {
         @Override
         void close();
+        void cancel();
```
aklish, line 159: Let's add javadoc for cancel (close isn't needed because it is an Override).

PR #1378, opened 2020-06-16
Enable query caching by default

john-karp:
## Description
For both spring and standalone setups, enable query caching with a Caffeine implementation by default, with a configurable maximum cache size.

## Motivation and Context
Reduce amount of work needed for users to take advantage of caching. (User will still need to satisfy requirements of query engine used for caching.)

Under Spring, metrics will be published since there is a Caffeine metrics binding for micrometer (micrometer is the metrics API used by Spring). However, since standalone uses dropwizard metrics, which only has a binding provided as a (long) example, for now metrics are not published in this context. (I might pursue a PR to make the caffeine-dropwizard binding available, and then open a follow-up PR here to use it.)

## How Has This Been Tested?
Will manually test, automated tests pass.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/cache/CaffeineCache.java
```diff
@@ -0,0 +1,38 @@
...
+
+import com.yahoo.elide.datastores.aggregation.query.QueryResult;
+
+import com.github.benmanes.caffeine.cache.Caffeine;
+
+/**
+ * A basic local-only cache.
+ */
+public class CaffeineCache implements Cache {
+    public static final int DEFAULT_CACHE_SIZE = 1024;
```
aklish, line 17: What is the size in terms of bytes?  Entries?  How about change the variable name to include units.

elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -92,6 +92,14 @@
             <version>${hibernate5.version}</version>
         </dependency>
 
+        <!-- Caffeine cache -->
+        <dependency>
+            <groupId>com.github.ben-manes.caffeine</groupId>
+            <artifactId>caffeine</artifactId>
+            <version>2.8.4</version>
```
wcekan, line 99: should this dependency be optional?

john-karp: The default configuration now uses Caffeine, so if we make it optional, won't that force the average user, who isn't specifying an alternative, to declare an extra dependency?

PR #1382, opened 2020-06-18
Dynamic Model schema validation

### LINKED ISSUE #1178: RFC : Elide Dynamic Model Configuration (Tables Only)

AvaniMakwana:
Authored-by: AvaniMakwana

One of the multiple PRs that will resolve #1178

Description
- Add Table, Security & variable config validation against schema

Motivation and Context
Please refer #1178

How Has This Been Tested?

License


elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -190,4 +228,45 @@ public static String readConfigFile(File configFile) {
...
+        }
+        return results.isSuccess();
+    }
+
+    private static JsonSchema loadSchema(String resource) {
+        ObjectMapper objectMapper = new ObjectMapper();
+        Reader reader = new InputStreamReader(DynamicConfigHelpers.class.getResourceAsStream(resource));
+        try {
+            return FACTORY.getJsonSchema(objectMapper.readTree(reader));
+        } catch (IOException e) {
```
moizarafat, line 263: we should let these exceptions propagate and fail.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -64,14 +89,19 @@ public static String formatFilePath(String basePath) {
...
+            throws JsonProcessingException, ProcessingException {
         String filePath = basePath + VARIABLE_CONFIG_PATH;
         File variableFile = new File(filePath);
         if (variableFile.exists()) {
             String jsonConfig = hjsonToJson(readConfigFile(variableFile));
-            return getModelPojo(jsonConfig, Map.class);
+            if (verifySchema(VARIABLE, jsonConfig)) {
+                return getModelPojo(jsonConfig, Map.class);
+            }
+            return null;
```
moizarafat, line 104: instead of multiple return we can store results in a variable,  and then have single return.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidatorTest.java
```diff
@@ -107,7 +107,7 @@ public void testBadSecurityRoleConfig() {
 
     @Test
     public void testBadTableConfigJoinType() {
-        assertThrows(ValueInstantiationException.class, () -> DynamicConfigValidator
+        assertThrows(NullPointerException.class, () -> DynamicConfigValidator
```
moizarafat, line 110: is this change needed?

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -190,4 +230,46 @@ public static String readConfigFile(File configFile) {
...
+    }
+
+    private static JsonSchema loadSchema(String resource) {
+        ObjectMapper objectMapper = new ObjectMapper();
+        Reader reader = new InputStreamReader(DynamicConfigHelpers.class.getResourceAsStream(resource));
+        try {
+            return FACTORY.getJsonSchema(objectMapper.readTree(reader));
+        } catch (IOException e) {
+            log.error("Error loading schema file " + resource + " to verify");
+            throw new RuntimeException(e.getMessage());
```
moizarafat, line 269: we can throw IllegalStateException instead RunTime else codacy complains.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidatorTest.java
moizarafat: i don't see any other related change in main class.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -190,4 +230,46 @@ public static String readConfigFile(File configFile) {
...
+        } catch (ProcessingException e) {
+            log.error("Error loading schema file " + resource + " to verify");
+            throw new RuntimeException(e.getMessage());
```
moizarafat, line 272: same as previous comment.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidatorTest.java
AvaniMakwana: the function throws NPE. hence the change.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidator.java
```diff
@@ -14,6 +14,7 @@
 import com.yahoo.elide.contrib.dynamicconfighelpers.model.Table;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.github.fge.jsonschema.core.exceptions.ProcessingException;
```
aklish, line 17: We probably don't want to leak the dependency on jsonschema outside the validation logic.

elide-contrib/elide-dynamic-config-helpers/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidatorTest.java
aklish: It looks very odd for a test to _expect_ a NPE.  NPE should never happen in code - so this test looks wrong or something is off.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -190,4 +230,46 @@ public static String readConfigFile(File configFile) {
...
+
+    private static String hjsonToJson(String hjson) {
+        return JsonValue.readHjson(hjson).toString();
+    }
+
+    private static <T> T getModelPojo(String jsonConfig, final Class<T> configPojo) throws JsonProcessingException {
+        return new ObjectMapper().readValue(jsonConfig, configPojo);
+    }
+
+    private static boolean verifySchema(String configType, String jsonConfig)
```
aklish, line 242: Do we have tests for this new verify logic?

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigSchemaValidator.java
```diff
@@ -0,0 +1,85 @@
...
+import com.github.fge.jsonschema.main.JsonSchemaFactory;
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.io.Reader;
+
+@Slf4j
+public class DynamicConfigSchemaValidator {
```
aklish, line 21: Missing javadoc.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigSchemaValidator.java
```diff
@@ -0,0 +1,85 @@
...
+     * @return whether config is valid
+     * @throws IOException
+     * @throws ProcessingException
+     */
+    public static boolean verifySchema(String configType, String jsonConfig) throws IOException, ProcessingException {
+        ProcessingReport results = null;
+        boolean isSuccess = false;
+
+        switch (configType) {
+        case DynamicConfigHelpers.TABLE :
```
aklish, line 51: Instead of configType being a string, we should make it an enum.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigSchemaValidator.java
```diff
@@ -0,0 +1,85 @@
...
+
+    /**
+     *  Verify config against schema.
+     * @param configType
+     * @param jsonConfig
+     * @return whether config is valid
+     * @throws IOException
+     * @throws ProcessingException
+     */
+    public static boolean verifySchema(String configType, String jsonConfig) throws IOException, ProcessingException {
```
aklish, line 46: Is there a reason all of this has to be static?  It will be easier to mock if it were not.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -36,6 +38,9 @@
     public static final String SECURITY_CONFIG_PATH = "security.hjson";
     public static final String VARIABLE_CONFIG_PATH = "variables.hjson";
     public static final String NEW_LINE = "\n";
+    public static final String TABLE = "table";
```
aklish, line 41: Make these an enum.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -121,8 +135,17 @@ public static ElideTableConfig getElideTablePojo(String basePath, Map<String, Ob
...
-        return getModelPojo(jsonConfig, ElideTableConfig.class);
+        try {
+            if (DynamicConfigSchemaValidator.verifySchema(TABLE, jsonConfig)) {
+                table = getModelPojo(jsonConfig, ElideTableConfig.class);
+            }
+        } catch (ProcessingException e) {
+            log.error("Error Validating Table config : " + e.getMessage());
+            throw new IOException(e);
+        }
+        return (table == null ? new ElideTableConfig() : table);
```
aklish, line 148: If you initialize table to new ElideTableConfig, you can remove this check for null at the end.

moizarafat:
Closed in favor of https://github.com/yahoo/elide/pull/1449

PR #1386, opened 2020-06-19
CARBON-212 TimeZone Serde

ayeswarya:

## Description
Add Serde for TimeZone that converts TimeZone into a scalar type (instead of a Java object) 
## Motivation and Context
TimeZone shows up as an object in swagger and we want it to show as a scalar value

## How Has This Been Tested?
Tested in Spring boot example project

## License

### DIFF
--- a/elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLScalarsTest.java
+++ b/elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLScalarsTest.java
@@ -11,6 +11,7 @@ import com.yahoo.elide.utils.coerce.CoerceUtil;
 import com.yahoo.elide.utils.coerce.converters.ISO8601DateSerde;
 import com.yahoo.elide.utils.coerce.converters.OffsetDateTimeSerde;
 import com.yahoo.elide.utils.coerce.converters.Serde;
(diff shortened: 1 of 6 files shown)


elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLScalarsTest.java
```diff
@@ -84,6 +85,16 @@ public void testGraphQLOffsetDateTimeDeserialize() {
         assertEquals(expectedDate, actualDate);
     }
 
+    @Test
+    public void testGraphQLTimeZoneDeserialize() {
+        TimeZone expectedTz = TimeZone.getTimeZone("EST");
+        String input = "EST";
+        TimeZoneSerde timeZoneScalar = new TimeZoneSerde();
```
aklish, line 92: Can we rename this variable to serde?

elide-graphql/src/test/java/com/yahoo/elide/graphql/GraphQLScalarsTest.java
```diff
@@ -84,6 +85,16 @@ public void testGraphQLOffsetDateTimeDeserialize() {
...
+        SerdeCoercing timeZoneSerde =
```
aklish, line 93: Can we rename to variable to serdeCoercing?

PR #1399, opened 2020-06-24
bump JUEL 2.2.7 to Jakarta EL 3.0.3

wcekan:
## Description
Replace hard coded JUEL with Jakarta EL while maintaining support for JUEL 2.7.
Uses interfaces instead of requiring JUEL classes.

## Motivation and Context
Product may need later version of EL.

## How Has This Been Tested?
Unit testing.  Local testing.

## License

### DIFF
--- a/elide-core/pom.xml
+++ b/elide-core/pom.xml
@@ -128,14 +128,9 @@
         </dependency>
 
         <dependency>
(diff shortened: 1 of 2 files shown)


aklish:
Won't Fix

PR #1400, opened 2020-06-24
Adding fetch joins only when required

suriyasundar92:
## Description
The HQL query builder was adding fetch join for all to one relations. This PR changes the behavior to fetch them only when they are explicitly requested in the entity projection.

## Motivation and Context
The Fetch join for relations that are not requested are not necessary. 

## How Has This Been Tested?
The Unit tests are updated to test for the new behavior

## License


elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/AbstractHQLQueryBuilder.java
```diff
@@ -98,6 +101,11 @@ public AbstractHQLQueryBuilder withPossiblePagination(final Optional<Pagination>
         return this;
     }
 
+    public AbstractHQLQueryBuilder withRelationsIncludedInProjection(final Set<String> relationsIncludedInProjection) {
```
aklish, line 104: Probably cleaner to just have withEntityProjection and remove all the withSorting, withPagination, and this new method.  Also, you can move the functionality to get the relationship names out of the EntityProjection and into this new builder function.

elide-datastore/elide-datastore-hibernate/src/test/java/com/yahoo/elide/datastores/hibernate/hql/SubCollectionFetchQueryBuilderTest.java
```diff
@@ -112,6 +118,78 @@ public void testSubCollectionFetchWithSorting() {
         assertEquals(expected, actual);
     }
 
+    @Test
+    public void testSubCollectionFetchWithIncludedToManyRelation() {
```
aklish, line 122: We should probably add at least one test where we do a FETCH JOIN when getting a sub collection

aklish: I would add another test with INcludedToOneRelationship

suriyasundar92: @aklish As per my understanding this test should serve the purpose.
https://github.com/yahoo/elide/pull/1459/files#diff-d507c81b32b694bbf433353645234239R118
Please let me know if I'm not understanding the comment.

Issue #1272, opened 2020-04-17
RFC: Download Async Query Result


moizarafat:
# Overview
Elide will support the downloading of the results of the async query execution. 

Also, the current implementation stores the results in a database table. This will work if the results are not too large. If the result size is large, it may be better to store the results in different storage like a filesystem, S3.  

## Supported Download Formats
The following formats will be supported.
- JSON (Already supported)
- CSV
- ~~TSV~~ (Not Needed for MVP)
- ~~XLSX~~ (Not Needed for MVP)

For JSON to CSV conversion, we can use jackson-dataformats-text library, earlier known as jackson-dataformat-csv. 

## Record Count filter
The MVP requirement says it should allow up to 10K records but for the long term it should be scalable. At this point, I don't see a reason to put in any limitations of 10K right now, we can go ahead and stream back all the records. 

## Supported Storage
Elide will have in-built support to store the results in a database table. But can be expanded to any other storage.

## ResultStorageEngine Interface
A new ResultStorageEngine Interface will be created. The interface will have methods:
- Store the results.  should return the URL to the results.
- retrieve the results by id 
- delete the results by id
- delete all the results 
- access permission check //Only the original requester can retrieve the results. No update/delete allowed to users.

### DatabaseResultStorageEngine
Elide will come in-built with a default implementation of ResultStorageEngine Interface, which will store/retrieve the results in a database table. 

A new Entity will be created for storing the results.  This will be used only for the default Implementation case (Database). **So no relationship with AsyncQueryResult established.**

```java
@Exclude //We don't want this to be exposed through the standard Elide APIs. 
@Entity
@Data 
public class asyncQueryResultStorage {
    @Id
    @Column(columnDefinition="varchar(36)")
    private UUID id; //Matches UUID in query.
    
    private java.sql.Blob result; // To allow expansion to XLSX? 
}

public enum ResultFormatType {
    JSON,
    CSV
}
```
Existing Model to be updated: 

``` java
@Embeddable
@Data
public class AsyncQueryResult extends AsyncBase  {
    private Integer contentLength;

    private Integer recordCount; // New Column to be added. No of records in the final result

    ResultType resultType; //EMBEDDED, DOWNLOAD
    String result;  //URL or Response body
    private Integer status; // HTTP Status
    Date completedOn; 

}

@Entity
@Include(type = "asyncQuery", rootLevel = true)
@ReadPermission(expression = "Principal is Owner")
@UpdatePermission(expression = "Prefab.Role.None")
[...]

PR #1404, opened 2020-06-26
Data Export Changes

chandrakaran:
Resolves #1272

## Description
i) Update the Async Models, per the RFC-1272.
ii) Update the Async processQuery method to calculate the count of result records from the responseBody and include it in the result.
iii) Unit Test Case update/IT test update.

i) Add the ResultStorageEngine Interface.
ii) Add the default implementation of ResultStorageEngine interface with the methods in elide-async.
iii) Create the ResultStore Table.
iv) Store the results in the new ResultStore table in Process Query.
v) Unit Test Case update/IT test update.
vi) New Controllers for Spring and Standalone.

## Motivation and Context
This change allows the implementation of DefaultResultStorageEngine and helps store the results in the asyncResultStorageEngine.

## How Has This Been Tested?
Existing test cases and new test cases pass.

## License

### DIFF
--- a/elide-async/pom.xml
+++ b/elide-async/pom.xml
@@ -88,6 +88,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.junit.platform</groupId>
+            <artifactId>junit-platform-launcher</artifactId>
+            <scope>test</scope>
+        </dependency>
+
         <!-- H2 -->
         <dependency>
             <groupId>com.h2database</groupId>
@@ -112,6 +118,32 @@
             <artifactId>javax.el-api</artifactId>
             <version>3.0.0</version>
             <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>com.github.opendevl</groupId>
+            <artifactId>json2flat</artifactId>
+            <version>1.0.3</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>com.jayway.jsonpath</groupId>
+                    <artifactId>json-path</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>${version.jackson}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>${version.jackson}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.jayway.jsonpath</groupId>
+            <artifactId>json-path</artifactId>
+            <version>2.4.0</version>
         </dependency>
     </dependencies>
 
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
@@ -47,8 +47,11 @@ public class AsyncExecutorService {
     private int maxRunTime;
     private AsyncQueryDAO asyncQueryDao;
     private static AsyncExecutorService asyncExecutorService = null;
+    private ResultStorageEngine resultStorageEngine;
+
     @Inject
-    private AsyncExecutorService(Elide elide, Integer threadPoolSize, Integer maxRunTime, AsyncQueryDAO asyncQueryDao) {
+    private AsyncExecutorService(Elide elide, Integer threadPoolSize, Integer maxRunTime, AsyncQueryDAO asyncQueryDao,
+            ResultStorageEngine resultStorageEngine) {
         this.elide = elide;
         runners = new HashMap();
 
@@ -60,6 +63,7 @@ public class AsyncExecutorService {
         executor = Executors.newFixedThreadPool(threadPoolSize == null ? defaultThreadpoolSize : threadPoolSize);
         updater = Executors.newFixedThreadPool(threadPoolSize == null ? defaultThreadpoolSize : threadPoolSize);
         this.asyncQueryDao = asyncQueryDao;
+        this.resultStorageEngine = resultStorageEngine;
     }
 
     /**
@@ -69,10 +73,13 @@ public class AsyncExecutorService {
      * @param threadPoolSize thread pool size
      * @param maxRunTime max run times in minutes
      * @param asyncQueryDao DAO Object
+     * @param resultStorageEngine ResultStorageEngine Object
      */
-    public static void init(Elide elide, Integer threadPoolSize, Integer maxRunTime, AsyncQueryDAO asyncQueryDao) {
+    public static void init(Elide elide, Integer threadPoolSize, Integer maxRunTime, AsyncQueryDAO asyncQueryDao,
+            ResultStorageEngine resultStorageEngine) {
         if (asyncExecutorService == null) {
-            asyncExecutorService = new AsyncExecutorService(elide, threadPoolSize, maxRunTime, asyncQueryDao);
+            asyncExecutorService = new AsyncExecutorService(elide, threadPoolSize, maxRunTime, asyncQueryDao,
+                    resultStorageEngine);
         } else {
             log.debug("asyncExecutorService is already initialized.");
         }
@@ -91,15 +98,19 @@ public class AsyncExecutorService {
      * @param queryObj Query Object
      * @param user User
      * @param apiVersion api version
+     * @param requestURL URL of the AsyncRequest pulled from Request Scope
      */
-    public void executeQuery(AsyncQuery queryObj, User user, String apiVersion) {
+    public void executeQuery(AsyncQuery queryObj, User user, String apiVersion,
+            String requestURL) {
 
         QueryRunner runner = runners.get(apiVersion);
         if (runner == null) {
             throw new InvalidOperationException("Invalid API Version");
         }
-        AsyncQueryThread queryWorker = new AsyncQueryThread(queryObj, user, elide, runner, asyncQueryDao, apiVersion);
+        AsyncQueryThread queryWorker = new AsyncQueryThread(queryObj, user, elide, runner, asyncQueryDao,
+                apiVersion, resultStorageEngine, requestURL);
         Future<AsyncQueryResult> task = executor.submit(queryWorker);
+
         try {
             queryObj.setStatus(QueryStatus.PROCESSING);
             AsyncQueryResult queryResultObj = task.get(queryObj.getAsyncAfterSeconds(), TimeUnit.SECONDS);
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
@@ -22,6 +22,7 @@ import lombok.extern.slf4j.Slf4j;
 
 import java.util.ArrayList;
 import java.util.Calendar;
+import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 
@@ -39,6 +40,7 @@ public class AsyncQueryCleanerThread implements Runnable {
     private Elide elide;
     private int queryCleanupDays;
     private AsyncQueryDAO asyncQueryDao;
+    private ResultStorageEngine resultStorageEngine;
 
     @Override
     public void run() {
@@ -55,7 +57,8 @@ public class AsyncQueryCleanerThread implements Runnable {
             Date cleanupDate = DateUtil.calculateFilterDate(Calendar.DATE, queryCleanupDays);
             PathElement createdOnPathElement = new PathElement(AsyncQuery.class, Long.class, "createdOn");
             FilterExpression fltDeleteExp = new LEPredicate(createdOnPathElement, cleanupDate);
-            asyncQueryDao.deleteAsyncQueryAndResultCollection(fltDeleteExp);
+            Collection<AsyncQuery> asyncQueryList = asyncQueryDao.deleteAsyncQueryAndResultCollection(fltDeleteExp);
+            resultStorageEngine.deleteResultsCollection(asyncQueryList);
         } catch (Exception e) {
             log.error("Exception in scheduled cleanup: {}", e);
         }
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
@@ -187,6 +187,7 @@ public class AsyncQueryThread implements Callable<AsyncQueryResult> {
         try {
             isError = (Integer) JsonPath.read(jsonStr, "$.errors.length()") >= 1;
         } catch (PathNotFoundException e) {
+            //ignore when not an error message
             log.error(e.getMessage());
         }
         return isError;
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
@@ -10,9 +10,14 @@ import com.yahoo.elide.ElideResponse;
 import com.yahoo.elide.async.models.AsyncQuery;
 import com.yahoo.elide.async.models.AsyncQueryResult;
 import com.yahoo.elide.async.models.QueryType;
+import com.yahoo.elide.async.models.ResultFormatType;
 import com.yahoo.elide.async.models.ResultType;
 import com.yahoo.elide.graphql.QueryRunner;
 import com.yahoo.elide.security.User;
+
+import com.github.opendevl.JFlat;
+import com.jayway.jsonpath.JsonPath;
+import com.jayway.jsonpath.PathNotFoundException;
 
 import org.apache.http.NameValuePair;
 import org.apache.http.NoHttpResponseException;
@@ -21,8 +26,11 @@ import org.apache.http.client.utils.URIBuilder;
 import lombok.Data;
 import lombok.extern.slf4j.Slf4j;
 
+import java.io.IOException;
 import java.net.URISyntaxException;
+import java.util.Arrays;
 import java.util.Date;
+import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.Callable;
 
@@ -36,7 +44,6 @@ import javax.ws.rs.core.MultivaluedMap;
  */
 @Slf4j
 @Data
-
 public class AsyncQueryThread implements Callable<AsyncQueryResult> {
 
     private AsyncQuery queryObj;
@@ -46,50 +53,79 @@ public class AsyncQueryThread implements Callable<AsyncQueryResult> {
     private final QueryRunner runner;
     private AsyncQueryDAO asyncQueryDao;
     private String apiVersion;
+    private ResultStorageEngine resultStorageEngine;
+    private String requestURL;
 
     @Override
-    public AsyncQueryResult call() throws NoHttpResponseException, URISyntaxException {
-         return processQuery();
+    public AsyncQueryResult call() throws URISyntaxException, IOException {
+        return processQuery();
     }
 
+    /**
+     * Constructor for AsyncQueryThread.
+     * @param queryObj AsyncQuery object
+     * @param user Elide User Object
+     * @param elide Elide Object
+     * @param runner QueryRunner Object for GraphQL executions
+     * @param asyncQueryDao AsyncQueryDAO implementation Object
+     * @param apiVersion Api Version
+     * @param resultStorageEngine ResultStorageEngine implementation Object
+     * @param requestURL URL of the AsyncRequest pulled from Request Scope
+     */
     public AsyncQueryThread(AsyncQuery queryObj, User user, Elide elide, QueryRunner runner,
-            AsyncQueryDAO asyncQueryDao, String apiVersion) {
+            AsyncQueryDAO asyncQueryDao, String apiVersion, ResultStorageEngine resultStorageEngine,
+            String requestURL) {
         this.queryObj = queryObj;
         this.user = user;
         this.elide = elide;
         this.runner = runner;
         this.asyncQueryDao = asyncQueryDao;
         this.apiVersion = apiVersion;
+        this.resultStorageEngine = resultStorageEngine;
+        this.requestURL = requestURL;
     }
 
-
-   /**
-    * This is the main method which processes the Async Query request, executes the query and updates
-    * values for AsyncQuery and AsyncQueryResult models accordingly.
-    * @return AsyncQueryResult
-    * @throws URISyntaxException
-    * @throws NoHttpResponseException
-    */
-    protected AsyncQueryResult processQuery() throws URISyntaxException, NoHttpResponseException {
+    /**
+     * This is the main method which processes the Async Query request, executes the query and updates
+     * values for AsyncQuery and AsyncQueryResult models accordingly.
+     * @return AsyncQueryResult
+     * @throws URISyntaxException URISyntaxException
+     * @throws IOException IOException
+     */
+    protected AsyncQueryResult processQuery() throws URISyntaxException, IOException {
         UUID requestId = UUID.fromString(queryObj.getRequestId());
 
         ElideResponse response = null;
+        Integer recCount = null;
+        String responseBody = null;
+        boolean isError = false;
         log.debug("AsyncQuery Object from request: {}", queryObj);
         if (queryObj.getQueryType().equals(QueryType.JSONAPI_V1_0)) {
             MultivaluedMap<String, String> queryParams = getQueryParams(queryObj.getQuery());
             log.debug("Extracted QueryParams from AsyncQuery Object: {}", queryParams);
 
-            //TODO - we need to add the baseUrlEndpoint to the queryObject.
+            //TODO - we need to add the baseURLEndpoint to the queryObject.
             response = elide.get("", getPath(queryObj.getQuery()), queryParams, user, apiVersion, requestId);
+            responseBody = response.getBody();
+            isError = checkJsonStrErrorMessage(responseBody);
             log.debug("JSONAPI_V1_0 getResponseCode: {}, JSONAPI_V1_0 getBody: {}",
-                    response.getResponseCode(), response.getBody());
+                    response.getResponseCode(), responseBody);
+
+            if (response.getResponseCode() == 200 && !isError) {
+                recCount = calculateRecordsJSON(responseBody);
+            }
         }
         else if (queryObj.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
-            //TODO - we need to add the baseUrlEndpoint to the queryObject.
-
+            //TODO - we need to add the baseURLEndpoint to the queryObject.
             response = runner.run("", queryObj.getQuery(), user, requestId);
+            responseBody = response.getBody();
+            isError = checkJsonStrErrorMessage(responseBody);
             log.debug("GRAPHQL_V1_0 getResponseCode: {}, GRAPHQL_V1_0 getBody: {}",
-                    response.getResponseCode(), response.getBody());
+                    response.getResponseCode(), responseBody);
+
+            if (response.getResponseCode() == 200 && !isError) {
+                recCount = calculateRecordsGraphQL(responseBody);
+            }
         }
         if (response == null) {
             throw new NoHttpResponseException("Response for request returned as null");
@@ -99,16 +135,96 @@ public class AsyncQueryThread implements Callable<AsyncQueryResult> {
 
         queryResultObj = new AsyncQueryResult();
         queryResultObj.setHttpStatus(response.getResponseCode());
-        queryResultObj.setResponseBody(response.getBody());
-        queryResultObj.setContentLength(response.getBody().length());
-        queryResultObj.setResultType(ResultType.EMBEDDED);
+        queryResultObj.setContentLength(responseBody.length());
+        queryResultObj.setRecordCount(recCount);
         queryResultObj.setCompletedOn(new Date());
+
+        String tempResult = queryObj.getResultFormatType() == ResultFormatType.CSV && !isError
+                ? convertJsonToCSV(responseBody) : responseBody;
+
+        if (queryObj.getResultType() == ResultType.EMBEDDED || isError) {
+            queryResultObj.setResponseBody(tempResult);
+        } else if (queryObj.getResultType() == ResultType.DOWNLOAD) {
+            if (resultStorageEngine == null) {
+                throw new IllegalStateException("Unable to store async results for download");
+            }
+            queryResultObj = resultStorageEngine.storeResults(queryResultObj, tempResult, queryObj.getId());
+            queryResultObj.setResponseBody(resultStorageEngine.generateDownloadUrl(requestURL,
+                    queryObj.getId()).toString());
+        }
 
         return queryResultObj;
     }
 
     /**
-     * This method parses the url and gets the query params.
+     * This method calculates the number of records from the response with JSON API.
+     * @param jsonStr is the response.getBody() we get from the response
+     * @return rec is the recordCount
+     */
+    protected Integer calculateRecordsJSON(String jsonStr) {
+        return JsonPath.read(jsonStr, "$.data.length()");
+    }
+
+    /**
+     * This method calculates the number of records from the response with GRAPHQL API.
+     * @param jsonStr is the response.getBody() we get from the response
+     * @return rec is the recordCount
+     * @throws IOException Exception thrown by JsonPath
+     */
+    protected Integer calculateRecordsGraphQL(String jsonStr) throws IOException {
+        List<Integer> countList = JsonPath.read(jsonStr, "$..edges.length()");
+        return countList.get(0);
+    }
+
+    /**
+     * This method checks if the json string is error message.
+     * @param jsonStr is the response.getBody() we get from the response
+     * @return is the message an error message
+     */
+    protected boolean checkJsonStrErrorMessage(String jsonStr) {
+        boolean isError = false;
+        try {
+            isError = (Integer) JsonPath.read(jsonStr, "$.errors.length()") >= 1;
+        } catch (PathNotFoundException e) {
+            log.error(e.getMessage());
+        }
+        return isError;
+    }
+
+    /**
+     * This method converts the JSON response to a CSV response type.
+     * @param jsonStr is the response.getBody() of the query
+     * @return retuns a string which nis in CSV format
+     * @throws IllegalStateException Exception thrown
+     */
+    protected String convertJsonToCSV(String jsonStr) {
+        if (jsonStr == null) {
+            return null;
+        }
+        StringBuilder str = new StringBuilder();
+        JFlat flatMe = new JFlat(jsonStr);
+        List<Object[]> json2csv;
+        try {
+            json2csv = flatMe.json2Sheet().headerSeparator("_").getJsonAsSheet();
+            for (Object[] obj : json2csv) {
+                String objString = Arrays.toString(obj);
+                if (objString != null) {
+                    objString = objString.substring(1, objString.length() - 1);
+                }
+                str.append(objString);
+                str.append(System.getProperty("line.separator"));
+            }
+        } catch (Exception e) {
+            log.debug("Exception while converting to CSV: {}", e.getMessage());
+            throw new IllegalStateException(e);
+        }
+
+        return str.toString();
+
+    }
+
+    /**
+     * This method parses the URL and gets the query params.
      * And adds them into a MultivaluedMap to be used by underlying Elide.get method
      * @param query query from the Async request
      * @throws URISyntaxException URISyntaxException from malformed or incorrect URI
@@ -125,7 +241,7 @@ public class AsyncQueryThread implements Callable<AsyncQueryResult> {
     }
 
     /**
-     * This method parses the url and gets the query params.
+     * This method parses the URL and gets the query params.
      * And retrieves path to be used by underlying Elide.get method
      * @param query query from the Async request
      * @throws URISyntaxException URISyntaxException from malformed or incorrect URI
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
@@ -104,7 +104,7 @@ public class AsyncQueryThread implements Callable<AsyncQueryResult> {
             MultivaluedMap<String, String> queryParams = getQueryParams(queryObj.getQuery());
             log.debug("Extracted QueryParams from AsyncQuery Object: {}", queryParams);
 
-            //TODO - we need to add the baseURLEndpoint to the queryObject.
+            //TODO - we need to add the baseUrlEndpoint to the queryObject.
             response = elide.get("", getPath(queryObj.getQuery()), queryParams, user, apiVersion, requestId);
             responseBody = response.getBody();
             isError = checkJsonStrErrorMessage(responseBody);
@@ -173,7 +173,8 @@ public class AsyncQueryThread implements Callable<AsyncQueryResult> {
      */
     protected Integer calculateRecordsGraphQL(String jsonStr) throws IOException {
         List<Integer> countList = JsonPath.read(jsonStr, "$..edges.length()");
-        return countList.get(0);
+        Integer count = countList.size() > 0 ? countList.get(0) : 0;
+        return count;
     }
 
     /**
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
@@ -188,7 +188,7 @@ public class AsyncQueryThread implements Callable<AsyncQueryResult> {
             isError = (Integer) JsonPath.read(jsonStr, "$.errors.length()") >= 1;
         } catch (PathNotFoundException e) {
             //ignore when not an error message
-            log.error(e.getMessage());
+            log.debug(e.getMessage());
         }
         return isError;
     }
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/DBUtil.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/DBUtil.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2020, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+
+package com.yahoo.elide.async.service;
+
+import static com.yahoo.elide.core.EntityDictionary.NO_VERSION;
+
+import com.yahoo.elide.ElideSettings;
+import com.yahoo.elide.core.DataStore;
+import com.yahoo.elide.core.DataStoreTransaction;
+import com.yahoo.elide.core.RequestScope;
+import com.yahoo.elide.jsonapi.models.JsonApiDocument;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+import java.util.UUID;
+
+import javax.ws.rs.core.MultivaluedHashMap;
+import javax.ws.rs.core.MultivaluedMap;
+
+/**
+ * Utility class which implements a static method executeInTransaction.
+ */
+@Slf4j
+public class DBUtil {
+
+    /**
+     * This method creates a transaction from the datastore, performs the DB action using
+     * a generic functional interface and closes the transaction.
+     * @param dataStore Elide datastore retrieved from Elide object
+     * @param action Functional interface to perform DB action
+     * @return Object Returns Entity Object (AsyncQueryResult or AsyncResult)
+     */
+    protected static Object executeInTransaction(ElideSettings elideSettings,
+        DataStore dataStore, Transactional action) {
+        log.debug("executeInTransaction");
+        Object result = null;
+        try (DataStoreTransaction tx = dataStore.beginTransaction()) {
+            JsonApiDocument jsonApiDoc = new JsonApiDocument();
+            MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
+            RequestScope scope = new RequestScope(null, "query", NO_VERSION, jsonApiDoc,
+                    tx, null, queryParams, UUID.randomUUID(), elideSettings);
+            result = action.execute(tx, scope);
+            tx.flush(scope);
+            tx.commit(scope);
+        } catch (IOException e) {
+            log.error("IOException: {}", e);
+            throw new IllegalStateException(e);
+        }
+        return result;
+    }
+}
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
@@ -5,31 +5,30 @@
  */
 package com.yahoo.elide.async.service;
 
-import static com.yahoo.elide.core.EntityDictionary.NO_VERSION;
-
-import com.yahoo.elide.Elide;
+import com.yahoo.elide.ElideSettings;
 import com.yahoo.elide.async.models.AsyncQuery;
 import com.yahoo.elide.async.models.AsyncQueryResult;
 import com.yahoo.elide.async.models.QueryStatus;
 import com.yahoo.elide.core.DataStore;
+<<<<<<< HEAD
 import com.yahoo.elide.core.DataStoreTransaction;
 import com.yahoo.elide.core.RequestScope;
+=======
+import com.yahoo.elide.core.EntityDictionary;
+import com.yahoo.elide.core.filter.dialect.ParseException;
+import com.yahoo.elide.core.filter.dialect.RSQLFilterDialect;
+>>>>>>> Adding record count to AsyncQueryResult and related changes
 import com.yahoo.elide.core.filter.expression.FilterExpression;
-import com.yahoo.elide.jsonapi.models.JsonApiDocument;
 import com.yahoo.elide.request.EntityProjection;
 
 import lombok.Getter;
 import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 
-import java.io.IOException;
 import java.util.Collection;
 import java.util.Iterator;
-import java.util.UUID;
 
 import javax.inject.Singleton;
-import javax.ws.rs.core.MultivaluedHashMap;
-import javax.ws.rs.core.MultivaluedMap;
 
 /**
  * Utility class which implements AsyncQueryDAO.
@@ -39,21 +38,22 @@ import javax.ws.rs.core.MultivaluedMap;
 @Getter
 public class DefaultAsyncQueryDAO implements AsyncQueryDAO {
 
-    @Setter private Elide elide;
+    @Setter private ElideSettings elideSettings;
     @Setter private DataStore dataStore;
 
     // Default constructor is needed for standalone implementation for override in getAsyncQueryDao
     public DefaultAsyncQueryDAO() {
     }
 
-    public DefaultAsyncQueryDAO(Elide elide, DataStore dataStore) {
-        this.elide = elide;
+    public DefaultAsyncQueryDAO(ElideSettings elideSettings, DataStore dataStore) {
+        this.elideSettings = elideSettings;
         this.dataStore = dataStore;
     }
 
     @Override
     public AsyncQuery updateStatus(String asyncQueryId, QueryStatus status) {
-        AsyncQuery queryObj = (AsyncQuery) executeInTransaction(dataStore, (tx, scope) -> {
+        AsyncQuery queryObj = (AsyncQuery) DBUtil.executeInTransaction(elideSettings,
+                dataStore, (tx, scope) -> {
             EntityProjection asyncQueryCollection = EntityProjection.builder()
                     .type(AsyncQuery.class)
                     .build();
@@ -70,7 +70,7 @@ public class DefaultAsyncQueryDAO implements AsyncQueryDAO {
             QueryStatus status) {
         return updateAsyncQueryCollection(filterExpression, (asyncQuery) -> {
             asyncQuery.setStatus(status);
-            });
+        });
     }
 
     /**
@@ -86,23 +86,23 @@ public class DefaultAsyncQueryDAO implements AsyncQueryDAO {
         log.debug("updateAsyncQueryCollection");
 
         Collection<AsyncQuery> asyncQueryList = null;
-         asyncQueryList = (Collection<AsyncQuery>) executeInTransaction(dataStore,
+        asyncQueryList = (Collection<AsyncQuery>) DBUtil.executeInTransaction(dataStore,
                     (tx, scope) -> {
-             EntityProjection asyncQueryCollection = EntityProjection.builder()
-                     .type(AsyncQuery.class)
-                     .filterExpression(filterExpression)
-                     .build();
+            EntityProjection asyncQueryCollection = EntityProjection.builder()
+                    .type(AsyncQuery.class)
+                    .filterExpression(filterExpression)
+                    .build();
 
-             Iterable<Object> loaded = tx.loadObjects(asyncQueryCollection, scope);
-             Iterator<Object> itr = loaded.iterator();
+            Iterable<Object> loaded = tx.loadObjects(asyncQueryCollection, scope);
+            Iterator<Object> itr = loaded.iterator();
 
-             while (itr.hasNext()) {
-                 AsyncQuery query = (AsyncQuery) itr.next();
-                 updateFunction.update(query);
-                 tx.save(query, scope);
-             }
-             return loaded;
-         });
+            while (itr.hasNext()) {
+                AsyncQuery query = (AsyncQuery) itr.next();
+                updateFunction.update(query);
+                tx.save(query, scope);
+            }
+            return loaded;
+        });
         return asyncQueryList;
     }
 
@@ -111,7 +111,7 @@ public class DefaultAsyncQueryDAO implements AsyncQueryDAO {
     public Collection<AsyncQuery> deleteAsyncQueryAndResultCollection(FilterExpression filterExpression) {
         log.debug("deleteAsyncQueryAndResultCollection");
         Collection<AsyncQuery> asyncQueryList = null;
-        asyncQueryList = (Collection<AsyncQuery>) executeInTransaction(dataStore, (tx, scope) -> {
+        asyncQueryList = (Collection<AsyncQuery>) DBUtil.executeInTransaction(dataStore, (tx, scope) -> {
             EntityProjection asyncQueryCollection = EntityProjection.builder()
                     .type(AsyncQuery.class)
                     .filterExpression(filterExpression)
@@ -132,10 +132,10 @@ public class DefaultAsyncQueryDAO implements AsyncQueryDAO {
     }
 
     @Override
-    public AsyncQuery updateAsyncQueryResult(AsyncQueryResult asyncQueryResult,
-            String asyncQueryId) {
+    public AsyncQuery updateAsyncQueryResult(AsyncQueryResult asyncQueryResult, String asyncQueryId) {
         log.debug("updateAsyncQueryResult");
-        AsyncQuery queryObj = (AsyncQuery) executeInTransaction(dataStore, (tx, scope) -> {
+        AsyncQuery queryObj = (AsyncQuery) DBUtil.executeInTransaction(elideSettings,
+                dataStore, (tx, scope) -> {
             EntityProjection asyncQueryCollection = EntityProjection.builder()
                     .type(AsyncQuery.class)
                     .build();
@@ -152,38 +152,13 @@ public class DefaultAsyncQueryDAO implements AsyncQueryDAO {
         return queryObj;
     }
 
-    /**
-     * This method creates a transaction from the datastore, performs the DB action using
-     * a generic functional interface and closes the transaction.
-     * @param dataStore Elide datastore retrieved from Elide object
-     * @param action Functional interface to perform DB action
-     * @return Object Returns Entity Object (AsyncQueryResult or AsyncResult)
-     */
-    protected Object executeInTransaction(DataStore dataStore, Transactional action) {
-        log.debug("executeInTransaction");
-        Object result = null;
-        try (DataStoreTransaction tx = dataStore.beginTransaction()) {
-            JsonApiDocument jsonApiDoc = new JsonApiDocument();
-            MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
-            RequestScope scope = new RequestScope("", "query", NO_VERSION, jsonApiDoc,
-                    tx, null, queryParams, UUID.randomUUID(), elide.getElideSettings());
-            result = action.execute(tx, scope);
-            tx.flush(scope);
-            tx.commit(scope);
-        } catch (IOException e) {
-            log.error("IOException: {}", e);
-            throw new IllegalStateException(e);
-        }
-        return result;
-    }
-
     @Override
     @SuppressWarnings("unchecked")
     public Collection<AsyncQuery> loadAsyncQueryCollection(FilterExpression filterExpression) {
         Collection<AsyncQuery> asyncQueryList = null;
         log.debug("loadAsyncQueryCollection");
         try {
-            asyncQueryList = (Collection<AsyncQuery>) executeInTransaction(dataStore, (tx, scope) -> {
+            asyncQueryList = (Collection<AsyncQuery>) DBUtil.executeInTransaction(dataStore, (tx, scope) -> {
 
                 EntityProjection asyncQueryCollection = EntityProjection.builder()
                         .type(AsyncQuery.class)
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
@@ -10,14 +10,6 @@ import com.yahoo.elide.async.models.AsyncQuery;
 import com.yahoo.elide.async.models.AsyncQueryResult;
 import com.yahoo.elide.async.models.QueryStatus;
 import com.yahoo.elide.core.DataStore;
-<<<<<<< HEAD
-import com.yahoo.elide.core.DataStoreTransaction;
-import com.yahoo.elide.core.RequestScope;
-=======
-import com.yahoo.elide.core.EntityDictionary;
-import com.yahoo.elide.core.filter.dialect.ParseException;
-import com.yahoo.elide.core.filter.dialect.RSQLFilterDialect;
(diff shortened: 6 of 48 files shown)


elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -48,7 +52,13 @@ protected void deleteAsyncQuery() {
 
         String filterExpression = "createdOn=le='" + cleanupDateFormatted + "'";
 
-        asyncQueryDao.deleteAsyncQueryAndResultCollection(filterExpression);
+        Collection<AsyncQuery> asyncQueryList = asyncQueryDao.deleteAsyncQueryAndResultCollection(filterExpression);
+        Iterator itr = asyncQueryList.iterator();
+
+        while (itr.hasNext()) {
+            AsyncQuery query = (AsyncQuery) itr.next();
+            resultStorageEngine.deleteResultsByID(query.getId());
```
moizarafat, line 60: I think its better to rename the function deleteResultsByID to deleteResultsCollection. It should take `Collection<AsyncQuery>` as an argument.  All of the iterator logic can be then moved to the function implementation. The reason being in our database interpretation we have to delete each row one by one but for a filesystem-based, we can delete multiple at the same time.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -72,35 +81,164 @@ public AsyncQueryThread(AsyncQuery queryObj, User user, Elide elide, QueryRunner
...
 
         queryResultObj = new AsyncQueryResult();
         queryResultObj.setHttpStatus(response.getResponseCode());
-        queryResultObj.setResponseBody(response.getBody());
         queryResultObj.setContentLength(response.getBody().length());
-        queryResultObj.setResultType(ResultType.EMBEDDED);
+        queryResultObj.setRecordCount(recCount);
         queryResultObj.setCompletedOn(new Date());
 
+        if (timeElapsed - asyncAfterMilliSeconds > 0) {
```
moizarafat, line 125: Lets do `>=`

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -72,35 +81,164 @@ public AsyncQueryThread(AsyncQuery queryObj, User user, Elide elide, QueryRunner
...
+            String jsonToCSV = convertJsonToCSV(response.getBody());
```
moizarafat, line 126: This should be in the if block at L-128

elide-async/src/main/java/com/yahoo/elide/async/service/EIT.java
```diff
@@ -0,0 +1,53 @@
...
+import java.io.IOException;
+
+import javax.ws.rs.core.MultivaluedHashMap;
+import javax.ws.rs.core.MultivaluedMap;
+
+/**
+ * Utility class which implements a static method executeInTransaction.
+ */
+@Slf4j
+public class EIT {
```
moizarafat, line 27: Lets call it `DBUtil`

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -72,35 +81,164 @@ public AsyncQueryThread(AsyncQuery queryObj, User user, Elide elide, QueryRunner
...
+            URL url = null;
+            if (queryObj.getResultFormatType() == ResultFormatType.CSV) {
+                url = resultStorageEngine.storeResults(queryObj.getId(), jsonToCSV);
```
moizarafat, line 129: we can create an AsyncResultStorage Object with Id and Blob. and pass that as argument to storeResults method.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,160 @@
...
+
+        URL url = null;
+        String buildURL = baseURL + "/AsyncQueryResultStorage/" + asyncQueryID;
+        try {
+            url = new URL(buildURL);
+        } catch (MalformedURLException e) {
+            log.error("Exception: {}", e);
+            throw new IllegalStateException(e);
+        }
+
```
moizarafat, line 70: the logic to store the results in db is missing.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    public DefaultResultStorageEngine(Elide elide, DataStore dataStore, String baseURL) {
+        this.elide = elide;
+        this.dataStore = dataStore;
+        this.baseURL = baseURL;
+    }
+
+    @Override
+    public URL storeResults(String asyncQueryID, String responseBody) {
```
moizarafat, line 49: We should change the method to take asyncQueryID and Byte[] as the input args. ProcessQuery should send byte[] instead of responseBody

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,160 @@
...
+            log.error("Exception: {}", e);
+            throw new IllegalStateException(e);
+        }
+
+        return url;
+
+    }
+
+    @Override
+    public AsyncQueryResultStorage getResultsByID(String asyncQueryID) {
```
moizarafat, line 76: Instead of returning AsyncQueryResultStorage we should return byte[] to keep the code extensible.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,160 @@
...
+                    dataStore, (tx, scope) -> {
+
+                EntityProjection asyncQueryCollection = EntityProjection.builder()
+                        .type(AsyncQueryResultStorage.class)
+                        .build();
+
+                Object loaded = tx.loadObject(asyncQueryCollection, asyncQueryID, scope);
+
+                return loaded;
+
```
moizarafat, line 92: remove extra line.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,160 @@
...
+        return url;
+
+    }
+
+    @Override
+    public AsyncQueryResultStorage getResultsByID(String asyncQueryID) {
+        log.debug("getResultsByID");
+
+        AsyncQueryResultStorage asyncQueryResultStorages = null;
```
moizarafat, line 79: asyncQueryResultStorages should be singular.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,160 @@
...
+
+            });
+        } catch (Exception e) {
+            log.error("Exception: {}", e);
+        }
+        return asyncQueryResultStorages;
+    }
+
+    @Override
+    public AsyncQueryResultStorage deleteResultsByID(String asyncQueryID) {
```
moizarafat, line 102: lets make the return type as void.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,160 @@
...
+                return loaded;
+            });
+        } catch (Exception e) {
+            log.error("Exception: {}", e);
+        }
+        return asyncQueryResultStorages;
+    }
+
+    @Override
+    public Collection<AsyncQueryResultStorage> deleteAllResults() {
```
moizarafat, line 131: return type as void

moizarafat: On 2nd thought, lets remove this method. I dont think we will have a use case for it anymore.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -105,4 +102,23 @@ public AsyncQueryDAO buildAsyncQueryDAO(Elide elide) {
         Elide asyncElide = new Elide(asyncElideSettings);
         return new DefaultAsyncQueryDAO(asyncElide, asyncElide.getDataStore());
     }
+
+    /**
+     * Configure the AsyncQueryDAO used by async query requests.
```
moizarafat, line 107: comment needs to be updated.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -105,4 +102,23 @@ public AsyncQueryDAO buildAsyncQueryDAO(Elide elide) {
...
+    @ConditionalOnProperty(prefix = "elide.async", name = "defaultResultStorageEngine", matchIfMissing = true)
+    public ResultStorageEngine buildResultStorageEngine(Elide elide) {
+        // Creating a new ElideSettings and Elide object for Async services
+        // which will have ISO8601 Dates. Used for DefaultAsyncQueryDAO.
+        ElideSettings asyncElideSettings = new ElideSettingsBuilder(elide.getDataStore())
+                .withEntityDictionary(elide.getElideSettings().getDictionary())
+                .withISO8601Dates("yyyy-MM-dd'T'HH:mm'Z'", TimeZone.getTimeZone("UTC"))
+                .build();
+        Elide asyncElide = new Elide(asyncElideSettings);
+        return new DefaultResultStorageEngine(asyncElide, asyncElide.getDataStore(), "http://localhost:8080");
```
moizarafat, line 122: `http://localhost:8080` should not be hardcoded. we should use `settings.getAsync().getBaseUrl()`;

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -39,4 +39,14 @@
...
 
+    /**
+     * Whether or not the controller is enabled.
+     */
+    private boolean enabled = false;
+
+    /**
+     * The URL path prefix for the controller.
+     */
+    private String baseURL = "/";
```
moizarafat, line 50: lets call this downloadBaseURL

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -105,4 +102,23 @@ public AsyncQueryDAO buildAsyncQueryDAO(Elide elide) {
...
+     * @param elide elideObject.
+     * @return an AsyncQueryDAO object.
+     */
+    @Bean
+    @ConditionalOnMissingBean
+    @ConditionalOnProperty(prefix = "elide.async", name = "defaultResultStorageEngine", matchIfMissing = true)
+    public ResultStorageEngine buildResultStorageEngine(Elide elide) {
+        // Creating a new ElideSettings and Elide object for Async services
+        // which will have ISO8601 Dates. Used for DefaultAsyncQueryDAO.
+        ElideSettings asyncElideSettings = new ElideSettingsBuilder(elide.getDataStore())
```
moizarafat, line 117: L117 to L121 is common between the methods in this class. Instead of duplicating we can move to a private method and use it.

elide-spring/elide-spring-boot-autoconfigure/src/test/resources/application.yaml
```diff
@@ -19,6 +19,7 @@ elide:
     cleanupEnabled: true
     queryCleanupDays: 7
     defaultAsyncQueryDAO: true
+    baseURL: http://localhost:8080
```
moizarafat, line 22: rename to downloadBaseURL

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -17,10 +17,7 @@
 import com.yahoo.elide.async.hooks.ExecuteQueryHook;
 import com.yahoo.elide.async.hooks.UpdatePrincipalNameHook;
 import com.yahoo.elide.async.models.AsyncQuery;
-import com.yahoo.elide.async.service.AsyncCleanerService;
-import com.yahoo.elide.async.service.AsyncExecutorService;
-import com.yahoo.elide.async.service.AsyncQueryDAO;
-import com.yahoo.elide.async.service.DefaultAsyncQueryDAO;
+import com.yahoo.elide.async.service.*;
```
moizarafat, line 20: Lets not use * , instead put individual dependencies.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -243,6 +244,24 @@ default AsyncQueryDAO getAsyncQueryDAO() {
...
+    default ResultStorageEngine getResultStorageEngine() {
+        return null;
+    }
+
+    /**
+     * Generates a default a baseURL.
+     *
+     * @return a URL in String format.
+     */
+    default String getBaseURL() {
```
moizarafat, line 261: rename to getAsyncDownloadBaseURL.

~/.vzmi_tool/vzmi_tool.log
```diff
@@ -0,0 +1 @@
+Activating /Users/cpatlolla/python_virtualenvs/oath_tools virtual environment
```
moizarafat, line 1: remove this file.

~/.vzmi_tools/vzmi_tools_update.log
```diff
@@ -0,0 +1,2 @@
+Updating bootstrapper package
```
moizarafat, line 1: remove this fuile

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -57,11 +57,11 @@
 
     @Override
     public AsyncQueryResult call() throws NoHttpResponseException, URISyntaxException {
-         return processQuery();
+        return processQuery();
     }
 
     public AsyncQueryThread(AsyncQuery queryObj, User user, Elide elide, QueryRunner runner,
-            AsyncQueryDAO asyncQueryDao, String apiVersion, ResultStorageEngine resultStorageEngine) {
+                            AsyncQueryDAO asyncQueryDao, String apiVersion, ResultStorageEngine resultStorageEngine) {
```
moizarafat, line 64: Instead of these many spaces to indent, 8 spaces from the beginning of the line will be fine to use here.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -65,10 +65,10 @@ public AsyncQuery updateStatus(String asyncQueryId, QueryStatus status) {
 
     @Override
     public Collection<AsyncQuery> updateStatusAsyncQueryCollection(String filterExpression,
-            QueryStatus status) {
+                                                                   QueryStatus status) {
```
moizarafat, line 68: 8 Spaces from the beginning should be fine here

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -80,31 +80,31 @@ public AsyncQuery updateStatus(String asyncQueryId, QueryStatus status) {
      */
     @SuppressWarnings("unchecked")
     private Collection<AsyncQuery> updateAsyncQueryCollection(String filterExpression,
-            UpdateQuery updateFunction) {
+                                                              UpdateQuery updateFunction) {
```
moizarafat, line 83: Same comment on spaces as others.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -147,9 +147,9 @@ public AsyncQuery updateStatus(String asyncQueryId, QueryStatus status) {
 
     @Override
     public AsyncQuery updateAsyncQueryResult(AsyncQueryResult asyncQueryResult,
-            String asyncQueryId) {
+                                             String asyncQueryId) {
```
moizarafat, line 150: Same as others

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -68,93 +69,117 @@ public URL storeResults(String asyncQueryID, String responseBody) {
...
+        AsyncQueryResultStorage aQRS = null;
+        try {
+            aQRS = (AsyncQueryResultStorage) DBUtil.executeInTransaction(elide, dataStore,
+                    (tx, scope) -> {
+
+                        EntityProjection asyncQueryCollection = EntityProjection.builder()
+                                .type(AsyncQueryResultStorage.class)
+                                .build();
+
+                        Object loaded = tx.loadObject(asyncQueryCollection, asyncQueryResultStorage.getId(), scope);
```
moizarafat, line 81: This is going to return null cos the AsyncQueryResultStorage object does not exist in the database. We should be creating a new AsyncQueryResultStorage with asyncQueryID and response. Then we should just call tx.save for that object.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -511,7 +511,7 @@ default QueryEngine getQueryEngine(MetaDataStore metaDataStore, EntityManagerFac
     }
 
     static Set<Class<?>> getDynamicClassesIfAvailable(Optional<ElideDynamicEntityCompiler> optionalCompiler,
-            Class<?> classz) {
+                                                      Class<?> classz) {
```
moizarafat, line 514: Same comment on indentation as others.

moizarafat: Please do the same for other methods in the class.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -95,30 +97,37 @@ public AsyncCleanerService buildAsyncCleanerService(Elide elide, ElideConfigProp
...
+        // which will have ISO8601 Dates. Used for DefaultResultStorageEngine.
+        Elide asyncElide = getElideSettings(elide);
+        return new DefaultResultStorageEngine(asyncElide, asyncElide.getDataStore(),
+                settings.getAsync().getDownloadBaseURL());
+    }
+
+    /**
+     * Creating a new ElideSettings and Elide object for Async services
+     * @param elide elideObject
+     * @return new ElideSettings and Elide object for Async services
```
moizarafat, line 123: Comment needs adjustment. It only returns Elide Object not ElideSettings.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -95,30 +97,37 @@ public AsyncCleanerService buildAsyncCleanerService(Elide elide, ElideConfigProp
...
+     */
+    private Elide getElideSettings(Elide elide) {
```
moizarafat, line 125: lets call it getAsyncElideInstance

elide-async/src/main/java/com/yahoo/elide/async/service/ResultStorageEngine.java
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Deletes all the records from the AsyncQuery Collection .
+     */
+    public void deleteResultsCollection(Collection<AsyncQuery> asyncQueryCollection);
+
+    /**
+     * Deletes the record with ID as AsyncQueryID in the AsyncQueryResultStorage table and returns the record.
+     * @param AsyncQueryID AsyncQueryID is the query ID of the AsyncQuery
+     */
+    public void deleteResultsByID(String AsyncQueryID);
```
moizarafat, line 43: I dont think we are making use of this method, so lets remove it.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -68,93 +69,117 @@ public URL storeResults(String asyncQueryID, String responseBody) {
...
 
-                return loaded;
+                        Object loaded = tx.loadObject(asyncQueryCollection, asyncQueryID, scope);
 
+                        return loaded;
+                    });
 
-            });
+            Blob result = asyncQueryResultStorage.getResult();
+            byteResult = result.getBytes(1, (int) result.length());
```
moizarafat, line 114: Why would this be needed?

moizarafat: (int) should be long.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -68,93 +69,117 @@ public URL storeResults(String asyncQueryID, String responseBody) {
...
-
+                        Iterator<AsyncQuery> itr = asyncQueryList.iterator();
+                        Object loaded = null;
                         while (itr.hasNext()) {
-                            AsyncQueryResultStorage query = (AsyncQueryResultStorage) itr.next();
-                            if (query != null) {
-                                tx.delete(query, scope);
+                            AsyncQuery query = itr.next();
+                            loaded = tx.loadObject(asyncQueryCollection, query.getId(), scope);
+                            Object result = loaded;
```
moizarafat, line 170: you can directly check if loaded != null in the next line.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -68,93 +69,117 @@ public URL storeResults(String asyncQueryID, String responseBody) {
...
+                    dataStore, (tx, scope) -> {
 
                         EntityProjection asyncQueryCollection = EntityProjection.builder()
                                 .type(AsyncQueryResultStorage.class)
                                 .build();
 
-                        Iterable<Object> loaded = tx.loadObjects(asyncQueryCollection, scope);
-                        Iterator<Object> itr = loaded.iterator();
-
+                        Iterator<AsyncQuery> itr = asyncQueryList.iterator();
```
moizarafat, line 165: Lets keep this iterator outside executeInTransaction block. we dont want to keep the transaction open for too long. We should iterate, take 1 asyncId, execute in transaction, then move on to next Id.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -78,7 +79,8 @@ public AsyncExecutorService buildAsyncExecutorService(Elide elide, ElideConfigPr
     @ConditionalOnMissingBean
     @ConditionalOnProperty(prefix = "elide.async", name = "cleanupEnabled", matchIfMissing = false)
     public AsyncCleanerService buildAsyncCleanerService(Elide elide, ElideConfigProperties settings,
-            AsyncQueryDAO asyncQueryDao, ResultStorageEngine resultStorageEngine) {
+                                                        AsyncQueryDAO asyncQueryDao,
```
moizarafat, line 82: Same comment on indenatation as others. Please adjust in the whole class.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryThreadTest.java
```diff
@@ -69,9 +71,32 @@ public void testProcessQueryGraphQl() throws NoHttpResponseException, URISyntaxE
...
     }
+
+    @Test
+    public void testConvertJsonToCSV() {
+
+        String csvStr = "[/key][\"value\"]";
+        String jsonStr = "{\"key\":\"value\"}";
+
+        AsyncQuery queryObj = new AsyncQuery();
+        ElideResponse response = new ElideResponse(200, "ResponseBody");
```
moizarafat, line 88: instead of ResponseBody should this be jsonStr?

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryThreadTest.java
```diff
@@ -69,9 +71,32 @@ public void testProcessQueryGraphQl() throws NoHttpResponseException, URISyntaxE
...
+        String query = "/group?sort=commonName&fields%5Bgroup%5D=commonName,description";
+        String id = "edc4a871-dff2-4054-804e-d80075cf827d";
+        queryObj.setId(id);
+        queryObj.setQuery(query);
+        queryObj.setQueryType(QueryType.JSONAPI_V1_0);
+        when(elide.get(anyString(), any(), any(), anyString())).thenReturn(response);
+        AsyncQueryThread queryThread = new AsyncQueryThread(queryObj, user, elide, runner, asyncQueryDao, "v1",
+                resultStorageEngine);
+
+        String jsonToCSV = queryThread.convertJsonToCSV(jsonStr);
```
moizarafat, line 98: If the intention is to just test convertJsonToCSV method. You can just create a mock queryThread object and test it. You can remove L-87 to L97

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -45,62 +53,195 @@
...
+        }
+        StringBuilder str = new StringBuilder();
+
+        try {
+            new JSONObject(jsonStr);
+            JFlat flatMe = new JFlat(jsonStr);
+            List<Object[]> json2csv = flatMe.json2Sheet().getJsonAsSheet();
+
+            for (Object[] obj : json2csv) {
+                str.append(Arrays.toString(obj));
```
moizarafat, line 232: Are we missing the comma delimiter?

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryThreadTest.java
```diff
@@ -69,9 +71,32 @@ public void testProcessQueryGraphQl() throws NoHttpResponseException, URISyntaxE
...
+                resultStorageEngine);
         queryResultObj = queryThread.processQuery();
         assertEquals(queryResultObj.getResponseBody(), "ResponseBody");
         assertEquals(queryResultObj.getHttpStatus(), 200);
     }
+
+    @Test
+    public void testConvertJsonToCSV() {
+
+        String csvStr = "[/key][\"value\"]";
```
moizarafat, line 84: when converted to String, should the value be after a new line?

moizarafat: Also we can use `.headerSeparator("_")` to remove `/` from beginning of key

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -48,7 +52,10 @@ protected void deleteAsyncQuery() {
 
         String filterExpression = "createdOn=le='" + cleanupDateFormatted + "'";
 
-        asyncQueryDao.deleteAsyncQueryAndResultCollection(filterExpression);
+        Collection<AsyncQuery> asyncQueryList = asyncQueryDao.deleteAsyncQueryAndResultCollection(filterExpression);
+        Iterator itr = asyncQueryList.iterator();
```
moizarafat, line 56: this is not needed.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Override
+    public URL storeResults(String asyncQueryID, byte[] byteResponse) {
+
+        Blob response = null;
+        AsyncQueryResultStorage aQRS = null;
+        try {
+            response = new SerialBlob(byteResponse);
+
+            Blob finalResponse = response;
```
moizarafat, line 57: this line is redundant.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -98,11 +95,37 @@ public AsyncCleanerService buildAsyncCleanerService(Elide elide, ElideConfigProp
...
+    public ResultStorageEngine buildResultStorageEngine(Elide elide, ElideConfigProperties settings) {
+        // Creating a new ElideSettings and Elide object for Async services
+        // which will have ISO8601 Dates. Used for DefaultResultStorageEngine.
+        Elide asyncElide = getAsyncElideInstance(elide);
+        return new DefaultResultStorageEngine(asyncElide, asyncElide.getDataStore(),
+                settings.getAsync().getDownloadBaseURL());
+    }
+
+    /**
+     * Creating a new ElideSettings and Elide object for Async services
```
moizarafat, line 119: remove ElideSettings

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
moizarafat: Discussed, this needs to be tested with the controller.

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+        asyncQueryDAO = new com.yahoo.elide.async.service.DefaultAsyncQueryDAO(elide, dataStore);
+        defaultResultStorageEngine = new com.yahoo.elide.async.service.DefaultResultStorageEngine(elide, dataStore, "http://localhost:8080");
+    }
+
+    @Test
+    public void testStoreResults() {
+        String responseBody = "responseBody";
+        byte[] testResponse = responseBody.getBytes();
+        Blob result = null;
```
moizarafat, line 81: I think L-81 to 87 is not needed.

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -0,0 +1,110 @@
...
+        try {
+            result = new SerialBlob(testResponse);
+        } catch (SQLException throwables) {
+            throwables.printStackTrace();
+        }
+
+        URL url = defaultResultStorageEngine.storeResults(
+                asyncQueryResultStorage.getId(), testResponse);
+
+        assertEquals(url.getClass(), URL.class);
```
moizarafat, line 92: You should actually verify the URL not just the class.

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -0,0 +1,110 @@
...
+        Blob result = null;
+
+        try {
+            result = new SerialBlob(testResponse);
+        } catch (SQLException throwables) {
+            throwables.printStackTrace();
+        }
+
+        URL url = defaultResultStorageEngine.storeResults(
+                asyncQueryResultStorage.getId(), testResponse);
```
moizarafat, line 90: shouldn't there be a when method for this? asyncQueryResultStorage is a mock object so get will return null.

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -0,0 +1,110 @@
...
+    public void testGetResultsByID() {
+        String id = "id";
+        AsyncQueryResultStorage n = new AsyncQueryResultStorage();
+        n.setId(id);
+        n.setResult(null);
+
+        Object loaded = n;
+        when(tx.loadObject(any(), any(), any())).thenReturn(loaded);
+
+        defaultResultStorageEngine.getResultsByID(n.getId());
```
moizarafat, line 105: you can skip line 97 to Line 102. and here just pass "id" as argument.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryThreadTest.java
```diff
@@ -69,10 +72,27 @@ public void testProcessQueryGraphQl() throws NoHttpResponseException, URISyntaxE
...
         assertEquals(queryResultObj.getResponseBody(), "ResponseBody");
         assertEquals(queryResultObj.getHttpStatus(), 200);
     }
+
+    @Test
+    public void testConvertJsonToCSV() throws URISyntaxException, NoHttpResponseException {
+
+        String csvStr = "[key][\"value\"]";
+        String jsonStr = "{\"key\":\"value\"}";
+        AsyncQueryThread queryThread = mock(AsyncQueryThread.class);
```
moizarafat, line 89: You can replace Line 89 to L 94 with 
```  AsyncQuery queryObj = new AsyncQuery();
        String query = "/group?sort=commonName&fields%5Bgroup%5D=commonName,description";
        String id = "edc4a871-dff2-4054-804e-d80075cf827d";
        queryObj.setId(id);
        queryObj.setQuery(query);
        queryObj.setQueryType(QueryType.JSONAPI_V1_0);

        AsyncQueryThread queryThread = new AsyncQueryThread(queryObj, user, elide, runner, asyncQueryDao, "v1",
                resultStorageEngine);```

File: elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java [hunk 47]
```diff
@@ -0,0 +1,146 @@
...
+        this.elide = elide;
+        this.dataStore = dataStore;
+        this.baseURL = baseURL;
+    }
+
+    @Override
+    public URL storeResults(String asyncQueryID, byte[] byteResponse) {
+
+        AsyncQueryResultStorage aQRS = null;
+        try {
```
moizarafat, line 53: I think you can remove the try-catch block here,

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -46,63 +54,198 @@
...
+        queryResultObj.setRecordCount(recCount);
         queryResultObj.setCompletedOn(new Date());
 
+        if (timeElapsed - asyncAfterMilliSeconds > 0) {
+
+            URL url = null;
+            if (queryObj.getResultFormatType() == ResultFormatType.CSV) {
+                String jsonToCSV = convertJsonToCSV(response.getBody());
+                byte[] temp = jsonToCSV.getBytes();
+                url = resultStorageEngine.storeResults(queryObj.getId(), temp);
```
moizarafat, line 133: This is common between if and else. This could be moved outside the if else blocks.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -46,63 +54,198 @@
...
+            }
+            else {
```
moizarafat, line 135: you can remove the else block completely. 
Before the if block. 
String tempResult = response.getBody()

In the if block
 tempResult = convertJsonToCSV(response.getBody())

After the if block, convert tempResult to byte[] and then store the results,

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -46,63 +54,198 @@
...
+        return rec;
+    }
+
+    /**
+     * This method calculates the number of records from the response with GRAPHQL API.
+     * @param response is the response.getBody() we get from the response
+     * @param table_name is the table from which we extract the data
+     * @return rec is the recordCount
+     */
+    protected Integer calculateRecordsGRAPHQL(String response, String table_name) {
```
moizarafat, line 173: GRAPHQL should be GraphQL

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public byte[] getResultsByID(String asyncQueryID) {
+        log.debug("getResultsByID");
+
+        AsyncQueryResultStorage asyncQueryResultStorage = null;
+        byte[] byteResult = null;
+
+        try {
```
moizarafat, line 93: Remove Try catch.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,146 @@
...
+
+    @Override
+    public void deleteResultsCollection(Collection<AsyncQuery> asyncQueryList) {
+        log.debug("deleteResultsCollection");
+
+        Collection<AsyncQueryResultStorage> asyncQueryResultStorage = null;
+        Iterator<AsyncQuery> itr = asyncQueryList.iterator();
+        while (itr.hasNext()) {
+            AsyncQuery query = itr.next();
+            try {
```
moizarafat, line 123: Remove try catch

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -0,0 +1,110 @@
...
+    private AsyncQuery asyncQuery;
+    private DataStoreTransaction tx;
+    private EntityDictionary dictionary;
+    private com.yahoo.elide.async.service.DefaultResultStorageEngine defaultResultStorageEngine;
+    private AsyncQueryResultStorage asyncQueryResultStorage;
+
+    @BeforeEach
+    public void setupMocks() {
+        dataStore = mock(DataStore.class);
+        asyncQuery = mock(AsyncQuery.class);
```
moizarafat, line 52: This can be removed.

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -0,0 +1,110 @@
...
+                .withJoinFilterDialect(new RSQLFilterDialect(dictionary))
+                .withSubqueryFilterDialect(new RSQLFilterDialect(dictionary))
+                .withISO8601Dates("yyyy-MM-dd'T'HH:mm'Z'", TimeZone.getTimeZone("UTC"))
+                .build();
+
+        elide = new Elide(elideSettings);
+
+        when(dataStore.beginTransaction()).thenReturn(tx);
+
+        asyncQueryDAO = new com.yahoo.elide.async.service.DefaultAsyncQueryDAO(elide, dataStore);
```
moizarafat, line 73: This can be removed.

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -0,0 +1,110 @@
...
+        n.setResult(null);
+
+        Object loaded = n;
+        when(tx.loadObject(any(), any(), any())).thenReturn(loaded);
+
+        defaultResultStorageEngine.getResultsByID(n.getId());
+
+        verify(tx, times(1)).loadObject(any(), any(), any());
+
+    }
```
moizarafat, line 109: Missing the test case for delete.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
chandrakaran: new SerialBlob() will throw an exception

moizarafat: Then when we catch any exception, we should throw it as IllegalStateException.

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -26,14 +26,9 @@
...
+import java.net.MalformedURLException;
 import java.net.URL;
-import java.sql.Blob;
-import java.sql.SQLException;
-import java.util.HashMap;
-import java.util.Map;
-import java.util.TimeZone;
-
-import javax.sql.rowset.serial.SerialBlob;
+import java.util.*;
```
moizarafat, line 31: use individual dependencies.

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -69,42 +63,43 @@ public void setupMocks() {
...
-            throwables.printStackTrace();
-        }
+        when(asyncQueryResultStorage.getId()).thenReturn("asyncQueryID");
+        URL testURL = new URL("http://localhost:8080/AsyncQueryResultStorage/asyncQueryID");
 
         URL url = defaultResultStorageEngine.storeResults(
                 asyncQueryResultStorage.getId(), testResponse);
 
+        assertEquals(url, testURL);
         assertEquals(url.getClass(), URL.class);
```
moizarafat, line 81: you can remove this assert. The URL class will be equal anyways.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -46,63 +54,195 @@
...
+     */
+    protected String convertJsonToCSV(String jsonStr) {
+        if (jsonStr == null) {
+            return null;
+        }
+        StringBuilder str = new StringBuilder();
+
+        try {
+            new JSONObject(jsonStr);
+            JFlat flatMe = new JFlat(jsonStr);
```
aklish, line 227: I'd like to see what the format of the output is for JFlat.  We may have specific requirements around how the JSON is flattened - so would also be good to see the customization options.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -46,63 +54,195 @@
...
+
+            for (Object[] obj : json2csv) {
+                str.append(Arrays.toString(obj));
+            }
+
+        } catch (JSONException e) {
+            log.error("Exception: {}", e);
+        } catch (PathNotFoundException e) {
+            log.error("Exception: {}", e);
+        } catch (Exception e) {
```
aklish, line 238: What exception are we trying to capture here?  I'm worried about swallowing NPE and other fatal errors we may want to let propagate.

elide-async/src/main/java/com/yahoo/elide/async/service/ResultStorageEngine.java
```diff
@@ -0,0 +1,39 @@
...
+     * @return it returns the URL from where we can access the result of the query
+     */
+    public URL storeResults(String asyncQueryID, byte[] byteResponse);
+
+    /**
+     * Searches for the query with ID as AsyncQueryID in the AsyncQueryResultStorage table and returns the record.
+     * @param AsyncQueryID is the query ID of the AsyncQuery
+     * @return returns the record associated with the AsyncQueryID
+     */
+    public byte[] getResultsByID(String AsyncQueryID);
```
aklish, line 32: Lowercase the 'A' in AsyncQueryID.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    @Override
+    public URL storeResults(String asyncQueryID, byte[] byteResponse) {
+
+        AsyncQueryResultStorage aQRS = null;
+        try {
+            Blob response = new SerialBlob(byteResponse);
+
+            aQRS = (AsyncQueryResultStorage) DBUtil.executeInTransaction(elide, dataStore,
```
aklish, line 56: Avoid acronyms like aQRS for variables.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,146 @@
...
+                        asyncQueryResultStorage.setResult(response);
+
+                        tx.save(asyncQueryResultStorage, scope);
+
+                        return asyncQueryResultStorage;
+                    });
+        } catch (SQLException e) {
+            log.error("Exception: {}", e);
+            throw new IllegalStateException(e);
+        } catch (Exception e) {
```
aklish, line 70: Let's not catch and swallow Exception (Too broad).

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,146 @@
...
+            log.error("Exception: {}", e);
+        }
+
+        URL url = null;
+        String buildURL = baseURL + "/AsyncQueryResultStorage/" + asyncQueryID;
```
aklish, line 75: Should 'AsyncQueryResultsStorage' be part of the baseURL?

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,146 @@
...
+
+                        Object loaded = tx.loadObject(asyncQueryCollection, asyncQueryID, scope);
+
+                        return loaded;
+                    });
+            if (asyncQueryResultStorage != null) {
+                Blob result = asyncQueryResultStorage.getResult();
+                byteResult = result.getBytes(1, (int) result.length());
+            }
+        } catch (Exception e) {
```
aklish, line 109: Too broad.

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
```diff
@@ -0,0 +1,108 @@
...
+                .withJoinFilterDialect(new RSQLFilterDialect(dictionary))
+                .withSubqueryFilterDialect(new RSQLFilterDialect(dictionary))
+                .withISO8601Dates("yyyy-MM-dd'T'HH:mm'Z'", TimeZone.getTimeZone("UTC"))
+                .build();
+
+        elide = new Elide(elideSettings);
+
+        when(dataStore.beginTransaction()).thenReturn(tx);
+        asyncQueryDAO = new DefaultAsyncQueryDAO(elide, dataStore);
+        defaultResultStorageEngine = new com.yahoo.elide.async.service.DefaultResultStorageEngine(elide, dataStore, "http://localhost:8080");
```
aklish, line 71: Can you import DefaultResultStorageEngine?

elide-async/src/test/java/com/yahoo/elide/async/service/DefaultResultStorageEngineTest.java
aklish: :+1:

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -17,10 +17,7 @@
 import com.yahoo.elide.async.hooks.ExecuteQueryHook;
 import com.yahoo.elide.async.hooks.UpdatePrincipalNameHook;
 import com.yahoo.elide.async.models.AsyncQuery;
-import com.yahoo.elide.async.service.AsyncCleanerService;
-import com.yahoo.elide.async.service.AsyncExecutorService;
-import com.yahoo.elide.async.service.AsyncQueryDAO;
-import com.yahoo.elide.async.service.DefaultAsyncQueryDAO;
+import com.yahoo.elide.async.service.*;
```
aklish, line 20: No asterisk imports.

elide-async/src/main/java/com/yahoo/elide/async/service/DBUtil.java
```diff
@@ -0,0 +1,54 @@
...
+public class DBUtil {
+
+    /**
+     * This method creates a transaction from the datastore, performs the DB action using
+     * a generic functional interface and closes the transaction.
+     * @param dataStore Elide datastore retrieved from Elide object
+     * @param action Functional interface to perform DB action
+     * @return Object Returns Entity Object (AsyncQueryResult or AsyncResult)
+     */
+    protected static Object executeInTransaction(Elide elide, DataStore dataStore, Transactional action) {
```
aklish, line 37: If you only need the Settings, don't pass through Elide object itself.  Change the signature (and all the related plumbing to pass in ElideSettings instead of Elide).

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
moizarafat: There are not many customization options except for changing header syntax and delimiters. https://github.com/opendevl/Json2Flat

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncExecutorServiceTest.java
```diff
@@ -85,15 +87,17 @@ public void testExecuteQueryFail() throws ExecutionException, TimeoutException,
     public void testExecuteQueryComplete() throws InterruptedException {
 
         AsyncQuery queryObj = mock(AsyncQuery.class);
+        when(resultStorageEngine.supportedResultTypes()).thenReturn(EnumSet.of(ResultType.EMBEDDED,
```
aklish, line 90: Do we have tests for the two result types: embedded and download?

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/framework/AsyncIntegrationTestApplicationResourceConfig.java
```diff
@@ -90,10 +95,11 @@ protected void configure() {
                         .build());
                 bind(elide).to(Elide.class).named("elide");
 
-                AsyncQueryDAO asyncQueryDao = new DefaultAsyncQueryDAO(elide, elide.getDataStore());
+                AsyncQueryDAO asyncQueryDao = new DefaultAsyncQueryDAO(elide.getElideSettings(), elide.getDataStore());
                 bind(asyncQueryDao).to(AsyncQueryDAO.class);
 
-                ResultStorageEngine resultStorageEngine = new DefaultResultStorageEngine(elide, elide.getDataStore(), "http://localhost:8080");
+                ResultStorageEngine resultStorageEngine = new DefaultResultStorageEngine(elide.getElideSettings(),
+                        elide.getDataStore(), "http://localhost:8080/AsyncQueryResultStorage/");
```
aklish, line 102: Hardcoding the port like this might make it not match the RestAssured configuration.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -259,7 +259,7 @@ default ResultStorageEngine getResultStorageEngine() {
      * @return a URL in String format.
      */
     default String getAsyncDownloadBaseURL() {
-        return "http://localhost:8080";
+        return "http://localhost:8080/AsyncQueryResultStorage/";
```
aklish, line 262: Localhost and Port should come from other configuration settings.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -150,93 +146,80 @@ else if (queryObj.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
...
     /**
      * This method calculates the number of records from the response with GRAPHQL API.
-     * @param response is the response.getBody() we get from the response
-     * @param table_name is the table from which we extract the data
+     * @param jsonStr is the response.getBody() we get from the response
      * @return rec is the recordCount
      */
-    protected Integer calculateRecordsGraphQL(String response, String table_name) {
-        Integer rec;
+    protected Integer calculateRecordsGraphQL(String jsonStr) {
```
aklish, line 168: All of these parsing methods need unit tests to cover all the scenarios.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -150,93 +146,80 @@ else if (queryObj.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
...
-        try {
-            JSONObject j = new JSONObject(jsonStr);
-            String s = (String) j.get("query");
-
-            Integer countBrackets = 0;
-            for (int i = 0; i < s.length(); i++) {
-                if (s.charAt(i) == '{' || s.charAt(i) == '}') {
-                    countBrackets++;
-                    if (countBrackets == 2) {
+                    jsonReader.beginObject();
```
aklish, line 183: I wonder if this would be simpler with JsonPath using a wildcard for the top node.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
```diff
@@ -103,11 +103,19 @@ public void executeQuery(AsyncQuery queryObj, User user, String apiVersion) {
         if (runner == null) {
             throw new InvalidOperationException("Invalid API Version");
         }
-        AsyncQueryThread queryWorker = new AsyncQueryThread(queryObj, user, elide, runner, asyncQueryDao,
-                apiVersion, resultStorageEngine);
-        Future<AsyncQueryResult> task = executor.submit(queryWorker);
+        Future<AsyncQueryResult> task = null;
```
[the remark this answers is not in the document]
moizarafat: Task object needs to be accessible in the Timedoutexception block.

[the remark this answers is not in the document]
moizarafat: We can probably just declare it but not initialize it with null.

elide-async/pom.xml
```diff
@@ -119,9 +119,14 @@
...
-            <artifactId>gson</artifactId>
-            <version>2.8.6</version>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>2.10.4</version>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.10.4</version>
```
moizarafat, line 129: replace with ${version.jackson}

elide-async/pom.xml
```diff
@@ -119,9 +119,14 @@
             <version>1.0.3</version>
         </dependency>
         <dependency>
-            <groupId>com.google.code.gson</groupId>
-            <artifactId>gson</artifactId>
-            <version>2.8.6</version>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+            <version>2.10.4</version>
```
moizarafat, line 124: replace with replace with ${version.jackson}

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -147,14 +147,21 @@ else if (queryObj.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
...
+            JsonNode jsonNode = objectMapper.readTree(jsonStr);
+            if (!jsonNode.isObject()) {
+                return null;
+            }
+             jsonNode = jsonNode.get("data");
+            if (jsonNode.isArray()) {
+                rec = jsonNode.size();
+            }
+
+        } catch (JsonMappingException e) {
```
moizarafat, line 161: no need to catch exceptions. just let them propagate.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -167,41 +174,47 @@ protected Integer calculateRecordsJSON(String jsonStr) {
...
-
-                if (JsonToken.BEGIN_OBJECT.equals(nextToken)) {
+            while (!parser.isClosed()) {
+                com.fasterxml.jackson.core.JsonToken jsonToken = parser.nextToken();
 
-                    jsonReader.beginObject();
-                    jsonObject = jsonTree.getAsJsonObject();
+                if (JsonToken.FIELD_NAME.equals(jsonToken)) {
+                    String fieldName = parser.getCurrentName();
+                    System.out.println(fieldName);
```
moizarafat, line 190: remove system.out.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -167,41 +174,47 @@ protected Integer calculateRecordsJSON(String jsonStr) {
...
+                    if (jsonNode.isObject()) {
+                        jsonNode = jsonNode.get(fieldName);
+                        System.out.println(jsonNode);
+                    } else {
+                        return null;
+                    }
                 }
             }
+
+        } catch (JsonParseException e) {
```
moizarafat, line 208: Let the exceptions propogate.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryThreadTest.java
```diff
@@ -42,37 +40,57 @@ public void setupMocks() {
...
+
+        String csvStr = "[key][\"value\"]";
+        String jsonStr = "{\"key\":\"value\"}";
+
+        AsyncQuery queryObj = new AsyncQuery();
+        AsyncQueryThread queryThread = new AsyncQueryThread(queryObj, user, elide, runner, asyncQueryDao, "v1",
+                resultStorageEngine);
+
+        assertEquals(queryThread.convertJsonToCSV(jsonStr), csvStr);
+    }
```
moizarafat, line 95: We need additional test cases for the calculateRecordCounts method with positive and negative scenarios.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryThreadTest.java
```diff
@@ -58,14 +111,42 @@ public void testProcessQueryJsonApi() throws Exception {
...
+        assertEquals(queryResultObj.getRecordCount(), null);
         assertEquals(queryResultObj.getResponseBody(), "ResponseBody");
-        assertEquals(queryResultObj.getHttpStatus(), 200);
+        });
+
     }
 
     @Test
     public void testProcessQueryGraphQl() throws Exception {
+        Exception exception = assertThrows(JsonParseException.class, () -> {
```
moizarafat, line 123: you dont need to store the results back in  `Exception exception`

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryThreadTest.java
```diff
@@ -45,6 +51,53 @@ public void setupMocks() {
...
+        when(elide.get(anyString(), any(), any(), anyString(), any())).thenReturn(response);
+        AsyncQueryThread queryThread = new AsyncQueryThread(queryObj, user, elide, runner, asyncQueryDao, "v1",
+                resultStorageEngine);
+        queryResultObj = queryThread.processQuery();
+        assertEquals(queryResultObj.getRecordCount(), 3);
+        assertEquals(queryResultObj.getResponseBody(), responseBody);
+        assertEquals(queryResultObj.getHttpStatus(), 200);
+    }
+
+    @Test
```
moizarafat, line 76: may be include a comment saying which particular method/scenario is getting tested here.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/controllers/DownloadController.java
```diff
@@ -0,0 +1,4 @@
+package com.yahoo.elide.spring.controllers;
+
```
moizarafat, line 2: this class should not be part of this PR.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryThreadTest.java
```diff
@@ -77,8 +158,10 @@ public void testProcessQueryGraphQl() throws Exception {
         AsyncQueryThread queryThread = new AsyncQueryThread(queryObj, user, elide, runner, asyncQueryDao, "v1",
                 resultStorageEngine);
         queryResultObj = queryThread.processQuery();
-        assertEquals(queryResultObj.getResponseBody(), "ResponseBody");
+        assertEquals(queryResultObj.getRecordCount(), 3);
+        assertEquals(queryResultObj.getResponseBody(), responseBody);
         assertEquals(queryResultObj.getHttpStatus(), 200);
+
     }
```
moizarafat, line 166: Please include some negative test cases for jsontocsv convert method, failure scenarios

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -95,14 +117,78 @@ else if (queryObj.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
...
+     * @return rec is the recordCount
+     */
+    protected Integer calculateRecordsGraphQL(String jsonStr) throws IOException {
+        Integer rec = null;
+
+        Configuration conf = Configuration.builder()
+                .options(Option.AS_PATH_LIST).build();
+
+        List<String> pathList = using(conf).parse(jsonStr).read("$..edges");
+        String modify = pathList.get(0).replace("[", ".");
```
moizarafat, line 164: L-164 to 166 can be changed as
`pathList.get(0).replace("[", ".").replace("'", "").replace("]", "") + ".length()"`

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -79,14 +84,12 @@ public AsyncQueryThread(AsyncQuery queryObj, User user, Elide elide, QueryRunner
      * @throws URISyntaxException
      * @throws NoHttpResponseException
      */
-    protected AsyncQueryResult processQuery() throws URISyntaxException, NoHttpResponseException {
+    protected AsyncQueryResult processQuery() throws Exception {
```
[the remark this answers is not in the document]
moizarafat: We are getting that exception from JFlat

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @Override
+    public URL storeResults(String asyncQueryID, byte[] byteResponse, String downloadBaseURL) {
+
+        try {
+            Blob response = new SerialBlob(byteResponse);
+
+            DBUtil.executeInTransaction(elideSettings,
+                    dataStore, (tx, scope) -> {
```
aklish, line 56: This may not be safe because we are asking an Elide transaction to manipulate a model that is not in the dictionary.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -46,50 +53,66 @@
...
-    public AsyncQueryResult call() throws NoHttpResponseException, URISyntaxException {
-         return processQuery();
+    public AsyncQueryResult call() throws URISyntaxException, IOException {
+        return processQuery();
     }
 
     public AsyncQueryThread(AsyncQuery queryObj, User user, Elide elide, QueryRunner runner,
-            AsyncQueryDAO asyncQueryDao, String apiVersion) {
+            AsyncQueryDAO asyncQueryDao, String apiVersion, ResultStorageEngine resultStorageEngine,
+            String baseURLEndPoint, String downloadBasePath) {
```
aklish, line 67: These need different names (and maybe helpful comments to distinguish them).  I suggested requestURL and downloadPath or something along those lines.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultResultStorageEngine.java
```diff
@@ -0,0 +1,148 @@
...
+    public DefaultResultStorageEngine() {
+    }
+
+    public DefaultResultStorageEngine(ElideSettings elideSettings, DataStore dataStore) {
+        this.elideSettings = elideSettings;
+        this.dataStore = dataStore;
+    }
+
+    @Override
+    public URL storeResults(String asyncQueryID, byte[] byteResponse, String downloadBaseURL) {
```
aklish, line 50: I think the contract for this function is odd.

The downloadBaseURL is probably a constructor argument.  Also, if this were S3, would we have a method that does two things:
1. Store results.
2. Return the URL.

I feel like mixing concerns in a single method may not be the right abstraction.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -39,4 +39,14 @@
...
 
+    /**
+     * Whether or not the controller is enabled.
+     */
+    private boolean enabled = false;
+
+    /**
+     * The URL path prefix for the controller.
+     */
+    private String downloadPath = "/";
```
aklish, line 50: We already have a way to configure controllers.  Should this be a subclass of the a controller?

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -253,6 +254,24 @@ default AsyncQueryDAO getAsyncQueryDAO() {
         return null;
     }
 
+    /**
+     * Implementation of ResultStorageEngine to use.
+     *
+     * @return ResultStorageEngine type object.
+     */
+    default ResultStorageEngine getResultStorageEngine() {
```
aklish, line 262: Maybe all of these individual async settings shoudl be consolidated into a configuration object that is returned by a single setting function.

PR #1406, opened 2020-06-26
Added Query Logger Implementation

dvarshney1:
## Description
This pull request added a basic query logger implementation to the Hibernate3-5, JPA datastore implementation for elide. In order to accomplish this, the following files were added to the package com.yahoo.elide.core:
 1. QueryDetail.java -> A public java class which contains information about the data model name and the underlying querytext that is being executed by the datastore
 2. QueryLogger.java -> An interface file that contains methods acceptQuery, processQuery, cancelQuery, completeQuery
 3. NoopQueryLogger.java -> The default implementation for the QueryLogger. Does Nothing, returns directly
 4. Slf4jQueryLogger.java -> The second implementation for the QueryLogger. Logs the QueryDetail along with necessary information such as the query execution time, the requesrID for the query etc. to the screen.

Major changes were also made to the following files to incorporate the change.
1. Elide.java -> the queryLogger accept and completeQuery function are called here.
2. PersistenceResource.java -> the queryLogger processQuery function is called here. (specifically in 3 places where the ltx.loadObject, tx.loadObjects and tx.getRelation) functions are called
3. DataStoreTransaction.java -> 2 new methods, public QueryDetail explain(EntityProjection projection, RequestScope scope) and public QueryDetail explain(Relationship relationship, RequestScope scope, Object entity) were added to surface the query back up to be logged by the QueryLogger
4. hibernate3/HibernateTransaction.java -> For the hibernate3 underlying datastore, a new member variable of type List <String> was added. Whenever elide used to call the build() method of AbstractHQLBuilder.java class to create a SQL statement, the returned queryString was added to this List.
5. hibernate5/HibernateTransaction.java & AbstractJPATransaction.class class. Both these classes had similar changes and for a similar reason to the hibernate3 changes mentioned above.
6. QueryRunner.java -> For handling GraphQL request and logging them, the executeGraphQLRequest function was modified inside this class. Here, the acceptQuery was called before the GraphQL request was executed and the completeQuery was called after the ElideResponse was built in the same functionscope (executeGraphQLRequest)

## Motivation and Context
This change will help in queryLogging and surface and log the underline HQL, JPQL, SQL queryText. Moreover, it can also be used in other  projects which use Elide. This changes can help users know the amount of time their query is taking in executing, the reponsecode they are getting etc.

## How Has This Been Tested?
Regression Testing was done to see if all the test cases till passed after adding these test cases. For each of the DataStoreIT test in hibernate3, hibernate5, the corresponding QueryText was logged out and it was checked whether the underlying hibernate SQL statement matches the HQL statement. The following changes were also tested on a demo application that uses the JPA datastore. The result of a simple GET request is attached in the screenshot below:

Screen Shot 2020-06-26 at 9 58 56 AM

## License


elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -57,8 +58,7 @@
 
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
moizarafat, line 61: instead of * , better to list individual dependencies.

elide-core/src/main/java/com/yahoo/elide/core/Slf4jQueryLogger.java
```diff
@@ -0,0 +1,95 @@
...
+    public void processQuery(UUID queryId, QueryDetail qd) {
+        if (!qd.toString().equals("")) {
+            log.info("Processing Query with RequestId: " + queryId + ". QueryText: " + qd.toString());
+        }
+    }
+
+    @Override
+    public void cancelQuery(UUID queryId) {
+        synchronized (this) {
+            startTime.remove(queryId);
```
aklish, line 64: YOu can remove the syncrhonized block.

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -309,6 +348,12 @@ protected ElideResponse handleRequest(boolean isReadOnly, User user,
                 requestScope.getPermissionExecutor().printCheckStats();
             }
 
+            //call complete query here
+            if (requestScope.getElideSettings() != null) {
+                QueryLogger ql = requestScope.getElideSettings().getQueryLogger();
+                ql.completeQuery(UUID.fromString(requestScope.getRequestId()), response);
```
aklish, line 354: This probably needs to be in a finally block to avoid leaks.

elide-core/src/main/java/com/yahoo/elide/core/Slf4jQueryLogger.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Override
+    public void acceptQuery(UUID queryId, Principal user, Map<String, String> headers, String apiVer, String apiQuery) {
+        synchronized (this) {
+            long start = System.currentTimeMillis();
+            startTime.put(queryId, start);
+        }
+
+        /**
```
aklish, line 35: Either keep the code or remove the comment.

elide-core/src/main/java/com/yahoo/elide/core/Slf4jQueryLogger.java
```diff
@@ -0,0 +1,95 @@
...
+    private volatile Map<UUID, Long> startTime;
+
+    public Slf4jQueryLogger() {
+        startTime = new ConcurrentHashMap<>();
+    }
+
+    @Override
+    public void acceptQuery(UUID queryId, Principal user, Map<String, String> headers, String apiVer, String apiQuery) {
+        synchronized (this) {
+            long start = System.currentTimeMillis();
```
aklish, line 31: Rather than use a map, we might just log the start and end times.  The map could leak memory over time if we don't clean this up correctly.

elide-core/src/main/java/com/yahoo/elide/core/Slf4jQueryLogger.java
```diff
@@ -0,0 +1,95 @@
...
+        synchronized (this) {
+            start = startTime.get(queryId);
+            startTime.remove(queryId);
+        }
+
+        long timeElapsed = endTime - start; //returns elapsed time in milliseconds
+        int responseCode = response.getResponseCode();
+
+        //log the result
+        log.info("Completed Query with RequestId: "
```
aklish, line 88: Make all infos debugs.

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/SubCollectionFetchQueryBuilder.java
```diff
@@ -78,37 +78,43 @@ public Query build() {
...
-        ));
+        }).orElse(getQuery(childAlias, parentName, parentAlias, relationshipName));
 
         query.setParameter(parentAlias, relationship.getParent());
 
         addPaginationToQuery(query);
         return query;
     }
+
+    private Query getQuery(String childAlias, String parentName, String parentAlias, String relationshipName) {
```
aklish, line 106: Why break this out as a separate functioN?

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -36,13 +37,12 @@
...
 
+import lombok.Getter;
 import lombok.extern.slf4j.Slf4j;
 
 import java.io.IOException;
 import java.io.Serializable;
-import java.util.Collection;
-import java.util.LinkedHashSet;
-import java.util.Optional;
+import java.util.*;
```
aklish, line 45: revert

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -299,4 +325,22 @@ public Integer getQueryLimit() {
     public void cancel() {
         session.cancelQuery();
     }
+
+    @Override
+    public QueryDetail explain(EntityProjection projection, RequestScope scope) {
+        Class<?> entityClass = projection.getType();
+        String modelName = entityClass.getName();
+        List<String> queryTextList = getQueryText();
```
aklish, line 333: I don't think this is correct.  The explain interface wants to explain _only_ this entity projection.  The getQueryText function is returning everything to date that has happened with this projection.

elide-datastore/elide-datastore-hibernate3/src/main/java/com/yahoo/elide/datastores/hibernate3/HibernateTransaction.java
```diff
@@ -55,6 +55,10 @@
     private final SessionWrapper sessionWrapper;
     private final LinkedHashSet<Runnable> deferredTasks = new LinkedHashSet<>();
     private final boolean isScrollEnabled;
+
+    @Getter
+    private List<String> queryText;
```
aklish, line 60: I don't think we can store state like this.  See comment below.

elide-datastore/elide-datastore-hibernate3/src/test/resources/logback.xml
```diff
@@ -0,0 +1,36 @@
...
+    </appender>
+    <appender name="ASYNC" class="ch.qos.logback.classic.AsyncAppender">
+        <appender-ref ref="FILE" />
+        <neverBlock>true</neverBlock>
+    </appender>
+
+    <logger name="com.yahoo.elide" level="TRACE" />
+
+    <root level="INFO">
+        <appender-ref ref="STDOUT" />
```
aklish, line 32: We don't want the build to output more log lines to the console.  Travis enforces limits (which we have hit in the past).

aklish: Do we really need this file?  Can we do without?

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -55,6 +55,10 @@
     private final SessionWrapper sessionWrapper;
     private final LinkedHashSet<Runnable> deferredTasks = new LinkedHashSet<>();
     private final boolean isScrollEnabled;
+
+    @Getter
+    private List<String> queryText;
```
aklish, line 60: See comments on hibernate3

elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/transaction/AbstractJpaTransaction.java
```diff
@@ -297,16 +312,36 @@ public Object getRelation(
...
+    public QueryDetail explain(EntityProjection projection, RequestScope scope) {
+        Class<?> entityClass = projection.getType();
+        String modelName = entityClass.getName();
+        List<String> queryTextList = getQueryText();
+        this.queryText = new ArrayList<>();
+        return new QueryDetail(modelName, queryTextList);
+    }
+
+    @Override
+    public QueryDetail explain(Relationship relationship, RequestScope scope, Object entity) {
```
aklish, line 340: See comments on hibernate3

elide-datastore/elide-datastore-jpa/src/test/resources/logback.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2015, Yahoo Inc.
+  ~ Licensed under the Apache License, Version 2.0
+  ~ See LICENSE file in project root for terms.
+  -->
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
```
aklish, line 8: See comments on hibernate 3.

elide-datastore/elide-datastore-search/src/test/resources/logback.xml
```diff
@@ -0,0 +1,36 @@
...
+  ~ See LICENSE file in project root for terms.
+  -->
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <filter
+                class="ch.qos.logback.classic.filter.ThresholdFilter">
+            <level>INFO</level>
+        </filter>
+        <encoder>
+            <pattern>%msg%n</pattern>
```
aklish, line 14: See comments on hibernate3

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -172,6 +172,45 @@ public ElideResponse get(String path, MultivaluedMap<String, String> queryParams
             requestScope.setEntityProjection(new EntityProjectionMaker(elideSettings.getDictionary(),
                     requestScope).parsePath(path));
             BaseVisitor visitor = new GetVisitor(requestScope);
+
+            //form the ApiQuery
+            Optional<MultivaluedMap<String, String>> queryParamsOptional = requestScope.getQueryParams();
```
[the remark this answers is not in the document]
rmihalko: We had talked about some complications with getting query text out of insert/update/delete operations, do we want the query logger to skip the `QueryLogger.processQuery()` function but still call `QueryLogger.acceptQuery()` and `QueryLogger.completeQuery()` for those operations?

elide-core/src/main/java/com/yahoo/elide/core/Slf4jQueryLogger.java
rmihalko: Just to be clear, do we want all QueryLogger messages in general to be set as DEBUG?

elide-core/src/main/java/com/yahoo/elide/ElideSettings.java
```diff
@@ -38,4 +39,5 @@
     @Getter private final int defaultPageSize;
     @Getter private final int updateStatusCode;
     @Getter private final Map<Class, Serde> serdes;
+    @Getter private QueryLogger queryLogger;
```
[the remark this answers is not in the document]
dvarshney1: Yep yep. Missed that before. Can surely be final!!

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1871,4 +1875,11 @@ private static void assignId(PersistentResource persistentResource, String id) {
     private static <T> T firstOrNullIfEmpty(final Collection<T> coll) {
         return CollectionUtils.isEmpty(coll) ? null : IterableUtils.first(coll);
     }
+
+    private static void queryLoggerProcessQuery(QueryDetail qd, RequestScope requestScope) {
+        if (requestScope.getElideSettings() != null) {
+            QueryLogger ql = requestScope.getElideSettings().getQueryLogger();
+            ql.processQuery(UUID.fromString(requestScope.getRequestId()), qd);
```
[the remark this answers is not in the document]
dvarshney1: Should we instead pass the String only (String uuid, instead of UUID uuid) because we are using the uuid for logging only?

elide-core/src/main/java/com/yahoo/elide/Elide.java
dvarshney1: Umm sorry but, I cannot see where the leaks might come from. Can you elaborate a bit more please?

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/SubCollectionFetchQueryBuilder.java
dvarshney1: We want to create the query from the querytext (session.createQuery) and set the value of the member variable from this querytext. Cannot do both inside the .orElse statement, hence created a function that does this. Cannot use lambda expressions as Query is not a functional intreface.

elide-datastore/elide-datastore-hibernate3/src/test/resources/logback.xml
dvarshney1: We used this file for testing. If we do not include the files the log will not show up in STDOUT or the log file (target/trace.log here)

dvarshney1: We can remove logging from the console and only to the given file?

dvarshney1:
Created another PR (#1426) with basic modifications.

PR #1407, opened 2020-06-30
Carbon 213 - Add Category to table and columns

ayeswarya:
Implements a feature in https://github.com/yahoo/elide/issues/1178 

## Description
Add Category to Table, Column and Metric function
## Motivation and Context
For NAvi integration

## How Has This Been Tested?
Tested in Spring boot example

## License


moizarafat:
LGTM

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/MetricFunction.java
```diff
@@ -30,6 +30,8 @@
 
     private String description;
 
+    private String category;
```
aklish, line 33: WHy do MetricFunctions need categories - when the associated metric already has one?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/integration/AggregationDataStoreIntegrationTest.java
```diff
@@ -749,6 +748,24 @@ public void metaDataTest() {
                 .body("data.attributes.expression",  equalTo("player.name"))
                 .body("data.relationships.table.data.id", equalTo("playerStats"));
 
+        given()
```
aklish, line 751: This looks like a separate test.  Let's not complicate existing tests by joining them together.  Also, there are already metadata tests - let's see how this new test relates to those.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/MetricFunction.java
ayeswarya: https://github.com/yahoo/elide/issues/1005#issuecomment-541196843 i saw this in the RFC

PR #1409, opened 2020-07-01
Async change to use onReadPreSecurity Hook

moizarafat:
## Description
i) Moved Async execute query hook to onReadPreSecurity Hook.
ii) Had to tweak the logic for values updated using `@PrePersist` annotation as they were overriding the values set in onReadPreSecurity hook.
iii) A fix for potential NPE in Async Models Operation Check.

## Motivation and Context
AsyncQuery submitted through Graphql end-point with asyncAfterSeconds 10 still comes back with embeddable results as null even though it finished before 10 seconds.
Based on debugging, its happening cos of [this](https://github.com/yahoo/elide/blob/d4901c6f07e57aa179c5afd640c9c67e90a8cdaf/elide-graphql/src/main/java/com/yahoo/elide/graphql/QueryRunner.java#L190). In GraphQL, the only part of the body that is lazily returned is the ID

## How Has This Been Tested?
Existing test cases passed.
Modified Graphql tests for Async
Verified that the example code worked.

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/hooks/ExecuteQueryHook.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/hooks/ExecuteQueryHook.java
@@ -7,6 +7,7 @@ package com.yahoo.elide.async.hooks;
 
 import com.yahoo.elide.annotation.LifeCycleHookBinding;
 import com.yahoo.elide.async.models.AsyncQuery;
+import com.yahoo.elide.async.models.QueryStatus;
 import com.yahoo.elide.async.service.AsyncExecutorService;
 import com.yahoo.elide.functions.LifeCycleHook;
 import com.yahoo.elide.security.ChangeSpec;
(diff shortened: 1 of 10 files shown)


elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -69,7 +69,10 @@
 
     @PrePersist
     public void prePersistStatus() {
-        status = QueryStatus.QUEUED;
+        if (status == null) {
```
aklish, line 72: why would status ever be null?

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncBase.java
```diff
@@ -28,7 +28,14 @@
 
     @PrePersist
     public void prePersist() {
-        createdOn = updatedOn = new Date();
+
```
aklish, line 31: Let's revisit how prePersist and preUpdate work.

elide-async/src/main/java/com/yahoo/elide/async/hooks/ExecuteQueryHook.java
```diff
@@ -28,6 +29,9 @@ public ExecuteQueryHook (AsyncExecutorService asyncExecutorService) {
     @Override
     public void execute(LifeCycleHookBinding.Operation operation, AsyncQuery query,
                         RequestScope requestScope, Optional<ChangeSpec> changes) {
-        asyncExecutorService.executeQuery(query, requestScope.getUser(), requestScope.getApiVersion());
+        if (query.getStatus() == QueryStatus.QUEUED && query.getResult() == null) {
+            asyncExecutorService.executeQuery(query, requestScope.getUser(), requestScope.getApiVersion());
+            query.prePersist();
```
aklish, line 34: I feel like this doesn't belong in this hook.

PR #1411, opened 2020-07-01
add queryParams variants for post(), patch() and delete()

### LINKED ISSUE #1410: POST, PATCH, DELETE operations do not make query parameters available from request scope.
Currently query parameters are **only** put into the request scope for GET operations.

## Expected Behavior
query parameters passed in for POST,PATCH and DELETE requests should be processed and made available in the request scope if present.
```java
    scope.getLoadFilterExpression(Foobar.class) // provides value
```

## Current Behavior
only GET requests provide 
```java
    scope.getLoadFilterExpression(Foobar.class) // Optional.empty()
```

## Possible Solution
Update `Elide.java` to provide queryParam variants of the post(), patch() and delete() methods.

## Context
We have a usage scenario where we need to bulk update a large number of entities. (in the region of thousands) To accomplish this we have some "virtual" entities which each map down to an underlying set of real entities which are defined by a filter expression. In our case we map through to an elastic search expression and perform the bulk update.

Currently we have had to derive our own Elide instance which passes down the query parameters for non-GET operations, but it would be better if query parameter support was provided native in the library.

## Your Environment
* Elide version used: Elide 4.6.5
* Environment name and version (Java 1.8.0_152): spring-boot/JDK-11

gmckerrell:
Resolves #1410

## Description
See issue for details

## Motivation and Context
query parameters currently not accessible from request scope when processing POST, PATCH or DELETE operations.

## License

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,17 @@
 # Change Log
 
+## 4.6.5
+**Fixes**
+   * [view commit](https://github.com/yahoo/elide/commit/bd516473fbeeb47ca05eaf510734f06432c4280a) Disabling Legacy Filter Dialect in Swagger Documentation (#1363) 
+   * [view commit](https://github.com/yahoo/elide/commit/d3cbcc23e9c7719b1a73f627121c6bc5fcb1bf23) Bumping Hibernate Validator Version (#1377) 
+   * [view commit](https://github.com/yahoo/elide/commit/6186da248b7249bcd858a4a433bfdbc50744af60) Bump classgraph from 4.8.78 to 4.8.86 (#1373) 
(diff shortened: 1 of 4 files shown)


gmckerrell:
just realised I need to look at elide-spring controllers as well.

gmckerrell:
I'll have a look at the extra endpoints you've identified.

gmckerrell:
Having had a quick look at the graphQL side of things, I'm not convinces this is applicable. The arguments to a graphQL request are all contained in the body of the JsonDocument so there is no need to pass the HTTP Query parameters downstream.

The only use for query parameters I could find for GraphQL was in specifying a `query` value which enables the passing of the JsonDocument in the URL for GET or POST requests, rather than in the body of a POST.

PR #1412, opened 2020-07-01
Issue756

### LINKED ISSUE #756: Support Composite IDs (EmbeddedId or idClass)
Support Composite IDs (EmbeddedId or idClass)

aklish:
Resolves #756

## Description
There are three mechanisms to support composite IDs:

1. JPA annotation `@IdClass` and multiple `@Id` annotations for each component of the composite ID.
2. JPA `@EmbeddedID` annotation 
3. Create an embeddable class which holds the composite elements, create an attribute in the model with this new embedded class, and decorate the attribute with `@Id`.

Elide now supports option 3:

For example, here is a new embeddable class containing the components of the composite ID:

```java
@Embeddable
@Data
@AllArgsConstructor
@NoArgsConstructor
public class Address implements Serializable {
    private long number;
    private String street;
    private long zipCode;
}
```
Here is the composite ID being used in an Elide model:
```java
@Include(rootLevel = true)
@Data
@Entity
public class Building {
    @Id
    Address address;
}
```

JSON-API requires all IDs to be serializable to a String.  To make this work, all composite ID classes need a registered Serde class.  For example, this Serde will convert the Address class to a base64 encoded string (and vice versa):

```java
@ElideTypeConverter(type = Address.class, name = "Address")
public class AddressSerde implements Serde<String, Address> {
    private static final Pattern ADDRESS_PATTERN =
            Pattern.compile("Address\\(number=(\\d+), street=([a-zA-Z0-9 ]+), zipCode=(\\d+)\\)");

    @Override
    public Address deserialize(String val) {
        byte[] decodedBytes = Base64.getDecoder().decode(val);
        String decodedString = new String(decodedBytes);

        Matcher matcher = ADDRESS_PATTERN.matcher(decodedString);
        if (! matcher.matches()) {
            throw new InvalidValueException(decodedString);
        }
        long number = Long.valueOf(matcher.group(1));
        String street = matcher.group(2);
        long zipCode = Long.valueOf(matcher.group(3));

        Address address = new Address(number, street, zipCode);

        return address;
    }

    @Override
    public String serialize(Address val) {
        return Base64.getEncoder().encodeToString(val.toString().getBytes());
    }
}
```

## License


aklish:
Rationale for approach:
https://stackoverflow.com/questions/212350/which-annotation-should-i-use-idclass-or-embeddedid

soberich:
@aklish 
How is two different except annotating embeddable field/prop with
`@Id` vs `@EmbeddedId`??

aklish:
There is an issue to support @EmbeddedId in Elide:
https://github.com/yahoo/elide/issues/2087

If you look at the above referenced stack overflow, the use of `@Id` plus `@Embedded` may have been an approach documented in early versions of some Hibernate/JPA books as a third (preferred) approach.  However, in practice, this does not appear to jive with the JPA specification (hence the above ticket).  Once that ticket is implemented, points 2 & 3 will work in the same way.

aklish:
@soberich #2087 is merged.

PR #1415, opened 2020-07-04
[README] - Fix the standalone README Java sample

3ygun:
## Description

The old code isn't valid Java. It has a stray `)` and isn't formatted correctly. Thus fix it.

Viewing without white space should make things easier

## Motivation and Context

It bothered me when reading an otherwise very nice readme.

## How Has This Been Tested?

Doesn't have issues when pasted into IntelliJ

## License


aklish:
Good catch.  Thanks for the submission!

PR #1417, opened 2020-07-06
Add support for ISO8601 to java.time.Instant serialization/deserialization

### LINKED ISSUE #1416: java.time.Instant is not (de)serialized "out of the box"
For non-SQL based datastores, the entity model may typically include `Instant` attributes. Builtin support for the (de)serialization of this standard types is desirable.

## Expected Behavior
ELIDE should provide serialization support for these standard data types out-of-the-box

## Current Behavior
ELIDE has to be explicitly set up to coerce these standard type to/from it's ISO8601 string representation "on the wire".
```java
        CoerceUtil.register(Instant.class, new ThirdPartyInstantSerde());
```

## Possible Solution
- add a new serde for `Instant` 

## Context
Using non-SQL datastore, and using JAVA-8 date/time types in the model.

## Your Environment
- JAVA-11
- ELIDE 4.6.5

gmckerrell:
Resolves #1416

## Description
See issue

## How Has This Been Tested?
Unittest added for new serde

## License


gmckerrell:
Hi @aklish 
Sorry to bother you, but I'm very confused as to why this PR is failing to build on travis. As far as I can see the class in question exists in the source tree but ElideSettingsBuilder fails to import the class. 

I've see some reports of earlier versions of Travis (2015) having such issues because of JDK versions, but not sure this is related or not.

I'm hoping this may be a quirk that you've come across... or that I'm missing something blindingly obvious but am too close the changes to see it...

Yours slightly confused
Graeme

gmckerrell:
I'm going to close this PR, as on reflection this is not a good fit with `with8601Dates()` which takes a format string (which would be ignored for these temporal types).

gmckerrell:
just thought better support could be via `withJavaTime()` method

gmckerrell:
> Hi @aklish
> Sorry to bother you, but I'm very confused as to why this PR is failing to build on travis. As far as I can see the class in question exists in the source tree but ElideSettingsBuilder fails to import the class.
> 
> I've see some reports of earlier versions of Travis (2015) having such issues because of JDK versions, but not sure this is related or not.
> 
> I'm hoping this may be a quirk that you've come across... or that I'm missing something blindingly obvious but am too close to the changes to see it...
> 
> Yours slightly confused
> Graeme

hmm just realised I forgot to make the source code a .java file!!!

gmckerrell:
There is no need to touch the Builder class. The serde can be auto registered using the `ElideTypeConverter` annotation

gmckerrell:
build currently failing due to CVE in tomcat version. Looks like dependabot has already got a PR open to bump tomcat version.

aklish:
Every artifact in the mono repo runs a CVE security scan to see if any package in the dependency list has an outstanding security CVE with a severity greater than 7.  

The build is failing because of a spring dependency on tomcat:

```
[ERROR] One or more dependencies were identified with vulnerabilities that have a CVSS score greater than or equal to '7.0': 
[ERROR] 
[ERROR] tomcat-embed-core-9.0.35.jar: CVE-2020-11996
```

I'll patch master so the build works again.  That should unblock this PR.

aklish:
I also think Elide should have out of box support for all the java.time classes.  Changes look good to me.  

The only downside I see is that Elide has no mechanism to override a Serde registered by annotation (ElideTypeConverter).   OffsetDateTimeSerde has the same issue.

We may want to add (probably in a separate ticket) some mechanism to shade the serdes that come bundled from _com.yahoo.elide_ in case the application wants to override the behavior.

PR #1426, opened 2020-07-09
Added NoopQueryLogger Implementation

dvarshney1:
## Description
This pull request added the NoopQueryLogger implementation for Elide:

New Files that are added:
QueryDetail.java -> A public java class which contains information about the data model name and the underlying querytext that is being executed by the datastore
QueryLogger.java -> An interface file that contains methods acceptQuery, processQuery, cancelQuery, completeQuery
NoopQueryLogger.java -> The default implementation for the QueryLogger. Does Nothing, returns directly

Major changes were also made to the following files to incorporate the change:
Elide.java -> the queryLogger acceptQuery, processQuery and completeQuery function are called here.
DataStoreTransaction.java -> 1 new method, public QueryDetail explain(EntityProjection projection, RequestScope scope) was added to surface the query back up to be logged by the QueryLogger
QueryRunner.java -> For handling GraphQL request and logging them, the executeGraphQLRequest function was modified inside this class. Here, the acceptQuery and completeQuery was called before the GraphQL request was executed and the completeQuery was called after the ElideResponse was built in the same functionscope (executeGraphQLRequest)-->

## Motivation and Context
This change will help in basic queryLogging for projects that use Elide underneath. This change can help users know the amount of time their query is taking in executing, the response code they are getting etc.

## How Has This Been Tested?
Regression Testing was done to see if all the test cases till passed after adding these test cases. 
For JSON-API Endpoint test, a new file (QueryLoggerTest.java) has been added to verify whether all QueryLogger methods are called appropriately at the same amount of time and with the correct parameters.
For GraphQL, 2 new tests were added in the GraphQLEndPointTest.java file which verify the same constraints as the JSON-API QueryLoggerTest for the standard GraphQL post requests.

## License


elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -169,9 +172,14 @@ public ElideResponse get(String path, MultivaluedMap<String, String> queryParams
...
             RequestScope requestScope = new RequestScope(path, apiVersion, jsonApiDoc,
                     tx, user, queryParams, elideSettings);
-            requestScope.setEntityProjection(new EntityProjectionMaker(elideSettings.getDictionary(),
-                    requestScope).parsePath(path));
+            EntityProjection projection = new EntityProjectionMaker(elideSettings.getDictionary(),
+                    requestScope).parsePath(path);
+            requestScope.setEntityProjection(projection);
             BaseVisitor visitor = new GetVisitor(requestScope);
+
+            //accept and process the query here
```
rmihalko, line 180: This comment string does not add much value as the two functions names describe exactly this comment.

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -190,9 +198,14 @@ public ElideResponse post(String path, String jsonApiDocument, User opaqueUser,
...
             RequestScope requestScope = new RequestScope(path, apiVersion,
                     jsonApiDoc, tx, user, null, elideSettings);
-            requestScope.setEntityProjection(new EntityProjectionMaker(elideSettings.getDictionary(),
-                    requestScope).parsePath(path));
+            EntityProjection projection = new EntityProjectionMaker(elideSettings.getDictionary(),
+                    requestScope).parsePath(path);
+            requestScope.setEntityProjection(projection);
             BaseVisitor visitor = new PostVisitor(requestScope);
+
+            //accept and process the query here
```
rmihalko, line 206: This comment string does not add much value as the two functions names describe exactly this comment.

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -217,6 +230,11 @@ public ElideResponse patch(String contentType, String accept,
                 try {
                     Supplier<Pair<Integer, JsonNode>> responder =
                             JsonApiPatch.processJsonPatch(dataStore, path, jsonApiDocument, requestScope);
+                    //accept and process the query here
```
rmihalko, line 233: Similar comment about usefulness of this comment.

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -419,4 +449,41 @@ public RequestScope getRequestScope() {
...
         }
     }
+
+    private static void elideAcceptQuery(RequestScope requestScope, User opaqueUser,
+                                         Optional<MultivaluedMap<String, String>> queryParams, String path) {
+        if (requestScope.getElideSettings() != null) {
+            final QueryLogger ql = requestScope.getElideSettings().getQueryLogger();
+            if (opaqueUser == null) {
+                ql.acceptQuery(UUID.fromString(requestScope.getRequestId()),
+                        null,
```
rmihalko, line 459: Instead of defining this same block twice, do the logic check on opaqueUser above and save the user principal (or null) in a local variable.

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -184,9 +188,13 @@ public ElideResponse get(String path, MultivaluedMap<String, String> queryParams
...
             RequestScope requestScope = new RequestScope(path, apiVersion, jsonApiDoc,
                     tx, user, queryParams, requestId, elideSettings);
-            requestScope.setEntityProjection(new EntityProjectionMaker(elideSettings.getDictionary(),
-                    requestScope).parsePath(path));
+            EntityProjection projection = new EntityProjectionMaker(elideSettings.getDictionary(),
+                    requestScope).parsePath(path);
+            requestScope.setEntityProjection(projection);
             BaseVisitor visitor = new GetVisitor(requestScope);
+
+            elideAcceptQuery(requestScope, opaqueUser, requestScope.getQueryParams(), path);
```
aklish, line 196: THe order here is wrong.  we have to accept the query before we build the entity projection.  THe entity projection includes parsing code that can throw an exception/error.  Those queries wont' get logged.

Also - accepting a query could move to the handleRequest function

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -380,6 +405,7 @@ protected ElideResponse handleRequest(boolean isReadOnly, User user,
                 requestScope.getPermissionExecutor().printCheckStats();
             }
 
+            elideCompleteQuery(requestScope, response);
```
aklish, line 408: This needs to happen in the finally block.  It might make sense for accept query to actually return a closeable of some kind so that we can accept the query inside the try statement itself (in the same way we create the transaction).

We can then call process and complete on the closeable returned from accept.

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -483,4 +509,33 @@ public RequestScope getRequestScope() {
             return requestScope;
         }
     }
+
+    private static void elideAcceptQuery(RequestScope requestScope, User opaqueUser,
+                                         Optional<MultivaluedMap<String, String>> queryParams, String path) {
+        if (requestScope.getElideSettings() != null) {
```
aklish, line 515: When is getElideSEttings null?  Is this only for tests?  This really shouldn't happen.

elide-core/src/main/java/com/yahoo/elide/core/QueryLogger.java
```diff
@@ -0,0 +1,66 @@
...
+                     Optional<MultivaluedMap<String, String>> queryParams, String path);
+
+    /**
+     * Processes and logs all the queries from QueryDetail to STDOUT and an external file (target/trace.log)
+     * @param queryId The RequestScope requestId.
+     * @param projection The Entity Projection of the current request
+     * @param scope The RequestScope of the current request
+     * @param tx DataStore Transaction for the underlying datastore
+     */
+    void processQuery(UUID queryId, EntityProjection projection, RequestScope scope, DataStoreTransaction tx);
```
aklish, line 52: Can we change this back to take QueryDetails instead of projections?  Also, the comment doesn't match.

elide-core/src/main/java/com/yahoo/elide/core/QueryLogger.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param headers Http Request Headers
+     * @param apiVer API Version
+     * @param queryParams QueryParams for the incoming JSON API query
+     * @param path The apiQuery endpoint path for the incoming query
+     */
+    void acceptQuery(UUID queryId, Principal user, Map<String, String> headers, String apiVer,
+                     Optional<MultivaluedMap<String, String>> queryParams, String path);
+
+    /**
+     * Processes and logs all the queries from QueryDetail to STDOUT and an external file (target/trace.log)
```
aklish, line 46: This is an interface so remove implementation description like STDOUT (which doesn't happen in some implementations) and trace.log.

elide-core/src/test/java/com/yahoo/elide/core/QueryLoggerTest.java
```diff
@@ -0,0 +1,140 @@
...
+        DataStore store = mock(DataStore.class);
+        DataStoreTransaction tx = mock(DataStoreTransaction.class);
+        QueryLogger ql = mock(QueryLogger.class);
+        FieldTestModel mockModel = mock(FieldTestModel.class);
+
+        Elide elide = getElide(store, dictionary, ql);
+
+        String body = "{\"data\": {\"type\":\"testModel\",\"id\":\"1\",\"attributes\": {\"field\":\"Foo\"}}}";
+
+        dictionary.setValue(mockModel, "id", "1");
```
aklish, line 129: Why are we calling setValue on a mock?

elide-graphql/src/main/java/com/yahoo/elide/graphql/QueryRunner.java
```diff
@@ -194,8 +196,9 @@ private ElideResponse executeGraphQLRequest(ObjectMapper mapper, User principal,
             if (jsonDocument.has(OPERATION_NAME) && !jsonDocument.get(OPERATION_NAME).isNull()) {
                 executionInput.operationName(jsonDocument.get(OPERATION_NAME).asText());
             }
-
             executionInput.variables(variables);
+            queryRunnerAcceptQuery(requestScope, principal, query);
```
aklish, line 200: Same comment I had earlier - this accept should move into the try statement itself (which means we need an interface change there).

elide-core/src/main/java/com/yahoo/elide/Elide.java
```diff
@@ -417,6 +440,7 @@ protected ElideResponse handleRequest(boolean isReadOnly, User user,
             log.error("Error or exception uncaught by Elide", e);
             throw e;
         } finally {
+            elideCompleteQuery(requestScope, response);
```
aklish, line 443: So the response is often set inside the exception handling logic.  For all exceptions, the response here will be null.

elide-core/src/main/java/com/yahoo/elide/core/QueryDetail.java
```diff
@@ -0,0 +1,40 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * QueryDetail implementation class. Stores the model name (root entity class) and
+ * the queryText that the underlying datastore will eventually run
+ */
+@Value
+public class QueryDetail {
```
aklish, line 18: You can add a lombok builder interface alternatively instead of defining all of the constructors.

elide-core/src/main/java/com/yahoo/elide/Elide.java
dvarshney1: Fixed in most recent commit

elide-core/src/main/java/com/yahoo/elide/Elide.java
dvarshney1: Fixed in most recent commit

elide-core/src/main/java/com/yahoo/elide/core/QueryDetail.java
dvarshney1: Fixed in most recent commit

dvarshney1:
Closing this PR in favor of the updated PR (#1448)

PR #1431, opened 2020-07-13
Updated ThreadLocal usage to recommended implementation.

gmckerrell:
## Description
The current `AuditLogger` current uses a `ThreadLocal<>` generic as a data member.

Given that typically there is only a single instance of an `AuditLogger`, this shouldn't cause a resource leak, however the documentation for [`ThreadLocal`](https://docs.oracle.com/javase/8/docs/api/java/lang/ThreadLocal.html) makes a strong suggestion that 
> ThreadLocal instances are typically private static fields in classes that wish to associate state with a thread (e.g., a user ID or Transaction ID).

This PR simply updates the implementation to make this a static field, as per the recommendation.

## Motivation and Context
As part of the static analysis of one of our projects, that uses the Elide library, we have a possible "singleton race condition" which has been raised via bytecode analysis, which suggests our usage of  the `elide.get()` method is somehow mutating the `elide` instance in a "race-like manner". Whilst looking at the code flow I noticed that the class instance does mutate the audit logger but it _should_ be safe because of its use of `ThreadLocal`.

Whilst reading up on this I noticed that the current usage is not "best practice" hence this PR.

## How Has This Been Tested?
Any regression should be identified by current `audit` unittests.

## License


gmckerrell:
hmm, not sure why all the commits are being listed above :-( I suspect I'm getting in a muddle trying to rebase/merge into my fork... (don't tend to use forks much in my day-job)

PR #1432, opened 2020-07-13
Raise IllegalArgumentException when DateTimeFormatter has a parsing failure.

gmckerrell:
## Description
Avoid "500 Internal Server Error" responses by handling DateTimeParseException.

### Current JSON-API behaviour
Current OffsetDateTime and Instant (de)serializers can raise the JAVA standard `DateTimeParseException` if the value passed in is of an incorrect format.

This can cause a 500 Internal Server error to occur when invalid values are passed in via a REST API.
```json
500 Internal Server Error
{
    "timestamp": "2020-07-13T11:36:07.924+00:00",
    "status": 500,
    "error": "Internal Server Error",
    "message": "Text '2020-07-10T11:19:52+0200' could not be parsed, unparsed text found at index 22",
    "path": "/pwStats"
}
```
### Updated JSON-API response
We convert the exception into an `IllegalArgumentException` which the CoreceUtil class handles.
```json
400 Bad Request
{
    "errors": [
        {
            "detail": "InvalidValueException: Invalid value: 2020-07-10T11:19:52+0200"
        }
    ]
}
```

## Motivation and Context
Should never allow user ingress data to cause `500 Internal Server Error`

## How Has This Been Tested?
Unittests updated to ensure `IllegalArgumentException` is raised.

## License


gmckerrell:
sorry I seem to be getting my cross-fork PRs in a muddle !!!

What the best practice here?
- create separate branches in forked repo, and don't merge them to master?

PR #1442, opened 2020-07-15
Using spring's dependency management

### LINKED ISSUE #1116: Use Spring Boot dependency managements versions rather than specific versions
It would be more consistent to use Spring Boot's dependency management in `elide-spring` at the very least. Right now the Spring Boot version is specified in both modules and reuses an specific Jetty version.

Spring Boot users adding the starter are more likely to use its dependency management and therefore the version that Spring Boot promotes. It would be better if tests were running against that arrangement.

One way would be to add the following to `elide-spring`

```xml
<dependencyManagement>
	<dependencies>
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-dependencies</artifactId>
			<version>${spring.boot.version}</version>
			<type>pom</type>
			<scope>import</scope>
		</dependency>
	</dependencies>
</dependencyManagement>
```

alongside

```
<properties>
	...
	<spring.boot.version>2.2.2.RELEASE</spring.boot.version>
</properties>
```

and then remove all the versions that Spring Boot provides.

moizarafat:
Resolves #1116

## Description
Use Spring Dependency Management

## Motivation and Context
see #1116. By using spring boot dependencies we won't have to provide versions for all the dependencies in autoconfigure and starter repo. 

## How Has This Been Tested?
Existing test case pass
Spring Example Code runs fine.

## License

### DIFF
--- a/elide-spring/elide-spring-boot-autoconfigure/pom.xml
+++ b/elide-spring/elide-spring-boot-autoconfigure/pom.xml
@@ -40,7 +40,6 @@
     </scm>
 
     <properties>
-        <spring.boot.version>2.3.1.RELEASE</spring.boot.version>
         <tomcat.version>9.0.37</tomcat.version>
         <project.build.sourceEncoding>utf-8</project.build.sourceEncoding>
         <min_jdk_version>1.8</min_jdk_version>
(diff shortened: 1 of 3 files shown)


elide-spring/elide-spring-boot-autoconfigure/pom.xml
```diff
@@ -158,8 +155,13 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-test</artifactId>
-            <version>${spring.boot.version}</version>
             <scope>test</scope>
+            <!--exclusions>
```
aklish, line 159: Remove commented out sections

PR #1444, opened 2020-07-17
Fix for Graphql Fragment Error in Debug Mode 

moizarafat:
## Description
Using Graphql fragments in debug mode gave error, `graphql.language.FragmentSpread cannot be cast to graphql.language.Field`. The error was caused by assuming the type is always Field during the getName operation. Put in a cast by checking the class before calling getName. 

## Motivation and Context
Resolving Error 

## How Has This Been Tested?
Tested with the app that was failing.

## License

### DIFF
--- a/elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
+++ b/elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
@@ -22,6 +22,7 @@ import com.yahoo.elide.request.Relationship;
 import com.google.common.collect.Sets;
 
 import graphql.language.Field;
+import graphql.language.FragmentSpread;
 import graphql.schema.DataFetcher;
 import graphql.schema.DataFetchingEnvironment;
 import graphql.schema.GraphQLType;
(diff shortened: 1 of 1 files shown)


elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -121,11 +122,22 @@ private void filterSortPaginateSanityCheck(Environment environment) {
...
+        if (children.size() > 0) {
+            children.stream().forEach(i -> { if (i.getClass().equals(Field.class)) {
+                    fieldName.add(((Field) i).getName());
+                } else {
+                    fieldName.add(((FragmentSpread) i).getName());
+                }
+            });
+        }
+
+        String requestedFields = environment.field.getName() + fieldName.toString();
```
aklish, line 139: What does this look like?  There is no space - I'd like to see an example log line.

elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
```diff
@@ -121,11 +122,22 @@ private void filterSortPaginateSanityCheck(Environment environment) {
...
                 : new ArrayList<>();
-        String requestedFields = environment.field.getName() + (children.size() > 0
-                ? "(" + children.stream().map(Field::getName).collect(Collectors.toList()) + ")" : "");
+
+        List<String> fieldName = new ArrayList<String>();
+        if (children.size() > 0) {
+            children.stream().forEach(i -> { if (i.getClass().equals(Field.class)) {
+                    fieldName.add(((Field) i).getName());
+                } else {
+                    fieldName.add(((FragmentSpread) i).getName());
```
aklish, line 134: This might not be safe (there are other types in graphQL besides fields and fragments).  Let's not bother logging the fragment spreads.

elide-graphql/src/main/java/com/yahoo/elide/graphql/PersistentResourceFetcher.java
moizarafat: ` DEBUG c.y.e.g.PersistentResourceFetcher.logContext - FETCH table[edges] fields with parent GraphQLObjectType<Query>`
`FETCH node[TableFragment] fields with parent GraphQLObjectType<TableEdge>`
`FETCH id[] fields with parent GraphQLObjectType<Table>`

PR #1447, opened 2020-07-17
Cancel long running or cancelled async queries

ayeswarya:

## Description
Cancel long running or cancelled async queries 

## How Has This Been Tested?
Tested in the Example Application

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
@@ -50,7 +50,7 @@ public class AsyncCleanerService {
         //Or maxRunTimeMinutes * 2 so that this process does not coincides with query
         //interrupt process.
 
-        cleaner.scheduleWithFixedDelay(cleanUpTask, 0, Math.max(defaultCleanupDelayMinutes,
+        cleaner.scheduleWithFixedDelay(cleanUpTask, initialDelayMinutes, Math.max(defaultCleanupDelayMinutes,
                 queryRunTimeThresholdMinutes), TimeUnit.MINUTES);
 
         //Setting up query cancel service that cancels long running queries based on status or runtime
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
@@ -50,7 +50,7 @@ public class AsyncCleanerService {
         //Or maxRunTimeMinutes * 2 so that this process does not coincides with query
         //interrupt process.
 
-        cleaner.scheduleWithFixedDelay(cleanUpTask, initialDelayMinutes, Math.max(defaultCleanupDelayMinutes,
+        cleaner.scheduleWithFixedDelay(cleanUpTask, 0, Math.max(defaultCleanupDelayMinutes,
                 queryRunTimeThresholdMinutes), TimeUnit.MINUTES);
 
         //Setting up query cancel service that cancels long running queries based on status or runtime
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
@@ -38,7 +38,7 @@ public class AsyncCleanerService {
         // Setting up query cleaner that marks long running query as TIMEDOUT.
         ScheduledExecutorService cleaner = Executors.newSingleThreadScheduledExecutor();
         AsyncQueryCleanerThread cleanUpTask = new AsyncQueryCleanerThread(queryRunTimeThresholdMinutes, elide,
-            queryCleanupDays, asyncQueryDao);
+            queryCleanupDays, asyncQueryDao, new DateUtil());
 
         // Since there will be multiple hosts running the elide service,
         // setting up random delays to avoid all of them trying to cleanup at the same time.
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
@@ -25,16 +25,15 @@ import javax.inject.Inject;
 public class AsyncCleanerService {
 
     private final int defaultCleanupDelayMinutes = 360;
-    private final int maxCleanupInitialDelayMinutes = 100;
-
+    private final int maxInitialDelayMinutes = 100;
     private static AsyncCleanerService asyncCleanerService = null;
 
     @Inject
-    private AsyncCleanerService(Elide elide, Integer maxRunTimeMinutes, Integer queryCleanupDays,
-            AsyncQueryDAO asyncQueryDao) {
+    private AsyncCleanerService(Elide elide, Integer maxRunTimeSeconds, Integer queryCleanupDays,
+            Integer cancelDelaySeconds, AsyncQueryDAO asyncQueryDao) {
 
         //If query is still running for twice than maxRunTime, then interrupt did not work due to host/app crash.
-        int queryRunTimeThresholdMinutes = maxRunTimeMinutes * 2;
+        int queryRunTimeThresholdMinutes = Math.round((maxRunTimeSeconds * 2) / 60);
 
         // Setting up query cleaner that marks long running query as TIMEDOUT.
         ScheduledExecutorService cleaner = Executors.newSingleThreadScheduledExecutor();
@@ -44,28 +43,38 @@ public class AsyncCleanerService {
         // Since there will be multiple hosts running the elide service,
         // setting up random delays to avoid all of them trying to cleanup at the same time.
         Random random = new Random();
-        int initialDelayMinutes = random.ints(0, maxCleanupInitialDelayMinutes).limit(1).findFirst().getAsInt();
+        int initialDelayMinutes = random.ints(0, maxInitialDelayMinutes).limit(1).findFirst().getAsInt();
         log.debug("Initial Delay for cleaner service is {}", initialDelayMinutes);
 
         //Having a delay of at least DEFAULT_CLEANUP_DELAY between two cleanup attempts.
         //Or maxRunTimeMinutes * 2 so that this process does not coincides with query
         //interrupt process.
+
         cleaner.scheduleWithFixedDelay(cleanUpTask, initialDelayMinutes, Math.max(defaultCleanupDelayMinutes,
                 queryRunTimeThresholdMinutes), TimeUnit.MINUTES);
+
+        //Setting up query cancel service that cancels long running queries based on status or runtime
+        ScheduledExecutorService cancellation = Executors.newSingleThreadScheduledExecutor();
+
+        AsyncQueryCancelThread cancelTask = new AsyncQueryCancelThread(maxRunTimeSeconds, elide, asyncQueryDao);
+
+        cancellation.scheduleWithFixedDelay(cancelTask, 0, cancelDelaySeconds, TimeUnit.SECONDS);
     }
 
     /**
      * Initialize the singleton AsyncCleanerService object.
      * If already initialized earlier, no new object is created.
      * @param elide Elide Instance
-     * @param maxRunTimeMinutes max run times in minutes
+     * @param maxRunTimeSeconds max run times in seconds
      * @param queryCleanupDays Async Query Clean up days
+     * @param cancelDelaySeconds Async Query Transaction cancel delay
      * @param asyncQueryDao DAO Object
      */
-    public static void init(Elide elide, Integer maxRunTimeMinutes, Integer queryCleanupDays,
-            AsyncQueryDAO asyncQueryDao) {
+    public static void init(Elide elide, Integer maxRunTimeSeconds, Integer queryCleanupDays,
+            Integer cancelDelaySeconds, AsyncQueryDAO asyncQueryDao) {
         if (asyncCleanerService == null) {
-            asyncCleanerService = new AsyncCleanerService(elide, maxRunTimeMinutes, queryCleanupDays, asyncQueryDao);
+            asyncCleanerService = new AsyncCleanerService(elide, maxRunTimeSeconds, queryCleanupDays,
+                    cancelDelaySeconds, asyncQueryDao);
         } else {
             log.debug("asyncCleanerService is already initialized.");
         }
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
@@ -27,7 +27,6 @@ import lombok.extern.slf4j.Slf4j;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Date;
-import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -53,7 +52,6 @@ public class AsyncQueryCancelThread implements Runnable {
 
     @Override
     public void run() {
-        System.out.println("CANCEL QUERY ***");
         cancelAsyncQuery();
     }
 
@@ -119,7 +117,6 @@ public class AsyncQueryCancelThread implements Runnable {
                 PathElement idPathElement = new PathElement(AsyncQuery.class, String.class, "id");
                 FilterExpression fltIdExpression =
                         new InPredicate(idPathElement, queryIDsToCancel);
-                System.out.println(fltIdExpression.toString());
                 asyncQueryDao.updateStatusAsyncQueryCollection(fltIdExpression, QueryStatus.CANCEL_COMPLETE);
             }
         } catch (Exception e) {
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
@@ -104,12 +104,15 @@ public class AsyncQueryCancelThread implements Runnable {
             //Cancel Transactions
             queryUUIDsToCancel.stream()
                .forEach((uuid) -> {
-                   JsonApiDocument jsonApiDoc = new JsonApiDocument();
-                   MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
-                   RequestScope scope = new RequestScope("", "query", NO_VERSION, jsonApiDoc,
-                           transactionRegistry.getRunningTransaction(uuid), null, queryParams,
-                           uuid, elide.getElideSettings());
-                   transactionRegistry.getRunningTransaction(uuid).cancel(scope);
+                   DataStoreTransaction runningTransaction = transactionRegistry.getRunningTransaction(uuid);
+                   if (runningTransaction != null) {
+                       JsonApiDocument jsonApiDoc = new JsonApiDocument();
+                       MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
+                       RequestScope scope = new RequestScope("", "query", NO_VERSION, jsonApiDoc,
+                               runningTransaction, null, queryParams,
+                               uuid, elide.getElideSettings());
+                       runningTransaction.cancel(scope);
+                   }
                });
 
             //Change queryStatus for cancelled queries
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
@@ -48,7 +48,7 @@ public class AsyncQueryCancelThread implements Runnable {
 
     private int maxRunTimeSeconds;
     private Elide elide;
-    AsyncQueryDAO asyncQueryDao;
+    private AsyncQueryDAO asyncQueryDao;
 
     @Override
     public void run() {
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
@@ -1,0 +1,129 @@
+/*
+ * Copyright 2020, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.async.service;
+
+import static com.yahoo.elide.core.EntityDictionary.NO_VERSION;
+
+import com.yahoo.elide.Elide;
+import com.yahoo.elide.async.models.AsyncQuery;
+import com.yahoo.elide.async.models.QueryStatus;
+import com.yahoo.elide.core.DataStoreTransaction;
+import com.yahoo.elide.core.Path.PathElement;
+import com.yahoo.elide.core.RequestScope;
+import com.yahoo.elide.core.TransactionRegistry;
+import com.yahoo.elide.core.filter.InPredicate;
+import com.yahoo.elide.core.filter.expression.FilterExpression;
+import com.yahoo.elide.jsonapi.models.JsonApiDocument;
+
+import com.google.common.collect.Sets;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Date;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
+
+import javax.ws.rs.core.MultivaluedHashMap;
+import javax.ws.rs.core.MultivaluedMap;
+
+/**
+ * Runnable thread for cancelling AsyncQuery transactions
+ * beyond the max run time or if it has status CANCELLED.
+ */
+@Slf4j
+@Data
+@AllArgsConstructor
+public class AsyncQueryCancelThread implements Runnable {
+
+    private int maxRunTimeSeconds;
+    private Elide elide;
+    AsyncQueryDAO asyncQueryDao;
+
+    @Override
+    public void run() {
+        System.out.println("CANCEL QUERY ***");
+        cancelAsyncQuery();
+    }
+
+    /**
+     * This method cancels queries based on threshold.
+     */
+    protected void cancelAsyncQuery() {
+
+        try {
+            TransactionRegistry transactionRegistry = elide.getTransactionRegistry();
+
+            Map<UUID, DataStoreTransaction> runningTransactionMap = transactionRegistry.getRunningTransactions();
+
+            //Running transaction UUIDs
+            Set<UUID> runningTransactionUUIDs = runningTransactionMap.keySet();
+
+            //Construct filter expression
+            PathElement statusPathElement = new PathElement(AsyncQuery.class, QueryStatus.class, "status");
+            List<QueryStatus> statusList = new ArrayList<QueryStatus>();
+            statusList.add(QueryStatus.CANCELLED);
+            statusList.add(QueryStatus.PROCESSING);
+            statusList.add(QueryStatus.QUEUED);
+            FilterExpression fltStatusExpression =
+                    new InPredicate(statusPathElement, statusList);
+
+            Collection<AsyncQuery> asyncQueryCollection =
+                    asyncQueryDao.loadAsyncQueryCollection(fltStatusExpression);
+
+            //Active AsyncQuery UUIDs
+            Set<UUID> asyncTransactionUUIDs = asyncQueryCollection.stream()
+                    .filter(query -> query.getStatus() == QueryStatus.CANCELLED
+                    || TimeUnit.SECONDS.convert(Math.abs(new Date(System.currentTimeMillis()).getTime()
+                            - query.getCreatedOn().getTime()), TimeUnit.MILLISECONDS) > maxRunTimeSeconds)
+                    .map(query -> UUID.fromString(query.getRequestId()))
+            .collect(Collectors.toSet());
+
+            //AsyncQuery UUIDs that have active transactions
+            Set<UUID> queryUUIDsToCancel = Sets.intersection(runningTransactionUUIDs, asyncTransactionUUIDs);
+
+            //AsyncQuery IDs that need to be cancelled
+            Set<String> queryIDsToCancel = queryUUIDsToCancel.stream()
+            .map(uuid -> {
+                return asyncQueryCollection.stream()
+                .filter(query -> query.getRequestId().equals(uuid.toString()))
+                .map(AsyncQuery::getId)
+                .findFirst().orElseThrow(IllegalStateException::new);
+            })
+            .collect(Collectors.toSet());
+
+            //Cancel Transactions
+            queryUUIDsToCancel.stream()
+               .forEach((uuid) -> {
+                   JsonApiDocument jsonApiDoc = new JsonApiDocument();
+                   MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
+                   RequestScope scope = new RequestScope("", "query", NO_VERSION, jsonApiDoc,
+                           transactionRegistry.getRunningTransaction(uuid), null, queryParams,
+                           uuid, elide.getElideSettings());
+                   transactionRegistry.getRunningTransaction(uuid).cancel(scope);
+               });
+
+            //Change queryStatus for cancelled queries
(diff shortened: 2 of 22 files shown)


elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -195,4 +199,32 @@ protected Object executeInTransaction(DataStore dataStore, Transactional action)
...
         return result;
     }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Collection<AsyncQuery> getActiveAsyncQueryCollection() {
+        Collection<AsyncQuery> asyncQueryList = null;
+
+        log.debug("getActiveAsyncQueryCollection");
+        String filterExpression = "status=in=(" + QueryStatus.CANCELLED.toString() + ","
```
moizarafat, line 209: we should expand this where clause to pull records based on create_time also as after sometime CANCELLED query list will expand. Something like create_time after current-date minus maxRunTimeSeconds;

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -195,4 +199,32 @@ protected Object executeInTransaction(DataStore dataStore, Transactional action)
...
+
+                EntityProjection asyncQueryCollection = EntityProjection.builder()
+                        .type(AsyncQuery.class)
+                        .filterExpression(filter)
+                        .build();
+                Iterable<Object> loaded = tx.loadObjects(asyncQueryCollection, scope);
+                return loaded;
+            });
+        } catch (Exception e) {
+            log.error("Exception: {}", e);
```
moizarafat, line 226: throw the exception as IllegalStateException.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * This method cancels queries based on threshold.
+     */
+    protected void cancelAsyncQuery() {
+
+        try {
+            TransactionRegistry transactionRegistry = elide.getTransactionRegistry();
+            Map<UUID, DataStoreTransaction> runningTransactionMap = transactionRegistry.getTransactionMap();
```
moizarafat, line 49: we should update the getRunningTransactions method to get this and we can remove the getter method on TransactionMap as its not used anyplace else.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
moizarafat: Also the FilterExpression should be the method argument as elide will be controlling the Filter, where as users can provide their implementation of getActiveAsyncQueryCollection method.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -0,0 +1,72 @@
...
+            for (Map.Entry<UUID, DataStoreTransaction> entry : runningTransactionMap.entrySet()) {
+                for (AsyncQuery obj : asyncQueryCollection) {
+                    if (obj.getRequestId().trim().equals(entry.getKey().toString().trim())) {
+                        Date currentDate = new Date(System.currentTimeMillis());
+                        long diffInMillies = Math.abs(obj.getUpdatedOn().getTime() - currentDate.getTime());
+                        long diffInSecs = TimeUnit.SECONDS.convert(diffInMillies, TimeUnit.MILLISECONDS);
+                        if (obj.getStatus().equals(QueryStatus.CANCELLED)) {
+                            log.debug("Async Query Cancelled: " + obj.getId());
+                            entry.getValue().cancel();
+                        } else if (diffInSecs > maxRunTimeSeconds) {
```
moizarafat, line 61: we can combine these two if statements with an OR

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -0,0 +1,72 @@
...
+                            log.debug("Async Query Cancelled: " + obj.getId());
+                            entry.getValue().cancel();
+                        }
```
moizarafat, line 64: a couple of things we can probably do to improve the performance here: 
i) When the match occurs in if clause, after processing the if block we should break the inner loop
ii) we should remove the object from asyncQueryCollection when a match happens so that the size keeps reducing as the transaction loop move on.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -234,6 +234,15 @@ default Integer getAsyncQueryCleanupDays() {
         return 7;
     }
 
+    /**
+     * Number of seconds  between async query cancellation checks.
+     *
+     * @return Default: 300
+     */
+    default Integer getAsyncQueryCancelDelaySeconds() {
```
moizarafat, line 242: i think getAsyncQueryCancelIntervalSeconds sounds better as Delay could be taken as differently.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -33,6 +33,11 @@
      */
     private int queryCleanupDays = 7;
 
+    /**
+     * Default cancel interval of async query.
+     */
+    private int queryCancellationSeconds = 300;
```
moizarafat, line 39: queryCancellationIntervalSeconds

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -0,0 +1,72 @@
...
+     * This method cancels queries based on threshold.
+     */
+    protected void cancelAsyncQuery() {
+
+        try {
+            TransactionRegistry transactionRegistry = elide.getTransactionRegistry();
+            Map<UUID, DataStoreTransaction> runningTransactionMap = transactionRegistry.getTransactionMap();
+            Collection<AsyncQuery> asyncQueryCollection = asyncQueryDao.getActiveAsyncQueryCollection();
+
+            for (Map.Entry<UUID, DataStoreTransaction> entry : runningTransactionMap.entrySet()) {
```
aklish, line 52: This is a bit of pseudocode, but I think the code is more readable with functional style and breaking this into sets:

```java

Set<UUID> runningQueries = runningTransctionMap.keySet();

Set<UUID> cancellingQueries = asyncQueryCollection.stream()
     .filter(query -> query.getStatus == QueryStatus.CANCELED)
     .map (AsyncQuery::getRequestId)
      .collect(Collectors.toSet());

Set<UUID> queriesToKill = Sets.intersection(runningQueries, cancellingQueries);

queriesToKill.stream().
   .map(transactionRegistry.getRunningTransaction)
   .forEach((tx) -> {
      tx.cancel();
   });

```

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryCancelThreadTest.java
```diff
@@ -0,0 +1,64 @@
...
+        assertEquals(elide, cancelThread.getElide());
+        assertEquals(asyncQueryDao, cancelThread.getAsyncQueryDao());
+        assertEquals(7, cancelThread.getMaxRunTimeSeconds());
+    }
+
+    @Test
+    public void testCancelAsyncQuery() {
+        cancelThread.cancelAsyncQuery();
+        verify(asyncQueryDao, times(1)).getActiveAsyncQueryCollection();
+        //verify(transactionRegistry, times(1)).getTransactionMap();
```
moizarafat, line 62: plz remove commented code

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -195,4 +199,32 @@ protected Object executeInTransaction(DataStore dataStore, Transactional action)
         }
         return result;
     }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Collection<AsyncQuery> getActiveAsyncQueryCollection() {
```
moizarafat, line 205: please include a test case in DefaultAsyncQueryDAOTest.java for this method.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
ayeswarya: I want to keep it similar to cleanup. cleanup  is called queryCleanupDays. So queryCancellationSeconds sounds correct to me.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -0,0 +1,103 @@
...
+        }
+    }
+
+    /**
+     * Evaluates and subtracts the amount based on the calendar unit and amount from current date.
+     * @param calendarUnit Enum such as Calendar.SECOND
+     * @param amount Amount of days to be subtracted from current time
+     * @return formatted filter date
+     */
+     private String evaluateFormattedFilterDate(int calendarUnit, int amount) {
```
moizarafat, line 93: This method is common between AsyncQueryCancelThread and AsyncQueryCleanerThread. We can move it to a common UTIL class.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -195,4 +199,30 @@ protected Object executeInTransaction(DataStore dataStore, Transactional action)
         }
         return result;
     }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Collection<AsyncQuery> getActiveAsyncQueryCollection(String filterExpression) {
```
aklish, line 205: The interface contract for the AsyncQueryDAO would be cleaner if the filterExpression for this method and others was an actual FilterExpression object instead of a String.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncExecutorServiceTest.java
```diff
@@ -85,9 +86,10 @@ public void testExecuteQueryComplete() throws InterruptedException {
         AsyncQuery queryObj = mock(AsyncQuery.class);
         String query = "/group?sort=commonName&fields%5Bgroup%5D=commonName,description";
         String id = "edc4a871-dff2-4054-804e-d80075cf827d";
+        UUID uuid = UUID.fromString("edc4a871-dff2-4054-804e-d80075cf827d");
```
aklish, line 89: Should we change this to UUID.fromString(id)?

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
```diff
@@ -235,6 +235,15 @@ default Integer getAsyncQueryCleanupDays() {
         return 7;
     }
 
+    /**
+     * Number of seconds  between async query cancellation checks.
+     *
+     * @return Default: 300
+     */
+    default Integer getAsyncQueryCancelSeconds() {
```
aklish, line 243: I wonder if this would read better with a more descriptive name (although longer):
getAsyncQueryCancelCheckIntervalSeconds

aklish: Maybe a better approach is to move all of the async query settings into a struct of some kind to namespace them (so they aren't so long)

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
aklish, line 39: I think the definition of this is vague.  What is this interval for?  Let's get a more descriptive name and a better description of its intended use.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -114,27 +106,26 @@ public AsyncQuery updateStatus(String asyncQueryId, QueryStatus status) {
                  }
                  return loaded;
              });
-        } catch (ParseException e) {
+        } catch (Exception e) {
```
aklish, line 109: This is too broad and is swallowing exceptions that need to be propagated.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -148,7 +139,7 @@ public AsyncQuery updateStatus(String asyncQueryId, QueryStatus status) {
                 }
                 return loaded;
             });
-        } catch (ParseException e) {
+        } catch (Exception e) {
```
aklish, line 142: The DAO should not swallow exceptions.  Utilities need to propagate exceptions and let the calling code handle them (or the utility will have limited reuse and can only be used in limited contexts).

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -61,13 +66,13 @@ protected void cancelAsyncQuery() {
...
             Map<UUID, DataStoreTransaction> runningTransactionMap = transactionRegistry.getRunningTransactions();
-
-            String filterDateFormatted = evaluateFormattedFilterDate(Calendar.SECOND, maxRunTimeSeconds);
-            String filterExpression = "status=in=(" + QueryStatus.CANCELLED.toString() + ","
-                    + QueryStatus.PROCESSING.toString() + ","
-                    + QueryStatus.QUEUED.toString() + ");createdOn=ge='" + filterDateFormatted + "'";
-
-            Collection<AsyncQuery> asyncQueryCollection = asyncQueryDao.getActiveAsyncQueryCollection(filterExpression);
+            String filterDateFormatted = evaluateFormattedFilterDate(Calendar.SECOND, 2 * maxRunTimeSeconds);
+            String filterExpression = "status=in=(" + QueryStatus.PROCESSING.toString() + ","
```
aklish, line 70: Relying on time to find the queries that need to be canceled has some race conditions and other issues with configuration that can cause problems.

Instead of filtering on time, I think we need to introduce another state - CANCEL_COMPLETE or something along those lines that marks which queries have been successfully canceled.

ayeswarya:
@aklish I have made changes per latest comments. Thanks

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -87,33 +85,20 @@ protected void cancelAsyncQuery() {
...
-                   transactionRegistry.getRunningTransaction(tx).cancel(scope);
+                       JsonApiDocument jsonApiDoc = new JsonApiDocument();
+                       MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
+                       RequestScope scope = new RequestScope("query", NO_VERSION, jsonApiDoc,
+                               transactionRegistry.getRunningTransaction(tx), null, queryParams,
+                               tx, elide.getElideSettings());
+                       transactionRegistry.getRunningTransaction(tx).cancel(scope);
+                       List<AsyncQuery> asyncQueryList = asyncQueryCollection.stream()
+                               .filter(query -> query.getRequestId().equals(tx)).collect(Collectors.toList());
+                       asyncQueryList.forEach(asyncQuery -> asyncQueryDao.updateStatus(asyncQuery.getId(),
```
aklish, line 96: It will be more efficient to call asyncQueryDao.updateAsyncQueryCollection which can do a bulk update

ayeswarya:
@aklish  made changes to call the update collection dao method.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -0,0 +1,112 @@
...
+            List<String> asyncQueryList = new ArrayList<String>();
+            queriesToCancel.stream()
+               .forEach((tx) -> {
+                   JsonApiDocument jsonApiDoc = new JsonApiDocument();
+                   MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
+                   RequestScope scope = new RequestScope("query", NO_VERSION, jsonApiDoc,
+                           transactionRegistry.getRunningTransaction(tx), null, queryParams,
+                           tx, elide.getElideSettings());
+                   transactionRegistry.getRunningTransaction(tx).cancel(scope);
+                   List<String> cancelledQuery = asyncQueryCollection.stream().filter(query
```
aklish, line 97: A couple of improvements on this code.  I think we should rename some of the local variables for clarity.  When dealing with ids, we shouldn't call them queries or transactions but rather identifiers - so we can distinguish the collections from each other.

Second, I'm not sure how much of an improvement this is, but I would probably have written the mapping from UUIDs to kill to queries to kill as follows:

```java
Set<Query> queriesToCancel = queryIdsToCancel.stream()
        .map(uuid -> { 
            asyncQueryCollection.stream()
               .filter(query -> query.getRequestId().equals(uuid))
               .findFirst().orElseThrow(IllegalStateException::new)
         })
         .collect(Collectors.toSet());
```

Because naming is hard, it is sometimes more readable to collapse the mapping into a single logical block without intermediate state.  Alternatively, we could write a separate function to do the same thing.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryDAO.java
```diff
@@ -48,7 +49,14 @@ public AsyncQuery updateAsyncQueryResult(AsyncQueryResult asyncQueryResult,
...
-    public Collection<AsyncQuery> updateStatusAsyncQueryCollection(String filterExpression,
+    public Collection<AsyncQuery> updateStatusAsyncQueryCollection(FilterExpression filterExpression,
             QueryStatus status);
+    /**
+     * This method gets a collection of AsyncQuery and objects from database with status QUEUED/PROCESSING/CANCELLED and
+     * returns the objects.
+     * @param filterExpression filter expression to cancel AsyncQuery Objects based on
+     * @return query object list updated
+     */
+    public Collection<AsyncQuery> getActiveAsyncQueryCollection(FilterExpression filterExpression);
```
aklish, line 60: The DAO will be more usable if we don't tie down the contract with constraints like limiting the status to QUEUED, PROCESSING, etc.

This can just be a load method that takes a filter expression.  Let the calling code determine what states or other bits to filter on.

elide-async/src/main/java/com/yahoo/elide/async/service/DefaultAsyncQueryDAO.java
```diff
@@ -195,4 +183,29 @@ protected Object executeInTransaction(DataStore dataStore, Transactional action)
         }
         return result;
     }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public Collection<AsyncQuery> getActiveAsyncQueryCollection(FilterExpression filterExpression) {
```
aklish, line 189: Change the name to reflect this is a simple loadQueries or something.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryCancelThreadTest.java
```diff
@@ -0,0 +1,64 @@
...
+    public void testAsyncQueryCancelThreadSet() {
+        assertEquals(elide, cancelThread.getElide());
+        assertEquals(asyncQueryDao, cancelThread.getAsyncQueryDao());
+        assertEquals(7, cancelThread.getMaxRunTimeSeconds());
+    }
+
+    @Test
+    public void testCancelAsyncQuery() {
+        cancelThread.cancelAsyncQuery();
+        verify(asyncQueryDao, times(1)).getActiveAsyncQueryCollection(any());
```
aklish, line 62: I feel like these tests don't test too much of the code that was added.

We have all this set interaction between running transactions and transactions that exist in the database.  We ought to be testing (with unit tests):

1. Async queries in the database that are COMPLETE, CANCEL_COMPLETE are skipped.
2. Only the async queries that reflect running queries in the transaction registry are canceled.
3. Queries that exceed their runtime are canceled (but only if they are in the transaction registry).
4. Queries that didn't exceed their runtime are not canceled.

That's not an exhaustive test - but that will verify the functionality works as we expect.  

1.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryCleanerThreadTest.java
```diff
@@ -59,14 +62,12 @@ public void testAsyncQueryCleanerThreadSet() {
     @Test
     public void testDeleteAsyncQuery() {
         cleanerThread.deleteAsyncQuery();
-
-        verify(asyncQueryDao, times(1)).deleteAsyncQueryAndResultCollection(anyString());
+        verify(asyncQueryDao, times(1)).deleteAsyncQueryAndResultCollection(any());
```
aklish, line 65: Similar to my above comment - stronger tests would have a list of mock queries with different conditions of staleness - and we verify that:

1. the ones we expect to be canceled are canceled
2. the ones that we expect to be deleted are deleted
3. the ones we expect to be ignored are ignored.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -0,0 +1,112 @@
...
+                   asyncQueryList.addAll(cancelledQuery);
+               });
+            if (!queriesToCancel.isEmpty()) {
+                String cancelledQueryStr = String.join(",", asyncQueryList);
+                String filterById = "id=in=(" + cancelledQueryStr + ")";
+                FilterExpression fltIdExpression = filterParser.parseFilterExpression
+                        (filterById, AsyncQuery.class, false);
+                asyncQueryDao.updateStatusAsyncQueryCollection(fltIdExpression, QueryStatus.CANCEL_COMPLETE);
+            }
+        } catch (ParseException e) {
```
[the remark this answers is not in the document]
ayeswarya: The way the code is now, cancellation happens before we mark the status of the query as CANCEL_COMPLETE. So if there were to be an exception, the query status will not change, so it can be picked up for cancellation in the next run of the cancel thread...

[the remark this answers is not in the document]
ayeswarya: I was testing the code in example application and discovered that the executor does not schedule another thread if it encounters an exception. Can I catch exceptions, log them?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
```diff
@@ -44,28 +43,38 @@ private AsyncCleanerService(Elide elide, Integer maxRunTimeMinutes, Integer quer
         // Since there will be multiple hosts running the elide service,
         // setting up random delays to avoid all of them trying to cleanup at the same time.
         Random random = new Random();
-        int initialDelayMinutes = random.ints(0, maxCleanupInitialDelayMinutes).limit(1).findFirst().getAsInt();
+        int initialDelayMinutes = random.ints(0, maxInitialDelayMinutes).limit(1).findFirst().getAsInt();
```
moizarafat, line 46: This variable is not used any more, can be removed.

elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneSettings.java
moizarafat: we can take care of this in the download pr

moizarafat: move it into a struct

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -0,0 +1,126 @@
...
+
+            //Cancel Transactions
+            queryUUIDsToCancel.stream()
+               .forEach((uuid) -> {
+                   JsonApiDocument jsonApiDoc = new JsonApiDocument();
+                   MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
+                   RequestScope scope = new RequestScope("", "query", NO_VERSION, jsonApiDoc,
+                           transactionRegistry.getRunningTransaction(uuid), null, queryParams,
+                           uuid, elide.getElideSettings());
+                   transactionRegistry.getRunningTransaction(uuid).cancel(scope);
```
moizarafat, line 112: may be we should store transactionRegistry.getRunningTransaction(uuid) in a variable and check for null before calling cancel.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -18,6 +18,11 @@
      */
     private int threadPoolSize = 5;
 
+    /**
+     * Default max query run time.
+     */
+    private int maxRunTimeSeconds = 3600;
+
     /**
      * Default max query run time.
```
moizarafat, line 27: maxRunTimeMinutes is not needed any more

elide-async/src/main/java/com/yahoo/elide/async/service/DateUtil.java
```diff
@@ -0,0 +1,34 @@
...
+@Slf4j
+public class DateUtil {
+
+    /**
+     * Calculated and subtracts the amount based on the calendar unit and amount from current date.
+     * @param calendarUnit Enum such as Calendar.DATE or Calendar.MINUTE
+     * @param amount Amount of days to be subtracted from current time
+     * @return filter date
+     */
+     protected static Date calculateFilterDate(int calendarUnit, int amount) {
```
aklish, line 26: Any reason not to make this public?

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryCancelThreadTest.java
```diff
@@ -0,0 +1,143 @@
...
+                1577883600000L, QueryStatus.QUEUED);
+        AsyncQuery asyncQuery2 = createAsyncQueryTestObject("edc4a871-dff2-4054-804e-d80075cf827d",
+                1577883600000L, QueryStatus.QUEUED);
+        Collection<AsyncQuery> asyncCollection = new ArrayList<AsyncQuery>();
+        asyncCollection.add(asyncQuery1);
+        asyncCollection.add(asyncQuery2);
+        when(cancelThread.getAsyncQueryDao().loadAsyncQueryCollection(any())).thenReturn(asyncCollection);
+        cancelThread.cancelAsyncQuery();
+        ArgumentCaptor<FilterExpression> filterCaptor = ArgumentCaptor.forClass(FilterExpression.class);
+        verify(asyncQueryDao, times(1)).updateStatusAsyncQueryCollection(filterCaptor.capture(), any());
```
aklish, line 84: You could also verify the status argument here.

aklish, line 84: Can we verify the load statement as well?

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryCancelThreadTest.java
```diff
@@ -0,0 +1,143 @@
...
+                1577883600000L, QueryStatus.CANCEL_COMPLETE);
+        AsyncQuery asyncQuery2 = createAsyncQueryTestObject("edc4a871-dff2-4054-804e-d80075cf827d",
+                1577883600000L, QueryStatus.CANCELLED);
+        AsyncQuery asyncQuery3 = createAsyncQueryTestObject("edc4a871-dff2-4054-804e-d80075cf826d",
+                1577883600000L, QueryStatus.PROCESSING);
+        Collection<AsyncQuery> asyncCollection = new ArrayList<AsyncQuery>();
+        asyncCollection.add(asyncQuery1);
+        asyncCollection.add(asyncQuery2);
+        asyncCollection.add(asyncQuery3);
+        when(cancelThread.getAsyncQueryDao().loadAsyncQueryCollection(any())).thenReturn(asyncCollection);
```
aklish, line 104: Can we verify the updateStatus statement as well?

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryCancelThreadTest.java
```diff
@@ -0,0 +1,143 @@
...
+        AsyncQuery asyncQuery3 = createAsyncQueryTestObject("edc4a871-dff2-4054-804e-d80075cf826d",
+                1577883600000L, QueryStatus.QUEUED);
+        Collection<AsyncQuery> asyncCollection = new ArrayList<AsyncQuery>();
+        asyncCollection.add(asyncQuery1);
+        asyncCollection.add(asyncQuery2);
+        asyncCollection.add(asyncQuery3);
+        when(cancelThread.getAsyncQueryDao().loadAsyncQueryCollection(any())).thenReturn(asyncCollection);
+        cancelThread.cancelAsyncQuery();
+        ArgumentCaptor<FilterExpression> filterCaptor = ArgumentCaptor.forClass(FilterExpression.class);
+        verify(asyncQueryDao, times(1)).updateStatusAsyncQueryCollection(filterCaptor.capture(), any());
```
aklish, line 130: Verify the status.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryCleanerThreadTest.java
```diff
@@ -58,15 +59,19 @@ public void testAsyncQueryCleanerThreadSet() {
 
     @Test
     public void testDeleteAsyncQuery() {
+        ArgumentCaptor<FilterExpression> filterCaptor = ArgumentCaptor.forClass(FilterExpression.class);
         cleanerThread.deleteAsyncQuery();
-
-        verify(asyncQueryDao, times(1)).deleteAsyncQueryAndResultCollection(anyString());
+        verify(asyncQueryDao, times(1)).deleteAsyncQueryAndResultCollection(filterCaptor.capture());
+        Long date = System.currentTimeMillis() - (cleanerThread.getQueryCleanupDays() * 24 * 3600 * 1000);
```
aklish, line 65: It's possible for this test to fail because of a 1ms gap between when this date is calculated and when the delete filter is generated.   We'll need to find a less brittle approach.

elide-async/src/test/java/com/yahoo/elide/async/service/AsyncQueryCleanerThreadTest.java
```diff
@@ -58,15 +59,19 @@ public void testAsyncQueryCleanerThreadSet() {
...
 
     @Test
-    public void timeoutAsyncQuery() {
+    public void testTimeoutAsyncQuery() {
+        ArgumentCaptor<FilterExpression> filterCaptor = ArgumentCaptor.forClass(FilterExpression.class);
         cleanerThread.timeoutAsyncQuery();
-
-        verify(asyncQueryDao, times(1)).updateStatusAsyncQueryCollection(anyString(), any(QueryStatus.class));
+        verify(asyncQueryDao, times(1)).updateStatusAsyncQueryCollection(filterCaptor.capture(), any(QueryStatus.class));
+        Long date = System.currentTimeMillis() - (cleanerThread.getMaxRunTimeMinutes() * 60 * 1000);
```
aklish, line 74: Same comment on date.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -33,6 +33,12 @@
      */
     private int queryCleanupDays = 7;
 
+    /**
+     * Default time interval for cancelling async query transactions
```
aklish, line 37: The other comment for standalone was:
"Number of seconds between async query cancellation checks."

Let's use the same comment for both standalone and spring and maybe change to:
"Polling interval to identify async queries that should be canceled."

Also, 300 seconds is way too long for this interval.  It should probably be somewhere in the 1-10 second range by default.

PR #1448, opened 2020-07-17
Aggregation DataStore QueryLogger Implementation

dvarshney1:
## Description
This pull request added the Aggregation datastore querylogger implementation for Elide:

New Files that are added:
QueryDetail.java -> A public java class which contains information about the data model name and the underlying querytext that is being executed by the datastore
QueryLogger.java -> An interface file that contains methods acceptQuery, processQuery, cancelQuery, completeQuery
NoopQueryLogger.java -> The default implementation for the QueryLogger. Does Nothing, returns directly
QueryResponse.java -> Class holding information about the query response that is to be surfaced to the user

Major changes were also made to the following files to incorporate the change:
AggregationDataStoreTransaction.java -> loadObject() function is updated to include these necessary changes
AggregationDataStore.java -> QueryLogger has been added as another member variable

## Motivation and Context
This change will help in basic queryLogging for projects that use AggregationDataStore for Elide underneath. This change can help users know the amount of time their query is taking in executing, the response code they are getting etc.

## How Has This Been Tested?
Regression Testing was done to see if all the test cases till passed after adding these test cases.
AggregationDataStoreTransactionTest.java was updated to test for the new changes that are made to the loadObjects(0 function, specifically seeing that the queryLogger functions are called appropriately

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
```diff
@@ -74,6 +76,11 @@ public void populateEntityDictionary(EntityDictionary dictionary) {
 
     @Override
     public DataStoreTransaction beginTransaction() {
-        return new AggregationDataStoreTransaction(queryEngine, cache);
+        return new AggregationDataStoreTransaction(queryEngine, cache, queryLogger);
+    }
+
+    public AggregationDataStore setupCustomQueryLogger(QueryLogger logger) {
```
aklish, line 82: Why not just set the logger in the constructor?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
```diff
@@ -61,27 +70,41 @@ public void createObject(Object entity, RequestScope scope) {
...
+            if (result == null) {
+                result = queryEngine.executeQuery(query, queryEngineTransaction);
+                if (cacheKey != null) {
+                    cache.put(cacheKey, result);
+                }
             }
+            if (entityProjection.getPagination() != null && entityProjection.getPagination().returnPageTotals()) {
+                entityProjection.getPagination().setPageTotals(result.getPageTotals());
+            }
+            response = new QueryResponse(200, result.getData(), null);
```
aklish, line 97: There are enums with the HTTP status codes (Like HttpStatus).

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -193,4 +194,12 @@ private void populateMetaData(MetaDataStore metaDataStore) {
...
         return tables.get(classAlias);
     }
+
+    /**
+     * Translates the client query into SQL.
+     *
+     * @param query the client query.
+     * @return the SQL query.
+     */
+    public abstract SQLQuery toSQL(Query query);
```
aklish, line 204: Not all query engines are SQL query engines.  QueryEngine cannot have a dependency on anything related to SQL.

aklish: Instead, make this method:
```java
public abstract String explain(Query query);
```

elide-core/src/main/java/com/yahoo/elide/core/QueryLogger.java
```diff
@@ -0,0 +1,61 @@
...
+     * @param queryId The RequestScope requestId.
+     * @param qd The QueryDetail Object
+     */
+    void processQuery(UUID queryId, QueryDetail qd);
+
+    /**
+     * Cancels all queries currently running for a particular requestId
+     * @param queryId The RequestScope requestId.
+     */
+    void cancelQuery(UUID queryId);
```
aklish, line 53: I don't think this is ever invoked.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
```diff
@@ -104,4 +127,20 @@ Query buildQuery(EntityProjection entityProjection, RequestScope scope) {
     public void cancel() {
         queryEngineTransaction.cancel();
     }
+
+    private void aggregationCompleteQuery(RequestScope scope, QueryResponse response) {
```
aklish, line 131: I think I might try to collapse these three functions into the main logic.  A few reasons:

- Single line functions are often not necessary.
- The isCached logic is tied to code outside the function - so this can't really be reused (the purpose of a function).
- Some of the variables seem like they might be common across the functions.

elide-core/src/main/java/com/yahoo/elide/core/QueryDetail.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * QueryDetail implementation class. Stores the model name (root entity class) and
+ * the queryText that the underlying datastore will eventually run
+ */
+@Value
+@AllArgsConstructor
+@NoArgsConstructor(force = true)
+public class QueryDetail {
+    @Getter private String modelName;
```
aklish, line 21: Instead of modelName - what if we changed this to the query itself (of type Query).

aklish: On second thought, It might be cleaner if we just made these arguments to the processQuery function.  QueryDetail is missing some fields (like the requestId) - which makes me think they all either need to come together - or we should keep them as separate arguments.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
dvarshney1: For navi or any other external application to instantiate their own querylogger, this function can be used to update the default QueryLogger from Noop to their custom querylogger.

elide-core/src/main/java/com/yahoo/elide/core/QueryLogger.java
dvarshney1: This is a member function of the QueryLogger class as indicated in the Carbon docs. Yep, you are right, this is never invoked.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
dvarshney1: Resolved in latest commit

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
dvarshney1: Resolved in latest commit

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
dvarshney1: Resolved in latest commit

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStore.java
aklish: That should happen in the constructor though - not a set function.  Ideally, most stuff should be immutable.

elide-core/src/main/java/com/yahoo/elide/core/QueryLogger.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    void acceptQuery(UUID queryId, Principal user, Map<String, String> headers, String apiVer,
+                     Optional<MultivaluedMap<String, String>> queryParams, String path);
+
+    /**
+     * Processes and logs all the queries from QueryDetail
+     * @param queryId The RequestScope requestId.
+     * @param qd The QueryDetail Object
+     */
+    void processQuery(UUID queryId, QueryDetail qd);
```
aklish, line 47: Let's remove the QueryDetail class - and instead pass the Query object here along with queryText and isCached flags.

elide-core/src/main/java/com/yahoo/elide/core/QueryLogger.java
aklish: So I think we should call this function from transaction.cancel.  In order for that to work, we probably need to change the contract of the DataStoreTransaction.cancel to include the RequestScope.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/core/QueryLogger.java
```diff
@@ -0,0 +1,65 @@
...
+     * @param queryId The RequestScope requestId.
+     * @param query The underlying Query
+     * @param apiQuery The output querytext
+     * @param isCached Whether the result came from a cache or not
+     */
+    void processQuery(UUID queryId, Query query, String apiQuery, boolean isCached);
+
+    /**
+     * Cancels all queries currently running for a particular requestId
+     * @param queryId The RequestScope requestId.
```
aklish, line 55: We should mention that this method must be thread safe.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/core/QueryLogger.java
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Accepts the incoming GraphQL query and notes the start time for the query
+     * @param queryId The RequestScope requestId.
+     * @param user The Principal user
+     * @param headers Http Request Headers
+     * @param apiVer API Version
+     * @param apiQuery QueryString of the requested URL request
+     */
+    void acceptQuery(UUID queryId, Principal user, Map<String, String> headers,
```
aklish, line 29: Do we need both of these acceptQuery methods or can we just have the one below?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/core/QueryLogger.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * Accepts the incoming JSON API query and notes the start time for the query
+     * @param queryId The RequestScope requestId.
+     * @param user The Principal user
+     * @param headers Http Request Headers
+     * @param apiVer API Version
+     * @param queryParams QueryParams for the incoming JSON API query
+     * @param path The apiQuery endpoint path for the incoming query
+     */
+    void acceptQuery(UUID queryId, Principal user, Map<String, String> headers, String apiVer,
```
aklish, line 41: I'm wondering if it would be more powerful to log the elide User object here instead of the Principal.  The user has some additional bits (like the role support).

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/core/QueryResponse.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright 2020, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.datastores.aggregation.core;
+//live inside aggregation ds
```
aklish, line 7: ?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransactionTest.java
```diff
@@ -50,8 +54,8 @@
     // inject our own query instead of using buildQuery impl
     private class MyAggregationDataStoreTransaction extends AggregationDataStoreTransaction {
 
-        public MyAggregationDataStoreTransaction(QueryEngine queryEngine, Cache cache) {
-            super(queryEngine, cache);
+        public MyAggregationDataStoreTransaction(QueryEngine queryEngine, Cache cache, QueryLogger queryLogger) {
+            super(queryEngine, cache, queryLogger);
```
aklish, line 58: I feel like we should have one test for query logging failures (where an exception is thrown).  Also, we should have one test for cancelation.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/core/QueryLogger.java
dvarshney1: We can just have the one below. Do not need the other one since Querylogger no longer lives in Elide. Will update this

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/core/QueryResponse.java
dvarshney1: Sorry about this. Modified

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransactionTest.java
```diff
@@ -222,4 +224,47 @@ public void loadObjectsBypassCache() {
...
+            AggregationDataStoreTransaction transaction =
+                    new MyAggregationDataStoreTransaction(queryEngine, cache, queryLogger);
+            EntityProjection entityProjection = EntityProjection.builder().type(PlayerStats.class).build();
+
+            transaction.loadObjects(entityProjection, scope);
+
+        } catch (Exception e) {
+                result = e.getMessage();
+        } finally {
+            Object finalResult = result;
```
aklish, line 249: I don't think we need this variable.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransactionTest.java
```diff
@@ -222,4 +224,47 @@ public void loadObjectsBypassCache() {
...
+            when(queryEngine.explain(query)).thenReturn(myQuery.toString());
+            AggregationDataStoreTransaction transaction =
+                    new MyAggregationDataStoreTransaction(queryEngine, cache, queryLogger);
+            EntityProjection entityProjection = EntityProjection.builder().type(PlayerStats.class).build();
+
+            transaction.loadObjects(entityProjection, scope);
+
+        } catch (Exception e) {
+                result = e.getMessage();
+        } finally {
```
aklish, line 248: We can do this without the finally block.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransactionTest.java
dvarshney1: The local variable that we are referencing from inside the lambda expression should be final. That is why we have this. Otherwise, we can also check for Mockito.verify() from inside the catch block. We will not need to have the `finalResult` local variable then.

PR #1449, opened 2020-07-19
Dynamic config schema validation; Read resources from jar

AvaniMakwana:
Authored-by: AvaniMakwana

One of the multiple PRs to support  #1178

Description

- Add Table, Security & variable config validation against schema
- Ability to read model config from resources

Motivation and Context
Please refer #1178

How Has This Been Tested?

License


elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigHelpers.java
```diff
@@ -190,4 +156,40 @@ public static String readConfigFile(File configFile) {
...
         return sb.toString();
     }
+
+    /**
+     * Read config from classpath.
+     * @param resourcePath : path to resource
+     * @return content of resource
+     * @throws IOException
+     */
+    public static String readResource(String resourcePath) throws IOException {
```
aklish, line 166: I think we can remove this function and also readConfigFile (see later comment).

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidator.java
```diff
@@ -53,11 +53,28 @@
...
     private ElideTableConfig elideTableConfig;
     private ElideSecurityConfig elideSecurityConfig;
     private Map<String, Object> variables;
     private String configDir;
+    private boolean configInClassPath;
+
+    public DynamicConfigValidator(String configDir) {
+        File config = new File(configDir);
+
+        if (config.exists()) {
```
aklish, line 71: I feel like this is the best place to construct the spring resource URL with either file:// or resource://.

this.configDir value could actually be a URL containing file:// or resource:// and then used for loading security, variables, and tables (all using spring resource loader).

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/Config.java
```diff
@@ -0,0 +1,47 @@
...
+
+import java.io.File;
+/**
+ * Dynamic Config enum.
+ */
+public enum Config {
+
+    TABLE("table",
+          "tables" + File.separator,
+          "/elideTableSchema.json"),
```
aklish, line 16: Should forward slash here also be File.separator?

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/DynamicConfigSchemaValidator.java
```diff
@@ -0,0 +1,83 @@
...
+            break;
+        case SECURITY :
+            results = this.securitySchema.validate(new ObjectMapper().readTree(jsonConfig));
+            break;
+        case VARIABLE :
+            results = this.variableSchema.validate(new ObjectMapper().readTree(jsonConfig));
+            break;
+        default :
+            log.error("Not a valid config type :" + configType);
+            return false;
```
aklish, line 60: A method like this should either return false or throw an exception but not both.  Otherwise the contract is confusing.

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidator.java
```diff
@@ -53,11 +56,29 @@
...
+    private static final String CLASSPATH_PATTERN = "classpath*:";
+    private static final String FILEPATH_PATTERN = "file:";
+    private static final String HJSON_EXTN = "**/*.hjson";
 
-    private ElideTableConfig elideTableConfig;
-    private ElideSecurityConfig elideSecurityConfig;
-    private Map<String, Object> variables;
+    private ElideTableConfig elideTableConfig = new ElideTableConfig();
+    private ElideSecurityConfig elideSecurityConfig = new ElideSecurityConfig();
+    private Map<String, Object> variables = Collections.<String, Object>emptyMap();
```
aklish, line 67: Why not just new HashMap<>?

PR #1459, opened 2020-07-22
Rebased fetch for included relations

suriyasundar92:
## Description
The HQL query builder was adding fetch join for all to one relations. This PR changes the behavior to fetch them only when they are explicitly requested in the entity projection.

## Motivation and Context
The Fetch join for relations that are not requested are not necessary. 

## How Has This Been Tested?
The Unit tests are updated to test for the new behavior

## License


elide-datastore/elide-datastore-hibernate/src/test/java/com/yahoo/elide/datastores/hibernate/hql/AbstractHQLQueryBuilderTest.java
```diff
@@ -179,20 +190,20 @@ public void testSettingQueryParams() {
...
 
-        Optional<Pagination> previousPagination = pagination;
-
         PaginationImpl paginationMock = mock(PaginationImpl.class);
         when(paginationMock.getLimit()).thenReturn(10);
         when(paginationMock.getOffset()).thenReturn(50);
 
-        pagination = Optional.of(paginationMock);
+        EntityProjection previousEntityProjection = entityProjection;
+        entityProjection = mock(EntityProjection.class);
```
aklish, line 198: Setting and resetting the mock this way (including the old code) seems fairly hacky.

I suggest we initialize this class with a mock entity projection.  Let's add a `@BeforeEach` or `AfterEach` annotation which resets the mocks.  

Each method will have to setup the projection the way that. makes the most sense.

elide-datastore/elide-datastore-hibernate/src/test/java/com/yahoo/elide/datastores/hibernate/hql/AbstractHQLQueryBuilderTest.java
```diff
@@ -54,15 +55,21 @@
...
-        super(new EntityDictionary(new HashMap<>()), new TestSessionWrapper());
+        super(getMockEntityProjection(), new EntityDictionary(new HashMap<>()), new TestSessionWrapper());
         dictionary.bindEntity(Author.class);
         dictionary.bindEntity(Book.class);
         dictionary.bindEntity(Chapter.class);
         dictionary.bindEntity(Publisher.class);
         dictionary.bindEntity(Left.class);
     }
 
+    private static  EntityProjection getMockEntityProjection() {
```
aklish, line 68: Do we need a method for this?

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/AbstractHQLQueryBuilder.java
```diff
@@ -88,16 +86,6 @@ public AbstractHQLQueryBuilder withPossibleFilterExpression(Optional<FilterExpre
         return this;
     }
 
-    public AbstractHQLQueryBuilder withPossibleSorting(final Optional<Sorting> possibleSorting) {
```
aklish, line 91: The entity projection includes the Filter as well (not just the sorting and the pagination)

elide-datastore/elide-datastore-hibernate/src/test/java/com/yahoo/elide/datastores/hibernate/hql/AbstractHQLQueryBuilderTest.java
suriyasundar92: I had to create this method, due to the fact that calling super should be the first statement of the constructor, where this method is used.

elide-datastore/elide-datastore-hibernate/src/main/java/com/yahoo/elide/core/hibernate/hql/SubCollectionFetchQueryBuilder.java
```diff
@@ -24,10 +26,11 @@
 
     private final Relationship relationship;
 
-    public SubCollectionFetchQueryBuilder(Relationship relationship,
+    public SubCollectionFetchQueryBuilder(EntityProjection entityProjection,
+                                          Relationship relationship,
```
aklish, line 30: This method should get the entityProjection from the relationship - so there is no parameter duplication

elide-datastore/elide-datastore-hibernate5/src/main/java/com/yahoo/elide/datastores/hibernate5/HibernateTransaction.java
```diff
@@ -223,26 +218,23 @@ public Object getRelation(
...
-                        relation.getName(),
                         entity,
-                        filteredVal);
+                        relation
+                );
 
                 if (pagination != null && pagination.returnPageTotals()) {
-                    pagination.setPageTotals(getTotalRecords(relationship,
-                            Optional.ofNullable(filterExpression), scope.getDictionary()));
+                    pagination.setPageTotals(getTotalRecords(
```
aklish, line 228: Same comment here.  getTotalRecords should only really need the relationship.

PR #1471, opened 2020-07-31
Async Filter expression check

AvaniMakwana:
Authored-by: AvaniMakwana

Description
Filter expression check implementation

How Has This Been Tested?

License


elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -64,7 +64,6 @@
     @Embedded
     private AsyncQueryResult result;
 
-    @Exclude
     private String principalName;
```
moizarafat, line 67: Lets include `@CreatePermission(expression = "Prefab.Role.None")` on principalName. We don't want it to be set by users, we will be setting it.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/AsyncIT.java
```diff
@@ -81,7 +81,7 @@
 
         private String id;
         private String query;
-
+        private String principalName;
```
aklish, line 84: Remove unused variable

elide-async/src/main/java/com/yahoo/elide/async/models/security/AsyncQueryFilterExpression.java
```diff
@@ -0,0 +1,56 @@
...
+import java.security.Principal;
+import java.util.Collections;
+
+/**
+ * FilterExpressions for AsyncQuery.
+ */
+public class AsyncQueryFilterExpression {
+    private static final String PRINCIPAL_NAME = "principalName";
+
+    static private FilterPredicate getPredicateOfPrincipalName(String principalName) {
```
aklish, line 26: Can we consolidate this logic into AsyncQueryInlineChecks.java ?

AvaniMakwana:
updates in https://github.com/yahoo/elide/pull/1505

PR #1479, opened 2020-08-07
ISO8601 Date Serde format for Async Elide Object

### LINKED ISSUE #1486: Date Serde conflict between Elide Settings and AsyncQueryDAO.
AsyncQueryDAO defines its own Date Serde. When ElideSettings also defines a Date Serde, the AsyncQueryDAO serde is taking precedence over Elide Settings.

## Expected Behavior
The Serdes should not conflict.

## Current Behavior
AsyncQueryDAO defines its own Date Serde. When ElideSettings also defines a Date Serde, the AsyncQueryDAO serde is taking precedence over Elide Settings.

## Context
When passing date arguments in the Graphql queries, the format mismatch error happens.

ayeswarya:
Resolves #1486 

## Motivation and Context
Currently in elide-standalone and elide-spring during initialization of Async Service we create a clone of Elide Object with an ISO8601 Date Serde format. This logic needs to be updated to check if the Main Elide Object has Serde provided then pull the date format from there
AsyncQueryCleanerThread and AsyncQueryCancelThread have their version of evaluateFormattedFilterDate. It can be moved into a common place and then format should come from the ElideSettings instead of being hardcoded.

We can during the initialization of AsyncCleanerService initialize this method with the format by extracting the format from the elideSettings.
## How Has This Been Tested?
Tested in Example Application

## License


elide-async/src/main/java/com/yahoo/elide/async/service/DateFilterUtil.java
```diff
@@ -0,0 +1,41 @@
...
+     * @param calendarUnit Enum such as Calendar.DATE or Calendar.MINUTE
+     * @param amount Amount of days to be subtracted from current time
+     * @return formatted filter date
+     */
+    protected static String evaluateFormattedFilterDate(Elide elide, int calendarUnit, int amount) {
+        Calendar cal = Calendar.getInstance();
+        cal.setTime(new Date());
+        cal.add(calendarUnit, -(amount));
+        Date filterDate = cal.getTime();
+        ISO8601DateSerde serde = (ISO8601DateSerde) elide.getElideSettings().getSerdes().get(Date.class);
```
aklish, line 34: I"m not sure this is safe.  They may not have ISO8601 serde configured.  There are other serdes for dates like epoch/longs.  I'm also not sure what was wrong with the original approach.  Async has its own elide settings so it should be able to use whatever serde it wants.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -98,9 +100,17 @@ public AsyncCleanerService buildAsyncCleanerService(Elide elide, ElideConfigProp
     public AsyncQueryDAO buildAsyncQueryDAO(Elide elide) {
         // Creating a new ElideSettings and Elide object for Async services
         // which will have ISO8601 Dates. Used for DefaultAsyncQueryDAO.
+        String serdePattern = "yyyy-MM-dd'T'HH:mm'Z'";
+        String serdeTimeZone = "UTC";
+        if (elide.getElideSettings().getSerdes().get(Date.class).getClass().getName()
```
moizarafat, line 105: I think if the serde is not defined for the main elide object, then `getClass().getName()` can throw a Null Pointer Exception.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAutoConfiguration.java
```diff
@@ -104,7 +104,7 @@ public Elide initializeElide(EntityDictionary dictionary,
                 .withDefaultPageSize(settings.getPageSize())
                 .withJoinFilterDialect(new RSQLFilterDialect(dictionary))
                 .withSubqueryFilterDialect(new RSQLFilterDialect(dictionary))
-                .withAuditLogger(new Slf4jLogger())
+                .withAuditLogger(new Slf4jLogger())//;
```
moizarafat, line 107: please remove commented code.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
ayeswarya: it gives me an EpochDateConverter... doesnt throw npe

PR #1481, opened 2020-08-08
escape quotes in quoted strings

wcekan:
## Description
GraphQL test helper should escape any quotes when quoting the string.

## Motivation and Context
GraphQL tests are failing for JSON strings.

## How Has This Been Tested?
Unit testing.  Local product testing with the fix.

## License

### DIFF
--- a/elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
+++ b/elide-contrib/elide-test-helpers/src/main/java/com/yahoo/elide/contrib/testhelpers/graphql/GraphQLDSL.java
@@ -486,7 +486,7 @@ public final class GraphQLDSL {
      */
     public static Argument argument(String name, Object value, boolean quoted) {
         if (value instanceof String) {
-            value = quoted ? String.format("\"%s\"", value) : value;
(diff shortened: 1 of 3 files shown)


elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
```diff
@@ -199,7 +199,6 @@ public String getPlaceholder() {
         }
 
         public String escapeMatching() {
-            // It is unclear why replaceAll here breaks our tests
```
wcekan, line 202: FYI: `replaceAll` is not a replacement for `replace`.  If the target is a literal always use `replace`.

elide-core/src/main/java/com/yahoo/elide/parsers/JsonApiParser.java
```diff
@@ -25,7 +26,7 @@
  */
 public class JsonApiParser {
 
-    private final static Pattern DUPLICATE_SEPARATOR_PATTERN = Pattern.compile("/+");
+    private final static Pattern DUPLICATE_SEPARATOR_PATTERN = Pattern.compile("//+");
```
wcekan, line 29: We want to replace 2 or more with a single slash.  No need to waste time matching single slashes.

PR #1482, opened 2020-08-08
Path matcher

wcekan:
## Description
General cleanup.
* normalizePath regex improvement
* Remove repeated conditional test
* use entrySet
* diamonds for new HashMap
* Unnecessary duplication in swagger pom

## Motivation and Context
Code quality

## How Has This Been Tested?
Unit testing

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
@@ -199,7 +199,6 @@ public class FilterPredicate implements FilterExpression, Function<RequestScope,
         }
 
         public String escapeMatching() {
-            // It is unclear why replaceAll here breaks our tests
             return value.toString().replace("%", "\\%");
         }
     }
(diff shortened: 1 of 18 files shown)


elide-core/src/main/java/com/yahoo/elide/core/filter/FilterPredicate.java
```diff
@@ -199,7 +199,6 @@ public String getPlaceholder() {
         }
 
         public String escapeMatching() {
-            // It is unclear why replaceAll here breaks our tests
```
wcekan, line 202: FYI: `replaceAll` is not a direct replacement for `replace`.  If the target is a literal String always use `replace`.

elide-core/src/main/java/com/yahoo/elide/parsers/JsonApiParser.java
```diff
@@ -25,7 +26,7 @@
  */
 public class JsonApiParser {
 
-    private final static Pattern DUPLICATE_SEPARATOR_PATTERN = Pattern.compile("/+");
+    private final static Pattern DUPLICATE_SEPARATOR_PATTERN = Pattern.compile("//+");
```
wcekan, line 29: We want to replace 2 or more slashes.  No need to waste time matching a single slash.

PR #1519, opened 2020-08-31
Introduce case insensitive equality rsql operator. Default is case sensitive

Chandrasekar-Rajasekar:
## Description
RSQL default operations are made case sensitive. Case Insensitive Equality operators are added. 

`==` and `=in=` are now Case Sensitive
`=ini=` is the case insensitive equivalent.
**Usage** :
 `=ini='val'` -> IN_INSENSITIVE
 `=ini='*val'` -> PREFIX_CASE_INSENSITIVE
 `=ini='*val*'` -> INFIX_CASE_INSENSITIVE
 `=ini='val*'` -> POSTFIX_CASE_INSENSITIVE
 `=in='val'` or `=='val'` -> IN
 `=in='*val'` or `=='*val'` -> PREFIX
 `=in='*val*'` or `=='*val*'` -> INFIX
 `=in='val*'` or `=='val*'` -> POSTFIX

## Motivation and Context
Having a default case insensitive filter bypasses the datastore index (if exist) thereby making queries slow. User needs to select whether they want CI or CS filter. 

## How Has This Been Tested?
Integration test and Unit Test

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
@@ -12,6 +12,7 @@ import com.yahoo.elide.core.EntityDictionary;
 import com.yahoo.elide.core.Path;
 import com.yahoo.elide.core.exceptions.InvalidValueException;
 import com.yahoo.elide.core.filter.FilterPredicate;
+import com.yahoo.elide.core.filter.InInsensitivePredicate;
 import com.yahoo.elide.core.filter.InPredicate;
 import com.yahoo.elide.core.filter.IsEmptyPredicate;
 import com.yahoo.elide.core.filter.IsNullPredicate;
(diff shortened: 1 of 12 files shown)


elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -59,6 +60,8 @@
     private static final String SINGLE_PARAMETER_ONLY = "There can only be a single filter query parameter";
     private static final String INVALID_QUERY_PARAMETER = "Invalid query parameter: ";
     private static final Pattern TYPED_FILTER_PATTERN = Pattern.compile("filter\\[([^\\]]+)\\]");
+    private static final ComparisonOperator INI = new ComparisonOperator("=ini=", true);
+    private static final ComparisonOperator NOT_INI = new ComparisonOperator("=outi=", true);
```
aklish, line 64: Looks good, but we should add all the insensitive operators here as well (PREFIXI, INFIXI, etc)

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -343,10 +348,16 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
                     .collect(Collectors.toList());
 
             if (op.equals(RSQLOperators.EQUAL) || op.equals(RSQLOperators.IN)) {
-                return equalityExpression(arguments.get(0), path, values);
+                return equalityExpression(arguments.get(0), path, values, true);
+            }
+            if (op.equals(INI)) {
```
aklish, line 353: Let's make these else if statements.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -343,10 +348,16 @@ public FilterExpression visit(ComparisonNode node, Class entityType) {
...
+                return equalityExpression(arguments.get(0), path, values, false);
             }
             if (op.equals(RSQLOperators.NOT_EQUAL) || op.equals(RSQLOperators.NOT_IN)) {
-                return new NotFilterExpression(equalityExpression(arguments.get(0), path, values));
+                return new NotFilterExpression(equalityExpression(arguments.get(0), path, values, true));
+            }
+            if (op.equals(NOT_INI)) {
```
aklish, line 359: I believe we need to add explicit conditions for INFIXI, PREFIX, etc.

PR #1527, opened 2020-09-10
Data export - Refactoring and Prep for using ResultStorageEngine

moizarafat:
Co-authored-by: chandrakaran

Resolves  #1272

## Description
Refactoring of existing classes and prep work for using ResultStorageEngine

## Motivation and Context
Refer #1272

## How Has This Been Tested?
Existing Test cases pass and new test cases.

## License

### DIFF
--- a/elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
+++ b/elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
@@ -11,7 +11,7 @@ import lombok.Data;
  * Extra properties for setting up async query support.
  */
 @Data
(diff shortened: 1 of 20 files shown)


elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -11,7 +11,7 @@
  * Extra properties for setting up async query support.
  */
 @Data
-public class AsyncProperties extends ControllerProperties {
+public class AsyncProperties {
```
moizarafat, line 14: Since AsyncProperties does not have use for `path` property, removed extends ControllerProperties.

aklish (changes requested):

aklish (approved):

Issue #1465, opened 2020-07-27
RFC - Design for Multi Data Source Support for SQL Query Engine


moizarafat:
## Overview
Elide SQL Query Engine should support connections to multiple databases at the same time for running analytical queries. Elide will also support a configuration file format for the Query Engine DB properties.

On startup, the Elide service will create Persistent Units and EntityManagerFactorys using the DB properties from the configuration files. If no configuration is provided, the existing functionality should remain functional (use the existing Entity/CRUD DB connection for aggregation datastore).  

## DB Configuration Format
The database connection properties will be provided in [HJSON](https://hjson.github.io/) format config file (used with Dynamic Model Config currently). There can be multiple config files or a single config file with an array of properties.

## Location of Config

Elide will reuse the URL configuration setting, _modelConfig_ used for dynamic config model file location (point to a web or filesystem root directory containing the configuration) for the DB configuration.  The configuration will be stored in a directory, _db_. There will be separate directories for each query engine like SQL/NoSQL.

Optionally, the DB file (tarball archive) can be signed with PGP to verify its integrity (Similar to Dynamic Model Config).

```
db
└───sql
    │   db1.hjson
    │   db2.hjson
└───nonsql
    │   db1.hjson
    │   db2.hjson
└───variables.hjson
└───db.asc
```

The DB hjson files will be handlebar templates, allowing substitution through a variables.hjson file (Similar to Dynamic Model Config).

## Database Properties (in HJSON Config)
Any compulsory parameters will have to be passed along as top-level param and optional params can be passed as a Map.

| Property                      | Required   | Description                               | Default Value |
| ----------------------------- | ---------- | ----------------------------------------- | ------------- |
| name                          | Yes        | The name of the database connection.  This will be used for the persistent unit name. | N/A |
| url                 | Yes        |  The JDBC URL for the database connection i.e. javax.persistence.jdbc.URL  .| N/A |
| driver                 | Yes        |  The  JDBC driver for the database connection i.e. javax.persistence.jdbc.driver.| N/A |
| user                 | Yes        |  The  user for the database connection i.e. javax.persistence.jdbc.user.| N/A |
| dialect                 | Yes        |  The  Elide Dialect to use for query generation.| N/A |
| propertyMap                 | No        |  A map of Hibernate properties and persistence properties like , hibernate.show_sql, etc.| N/A |

[...]

PR #1552, opened 2020-09-23
Update timegrain serde to return java.sql.Date instead of java.util.Date

rishi-aga:
One of the multiple PRs that will resolve https://github.com/yahoo/elide/issues/1465

## Description
Update: Timegrain serde to return java.sql.Date instead of java.util.Date
Improve regex for Parameter name matching

## Motivation and Context
After switching to use PreparedStatement, found an issue that java.sql.Date is not supported by PreparedStatement of PrestoDB driver https://github.com/prestodb/presto/blob/master/presto-jdbc/src/main/java/com/facebook/presto/jdbc/PrestoPreparedStatement.java. So updating Timegrain serdes to return java.sql.Date instead of java.util.Date

## How Has This Been Tested?
Existing Tests Pass

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/DateTime.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/DateTime.java
@@ -5,7 +5,7 @@
  */
 package com.yahoo.elide.datastores.aggregation.timegrains;
 
-import java.util.Date;
+import java.sql.Date;
 
 /**
  * Time Grain class for DATETIME("yyyy-MM-dd HH:mm:ss").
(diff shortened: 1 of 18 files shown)


aklish (approved):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/DateTime.java
```diff
@@ -5,7 +5,7 @@
  */
 package com.yahoo.elide.datastores.aggregation.timegrains;
 
-import java.util.Date;
+import java.sql.Date;
```
moizarafat, line 8: We should use sql.Time or Timestamp because java.sql.date does not store Time Portion of the date.

moizarafat (changes requested):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/DateTime.java
rishi-aga: updated.

moizarafat (approved):

moizarafat (approved):

aklish (approved):

PR #1561, opened 2020-09-29
ResultType Default Value

moizarafat:
Setting Default Value for Result Type in AsyncQuery

## Description
Setting Default Value for Result Type in AsyncQuery

## Motivation and Context
Setting Default Value for Result Type in AsyncQuery

## How Has This Been Tested?
Existing Test Cases pass

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
@@ -69,8 +69,7 @@ public class AsyncQuery extends AsyncBase implements PrincipalOwned {
     private QueryStatus status = QueryStatus.QUEUED;
 
     @Enumerated(EnumType.STRING)
-    @NotNull
-    private ResultType resultType; //EMBEDDED, DOWNLOAD
+    private ResultType resultType = ResultType.EMBEDDED; //EMBEDDED, DOWNLOAD
 
     @Embedded
     private AsyncQueryResult result;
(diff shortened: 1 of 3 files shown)


jkusa (approved):

elide-async/src/test/java/com/yahoo/elide/async/hooks/ExecuteQueryHookTest.java
```diff
@@ -66,22 +66,6 @@ public void testWithResultFormatTypeNull() {
         });
     }
 
-    @Test
```
jkusa, line 69: Do we already have a test to assert the default behavior?

moizarafat: Added one now.

jkusa (approved):

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -69,8 +68,7 @@
     private QueryStatus status = QueryStatus.QUEUED;
 
     @Enumerated(EnumType.STRING)
-    @NotNull
```
aklish, line 72: Why remove NotNull?

moizarafat: I thought since it's going to have defaulted value anyways, a check for NotNull could be skipped.

aklish (approved):

PR #1562, opened 2020-09-29
Table Export: Supporting Interfaces

moizarafat:
Resolves #1272

## Description
i) Interfaces that will be supporting Table Export Async Requests i.e. TableExportParser and Validator. 
ii) GraphqlParser - GraphQL specific implementation of TableExportParser.
iii) QueryRunner - Refactoring for breaking out methods.

## Motivation and Context
Table Export feature for Async API.

## License


elide-async/src/main/java/com/yahoo/elide/async/export/TableExportParser.java
```diff
@@ -0,0 +1,24 @@
...
+package com.yahoo.elide.async.export;
+
+import com.yahoo.elide.async.models.AsyncQuery;
+import com.yahoo.elide.core.exceptions.BadRequestException;
+import com.yahoo.elide.request.EntityProjection;
+
+/**
+ * Interface for parsing TableExport request.
+ */
+public interface TableExportParser {
```
aklish, line 15: I would make this a class instead of an interface.

aklish (approved):

PR #1564, opened 2020-09-30
Added item expiration to AggregationDataStore cache

aklish:
## Description
PR adds a default item expiration to the Aggregation Data Store query cache.

## Motivation and Context
The cache was only leveraged if a particular table supported a version query.  This change allows the cache to work in a more primitive context - simply based on item expiration.

## How Has This Been Tested?
Updated unit tests.

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
@@ -82,11 +82,12 @@ public class AggregationDataStoreTransaction implements DataStoreTransaction {
             Query query = buildQuery(entityProjection, scope);
             if (cache != null && !query.isBypassingCache()) {
                 String tableVersion = queryEngine.getTableVersion(query.getTable(), queryEngineTransaction);
-                if (tableVersion != null) {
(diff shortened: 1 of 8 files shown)


moizarafat (approved):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
```diff
@@ -82,11 +82,12 @@ public void createObject(Object entity, RequestScope scope) {
             Query query = buildQuery(entityProjection, scope);
             if (cache != null && !query.isBypassingCache()) {
                 String tableVersion = queryEngine.getTableVersion(query.getTable(), queryEngineTransaction);
-                if (tableVersion != null) {
-                    cacheKey = tableVersion + ';' + QueryKeyExtractor.extractKey(query);
-                    result = cache.get(cacheKey);
-                }
+                tableVersion = tableVersion == null ? "" : tableVersion;
```
jkusa, line 85: Will the null case happen when a cache query is not provided?

Would something like this help with debugging?
```suggestion
                tableVersion = tableVersion == null ? "__UNSET_TABLE_VERSION__" : tableVersion;
```

aklish: null happens when no VersionQuery annotation exists.  The ";" I feel is sufficient to know there was nothing set.

PR #1568, opened 2020-10-01
Resolves #1566

### LINKED ISSUE #1566: Aggregation Store : Physical column names in SQL should be preceded with a table or alias name
If a dimension or metric definition references a physical column by name, it is copied directly _as is_ into the resulting SQL.   This may not work well if joining to a table with the same column name (even if the other column is prefixed with its join alias).  

## Expected Behavior
A column definition like 'createdOn' should instead be referenced as 'tableAlias.createdOn'.   

This only appears to be missing for metric and dimension formulas.

## Possible Solution
Ideally, we need a way to note which keywords in a native SQL fragment should have this substitution occur.  

We could leverage a special syntax to denote which columns need special handling: {[physicalColumn]}.

Alternatively, we could leverage the existing syntax but with a reserved word:

{{$tableName}}

Alternatively, we could leverage the existing syntax but with the logical table name:

{{tableName}}

aklish:
Resolves #1566 

{{column}} can now refer to logical column or a physical column.  

## License


elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/core/SqlReferenceVisitorTest.java
```diff
@@ -0,0 +1,106 @@
...
+        //Logical name does not match physical name
+        @DimensionFormula("{{someColumn}}")
+        private String dimension2;
+
+        //Test a join to another table with a matching column name.
+        @DimensionFormula("{{joinModel.dimension3}}")
+        private String dimension3;
+    }
+
+    @FromTable(name = "join_model")
```
moizarafat, line 54: Looking at the test case, should this be join_table?

moizarafat (approved):

Issue #1548, opened 2020-09-22
Add test for time dimension with complex definition


aklish:
We should add an explicit test for a time dimension that has a SQL definition that is complex (involves some kind of computation or possibly a join to another table).  We need to verify that the grain SQL fragment works in conjunction with the SQL definition.

A good example would be:

{
    name : createdOn
    type : TIME
    definition : '{{someOtherTable.createdOn}}'
    grain:
    {
        type :  SIMPLEDATE
        sql :  '''
        PARSEDATETIME(FORMATDATETIME({{}}, 'yyyy-MM-dd'), 'yyyy-MM-dd')
        '''
    }
}

## Expected Behavior
The generated SQL should expand the definition inside the grain fragment.

PR #1575, opened 2020-10-05
Standardize Template Substitutions for Join definition

rishi-aga:
## Description
Trim column references in HJSON config for field definitions.
Validate join definition in hjson config
Use {{..}} for column references in join definition i.e. `{{col1}} = {{joinField.col2}}` instead of `%from.col1 = %join.col2`
Allow physical column references from referred table in parent table.
Also resolves: https://github.com/yahoo/elide/issues/1548

## How Has This Been Tested?
Existing Tests Pass.
Additional Unit Tests

## License


aklish (changes requested):
Looks good but it would be great if all of the handlebar templating used {{ and }} instead of joins using %to and %from.

Also, the names join and from are confusing I think.  I think better names would be helpful here.

elide-contrib/elide-dynamic-config-helpers/src/test/resources/validator/missing_configs/models/tables/table2.hjson.bkp
```diff
@@ -14,7 +14,7 @@
                 name: playerCountry
                 to: country
                 type: toOne
-                definition: '${to}.id = ${from}.country_id'
+                definition: '%join.id = %from.country_id'
```
moizarafat, line 17: should we remove this bkp file?

rishi-aga: removed.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/query/QueryTranslator.java
```diff
@@ -268,14 +270,29 @@ private String extractJoinClause(Class<?> fromClass,
...
+     * @param fromClass parent class
+     * @param fromAlias parent alias
+     * @param joinClass join class
+     * @param joinAlias join alias
+     * @param expr unresolved ON clause
+     * @return string resolved ON clause
+     */
+    private String getJoinClause(Class<?> fromClass, String fromAlias, Class<?> joinClass, String joinAlias,
+                    String expr) {
+        SQLReferenceVisitor visitor = new SQLReferenceVisitor(referenceTable.getMetaDataStore(), fromAlias);
```
aklish, line 293: This works a little different than how other references are resolved.  The SQLReferenceVisitor is normally called at service start and all the references are saved off in the SQLReferenceTable.  From that point onward, only lookups happen in the reference table.  

Is there any reason we can't follow the same approach here?

elide-core/src/main/java/com/yahoo/elide/core/Path.java
```diff
@@ -58,6 +58,9 @@ public PathElement(Class type, Class fieldType, String fieldName) {
         }
     }
 
+    public Path() {
```
aklish, line 61: Make this protected.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/core/JoinPath.java
```diff
@@ -61,4 +63,43 @@ public JoinPath extend(String extensionPath, EntityDictionary dictionary) {
...
+            }
+        }
+
+        return ImmutableList.copyOf(elements);
+    }
+
+    private PathElement resolvePathAttribute(Class<?> entityClass,
+                                             EntityDictionary dictionary,
+                                             String fieldName) {
+        Class<?> attributeClass = null;
```
aklish, line 98: Since we don't know the field type, we could make it Object.class.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceVisitor.java
```diff
@@ -81,16 +81,25 @@ protected String visitFormulaDimension(ColumnProjection dimension) {
...
-        String expr = column.getExpression();
+    /**
+     * Resolve references.
+     *
+     * @param source
+     * @param expr expression with unresolved references
+     * @param logicalName logical column name
+     * @return expression with resolved references
+     */
+    public String resolveReferences(Queryable source, String expr, String logicalName) {
```
aklish, line 95: Should this be private?

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceVisitor.java
rishi-aga: No, Its called from QueryTranslator's getJoinClause method

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/core/JoinPath.java
rishi-aga: updated.

elide-core/src/main/java/com/yahoo/elide/core/Path.java
rishi-aga: updated.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceVisitor.java
aklish: I'd prefer is this class was kept private to the construction of the SQLReferenceTable.

aklish (changes requested):

aklish (approved):

PR #1576, opened 2020-10-06
Table Exporter:  Initial Version

moizarafat:
Resolves #1272 

## Description
TableExporter Initial Version using PersistentResource.loadObjects(...)

## Motivation and Context
#1272

## How Has This Been Tested?
Existing Test Case pass. New Test case added.

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/export/TableExporter.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/export/TableExporter.java
@@ -7,8 +7,11 @@ package com.yahoo.elide.async.export;
 
 import com.yahoo.elide.Elide;
 import com.yahoo.elide.async.models.AsyncQuery;
+import com.yahoo.elide.async.models.QueryType;
 import com.yahoo.elide.core.DataStoreTransaction;
 import com.yahoo.elide.core.PersistentResource;
+import com.yahoo.elide.core.RequestScope;
+import com.yahoo.elide.core.exceptions.InvalidValueException;
 import com.yahoo.elide.graphql.GraphQLRequestScope;
 import com.yahoo.elide.request.EntityProjection;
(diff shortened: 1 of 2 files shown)


elide-async/src/main/java/com/yahoo/elide/async/export/TableExporter.java
```diff
@@ -0,0 +1,98 @@
...
+            tx.flush(requestScope);
+
+            requestScope.runQueuedPreCommitTriggers();
+
+            elide.getAuditLogger().commit();
+            tx.commit(requestScope);
+
+            requestScope.runQueuedPostCommitTriggers();
+        } catch (IOException e) {
+            throw new IllegalStateException(e);
```
aklish, line 90: IOExceptions are handled by throwing a TransactionException in Elide.  We should be consistent here.

elide-async/src/test/java/com/yahoo/elide/async/export/TableExporterTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+    @BeforeEach
+    public void beforeTest() {
+        reset(tx);
+        elide = mock(Elide.class);
+        user = mock(User.class);
+        asyncQuery = mock(AsyncQuery.class);
+        graphQLParser = mock(GraphQLParser.class);
+
+        TransactionRegistry transactionRegistry = mock(TransactionRegistry.class);
```
aklish, line 54: I feel like mocking all of these objects make the tests brittle.  Also, most mocks are likely not global and should be created per test.  I would focus on mocking only the challenging objects (like the data store or something).  Everything else should use real objects.

aklish (changes requested):

aklish (approved):

PR #1584, opened 2020-10-08
Async model Spec Changes 

moizarafat:
Resolves #1272 

## Description
i) Changes to Async Model to support splitting into 2 models:
 AsyncQuery: Regular Async Query Requests
 TableExport: Table Export Async Requests (To Be Added in a separate PR)
ii) Combined the Hooks into single class.
iii) Use of ThreadLocal for maintaining thread state instead of transient field in AsyncAPI.
iv) Class Name and Comments updated to reflect model names.

## How Has This Been Tested?
Existing Test case pass

## License


elide-async/src/main/java/com/yahoo/elide/async/models/AsyncQuery.java
```diff
@@ -41,48 +43,93 @@
...
+    /**
+     * This method parses the url and gets the query params.
+     * And retrieves path to be used by underlying Elide.get method
+     * @param query query from the Async request
+     * @throws URISyntaxException URISyntaxException from malformed or incorrect URI
+     * @return Path extracted from URI
+     */
+    private String getPath(String query) throws URISyntaxException {
+        URIBuilder uri;
+        uri = new URIBuilder(query);
```
aklish, line 103: nitpick but we parse the query URL twice - once for the path and once for the headers.  We could combine these in a single parse.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncAPI.java
```diff
@@ -0,0 +1,123 @@
...
+import javax.validation.constraints.Pattern;
+
+/**
+ * Base Model Class for Async Query.
+ */
+@Entity
+@Inheritance(strategy = InheritanceType.SINGLE_TABLE)
+@DiscriminatorColumn
+@Data
+public abstract class AsyncAPI implements PrincipalOwned {
```
aklish, line 44: We could make this non-abstract and expose it also as an Elide entity.  Benefits include:
1. Being able to see all the queries in one place
2. We can consolidate cleaners and cancel threads because they can query the base class (instead of the subclass).

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncAPI.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private Date updatedOn = new Date();
+
+    @Transient
+    @Max(10)
+    @ComputedAttribute
+    private Integer asyncAfterSeconds = 10;
+
+    @Exclude
+    protected String naturalKey = UUID.randomUUID().toString();
```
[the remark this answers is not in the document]
moizarafat: will use `id` instead.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
```diff
@@ -37,8 +38,10 @@ private AsyncCleanerService(Elide elide, Integer maxRunTimeSeconds, Integer quer
 
         // Setting up query cleaner that marks long running query as TIMEDOUT.
         ScheduledExecutorService cleaner = Executors.newSingleThreadScheduledExecutor();
-        AsyncQueryCleanerThread cleanUpTask = new AsyncQueryCleanerThread(queryRunTimeThresholdMinutes, elide,
-            queryCleanupDays, asyncQueryDao, new DateUtil());
+        AsyncQueryCleanerThread<AsyncQuery> cleanUpTask = new AsyncQueryCleanerThread<AsyncQuery>(
+                queryRunTimeThresholdMinutes, elide, queryCleanupDays, asyncQueryDao, new DateUtil(),
```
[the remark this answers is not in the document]
moizarafat: changed.

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncAPI.java
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Base Model Class for Async Query.
+ */
+@Entity
+@Include(type = "asyncAPI")
+@Inheritance(strategy = InheritanceType.SINGLE_TABLE)
+@DiscriminatorColumn
+@Data
+public class AsyncAPI implements PrincipalOwned {
```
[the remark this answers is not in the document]
moizarafat: Since this class is not an Entity anymore, we don't need it.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncAPI.java
```diff
@@ -0,0 +1,126 @@
...
+import javax.persistence.PreUpdate;
+import javax.persistence.Transient;
+import javax.validation.constraints.Max;
+import javax.validation.constraints.Pattern;
+
+/**
+ * Base Model Class for Async Query.
+ */
+@Entity
+@Include(type = "asyncAPI")
```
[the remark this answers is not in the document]
moizarafat: This will not be an entity anymore. Navi is already using AsyncQuery and TableExport as the model names.

elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncHook.java
```diff
@@ -0,0 +1,61 @@
...
+    protected void executeAsync(AsyncAPI query, Callable<AsyncAPIResult> callable) {
+        asyncExecutorService.executeQuery(query, callable);
+    }
+
+    /**
+     * Update Principal Name.
+     * @param query AsyncAPI object to complete.
+     * @param requestScope RequestScope object.
+     */
+    protected void updatePrincipalName(AsyncQuery query, RequestScope requestScope) {
```
aklish, line 55: Should the first param be AsyncAPI type?

elide-async/src/main/java/com/yahoo/elide/async/hooks/UpdateQueryPrincipalNameHook.java
```diff
@@ -11,17 +11,13 @@
 import com.yahoo.elide.security.ChangeSpec;
 import com.yahoo.elide.security.RequestScope;
 
-import java.security.Principal;
 import java.util.Optional;
 
-public class UpdatePrincipalNameHook implements LifeCycleHook<AsyncQuery> {
+public class UpdateQueryPrincipalNameHook extends AsyncHook implements LifeCycleHook<AsyncQuery> {
```
aklish, line 16: I wonder if we should consolidate these three hooks into a single class.  A single hook can be called with different operation and phase arguments.  It might be more readable to have all the hook logic in one place.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncAPIThread.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Callable thread for executing the query provided in AsyncQuery and TableExport.
+ * It will also update the query status and result object at different
+ * stages of execution.
+ */
+@Data
+public abstract class AsyncAPIThread implements Callable<AsyncAPIResult> {
+
+    @Override
+    public AsyncAPIResult call() throws NoHttpResponseException, URISyntaxException {
```
aklish, line 26: You could make this method abstract instead of calling another method with the same signature.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCancelThread.java
```diff
@@ -8,6 +8,7 @@
 import static com.yahoo.elide.core.EntityDictionary.NO_VERSION;
 
 import com.yahoo.elide.Elide;
+import com.yahoo.elide.async.models.AsyncAPI;
```
aklish, line 11: Is this class will be common for both AsyncQuery and TableExport - we need to name all the common classes with the same convention - AsyncAPI or AsyncBase or something along those lines.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryCleanerThread.java
```diff
@@ -6,6 +6,7 @@
 package com.yahoo.elide.async.service;
 
 import com.yahoo.elide.Elide;
+import com.yahoo.elide.async.models.AsyncAPI;
```
aklish, line 9: Same comment on the class name here.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -17,91 +18,76 @@
...
+    /**
+     * Execute Async Request.
+     * @param service AsyncExecutorService Instance.
+     * @param user Elide User.
+     * @param apiVersion Api Version.
+     * @return Base Result Object.
+     * @throws URISyntaxException URISyntaxException
+     * @throws NoHttpResponseException NoHttpResponseException
+     */
+    private AsyncAPIResult executeRequest()
```
aklish, line 64: I would collapse this into the above processQuery method.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -112,30 +98,33 @@ else if (queryObj.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
...
-        URIBuilder uri;
-        uri = new URIBuilder(query);
+    private String getPath(URIBuilder uri) {
         return uri.getPath();
     }
 
-    private ElideResponse executeJsonApiRequest(AsyncQuery queryObj, UUID requestId) throws URISyntaxException {
-        MultivaluedMap<String, String> queryParams = getQueryParams(queryObj.getQuery());
+    private ElideResponse executeJsonApiRequest(Elide elide, User user, String apiVersion) throws URISyntaxException {
+        UUID requestUUId = UUID.fromString(queryObj.getRequestId());
```
aklish, line 110: typo in variable name.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -112,30 +98,33 @@ else if (queryObj.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
...
+        ElideResponse response = elide.get("", getPath(uri), queryParams, user, apiVersion, requestUUId);
         log.debug("JSONAPI_V1_0 getResponseCode: {}, JSONAPI_V1_0 getBody: {}",
                 response.getResponseCode(), response.getBody());
         return response;
     }
 
-    private ElideResponse executeGraphqlRequest(AsyncQuery queryObj, UUID requestId) throws URISyntaxException {
+    private ElideResponse executeGraphqlRequest(Map<String, QueryRunner> runners, User user, String apiVersion)
+            throws URISyntaxException {
+        UUID requestUUId = UUID.fromString(queryObj.getRequestId());
```
aklish, line 124: Any reason we moved this logic (requestUUId assignment) inside each executeX method?

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncAPI.java
```diff
@@ -0,0 +1,98 @@
...
+
+    private Date updatedOn = new Date();
+
+    @Transient
+    @Max(10)
+    @ComputedAttribute
+    private Integer asyncAfterSeconds = 10;
+
+    @Transient
+    private AsyncAPIUpdateThread queryUpdateWorker = null;
```
aklish, line 67: I think with the new design, we should be able to remove this from the model - and tie it together elsewhere.

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/hooks/UpdateQueryPrincipalNameHook.java
moizarafat: We don't have the phase available during execute to allow consolidation.

elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncAPIHook.java
```diff
@@ -0,0 +1,101 @@
...
+import lombok.Data;
+
+import java.security.Principal;
+import java.util.concurrent.Callable;
+
+/**
+ * AsyncAPI Base Hook methods.
+ */
+@Data
+public abstract class AsyncAPIHook {
```
aklish, line 31: I think this should implement LifeCycleHook to create a "Is A" relationship here.

elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncAPIHook.java
```diff
@@ -0,0 +1,101 @@
...
+    private AsyncExecutorService asyncExecutorService;
+
+    public AsyncAPIHook() { };
```
aklish, line 34: Can we remove this?

elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncAPIHook.java
```diff
@@ -0,0 +1,101 @@
...
+
+import java.security.Principal;
+import java.util.concurrent.Callable;
+
+/**
+ * AsyncAPI Base Hook methods.
+ */
+@Data
+public abstract class AsyncAPIHook {
+    private AsyncExecutorService asyncExecutorService;
```
aklish, line 32: Mark final.

elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncAPIHook.java
```diff
@@ -0,0 +1,101 @@
...
+            AsyncAPI query, RequestScope requestScope, AsyncAPIThread queryWorker) {
+        if (operation.equals(READ) && phase.equals(PRESECURITY)) {
+            validateOptions(query, requestScope);
+            executeAsync(query, queryWorker);
+        } else if (operation.equals(CREATE)) {
+            if (phase.equals(POSTCOMMIT)) {
+                completeAsync(query, requestScope);
+            } else if (phase.equals(PRESECURITY)) {
+                updatePrincipalName(query, requestScope);
+            }
```
aklish, line 64: If the phase is not presecurity or post commit, this will do nothing - but I think it ought to throw the exception below.

I would restructure this a bit like:
```java
if (op is READ and PRESECURITY) {
   return ...
} else if (op is CREATE) {
   if (phase is POSTCOMMIT) {
       return ...
   } else if (phase is PRESECURITY) {
       return ...
   }
}
throw exception
```

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncAPI.java
```diff
@@ -0,0 +1,100 @@
...
+
+    @CreatePermission(expression = "Prefab.Role.None")
+    private String principalName;
+
+    @UpdatePermission(expression = "(Principal is Admin OR Principal is Owner) AND value is Cancelled")
+    @CreatePermission(expression = "value is Queued")
+    @Enumerated(EnumType.STRING)
+    private QueryStatus status = QueryStatus.QUEUED;
+
+    private Date createdOn = new Date();
```
aklish, line 56: It looks like this can be created by the client.  We probably want to prohibit that.

elide-async/src/main/java/com/yahoo/elide/async/models/AsyncAPI.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private Date updatedOn = new Date();
+
+    @Transient
+    @Max(10)
+    @ComputedAttribute
+    private Integer asyncAfterSeconds = 10;
+
+    @Override
```
aklish, line 65: I think you can remove this because of lombok.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncAPIThread.java
```diff
@@ -0,0 +1,34 @@
...
+import java.net.URISyntaxException;
+import java.util.concurrent.Callable;
+
+/**
+ * Callable thread for executing the query provided in AsyncQuery and TableExport.
+ * It will also update the query status and result object at different
+ * stages of execution.
+ */
+@Data
+public abstract class AsyncAPIThread implements Callable<AsyncAPIResult> {
```
aklish, line 23: This is an abstract class with no members and only a single method that implements an existing interface.  Do we need this separate class?

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
```diff
@@ -133,10 +127,12 @@ public void executeQuery(AsyncQuery queryObj, User user, String apiVersion) {
      * @param user User
      * @param apiVersion API Version
      */
-    public void completeQuery(AsyncQuery query, User user, String apiVersion) {
-        if (query.getQueryUpdateWorker() != null) {
+    public void completeQuery(AsyncAPI query, User user, String apiVersion) {
+        AsyncAPIUpdateThread queryUpdateWorker = futureTask.get();
+        if (queryUpdateWorker != null) {
```
aklish, line 132: It won't be safe to use null/not-null here to signify anything.  After the first threadlocal is initialized by a timeout, it will be non-null forever thereafter.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
```diff
@@ -133,10 +127,12 @@ public void executeQuery(AsyncQuery queryObj, User user, String apiVersion) {
...
             log.debug("Task has not completed");
-            updater.execute(query.getQueryUpdateWorker());
+            updater.execute(queryUpdateWorker);
+            futureTask.remove();
```
aklish, line 135: This could be missed because of an exception - and then Elide will remain in an invalid state.

Instead of using null/non-null - you'll need to keep track of something inside the threadlocal itself.

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
```diff
@@ -45,12 +47,23 @@
...
     private static AsyncExecutorService asyncExecutorService = null;
     private ResultStorageEngine resultStorageEngine;
+    private ThreadLocal<AsyncAPIResultFuture> asyncResultFutureThreadLocal = new ThreadLocal<>();
+
+    /**
+     * A Future with Synchronous Execution Complete Flag.
+     */
+    @Data
+    @Builder
+    public static class AsyncAPIResultFuture {
```
aklish, line 60: Maybe make this private and non-static if you can.

aklish (changes requested):

aklish (approved):

PR #1585, opened 2020-10-09
Elide 5: Decouple query and metadata

aklish:
Phase 2 refactor to decouple the QueryEngine Query models and the Metadata models.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/TimeDimension.java
```diff
@@ -23,7 +23,7 @@
 @EqualsAndHashCode(callSuper = true)
 @Include(rootLevel = false, type = "timeDimension")
 @Value
-public class TimeDimension extends Dimension {
+public class TimeDimension extends Column {
```
moizarafat, line 26: The javadoc for the class says TimeDimension is type of Dimension, should that be updated to reflect this change?

moizarafat (changes requested):

moizarafat (approved):

PR #1592, opened 2020-10-13
enum change for cardinality size

ayeswarya:

## Description
Make the cardinality sizes consistent between dynamic config and data aggregation
## How Has This Been Tested?

Existing test cases
## License


moizarafat (approved):

elide-contrib/elide-dynamic-config-helpers/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/model/Table.java
```diff
@@ -111,11 +111,9 @@ public String getDescription() {
 
     public enum Cardinality {
 
-        TINY("tiny"),
         SMALL("small"),
         MEDIUM("medium"),
-        LARGE("large"),
-        HUGE("huge");
+        LARGE("large");
```
aklish, line 116: We don't want to remove labels here.  We want to add TINY and HUGE to: com.yahoo.elide.datastores.aggregation.annotation.Cardinality.

aklish (approved):

PR #1594, opened 2020-10-15
Elide 5: Sketch out support for nested query plans.

aklish:
## Description
1. Each metric has an optional query plan resolver that can produce a custom query plan for that metric.
2. Query plans can be nested.
3. Adds support for merging nested and unnested query plans.
4. Adds a translator which takes the merged query plan and the client query and constructs a final query.
5. Fixes a few issues with Time data type serdes.

## How Has This Been Tested?
More tests are coming in final phase.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/query/QueryPlanTranslator.java
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    @Override
+    public Query.QueryBuilder visitQuery(Query query) {
+        throw new UnsupportedOperationException("Visitor does not visit queries");
+    }
+
+    @Override
+    public Query.QueryBuilder visitQueryable(Queryable plan) {
+        if (invoked == false) {
```
moizarafat, line 47: This could be written as `if (!invoked)`

moizarafat (approved):

PR #1602, opened 2020-10-19
Added SQL Generation Tests for Nested Aggregation Queries

aklish:
Added tests for H2, Hive, and Presto.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AndExpression.java
+++ b/elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AndExpression.java
@@ -10,7 +10,6 @@ import static com.yahoo.elide.security.permissions.ExpressionResult.FAIL;
 import static com.yahoo.elide.security.permissions.ExpressionResult.PASS;
 
 import com.yahoo.elide.security.permissions.ExpressionResult;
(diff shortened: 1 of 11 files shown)


elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/AndExpression.java
```diff
@@ -10,6 +10,7 @@
 import static com.yahoo.elide.security.permissions.ExpressionResult.PASS;
 
 import com.yahoo.elide.security.permissions.ExpressionResult;
+import lombok.EqualsAndHashCode;
```
moizarafat, line 13: Looks like this is unused.

elide-core/src/main/java/com/yahoo/elide/security/permissions/expressions/OrExpression.java
```diff
@@ -10,6 +10,7 @@
 import static com.yahoo.elide.security.permissions.ExpressionResult.PASS;
 
 import com.yahoo.elide.security.permissions.ExpressionResult;
+import lombok.EqualsAndHashCode;
```
moizarafat, line 13: Looks like this is unused.

moizarafat (approved):

PR #1604, opened 2020-10-20
Removed entity name from Entity Binding

aklish:
Cleans up the logic that extracts the model names in elide 5.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
@@ -31,6 +31,7 @@ import com.yahoo.elide.utils.ClassScanner;
 import com.yahoo.elide.utils.coerce.CoerceUtil;
 import com.yahoo.elide.utils.coerce.converters.Serde;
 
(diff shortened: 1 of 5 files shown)


moizarafat (approved):

rishi-aga (approved):

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1707,4 +1685,27 @@ public static String getModelVersion(Class<?> modelClass) {
...
         return (apiVersionAnnotation == null) ? NO_VERSION : apiVersionAnnotation.version();
     }
+
+    /**
+     * Looks up the API model name for a given class.
+     * @param modelClass The model class to lookup.
+     * @return the API name for the model class.
+     */
+    public static String getEntityName(Class<?> modelClass) {
+        Class<?> declaringClass = lookupAnnotationDeclarationClass(modelClass, Include.class);
```
wcekan, line 1695: @aklish This caused a bug because Include is not necessarily on an Entity bean, but can be instead in the package-info.

PR #1619, opened 2020-10-29
Add micrometer dependency

ayeswarya:

## Description
Add micrometer dependency to spring boot starter pom.xml

## Motivation and Context
Applications that use Elide should not have to add micrometer dependency explicitly
## How Has This Been Tested?
Tested in an application that uses elide, by removing the explicitly added micrometer dependency from its pom.xml
## License

### DIFF
--- a/elide-spring/elide-spring-boot-starter/pom.xml
+++ b/elide-spring/elide-spring-boot-starter/pom.xml
@@ -112,7 +112,7 @@
         <dependency>
             <groupId>io.micrometer</groupId>
             <artifactId>micrometer-core</artifactId>
-            <version>1.5.1</version>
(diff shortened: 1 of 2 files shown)


moizarafat (approved):

elide-spring/elide-spring-boot-starter/pom.xml
```diff
@@ -109,6 +109,12 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>io.micrometer</groupId>
+            <artifactId>micrometer-core</artifactId>
+            <version>1.5.1</version>
```
aklish, line 115: Latest version is 1.5.6

Issue #1567, opened 2020-09-30
Aggregation Store : Column and table names should be quoted in SQL queries.


aklish:
To avoid conflicts with SQL reserved words, all table and column names should be quoted.

PR #1621, opened 2020-10-29
Quote all identifiers in SQL query

rishi-aga:
Resolves: https://github.com/yahoo/elide/issues/1567

## Description
1) Quote all the identifiers with in SQL.

2) HJSON newline test
3) Add dialect for MySQL.

## How Has This Been Tested?
Modified existing tests to pass.

## License


aklish (changes requested):
Looks good.  One comment.  We added a MySQL Dialect - but no tests for it.  I'm fine if we do that in another PR - but we need to test the generated SQL.

aklish (approved):

PR #1624, opened 2020-10-30
Migrate metadatacalls to metadata store

aklish:
Small cleanup for QueryEngine to remove a metadata method that belongs in the MetadataStore.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/AggregationDataStoreTransaction.java
```diff
@@ -37,13 +38,15 @@
...
     private final QueryLogger queryLogger;
+    private final MetaDataStore metaDataStore;
 
     public AggregationDataStoreTransaction(QueryEngine queryEngine, Cache cache,
                                            QueryLogger queryLogger) {
         this.queryEngine = queryEngine;
         this.cache = cache;
         this.queryEngineTransaction = queryEngine.beginTransaction();
         this.queryLogger = queryLogger;
+        this.metaDataStore = queryEngine.metaDataStore;
```
rishi-aga, line 49: should we be using getter method for metaDataStore incase we move the classes in different packages later ?
Also do we require to add metaDataStore as private member here ?

rishi-aga (approved):

rishi-aga (approved):

PR #1629, opened 2020-11-02
Provide Default time grain SQL expression

ayeswarya:
## Description
Default time grain SQL expression.
TimeGrainDefinition.java default expression should be {{}}
Make type and SQLoptional in HJSON config for TimeGrains as the Time Grain Defaults to Day and default expression will be
{{}}
Temporal.java has a default timegrain definition that is updated too.

## How Has This Been Tested?
New and existing test cases.
## License


elide-datastore/elide-datastore-aggregation/src/test/resources/configs/models/tables/SalesView.hjson
```diff
@@ -165,6 +163,16 @@
             sql: PARSEDATETIME(FORMATDATETIME({{}}, 'yyyy'), 'yyyy')
           }
         }
+        {
+          name: deliveryDefault
+          type: TIME
+          // Logical Column Reference in same table, which references another Logical Column in referred table, which references another Logical column in referred table, which references another Logical column in referred table, which references Physical column in referred table
+          definition: '{{delivery.time}}'
+          readAccess: Principal is guest user
+          grain:
```
aklish, line 172: What happens if we remove the grain definition here?  I would expect the developer to not put it here empty like this.

elide-model-config/src/main/resources/templates/table.hbs
```diff
@@ -74,7 +74,7 @@ public class <%#capitalizeFirstLetter name%><%/capitalizeFirstLetter%> <%#if ext
 <%/if%>
 <%#ifTypeMatches type "TIME"%>
 <%#with grain%>
-    @Temporal(grain = @TimeGrainDefinition(grain = TimeGrain.<%type%>, expression = "<%sql%>"), timeZone = "UTC")
+    @Temporal(grain = @TimeGrainDefinition(<%#if type%>grain = TimeGrain.<%type%><%/if%><%#if type%><%#if sql%>, <%/if%><%/if%><%#if sql%>expression = "<%sql%>"<%/if%>), timeZone = "UTC")
```
aklish, line 77: SQL can be provided even if type is not.

elide-model-config/src/test/java/com/yahoo/elide/contrib/dynamicconfighelpers/parser/handlebars/HandlebarsHydratorTest.java
```diff
@@ -275,6 +275,183 @@
             + "\n"
             + "}\n";
 
+    private static final String VALID_TABLE_GRAIN_JAVA_NAME = "PlayerStatsGrain";
+    private static final String VALID_TABLE_GRAIN_CLASS = "/*\n"
```
aklish, line 279: Do we need to create such a complex model to test this?  THere is a lot of things unrelated to the test in this model.

aklish (changes requested):

elide-model-config/src/main/resources/templates/table.hbs
moizarafat: Can we can move this `<%#if sql%>, <%/if%>` within `<%#if type%>grain = TimeGrain.<%type%><%/if%>`?

elide-datastore/elide-datastore-aggregation/src/test/resources/configs/models/tables/SalesView.hjson
aklish: Let's remove the grain here.

aklish (changes requested):

aklish (approved):

PR #1633, opened 2020-11-03
Attempt to run explain query tests against H2 in compliance mode agai…

aklish:
H2 supports database compatibility modes for a small set of databases.  When available, this PR will run Aggregation Store explain query tests against the targeted database as an additional validation.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -255,7 +255,9 @@ private long getPageTotal(Query query, SQLQuery sql, SqlTransaction sqlTransacti
         supplyFilterQueryParameters(query, stmt);
 
         // Run the Pagination query and log the time spent.
-        return CoerceUtil.coerce(runQuery(stmt, paginationSQL, SINGLE_RESULT_MAPPER), Long.class);
+        Long result = CoerceUtil.coerce(runQuery(stmt, paginationSQL, SINGLE_RESULT_MAPPER), Long.class);
+
+        return (result != null) ? result : 0;
```
aklish, line 260: Fixes a bug where the result set can come back empty.

moizarafat (approved):

PR #1634, opened 2020-11-04
ValueSourceType should not be NONE for some of @DimensionFormula

rishi-aga:
TableSource should have default value of `NONE` if not provided.

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/ColumnVisitor.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/ColumnVisitor.java
@@ -78,7 +78,7 @@ public abstract class ColumnVisitor<T> {
      * @param formula formula expression
      * @return references appear in the formula.
      */
-    protected static List<String> resolveFormulaReferences(String formula) {
+    public static final List<String> resolveFormulaReferences(String formula) {
         Matcher matcher = REFERENCE_PARENTHESES.matcher(formula);
         List<String> references = new ArrayList<>();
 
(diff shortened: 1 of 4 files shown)


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Column.java
```diff
@@ -153,13 +153,22 @@ private static String getTableSource(ColumnMeta meta, Column column) {
             }
         }
 
-        if (column.getColumnType() == REFERENCE) {
+        if (referencesAnotherModel(column.getExpression())) {
```
aklish, line 156: tableSource is not something we want to infer. It must be provided by the developer in the HJSON file.  TableSource of NONE means no type ahead search should be provided by the API.  This change basically removes that option.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/ColumnVisitor.java
```diff
@@ -78,7 +78,7 @@ protected T visitPhysicalReference(String reference) {
      * @param formula formula expression
      * @return references appear in the formula.
      */
-    protected static List<String> resolveFormulaReferences(String formula) {
+    public static final List<String> resolveFormulaReferences(String formula) {
```
aklish, line 81: See comment below - this change can be reverted based on that comment.

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Column.java
```diff
@@ -79,11 +81,15 @@ protected Column(Table table, String fieldName, EntityDictionary dictionary) {
             this.category = meta.category();
             this.values = new HashSet<>(Arrays.asList(meta.values()));
             this.tags = new HashSet<>(Arrays.asList(meta.tags()));
+            this.tableSource = (meta.tableSource().trim().isEmpty()) ? NONE : meta.tableSource();
```
aklish, line 84: I believe the UI will interpret this as an actual ID and look it up in the metadata API.  Better to just let it be null.

aklish (changes requested):

aklish (approved):

PR #1636, opened 2020-11-04
Entity Models support in MetaDataStore

moizarafat:
## Description
Binding `@Entity` models to Metadatastore to allow joins to use them. Only `@Include` annotated models will be supported.

## Motivation and Context
Allow joins to use Entity annotated models.

## How Has This Been Tested?
Existing Test cases pass.
New Test cases pass.

## License


elide-datastore/elide-datastore-aggregation/src/test/resources/create_tables.sql
```diff
@@ -27,6 +27,12 @@ CREATE TABLE IF NOT EXISTS players
       name VARCHAR(255)
     ) AS SELECT * FROM CSVREAD('classpath:player.csv');
 
+CREATE TABLE IF NOT EXISTS playerRanking
```
rishi-aga, line 30: I think this file is no longer used, so can be deleted.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/MetaDataStore.java
```diff
@@ -80,12 +84,29 @@ public HashMapDataStore apply(String key) {
...
+    }
+
+    /**
+     * get all MetaDataStore supported annotated classes.
+     * @return Set of Class with specific annotations.
+     */
+    private static Set<Class<?>> getAllAnnotatedClasses() {
+        Set<Class<?>> metadataStoreResult = ClassScanner.getAnnotatedClasses(METADATA_STORE_ANNOTATIONS);
+
+        Set<Class<?>> entityResult = ClassScanner.getAnnotatedClasses(METADATA_STORE_ENTITY_ANNOTATION);
```
aklish, line 97: Class scanning is really time consuming.  We don't want to scan 100% of the classes three times to do this.  In the past, this has caused build times to go through the roof.  We should make maybe a new method in ClassScanner that looks for the classes we need in one pass.

aklish (changes requested):

elide-core/src/main/java/com/yahoo/elide/utils/ClassScanner.java
```diff
@@ -83,4 +96,12 @@
                     .collect(Collectors.toSet());
         }
     }
+
+    /**
+     * Function which will be invoked for deciding to include the class in final results.
+     */
+    @FunctionalInterface
```
aklish, line 103: nice

aklish (approved):

PR #1637, opened 2020-11-04
Jpa entity graph

Chandrasekar-Rajasekar:
## Description
Prefetch an attribute that has a to-Many relationship with the queried Entity in order to avoid n + 1 problem. This operation is done by passing a JPA Entity Graph containing the to-many relationship attribute along with the query. A relationship is added to the JPA Entity Graph if the following conditions are satisfied. 

- The relationship has to be present in Entity Projection
- The relationship type should be ToMany and not computed relationship
- The relationship Entity should not be present in dirtyResource. (no mutable operation is done on the relationship entity in this request)
- If the relationship Entity is present in dirtyResource, then the relationships should have CASCADE.ALL cascade type.

## Motivation and Context
This change solves n + 1 problem.

## How Has This Been Tested?
IT test

## License


elide-core/src/main/java/com/yahoo/elide/ElideSettings.java
```diff
@@ -43,4 +43,5 @@
     @Getter private final int updateStatusCode;
     @Getter private final Map<Class, Serde> serdes;
     @Getter private final boolean enableJsonLinks;
+    @Getter private final boolean enableJPAFetchToManyRelationship;
```
aklish, line 46: maybe change to useJpaEntityGraphHint

elide-core/src/main/java/com/yahoo/elide/core/EntityDictionary.java
```diff
@@ -1368,6 +1368,24 @@ public boolean cascadeDeletes(Class<?> targetClass, String fieldName) {
...
+     * @param targetClass The class which owns the relationship.
+     * @param fieldName The relationship
+     * @return true or false
+     */
+    public boolean cascadeAll(Class<?> targetClass, String fieldName) {
+        CascadeType [] cascadeTypes =
+                getEntityBinding(targetClass).relationshipToCascadeTypes.getOrDefault(fieldName, new CascadeType[0]);
+
+        for (CascadeType cascadeType : cascadeTypes) {
+            if (cascadeType == CascadeType.ALL) {
```
aklish, line 1382: Do we need cascade all or can this also work with other cascade types?

elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/transaction/AbstractJpaTransaction.java
```diff
@@ -179,7 +182,19 @@ public Object loadObject(EntityProjection projection,
                     (QueryWrapper) new RootCollectionFetchQueryBuilder(projection, dictionary, emWrapper)
                             .build();
 
-            return query.getQuery().getSingleResult();
+            EntityGraph entityGraph = createEntityGraph(scope, projection);
+
+            if (scope.getElideSettings() != null
+                    && scope.getElideSettings().isEnableJPAFetchToManyRelationship()
+                    && entityGraph != null) {
```
aklish, line 189: Move construction of the graph inside this if block.

elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/transaction/AbstractJpaTransaction.java
```diff
@@ -295,4 +321,45 @@ public Object getRelation(
...
+     */
+    private EntityGraph createEntityGraph(RequestScope scope, EntityProjection projection) {
+        EntityDictionary dictionary = scope.getDictionary();
+        Class<?> entityClass = projection.getType();
+        List<String> relationshipNames = dictionary.getRelationships(entityClass);
+        for (String relationshipName : relationshipNames) {
+            if (!projection.getIncludedRelationsName().contains(relationshipName)) {
+                continue;
+            }
+            RelationshipType type = dictionary.getRelationshipType(entityClass, relationshipName);
```
aklish, line 346: Rather than looking up this information the dictionary (name and class of the relationship), you can get it from the entity projection.

elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/transaction/AbstractJpaTransaction.java
```diff
@@ -295,4 +321,45 @@ public Object getRelation(
...
+
+            if (type.isToMany() && !type.isComputed()) {
+                // To Many Relationship will always be collection type.
+                String relationshipClassName = dictionary.getJsonAliasFor(
+                        dictionary.getParameterizedType(entityClass, relationshipName));
+                boolean isDirtyResource = StreamSupport.stream(scope.getDirtyResources().spliterator(), false)
+                        .anyMatch(resource -> resource.getType().equals(relationshipClassName));
```
aklish, line 353: Maybe add a method to persistentResource isDirty.

elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/transaction/AbstractJpaTransaction.java
```diff
@@ -295,4 +321,45 @@ public Object getRelation(
...
+
+                if (!isDirtyResource || dictionary.cascadeAll(entityClass, relationshipName)) {
```
aklish, line 355: Why do we need cascade all again?

elide-datastore/elide-datastore-jpa/src/main/java/com/yahoo/elide/datastores/jpa/transaction/AbstractJpaTransaction.java
```diff
@@ -295,4 +321,45 @@ public Object getRelation(
...
+                    EntityGraph entityGraph = em.createEntityGraph(entityClass);
+                    entityGraph.addAttributeNodes(relationshipName);
```
aklish, line 357: Why not add all the projection attributes and relationships?  Also, can this recurse?

aklish (changes requested):

thaingo:
May I know if this is still in Elide 5 roadmap @aklish ?

aklish:
@thaingo Unfortunately, the entity graph didn't work in the general case. @Chandrasekar-Rajasekar will post some details to the ticket.  

We've tried a few other options, but I no longer believe Hibernate can support this use case.  It's possible another ORM has better support.  Alternatively, I updated the contract for the `DataStoreTransaction` in Elide 5 that would enable some other workarounds.

One of them is to always use the proxy object (and filter, sort, & paginate) in memory whenever we are fetching a collection of collections.  This would be a very small change we could put behind a feature flag.

Chandrasekar-Rajasekar:
The problem we faced with JPA Entity Graph is that we can't tell the JPA to fetch a relationship based on a filter. Without this, there is a risk of loading a large number of records in memory even though we only need a small chunk of it. 

Another problem is we can only add 1 to-Many relationship to an entity graph, to avoid either MultipleBagFetchError or explosion of data due to cartesian product.

thaingo:
Many thanks to @aklish and @Chandrasekar-Rajasekar for the detailed information.

> One of them is to always use the proxy object (and filter, sort, & paginate) in memory whenever we are fetching a collection of collections. This would be a very small change we could put behind a feature flag.

Looking forward to this workaround.

aklish:
> Many thanks to @aklish and @Chandrasekar-Rajasekar for the detailed information.
> 
> > One of them is to always use the proxy object (and filter, sort, & paginate) in memory whenever we are fetching a collection of collections. This would be a very small change we could put behind a feature flag.
> 
> Looking forward to this workaround.

#1876 has the workaround. @thaingo

thaingo:
> #1876 has the workaround. @thaingo
Its great and looks promising. Thank you @aklish

aklish:
Fixed another way: #1876

PR #1641, opened 2020-11-05
Collapse Cardinality into ColumnMeta and TableMeta.

rishi-aga:
Remove `@Cardinality` annotation.
HJSON cardinality support for dimension columns.

## How Has This Been Tested?
Existing tests pass.
Additional tests included.

## License


moizarafat (approved):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Column.java
```diff
@@ -79,11 +82,13 @@ protected Column(Table table, String fieldName, EntityDictionary dictionary) {
             this.category = meta.category();
             this.values = new HashSet<>(Arrays.asList(meta.values()));
             this.tags = new HashSet<>(Arrays.asList(meta.tags()));
+            this.cardinality = (meta.size().equals(CardinalitySize.UNKNOWN)) ? null : meta.size();
```
aklish, line 85: Why not just set it to unknown here instead of null?

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Column.java
rishi-aga: before this change if cardinality was not provided by end user, it was set as NULL so wanted to keep same expectation.
Changing default to `UNKNOWN` instead of null

aklish (approved):

PR #1657, opened 2020-11-11
enums should be case insensitive in hjson configs.

rishi-aga:
Updated cardinality, join type, grain type and column type to be case insensitive in hjson configs.
Improve error messages for json schema validation failures.

## License


elide-model-config/src/main/resources/elideTableSchema.json
```diff
@@ -140,15 +140,7 @@
...
-        "type": {
-          "oneOf":[
-            {"$ref": "#/definitions/enumtype"},
-            {"enum": [
-              "TIME"
-            ]}
-          ],
-          "default": "INTEGER"
-        },
+        "type": { "$ref": "#/definitions/columnTypes" },
```
rishi-aga, line 143: `measure` doesn't support `TIME` type so removing this.

elide-model-config/src/main/resources/elideTableSchema.json
```diff
@@ -229,20 +221,7 @@
...
-                {
-                  "enum": [
-                    "RELATIONSHIP",
-                    "ID"
-                  ]
-                }
-              ],
-              "default": "TEXT"
-            },
+            "type": { "$ref": "#/definitions/columnTypes" },
```
rishi-aga, line 224: `dimension` doesn't support `RELATIONSHIP` and `ID` types so removing this.

aklish (approved):

PR #1658, opened 2020-11-11
Remove Id field from dimensions for fact models

rishi-aga:
Remove Id field from dimensions for fact models

## How Has This Been Tested?
Existing tests pass.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Table.java
```diff
@@ -170,8 +169,8 @@ private boolean isFact(Class<?> cls, TableMeta meta) {
                 })
                 .collect(Collectors.toSet());
 
-        // add id field if exists
-        if (dictionary.getIdFieldName(cls) != null) {
+        // add id field if exists and this is not a fact model
+        if (!this.isFact() && dictionary.getIdFieldName(cls) != null) {
```
rishi-aga, line 173: Assuming if a model has `FromTable` annotation but isFact is set to false, we want `id` included in dimensions.

aklish (approved):

PR #1660, opened 2020-11-12
Switched command line schema validator to use system.exit instead of …

aklish:
…throwing exceptions

## License


elide-model-config/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/validator/DynamicConfigValidator.java
```diff
@@ -99,25 +97,31 @@ public DynamicConfigValidator(String configDir) {
...
+
+            DynamicConfigValidator dynamicConfigValidator = new DynamicConfigValidator(configDir);
+            dynamicConfigValidator.readAndValidateConfigs();
+        } catch (Exception e) {
+            System.err.println(e.getMessage());
+            System.exit(2);
+        }
 
-        log.info("Configs Validation Passed!");
+        System.out.println("Configs Validation Passed!");
```
moizarafat, line 124: log.info

moizarafat (approved):

rishi-aga (approved):

PR #1668, opened 2020-11-17
Make Dynamic Compile work without shade plugin

moizarafat:
## Description
InMemoryCompiler can now locate jar and classes during compile time instead of relying on the Shade plugin.
 
## Motivation and Context
For dynamic compile to work we had to use the shade plugin. Though this was OK for standalone-example, for Spring we had to stop using `spring-boot-maven-plugin`. This might discourage some of our users who use `Spring` as the shade plugin will throw a lot of warnings. 

The reason for using the shade plugin was that dynamic compile would not be able to read classes from jar files that are placed in the `BOOT-INF` location by Spring.

## How Has This Been Tested?
Elide Spring and Standalone Example repos have been verified.

## License


elide-model-config/src/main/java/com/yahoo/elide/contrib/dynamicconfighelpers/compile/ElideDynamicEntityCompiler.java
```diff
@@ -125,9 +125,24 @@ public ElideDynamicEntityCompiler(String path, DBPasswordExtractor dbPasswordExt
...
+        }
+
+        String classpath = sb.toString();
+
+        if (classpath != null && !classpath.equals("")) {
+            log.debug("Classpath = " + classpath);
+            compiler.useOptions("-classpath", sb.toString());
+        }
+
+        compiler.useParentClassLoader(Thread.currentThread().getContextClassLoader());
```
aklish, line 144: There are some strong recommendations to not use the context class loader and instead just this.class.getClassLoader instead:

https://stackoverflow.com/questions/1771679/difference-between-threads-context-class-loader-and-normal-classloader#:~:text=The%20thread%20context%20classloader%20is,currentThread().

aklish (changes requested):

aklish (approved):

PR #1674, opened 2020-11-19
Tweak Cleaner Thread Delays

moizarafat:
## Description
Changes the default run time for the Cleaner thread from 360 Minutes to 120 Minutes. The cleaner thread is responsible for updating the status of interrupted async queries (due to app crash or restart) to TIMEDOUT from PROCESSING. It is also responsible for cleaning up history. 

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
@@ -44,7 +44,7 @@ public class AsyncCleanerService {
         // Since there will be multiple hosts running the elide service,
         // setting up random delays to avoid all of them trying to cleanup at the same time.
         Random random = new Random();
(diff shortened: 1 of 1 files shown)


elide-async/src/main/java/com/yahoo/elide/async/service/AsyncCleanerService.java
```diff
@@ -25,7 +25,7 @@
 @Slf4j
 public class AsyncCleanerService {
 
-    private final int defaultCleanupDelayMinutes = 360;
+    private final int defaultCleanupDelayMinutes = 120;
```
moizarafat, line 28: The cleaner thread is responsible for updating the status of interrupted async queries (due to app crash or restart) to TIMEDOUT from PROCESSING. It is also responsible for cleaning up history. Tweaking to lower the value.

aklish (approved):

PR #1692, opened 2020-11-24
Added validation logic to check for inheritance cycles

aklish:
## License


elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -140,11 +140,34 @@ public void readAndValidateConfigs() throws IOException {
...
+        });
+    }
+
+    private void validateInheritanceCycles(ElideTableConfig tables, Table table, Set<Table> visited) {
+        visited.add(table);
+
+        if (!table.hasParent()) {
+            return;
+        }
+        Table parent = table.getParent(tables);
```
rishi-aga, line 161: table.hasParent() doesn't verify if parent model is actually defined so parent can become `null` here.

rishi-aga: may be move this logic to `validateExtend` method.

rishi-aga (approved):

PR #1693, opened 2020-11-25
Optionally compile hjson configs while running CLI

rishi-aga:
Validate readAccess attribute provided for dynamic models.
Optionally compile hjson configs while running CLI.

## How Has This Been Tested?
Existing tests pass.

## License


elide-model-config/src/main/java/com/yahoo/elide/modelconfig/compile/ElideDynamicEntityCompiler.java
```diff
@@ -5,89 +5,30 @@
...
 import java.util.stream.Collectors;
-import javax.sql.DataSource;
 
 /**
  * Compiles dynamic model pojos generated from hjson files.
  */
-@Slf4j
 public class ElideDynamicEntityCompiler {
 
-    public static ArrayList<String> classNames = new ArrayList<String>();
```
rishi-aga, line 44: added getClassName method for this at end.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/compile/ElideDynamicEntityCompiler.java
```diff
@@ -189,15 +72,6 @@ public ClassLoader getClassLoader() {
         return compiler.getClassloader();
     }
 
-    /**
-     * Get the class from compiled class lists.
-     * @param name name of the class
-     * @return Class
-     */
-    public Class<?> getCompiled(String name) {
```
rishi-aga, line 197: unused method, so removed.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/parser/handlebars/HandlebarsHydrator.java
```diff
@@ -36,6 +37,23 @@
...
+        this.helper = new HandlebarsHelper(staticModelDetails);
+        this.handlebars = new Handlebars(loader).with(MY_ESCAPING_STRATEGY);
+        this.handlebars.registerHelpers(ConditionalHelpers.class);
+        this.handlebars.registerHelper(AssignHelper.NAME, AssignHelper.INSTANCE);
+        this.handlebars.registerHelpers(helper);
+    }
+
+    public HandlebarsHydrator() {
+        this.handlebars = new Handlebars();
+        this.helper = null;
```
aklish, line 54: What is the use case for this constructor?

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/parser/handlebars/HandlebarsHelper.java
```diff
@@ -24,7 +25,7 @@
  */
 public class HandlebarsHelper {
 
-    private static final String EMPTY_STRING = "";
+    public static final String EMPTY_STRING = "";
```
aklish, line 28: It looks like this is not accessed outside this class.

elide-model-config/src/test/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidatorTest.java
```diff
@@ -91,18 +91,60 @@ public void testMissingArgumentValue() throws Exception {
...
+            assertEquals(0, exitStatus);
+        });
+    }
+
+    @Test
+    public void testValidConfigDirWithCompile() throws Exception {
+        String error = tapSystemErr(() -> {
+            int exitStatus = catchSystemExit(() ->
+                    DynamicConfigValidator.main(new String[] { "--configDir", "src/test/resources/validator/valid" }));
+            assertEquals(2, exitStatus);
```
aklish, line 121: This looks confusing.  Why do we expect this to fail since the directory is valid?

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -122,13 +158,36 @@ public static void main(String[] args) {
...
+
+            if (cli.hasOption("nomodelcompile")) {
+                System.out.println("Skipped compilation for Model configs");
+            } else {
+                System.out.println("Compiling Model configs (Use '--nomodelcompile' to skip this step).");
+                dynamicConfigValidator.hydrateAndCompileModelConfigs();
+                System.out.println("Model Configs Compilation Passed!");
+            }
+
+            if (cli.hasOption("nodbcompile")) {
```
aklish, line 176: What is the use case for compiling models but not compiling db configs?

aklish (changes requested):

elide-model-config/src/test/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidatorTest.java
rishi-aga: hjson compilation will fail as it wont find classes defined in other modules.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
rishi-aga: basically if connection to db is not possible then it can be useful.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/parser/handlebars/HandlebarsHydrator.java
rishi-aga: hydrateTableTemplate and hydrateSecurityTemplate methods use helper methods and I had to pass staticModelDetails to helper class. 
Default constructor is for when to use just hydrateConfigTemplate method that doesn't use helper class.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/parser/handlebars/HandlebarsHelper.java
rishi-aga: used in DynamicConfigValidator.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/framework/SQLUnitTest.java
```diff
@@ -385,11 +385,11 @@ public static void init(String sqlDialect) {
         playerStatsTable = (SQLTable) metaDataStore.getTable("playerStats", NO_VERSION);
     }
 
-    private static String getCompatabilityMode(String dialect) {
-        if (dialect.equals(SQLDialectFactory.getMySQLDialect().getDialectType())) {
-            return "MODE=MySQL;DATABASE_TO_LOWER=TRUE";
-        } else if (dialect.equals(SQLDialectFactory.getPostgresDialect().getDialectType())) {
-            return "MODE=PostgreSQL;DATABASE_TO_LOWER=TRUE";
+    private static String getCompatabilityMode(String dialectType) {
+        if (dialectType.equals(SQLDialectFactory.getMySQLDialect().getDialectType())) {
```
rishi-aga, line 389: Hikari doesn't support `DATABASE_TO_LOWER` property. Earlier `if` block was not getting executed as we were trying to compare class name with dialect type.

elide-model-config/src/test/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidatorTest.java
```diff
@@ -37,9 +36,18 @@ public void testValidInheritanceConfig() throws Exception {
     }
 
     @Test
-    public void testHelpArgumnents() {
-        assertDoesNotThrow(() -> DynamicConfigValidator.main(new String[] { "-h" }));
-        assertDoesNotThrow(() -> DynamicConfigValidator.main(new String[] { "--help" }));
+    public void testHelpArgumnents() throws Exception {
+        tapSystemErr(() -> {
```
rishi-aga, line 40: test cases were getting stuck if there is an error so updated these.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/parser/handlebars/HandlebarsHelper.java
rishi-aga: reverted now.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -371,23 +455,73 @@ private static boolean validateTableConfig(ElideTableConfig elideTableConfig) {
...
+
+    private static void extractChecksFromExpr(String readAccess, Set<String> extractedChecks) {
+
+        if (!isNullOrEmpty(readAccess)) {
+            ParseTree root = EntityPermissions.parseExpression(readAccess);
+            extractChecksFromTree(root, extractedChecks);
+        }
+    }
+
+    private static void extractChecksFromTree(ParseTree root, Set<String> extractedChecks) {
```
aklish, line 506: Check out PermissionExpressionVisitor.

We should create a visitor to extract the check strings from the expression ParseTree rather than trying to manipulate the tree directly as we do here.

elide-model-config/src/test/resources/validator/valid/models/tables/player_stats.hjson
```diff
@@ -14,7 +14,7 @@
       category: Table Category
       cardinality : lARge
       hidden : false
-      readAccess : A user is admin or is a player in the game
+      readAccess : (Principal is user AND Principal is member) OR (Principal is admin AND NOT Principal is guest)
```
aklish, line 17: I"m not sure how friendly this is to expect HJSON configuration to type in "Principal is" for each of the defined roles.  We should discuss this.

aklish (changes requested):

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/parser/handlebars/HandlebarsHelper.java
```diff
@@ -98,12 +84,17 @@ public String toUpperCase(Object obj) {
...
-     * @return string converted to upper case and spaces replaced
+     * Capitalize first letter, replace space with underscore and dot with euro sign.
+     * @param obj Object representation of the string.
+     * @param toUpperCase Change case to upper for converted string.
+     * @return converted string.
      */
-    public String toUpperCaseReplaceSpace(Object obj) {
-        return toUpperCase(obj).replace(SPACE, UNDERSCORE);
+    public String createSecurityIdenitfier(Object obj, boolean toUpperCase) {
+        String id = capitalizeFirstLetter(obj.toString()).replace(SPACE, UNDERSCORE).replace(DOT, DOLLAR_SIGN);
```
rishi-aga, line 93: Replacing `.` with `$` sign so that dynamic compiler doesn't complain about file and class name.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/PermissionExpressionVisitor.java
```diff
@@ -0,0 +1,52 @@
...
+
+import lombok.AllArgsConstructor;
+
+import java.util.function.Function;
+
+/**
+ * Expression Visitor.
+ */
+@AllArgsConstructor
+public class PermissionExpressionVisitor extends ExpressionBaseVisitor<Expression> {
```
aklish, line 23: The visitor doesn't have to return an expression.  The visitor can construct anything really.  I thought you might construct a `Set<String>` of permission rules - that then could be iterated over and validated against the roles or existing security checks.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -373,23 +470,34 @@ private static boolean validateTableConfig(ElideTableConfig elideTableConfig) {
...
+            validateCheckExpr(table.getReadAccess(), visitor);
         }
 
         return true;
     }
 
+    private void validateCheckExpr(String readAccess, PermissionExpressionVisitor visitor) {
+        if (!isNullOrEmpty(readAccess)) {
+            ParseTree root = EntityPermissions.parseExpression(readAccess);
+            visitor.visit(root).evaluate(null);
```
aklish, line 497: This is probably not safe to evaluate checks like this.  Checks can do all kinds of crazy things (since they are developer provided code).

You can instead:
1.  Have the visitor extract a list/set of check names.
2. Lookup the check name either in the list of roles in the security.hjson file or check the dictionary (getCheckMappings)

aklish (changes requested):

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/PermissionExpressionVisitor.java
rishi-aga: Updated.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/parser/handlebars/HandlebarsHelper.java
```diff
@@ -91,12 +84,17 @@ public String toUpperCase(Object obj) {
...
-     * @param obj Object representation of the string
-     * @return string converted to upper case and spaces replaced
+     * Capitalize first letter, replace space with underscore and dot with euro sign.
+     * @param obj Object representation of the string.
+     * @param toUpperCase Change case to upper for converted string.
+     * @return converted string.
      */
-    public String toUpperCaseReplaceSpace(Object obj) {
-        return toUpperCase(obj).replace(SPACE, UNDERSCORE);
+    public String createSecurityIdenitfier(Object obj, boolean toUpperCase) {
```
aklish, line 92: Typo in function name.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/parser/handlebars/HandlebarsHydrator.java
aklish: Seems like it is safer to only have one fully initialized constructor.  Why not just use the same one throughout (even if config template doesn't need the static info)?

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -133,21 +179,57 @@ public static void main(String[] args) {
...
         validateNameUniqueness(this.elideSQLDBConfig.getDbconfigs());
         validateNameUniqueness(this.elideTableConfig.getTables());
         validateInheritance(this.elideTableConfig);
         populateInheritance(this.elideTableConfig);
-        validateTableConfig(this.elideTableConfig);
+        validateTableConfig();
         validateJoinedTablesDBConnectionName(this.elideTableConfig);
     }
 
+    public void hydrateAndCompileModelConfigs() throws Exception {
```
aklish, line 197: What calls this method?  Why does elide compiler have a singleton method here?

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -133,21 +179,57 @@ public static void main(String[] args) {
...
+        hydrateAndCompileModelConfigs(ElideDynamicInMemoryCompiler.newInstance().ignoreWarnings());
+    }
+
+    public void hydrateAndCompileModelConfigs(ElideDynamicInMemoryCompiler compiler) throws Exception {
+        HandlebarsHydrator hydrator = new HandlebarsHydrator(staticModelDetails);
+        Map<String, String> tableClasses = hydrator.hydrateTableTemplate(this.elideTableConfig);
+        Map<String, String> securityClasses = hydrator.hydrateSecurityTemplate(this.elideSecurityConfig);
+
+        compiler.useParentClassLoader(getClass().getClassLoader());
```
aklish, line 206: Seems like this belongs elsewhere (like maybe in the compiler constructor).

elide-model-config/src/test/resources/validator/duplicate_security_role/models/security.hjson
```diff
@@ -0,0 +1,6 @@
+{
+    roles:  [
+        Admin User
+        prefab.role.all
```
aklish, line 4: I want to talk about this test.  I'm a bit confused here about prefab.role.all.  Are we trying to define this is a role or reference the "Prefab.Role.All" that exists as prefabricated security role?

I assume the former, but I'm not sure what the goal of this is.  This looks like something nobody should ever try to do.

aklish (changes requested):

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/parser/handlebars/HandlebarsHelper.java
rishi-aga: fixed.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
rishi-aga: moved.

elide-model-config/src/test/resources/validator/duplicate_security_role/models/security.hjson
rishi-aga: this is negative test to ensure validation fails if someone tries to define the already defined roles. Role "Prefab.Role.All" already exists so validation must complain.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
rishi-aga: this is called from static main method required for cli. Removed this method and creating compiler object in main method.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/integration/AggregationDataStoreIntegrationTest.java
```diff
@@ -86,6 +91,7 @@ protected void configure() {
 
                     try {
                         dictionary.addSecurityChecks(COMPILER.findAnnotatedClasses(SecurityCheck.class));
+                        dictionary.addSecurityChecks(Collections.singleton(OperatorCheck.class));
```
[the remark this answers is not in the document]
rishi-aga: not calling scanForSecurityChecks method on this dictionary. Should I add this check to `TestCheckMappings.MAPPINGS` ?

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
rishi-aga: also removed singleton method for compiler class.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/integration/AggregationDataStoreIntegrationTest.java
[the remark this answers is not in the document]
aklish: Would probably be cleaner to do it that way (consistent with what is there).

aklish (approved):

PR #1696, opened 2020-11-25
Elide 5.x MetaDataStore enable/disable flag

AvaniMakwana:
## Description
Added Enable flag(default false) for MetaDataStore API 

## Motivation and Context
A new flag in Aggregation store properties which can be used to enable/disable MetaDataStore endpoints like /tables, /dimensions etc.

## How Has This Been Tested?
Unit tests included

## License


elide-model-config/src/test/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidatorTest.java
```diff
@@ -140,7 +140,7 @@ public void testInheritanceCycle() throws Exception {
             assertEquals(2, exitStatus);
         });
 
-        assertEquals("Inheriting from table 'B' creates an illegal cyclic dependency.\n", error);
+        assertEquals("Inheriting from table 'A' creates an illegal cyclic dependency.\n", error);
```
moizarafat, line 143: Just curious, what change in this PR would need this test case to be updated.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/framework/SQLUnitTest.java
```diff
@@ -348,7 +348,7 @@ public static void init(String sqlDialect) {
             throw new IllegalStateException(e);
         }
 
-        metaDataStore = new MetaDataStore(ClassScanner.getAllClasses("com.yahoo.elide.datastores.aggregation.example"));
+        metaDataStore = new MetaDataStore(ClassScanner.getAllClasses("com.yahoo.elide.datastores.aggregation.example"), true);
```
moizarafat, line 351: Can we set this to false? I think this does not need to be true for the test cases relying on SQLUnitTest to work. If it can work with false, then it will test the other scenario that we can do analytical queries without populating /table, /column etc.

moizarafat: If it does not work then we need test cases which verify that we can do analytical queries without populating /table, /column etc.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/DisableMetaDataStoreControllerTest.java
```diff
@@ -0,0 +1,30 @@
...
+import static io.restassured.RestAssured.when;
+import static org.hamcrest.Matchers.containsInAnyOrder;
+import com.yahoo.elide.core.exceptions.HttpStatus;
+import org.junit.jupiter.api.Test;
+import org.springframework.test.context.ActiveProfiles;
+
+/**
+ * Executes Controller tests with Aggregation Store disabled.
+ */
+@ActiveProfiles("disableMetaDataStore")
```
aklish, line 17: nice

aklish (approved):

PR #1706, opened 2020-12-02
Elide boot timing fix by allowing Spring Boot Apps to specify Packages to look for Models

moizarafat:
## Description
Spring Boot Applications use ClassScanner to find the models in the whole project including lib and jars. This change allows users to specify packagename to look for models. That will improve the boot time for application. 

## How Has This Been Tested?
Existing test case pass. Tested through the example app.

## License


moizarafat:
Moved to draft as we are going to try other options.

aklish:
Closing - but keeping the branch.  This didn't seem to make a big difference.

PR #1707, opened 2020-12-03
Dynamic Model Extends: Inherit Table Properties

moizarafat:
## Description
When extending Models in Dynamic Config, we need to inherit Table Properties.

## How Has This Been Tested?
Existing and new test cases.

## License


elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -174,57 +175,114 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
-        return dimensions;
-    }
 
-    private Map<String, Join> getInheritedJoins(Table table, Map<String, Join> joins) {
-        table.getJoins().forEach(join -> {
-            if (!joins.containsKey(join.getName())) {
-                joins.put(join.getName(), join);
-            }
-        });
+    private Object getInheritedAttribute(Table table, Object result, Inheritance action, boolean recurseOnlyIfNull) {
```
aklish, line 270: Don't use object here.  We should use generics.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -174,57 +175,114 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
     private void populateInheritance(ElideTableConfig elideTableConfig) {
-        for (Table table : elideTableConfig.getTables()) {
+        //ensures validation is run before populate always.
+        validateInheritance(this.elideTableConfig);
+
+        List<Table> tables = elideTableConfig.getTables().stream().collect(Collectors.toList());
+        while (!tables.isEmpty()) {
+            Table table = tables.remove(0);
             if (table.hasParent()) {
-                Map<String, Measure> measures = getInheritedMeasures(table, new HashMap<String, Measure>());
```
aklish, line 180: We should leave the methods getInheritedMeasures and getInheritedDimensions, etc.  They improve readability.  Those functions can return the correct type and in turn call getInheritedAttribute with the appropriate Inherit method.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -174,57 +175,114 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
+
+                Table parent = table.getParent(elideTableConfig);
+                // If parent also extends, ensure parent is processed first.
+                if (parent.getExtend() != null && parent.getSql() == null && parent.getTable() == null) {
+                    tables.add(table);
```
aklish, line 191: There are two standard ways to ensure ancestors are processed before decedents:
 - Standard DFS algorithm with a visited set and toVisit stack.
 - Tail Recursion. 

Let's not do this with a single data structure that we both add and remove from.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -174,57 +175,114 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
+
+        List<Table> tables = elideTableConfig.getTables().stream().collect(Collectors.toList());
+        while (!tables.isEmpty()) {
+            Table table = tables.remove(0);
             if (table.hasParent()) {
-                Map<String, Measure> measures = getInheritedMeasures(table, new HashMap<String, Measure>());
+
+                Table parent = table.getParent(elideTableConfig);
+                // If parent also extends, ensure parent is processed first.
+                if (parent.getExtend() != null && parent.getSql() == null && parent.getTable() == null) {
```
aklish, line 190: Why not just call parent.hasParent here?  It improves readability

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -174,57 +175,114 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
+                    tables.add(table);
+                    continue;
+                }
+
+                Map<String, Measure> measures = (Map<String, Measure>) getInheritedAttribute(table,
+                        new HashMap<String, Measure>(), (tab, result) -> {
```
aklish, line 196: You technically don't need to add a table parameter (tab) here because it is available through closure.

aklish: Having both tab and table looks odd.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -174,57 +175,114 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
-            if (!measures.containsKey(m.getName())) {
-                measures.put(m.getName(), m);
+                String schema = (String) getInheritedAttribute(table, null,
+                        (tab, result) -> {
+                            return tab.getSchema();
+                        }, true
+                );
+                table.setSchema(schema);
+
+                String dbConnectionName = (String) getInheritedAttribute(table, null,
```
aklish, line 238: These should all be separate functions.  Having embedded lambdas all together like this in one function makes this hard to read.

aklish: null as second parameter seems like the abstraction to commonize this code is not working.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
aklish, line 270: The last parameter recurseOnlyIfNull is a tell that this really ought to be two separate functions.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -174,57 +175,114 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
+                // If parent also extends, ensure parent is processed first.
+                if (parent.getExtend() != null && parent.getSql() == null && parent.getTable() == null) {
+                    tables.add(table);
+                    continue;
+                }
+
+                Map<String, Measure> measures = (Map<String, Measure>) getInheritedAttribute(table,
+                        new HashMap<String, Measure>(), (tab, result) -> {
+                            tab.getMeasures().forEach(measure -> {
+                                if (!((Map<String, Measure>) result).containsKey(measure.getName())) {
```
aklish, line 198: Lots of casting like this makes the code hard to read (and generally can be removed through other means).

elide-model-config/src/test/resources/validator/infinite_loop_cyclic_dependency_inheritance/models/tables/tableA.hjson
```diff
@@ -0,0 +1,18 @@
+{
+  tables:
+  [
+    {
+      name: A
+      extend: B
+      description: A extends B (which extends A) - an invalid cycle.
```
aklish, line 7: I thought there was already a test for this.

elide-model-config/src/test/resources/validator/infinite_loop_cyclic_dependency_inheritance/models/tables/tableB.hjson
```diff
@@ -0,0 +1,18 @@
+{
+  tables:
+  [
+    {
+      name: B
+      extend: A
+      description: B extends A (which extends B) - an invalid cycle.
```
aklish, line 7: Same comment.

elide-model-config/src/test/resources/validator/infinite_loop_missing_parent_inheritance/models/tables/tableA.hjson
```diff
@@ -0,0 +1,18 @@
+{
+  tables:
+  [
+    {
+      name: A
+      extend: B
+      description: A extends B (which doesn't exist)
```
aklish, line 7: Same comment.

elide-model-config/src/test/resources/validator/table_sql_both_inheritance/models/tables/player_stats_extends.hjson
```diff
@@ -0,0 +1,36 @@
+{
+  tables:
+  [
+    {
+      name: PlayerStatsChild
+      extend: PlayerStats
```
aklish, line 6: Don't we already have an extends test?  I'm confused why we needed to create so many new HJSON test files.

aklish (changes requested):

elide-model-config/src/test/resources/validator/table_sql_both_inheritance/models/tables/player_stats_extends.hjson
moizarafat: These tests are the same as those, but I wanted to have a test case specifically testing the infinite loop and issues related to that in case someone edits those validation methods and test files to work differently.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -252,56 +251,156 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
+        Set<Table> processed = new HashSet<Table>();
+        elideTableConfig.getTables().stream().forEach(table -> {
+            populateInheritance(table, processed);
+        });
+    }
 
-                Map<String, Join> joins = getInheritedJoins(table, new HashMap<String, Join>());
-                table.setJoins(new ArrayList<Join>(joins.values()));
+    private void populateInheritance(Table table, Set<Table> processed) {
+        if (table.hasParent() && !processed.contains(table)) {
```
aklish, line 264: Nitpick - but I would invert this if condition and return early if the opposite is true.  Then the entire function doesn't have to be nested inside this if block (it helps with line length and readability).

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -252,56 +251,156 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
+            Table parent = table.getParent(this.elideTableConfig);
+            // If parent also extends, ensure parent is processed first.
+            if (parent.hasParent()) {
+                populateInheritance(parent, processed);
             }
+
+            Map<String, Measure> measures = getInheritedMeasures(table, new HashMap<String, Measure>());
```
aklish, line 271: shouldn't this be called on the parent rather than table?  Similar comment for other calls below.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
```diff
@@ -252,56 +251,156 @@ private static void validateInheritance(ElideTableConfig tables, Table table, Se
...
+        }
+        return schema;
+    }
+
+    private String getInheritedConnection(Table table) {
+        Inheritance action = () -> {
+            return table.getDbConnectionName();
+        };
+
+        String conn = (String) action.inherit();
```
aklish, line 371: Lines 371-77 and below could be consolidated into a getInheritedProperty function that all the other property functions could leverage.

aklish, line 371: Do we really need to cast to String here?

aklish (changes requested):

elide-spring/elide-spring-boot-autoconfigure/pom.xml
```diff
@@ -209,6 +209,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
```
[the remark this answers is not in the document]
moizarafat: Earlier JUnit was pulled from an artifact within `spring-boot-starter-test`. But with latest version of `spring-boot-dependencies` that's not the case anymore.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidator.java
moizarafat: We first retrieve the measure list from current table and populate in the new HashMap in the `getInheritedMeasures` method. And in the same method we call it for `Parent` class too. Refer L-315.

aklish: But we shouldn't need to _add_ the measures for the current table (as they are already defined in the table).  We only need to add the inherited measures.  This reads like we are duplicating what's already defined in the current table.

aklish (changes requested):

aklish (approved):

aklish (approved):

PR #1709, opened 2020-12-04
Dynamic Model tests to AggregationDataStoreIntegrationTest

AvaniMakwana:
Moved common testcases for Dynamic Model Config from Spring & standalone

## License


AvaniMakwana:
Dynamic Model Inheritance tests will be added after https://github.com/yahoo/elide/pull/1707 is merged.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/DynamicConfigTest.java
```diff
@@ -1,113 +0,0 @@
...
-@Sql(executionPhase = Sql.ExecutionPhase.AFTER_TEST_METHOD,
-statements = "DROP TABLE PlayerStats; DROP TABLE PlayerCountry;")
-public class DynamicConfigTest extends IntegrationTest {
-    /**
-     * This test demonstrates an example test using the JSON-API DSL.
-     * @throws InterruptedException
-     */
-
-    @Test
-    public void jsonApiGetTest() {
```
moizarafat, line 50: We can move the filter test cases to Aggregation store IT test. But I think there may be some merit in running `DynamicConfigTest.jsonApiGetTest()` in Spring separately. It will help us in testing and making sure that any changes to bean configurations do not break Spring Applications. 

@aklish Thoughts?

moizarafat: Instead of a separate java file, may be it could just run as part of ControllerTest or AggregationStoreTest

aklish (approved):

PR #1736, opened 2020-12-24
Async cache setting

### LINKED ISSUE #1537: Elide 5: Add setting to turn on/off caching for async downloads.
Elide async needs a configuration setting to disable caching for queries that involve large downloads.  Some caching strategies (like in-memory caffeine cache) will not work with large results.

AvaniMakwana:
Resolves #1537  (if appropriate)

## Description
Add 'bypassCache' property to Async to enable/disable caching.

## Motivation and Context
Enable / Disable Caching via Async properties

## How Has This Been Tested?
Unit tests included.

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
@@ -19,7 +19,10 @@ import org.apache.http.client.utils.URIBuilder;
 import lombok.extern.slf4j.Slf4j;
 
 import java.net.URISyntaxException;
+import java.util.Arrays;
 import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.Callable;
@@ -37,6 +40,7 @@ public class AsyncQueryThread implements Callable<AsyncAPIResult> {
     private AsyncExecutorService service;
     private String apiVersion;
 
+
     public AsyncQueryThread(AsyncAPI queryObj, User user, AsyncExecutorService service, String apiVersion) {
         this.queryObj = queryObj;
         this.user = user;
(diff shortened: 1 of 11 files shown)


elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideStandaloneAsyncSettings.java
```diff
@@ -95,4 +95,13 @@ default AsyncAPIDAO getAPIDAO() {
...
         return null;
     }
+
+    /**
+     * Bypass Query result caching.
+     *
+     * @return Default: true
+     */
+    default boolean bypassCache() {
+        return true;
```
moizarafat, line 105: Default should be false.

moizarafat (approved):

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -49,10 +53,13 @@ public AsyncAPIResult call() throws URISyntaxException, NoHttpResponseException
         ElideResponse response = null;
         log.debug("AsyncQuery Object from request: {}", this);
         UUID requestUUID = UUID.fromString(queryObj.getRequestId());
+        Map<String, List<String>> bypassCacheMap = new HashMap<String, List<String>>();
+        bypassCacheMap.put("bypasscache", Arrays.asList(String.valueOf(service.isBypassCache())));
```
aklish, line 57: Won't this always override the cache?  I don't think this is the behavior that we want.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncQueryThread.java
```diff
@@ -49,10 +53,13 @@ public AsyncAPIResult call() throws URISyntaxException, NoHttpResponseException
...
+
         if (queryObj.getQueryType().equals(QueryType.JSONAPI_V1_0)) {
-            response = executeJsonApiRequest(service.getElide(), user, apiVersion, requestUUID);
+            response = executeJsonApiRequest(service.getElide(), user, apiVersion, requestUUID, bypassCacheMap);
```
aklish, line 60: Instead of naming the variable bypassCacheMap, call it headers.  It is a better reflection of what the map represents.

aklish (changes requested):

AvaniMakwana:
Closing since we are going to enable / disable caching based on Max cache size property, not on/off property

PR #1739, opened 2020-12-28
ISO Support for Time Grains

### LINKED ISSUE #1545: Support ISO Date Formats in Time Grains
Support ISO Date Formats as input String for Time Grains

## Expected Behavior
Input String format can be the same as the Time Grains input format and ISO Date Format. Elide should be able to parse both. Also for Weekly Time Grain, It should auto-convert to Monday.

-->

## Current Behavior
The input string has to match the format of the time grain. For Ex. Monthly Grain expects YYYY-MM etc.

AvaniMakwana:
Resolves #1545 

## Description
ISO Support for time grains 

## How Has This Been Tested?
Unit tests included

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/Day.java
```diff
@@ -34,10 +37,13 @@ public Day deserialize(Object val) {
...
-                    date = new Day(new Timestamp(FORMATTER.parse((String) val).getTime()));
-                } else {
-                    date = new Day(FORMATTER.parse(FORMATTER.format(val)));
+                    try {
+                        date = new Day(new Timestamp(FORMATTER.parse((String) val).getTime()));
+                    } catch (ParseException pe) {
+                        date = new Day(new Timestamp(ISO_FORMATTER.parse((String) val).getTime()));
+                    }
                 }
+                date = new Day(FORMATTER.parse(FORMATTER.format(val)));
```
moizarafat, line 46: This should be inside an else block otherwise it will get executed even when val is instanceof string

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/Day.java
```diff
@@ -9,14 +9,13 @@
...
 import java.sql.Date;
-import java.sql.Timestamp;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 
 /**
  * Time Grain class for Day.
  */
-public class Day extends Date {
+public class Day extends Date implements TimeGrainFormatter {
```
aklish, line 18: Why is a Day an instance of TimeGrainFormatter?  Can we remove this association?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/Day.java
```diff
@@ -34,14 +33,13 @@ public Day deserialize(Object val) {
 
             try {
                 if (val instanceof String) {
-                    date = new Day(new Timestamp(FORMATTER.parse((String) val).getTime()));
-                } else {
-                    date = new Day(FORMATTER.parse(FORMATTER.format(val)));
+                    date = new Day(TimeGrainFormatter.formatDateString(FORMATTER, (String) val));
                 }
+                date = new Day(FORMATTER.parse(FORMATTER.format(val)));
```
aklish, line 38: Don't we still need the else case?  Otherwise this line stomps on line 36.

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/Day.java
moizarafat: TimeGrainFormatter has common logic that will be used by all the Grains for ISO handling. So instead of creating a  Util class with that supporting method, we went the implement/interface route.

moizarafat: maybe DaySerde can implement it instead of Day.

aklish: We are calling a static method on the interface.  Technically nothing needs to implement it.  DaySerde would make more sense though if we do.

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/Day.java
AvaniMakwana: This line will catch ISO formatted date object/ dateString. It will ensure the output is always 'yyyy-MM-dd' format.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/TimeGrainFormatter.java
```diff
@@ -0,0 +1,25 @@
...
+package com.yahoo.elide.datastores.aggregation.timegrains;
+
+import java.sql.Timestamp;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+
+/** Interface for ISO timegrain Support. */
+public interface TimeGrainFormatter {
+
+    String ISO_FORMAT = "yyyy-MM-dd'T'HH:mm:ssZ";
```
moizarafat, line 15: It should be `"yyyy-MM-dd'T'HH:mm:ss'Z'"`, single quotes around `Z`

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/timegrains/serde/DaySerdeTest.java
```diff
@@ -17,6 +18,7 @@
 
 public class DaySerdeTest {
     private SimpleDateFormat formatter = new SimpleDateFormat("yyyy-MM-dd");
+    private SimpleDateFormat isoFormatter = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ssZ");
```
moizarafat, line 21: Does not look like isoFormatter is used anywhere in the class.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/timegrains/serde/DaySerdeTest.java
```diff
@@ -59,4 +61,14 @@ public void testDeserializeDateInvalidFormat() throws ParseException {
             serde.deserialize(dateInString);
         });
     }
+
+    @Test
+    public void testISODateString() throws ParseException {
+        String dateInString = "2020-01-01T00:00:00-0500";
+        Day expectedDate = new Day(formatter.parse(dateInString));
```
moizarafat, line 68: This should be isoFormatter instead of formatter. Also might be a good idea to update dateInString to have different hour minutes.

PR #1741, opened 2020-12-30
Closes #1740 More meaningful handling of constraint violations

### LINKED ISSUE #1740: More meaningful handling of constraint violations
Right now, Elide is not leveraging possibility of converting multiple constraint violations into list of response errors. It takes first occurrence and prints only message, which is not really helpful most of the time without information about which property is problematic

## Expected Behavior
In case of multiple constraint violations, they are all present in response. In error response there is also property path exposed
```
{
  "errors": [
    {
      "detail": "must not be null",
      "source": {
        "property": "reportStatus"
      }
    },
    {
      "detail": "must not be null",
      "source": {
        "property": "reporter"
      }
    }
  ]
}
```

## Current Behavior
Currently we get this response for the same situation as above, which is not really helpful
```
{
  "errors": [
    {
      "detail": "must not be null"
    }
  ]
}
```

## Possible Solution
Quick fix could be leveraging of CustomErrorException when converting ConstraintViolationException

## Your Environment
* Elide version used: 5.0.0-pr29 but I think it applies to all versions
* Environment name and version (Java 1.8.0_152):
* Operating System and version:
* Link to your project:

bukajsytlos:
Resolves #1740

## Description
Convert ConstraintViolationException into CustomErrorException

## Motivation and Context
see issue

## How Has This Been Tested?
Manually :(

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -12,11 +12,12 @@ import com.yahoo.elide.core.datastore.DataStore;
 import com.yahoo.elide.core.datastore.DataStoreTransaction;
 import com.yahoo.elide.core.datastore.inmemory.InMemoryDataStore;
 import com.yahoo.elide.core.dictionary.Injector;
+import com.yahoo.elide.core.exceptions.CustomErrorException;
+import com.yahoo.elide.core.exceptions.ErrorObjects;
 import com.yahoo.elide.core.exceptions.ForbiddenAccessException;
 import com.yahoo.elide.core.exceptions.HttpStatus;
 import com.yahoo.elide.core.exceptions.HttpStatusException;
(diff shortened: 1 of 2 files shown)


aklish (dismissed):

aklish:
Code looks good and a nice improvement.  The screwdriver build is broken because some dependencies need to be upgraded for OWASP security check.  I can look at that when I get back from break in a few days or alternatively you can fix them on this branch.  

Happy new year.

bukajsytlos:
thanks, happy new year to you too.
I think there's a cve for hibernate 3 but elide is using latest version, so I don't know what to do with it.

aklish:
@bukajsytlos You can add a suppression here for 6 months or so:

https://github.com/yahoo/elide/blob/master/suppressions.xml

aklish (approved):

PR #1747, opened 2021-01-06
Initial Table Export Thread

moizarafat:
Co-authored-by: chandrakaran

## Description
Initial Code for TableExportThread

## Motivation and Context
Table Export Support

## How Has This Been Tested?
Existing Test case pass and new test case were added.

## License


elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -0,0 +1,42 @@
...
+            Optional<ChangeSpec> changes) {
+    	AsyncExecutorService service = getAsyncExecutorService();
+    	TableExporter exporter =new TableExporter(service.getElide(), requestScope.getApiVersion(), requestScope.getUser());
+        TableExportThread queryWorker = new TableExportThread(query, service.getResultStorageEngine(), exporter);
+        executeHook(operation, phase, query, requestScope, queryWorker);
+    }
+
+    @Override
+    public void validateOptions(AsyncAPI query, RequestScope requestScope) {
+        super.validateOptions(query, requestScope);
```
moizarafat, line 40: Will add additional validation in next PR.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
```diff
@@ -0,0 +1,152 @@
...
+
+            JFlat flat = new JFlat(jsonStr);
+
+            json2Csv = flat.json2Sheet().headerSeparator("_").getJsonAsSheet();
+
+            int index = 0;
+
+            for (Object[] obj : json2Csv) {
+                // Skip Header record from 2nd time onwards.
+                if (index++ == 0 && downloadRecordCount != 0) {
```
aklish, line 123: Why do we check both index and downloadRecordCount.  Isn't the latter sufficient?

elide-async/src/test/java/com/yahoo/elide/async/service/thread/TableExportThreadTest.java
```diff
@@ -0,0 +1,61 @@
...
+        queryObj.setQuery(query);
+        queryObj.setQueryType(QueryType.GRAPHQL_V1_0);
+        queryObj.setResultType(ResultType.CSV);
+
+        when(exporter.export(any())).thenReturn(Observable.empty());
+        when(engine.storeResults(any(), any())).thenReturn(queryObj);
+
+        TableExportThread queryThread = new TableExportThread(queryObj, engine, exporter);
+        TableExportResult queryResultObj = (TableExportResult) queryThread.call();
+        assertEquals(queryResultObj.getHttpStatus(), 200);
```
aklish, line 56: We should also test what the body looks like for both JSON-API and CSV.

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
moizarafat: convertToCSV is called once for each PersistentResource in the observable. json2Csv always has 2 entries. 0th index is the header so we need to skip the header from 2nd time this method is called. That's why we need to use a combination of index and downloadRecordCount.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
```diff
@@ -0,0 +1,152 @@
...
+            results =  resources.map(resource -> convertToCSV(resource));
+        } else if (queryObj.getResultType() == ResultType.JSON) {
+            results = resources.map(resource -> resourceToJsonStr(resource));
+        }
+        return results;
+    }
+
+    private String resourceToJsonStr(PersistentResource resource) throws IOException {
+        incrementRecordCount();
+        //TODO Additional logic to create an array of records???
```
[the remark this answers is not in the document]
moizarafat: Sure. Will include this in the next PR to avoid increasing the PR size.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
```diff
@@ -0,0 +1,152 @@
...
+
+            for (Object[] obj : json2Csv) {
+                // Skip Header record from 2nd time onwards.
+                if (index++ == 0 && downloadRecordCount != 0) {
+                    continue;
+                }
+
+                String objString = Arrays.toString(obj);
+                if (objString != null) {
+                    objString = objString.substring(1, objString.length() - 1);
```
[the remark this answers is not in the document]
moizarafat: The arrays.toString returns o/p with [ and ] at the beginning and end. So need to exclude those 2 chars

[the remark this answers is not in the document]
moizarafat: For example: 
["edc4a871-dff2-4054-804e-d80075cf827d", "/group?sort=commonName&fields%5Bgroup%5D=commonName,description", "GRAPHQL_V1_0", "1300fa60-0bda-454c-882a-5f1f1861485d", null, "QUEUED", 1.610058468658E12, 1.610058468658E12, 10.0, "CSV", null]

elide-async/src/test/java/com/yahoo/elide/async/service/thread/TableExportThreadTest.java
moizarafat: Added test cases to verify how outputs will be for JSON and CSV.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
```diff
@@ -0,0 +1,157 @@
...
+            JFlat flat = new JFlat(jsonStr);
+
+            json2Csv = flat.json2Sheet().headerSeparator("_").getJsonAsSheet();
+
+            int index = 0;
+
+            for (Object[] obj : json2Csv) {
+                // Skip Header record from 2nd time onwards.
+                // TODO Flag to Skip Headers
+                if (index++ == 0 && downloadRecordCount != 1) {
```
moizarafat, line 129: aklish
Why do we check both index and downloadRecordCount. Isn't the latter sufficient?

 
moizarafat
convertToCSV is called once for each PersistentResource in the observable. json2Csv always has 2 entries. 0th index is the header so we need to skip the header from 2nd time this method is called. That's why we need to use a combination of index and downloadRecordCount.

aklish: We need to add more comments here then.  Variables like index don't have enough description to understand that nuance.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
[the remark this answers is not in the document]
aklish: Again - add a comment that we are stripping off those brackets.

elide-async/src/test/java/com/yahoo/elide/async/service/thread/TableExportThreadTest.java
```diff
@@ -0,0 +1,106 @@
...
+        assertEquals(queryResultObj.getRecordCount(), 0);
+    }
+
+    @Test
+    public void testJsonToCSV() {
+        String row1 = "id, query, queryType, requestId, principalName, status, createdOn, updatedOn, asyncAfterSeconds, resultType, result";
+        String row2 = "\"edc4a871-dff2-4054-804e-d80075cf827d\", \"/group?sort=commonName&fields%5Bgroup%5D=commonName,description\", \"GRAPHQL_V1_0\"";
```
aklish, line 65: It is a little confusing that the rows don't match the query.  Can we make them consistent?

aklish (changes requested):

aklish (approved):

PR #1761, opened 2021-01-11
add druid dialect with new pagination query structure

Chandrasekar-Rajasekar:
## Description
Create Druid dialect in aggregation datastore. 

## How Has This Been Tested?
unit test

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
@@ -240,6 +240,12 @@ public class SQLQueryEngine extends QueryEngine {
         SQLDialect dialect = details.getDialect();
         String paginationSQL = toPageTotalSQL(query, sql, dialect).toString();
 
+        if (paginationSQL == null) {
+            // The query returns the aggregated metric without any dimension.
+            // Only 1 record will be returned.
+            return 1;
+        }
+
         NamedParamPreparedStatement stmt = sqlTransaction.initializeStatement(paginationSQL, dataSource);
 
         // Supply the query parameters to the query
@@ -297,7 +303,8 @@ public class SQLQueryEngine extends QueryEngine {
 
         Pagination pagination = query.getPagination();
         if (returnPageTotals(pagination)) {
-            queries.add(toPageTotalSQL(expandedQuery, sql, dialect).toString());
+            SQLQuery paginationSql = toPageTotalSQL(expandedQuery, sql, dialect);
+            queries.add(paginationSql == null ? "" : paginationSql.toString());
         }
         queries.add(sql.toString());
         return queries;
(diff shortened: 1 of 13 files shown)


aklish (approved):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -398,15 +398,20 @@ private SQLQuery toPageTotalSQL(Query query, SQLQuery sql, SQLDialect sqlDialect
...
-        String projectionClause = sqlDialect.generateCountDistinctClause(groupByDimensions);
-
-        return SQLQuery.builder()
-                .projectionClause(projectionClause)
+        SQLQuery innerQuery =  SQLQuery.builder()
+                .projectionClause(groupByDimensions)
                 .fromClause(sql.getFromClause())
                 .joinClause(sql.getJoinClause())
                 .whereClause(sql.getWhereClause())
+                .groupByClause(String.format("GROUP BY %s", groupByDimensions))
```
aklish, line 406: Is there any reason we can't simply do:
.groupByClause(sql.getGroupByClause)
and avoid this extra bit of code here?
What if there are no group by dimensions?

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -389,21 +387,12 @@ private void supplyFilterQueryParameters(Query query, NamedParamPreparedStatemen
...
-        String groupByDimensions =
-                query.getAllDimensionProjections()
-                        .stream()
-                        .map(SQLColumnProjection.class::cast)
-                        .map((column) -> column.toSQL(queryReferenceTable))
-                        .collect(Collectors.joining(", "));
-
         SQLQuery innerQuery =  SQLQuery.builder()
-                .projectionClause(groupByDimensions)
+                .projectionClause(sql.getProjectionClause())
```
aklish, line 391: Thinking about this, we ought to remove the metrics if possible or replace them with count or something simple.  Imagine if the metric was a data sketch aggregation.  We don't want to do some expensive operation.

Chandrasekar-Rajasekar:
Couldn't use `groupByClause(sql.getGroupByClause)` for pagination because the main SQL query will not contain group by when a user queries the only dimension and will be in the form 
```
select distinct dims from table 
```
We are generating group by queries for the above cases in pagination.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
```diff
@@ -297,7 +303,8 @@ public String getTableVersion(Table table, Transaction transaction) {
 
         Pagination pagination = query.getPagination();
         if (returnPageTotals(pagination)) {
-            queries.add(toPageTotalSQL(expandedQuery, sql, dialect).toString());
+            SQLQuery paginationSql = toPageTotalSQL(expandedQuery, sql, dialect);
+            queries.add(paginationSql == null ? "" : paginationSql.toString());
```
aklish, line 307: We don't want to add a query here if we are not running one.  If we return "1" as the rowcount and are not running a query, we'll want to skip adding to the list of queries.

aklish (changes requested):

aklish (approved):

PR #1766, opened 2021-01-12
Table Export JSON Array + URL Generate + Generate Header Empty CSV

moizarafat:
Co-authored-by: chandrakaran

## Description
Table Export
i) JSON Array Output Generation. 
ii) Download URL Generation logic.
iii) Generate Header when no CSV results found.

## Motivation and Context
Table Export functionality

## How Has This Been Tested?
Existing Test case pass. New Test cases added.

## License


elide-standalone/src/main/java/com/yahoo/elide/standalone/config/ElideResourceConfig.java
```diff
@@ -142,8 +142,9 @@ protected void configure() {
 
                     // TODO: If null, initialize with FileResultStorageEngine
                     ResultStorageEngine resultStorageEngine = asyncProperties.getResultStorageEngine();
+                    // TODO: Add logic to generate baseURLPresent Flag
```
moizarafat, line 145: Will do separate PR for Standalone to set baseURLPresent flag and binding of Table Export Hooks.

elide-async/src/main/java/com/yahoo/elide/async/export/TableExporter.java
```diff
@@ -26,44 +26,39 @@
 
 /**
  * Class for Table Export functionality.
+ * Parses a TableExport request and returns an Observable of PersistentResource.
  */
 @Slf4j
 public class TableExporter {
 
     private Elide elide;
```
moizarafat, line 34: AsyncExecutorService seems to be a better place to keep the downloadURL instead of TableExporter. TableExporter will not have any use of downloadURL.

elide-async/src/main/java/com/yahoo/elide/async/export/TableExporter.java
```diff
@@ -26,44 +26,39 @@
...
         try (DataStoreTransaction tx = elide.getDataStore().beginTransaction()) {
             elide.getTransactionRegistry().addRunningTransaction(requestId, tx);
 
-            EntityProjection projection = null;
             RequestScope requestScope = null;
 
+            TableExportParser parser = getParser(query);
+
+            EntityProjection projection = parser.parse(query, apiVersion);
+
```
aklish, line 60: The TableExportParser hides away the differences for different protocols (JSON-API, GraphQL, etc).  However, we still are check the queryType below to build the request scope.

We might want a different abstraction here.  Maybe the parser should return a function that runs the query - rather than a projection - so we can move together all the query type specific logic.

elide-async/src/main/java/com/yahoo/elide/async/export/TableExporter.java
```diff
@@ -99,4 +94,16 @@ public TableExporter(Elide elide, String apiVersion, User user, GraphQLParser gr
 
         return results;
     }
+
+    public TableExportParser getParser(TableExport query) {
```
aklish, line 98: Maybe instead of returning a parser, this should return a load function which returns an Observable of records.

elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -15,29 +15,69 @@
...
 
 /**
  * LifeCycle Hook for execution of TableExpoer.
  */
+@Slf4j
 public class TableExportHook extends AsyncAPIHook<TableExport> {
+    private boolean baseURLPresent = true;
 
-    public TableExportHook (AsyncExecutorService asyncExecutorService, Integer maxAsyncAfterSeconds) {
+    public TableExportHook (AsyncExecutorService asyncExecutorService, Integer maxAsyncAfterSeconds,
```
aklish, line 32: baseUrlPresent is an odd contract.  Also, it doesn't really belong in the hook.

baseUrl can be extracted from the requestScope.elideSettings.

Constructing the downloadURL should be a utility function that is shared with the export controller and this class.

elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -15,29 +15,69 @@
...
         AsyncExecutorService service = getAsyncExecutorService();
-        TableExporter exporter = new TableExporter(service.getElide(), requestScope.getApiVersion(),
-                requestScope.getUser());
-        TableExportThread queryWorker = new TableExportThread(query, service.getResultStorageEngine(), exporter);
+
+        // TODO: Make it a Bean
+        TableExporter exporter = new TableExporter(service.getElide());
+
+        String downloadURL = retrieveDownloadURL(requestScope.getBaseUrlEndPoint());
+        TableExportThread queryWorker = new TableExportThread(query, exporter, downloadURL,
```
aklish, line 47: The AsyncExecutorService should not have a function called isSkipCSVHeader.  That is specific configuration for TableExport.

elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -15,29 +15,69 @@
...
+                service.isSkipCSVHeader());
         executeHook(operation, phase, query, requestScope, queryWorker);
     }
 
     @Override
     public void validateOptions(AsyncAPI query, RequestScope requestScope) {
         super.validateOptions(query, requestScope);
     }
+
+    private String retrieveDownloadURL(String baseURL) {
```
aklish, line 58: Let's move this logic.

elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -15,29 +15,69 @@
...
+            downloadURL = getAsyncExecutorService().getDownloadURL();
+        } else {
+            downloadURL = getBasePath(baseURL) + getAsyncExecutorService().getDownloadURL() + "/";
+        }
+
+        return downloadURL;
+    }
+
+    private String getBasePath(String URL) {
+        URIBuilder uri;
```
aklish, line 70: Same comment - this doesn't belong in the hook.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
```diff
@@ -50,6 +50,9 @@
     private static AsyncExecutorService asyncExecutorService = null;
     private ResultStorageEngine resultStorageEngine;
     private ThreadLocal<AsyncAPIResultFuture> asyncResultFutureThreadLocal = new ThreadLocal<>();
+    private String downloadURL;
+    // TODO Update constructor to populate these.
+    private boolean skipCSVHeader;
```
aklish, line 55: Neither of these new variables belong in the generic AsyncExecutorService.

elide-async/src/main/java/com/yahoo/elide/async/service/AsyncExecutorService.java
```diff
@@ -84,10 +88,10 @@ private AsyncExecutorService(Elide elide, Integer threadPoolSize, AsyncAPIDAO as
      * @param asyncAPIDao DAO Object
      */
     public static void init(Elide elide, Integer threadPoolSize, AsyncAPIDAO asyncAPIDao,
-            ResultStorageEngine resultStorageEngine) {
+            ResultStorageEngine resultStorageEngine, String downloadURL) {
```
aklish, line 91: The async executor shouldn't need the downloadURL.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/AsyncQueryThread.java
```diff
@@ -62,10 +69,31 @@ public AsyncAPIResult call() throws URISyntaxException, NoHttpResponseException
...
+     * @param response is the ElideResponse
+     * @param queryType is the query type (GraphQL or JSON).
+     * @return Record Count
+     * @throws IOException Exception thrown by JsonPath
+     */
+    protected Integer calculateRecordCount(ElideResponse response, QueryType queryType)
+            throws IOException {
+        Integer count = null;
+        if (response.getResponseCode() == 200) {
+            if (queryType.equals(QueryType.GRAPHQL_V1_0)) {
```
aklish, line 87: The code has these query type checks sprinkled in many different places.  
This suggests to me that we ought to have an interface (and concrete functions) with a set of functions that we need each query type to perform.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
```diff
@@ -36,33 +41,68 @@
  */
 @Slf4j
 public class TableExportThread implements Callable<AsyncAPIResult> {
+    private static final String FORWARD_SLASH = "/";
```
aklish, line 44: Why do we have both TableExportThread and AsyncQuerythread?  Do we need both concepts?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/AsyncProperties.java
```diff
@@ -54,4 +54,9 @@
      * Whether or not the async feature is enabled.
      */
     private boolean enabled = false;
+
+    /**
+     * Settings for the export controller.
+     */
+    private ControllerProperties export = new ControllerProperties();
```
aklish, line 61: There shouldn't be a call to new here.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -50,8 +51,12 @@
     public AsyncExecutorService buildAsyncExecutorService(Elide elide, ElideConfigProperties settings,
             AsyncAPIDAO asyncQueryDao, EntityDictionary dictionary,
             @Autowired(required = false) ResultStorageEngine resultStorageEngine) {
+        boolean baseURLPresent = settings.getBaseUrl() != null && !settings.getBaseUrl().isEmpty();
+        String downloadURL = baseURLPresent ? settings.getBaseUrl() + settings.getAsync().getExport().getPath()
```
aklish, line 55: This doesn't really mirror how we initialized JSONApiLinks.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
```diff
@@ -36,33 +41,68 @@
...
-        //TODO Add URL generation logic and set
-        //queryResult.setUrl(new URL("url"));
+        queryResult.setUrl(generateDownloadURL());
         queryResult.setRecordCount(downloadRecordCount);
         return queryResult;
     }
 
+    private URL generateDownloadURL() {
+        log.debug("generateDownloadUrl");
+        String asyncQueryID = queryObj.getId();
```
aklish, line 92: I think there are three places currently where we attempt to generate the downloadURL.  We need to consolidate this to one.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
```diff
@@ -157,4 +210,33 @@ protected String convertToCSV(PersistentResource resource) {
...
         this.downloadRecordCount++;
     }
+
+    /**
+     * Generate CSV Header when Observable is Empty.
+     * @param parser TableExportParser object.
+     * @return returns Header string which is in CSV format.
+     */
+    protected Observable<String> generateCSVHeader(TableExportParser parser) {
+        EntityProjection projection = parser.parse(queryObj, apiVersion);
```
aklish, line 220: Is this the 2nd time we construct the entity projection?  Ideally we should do this once.

elide-async/src/main/java/com/yahoo/elide/async/service/thread/TableExportThread.java
```diff
@@ -78,25 +118,39 @@ protected TableExport storeResults(Observable<String> result) {
      * @return result as Observable of String
      */
     protected Observable<String> processObservablePersistentResource(Observable<PersistentResource> resources) {
-        Observable<String> results = Observable.just("No Records Generated");
+        Observable<String> results = Observable.empty();
```
aklish, line 121: We should generate the CSV header using the same code regardless of what the record count is.

aklish (changes requested):

PR #1768, opened 2021-01-15
Elide 5 - Support new Type system

aklish:
## Description
This PR creates a parallel type system to Java's class, method, field, etc.  Elide 5 migrates to use this new parallel type system rather than native type system.

## Motivation and Context
Some data stores will allow the creation of new types on the fly (while the service is running and not simply at startup).  Elide's type system relies on Java's type system which is hardwired into class compilation and cannot be extended through inheritance.

## How Has This Been Tested?
All existing tests pass.

## License


elide-core/src/main/java/com/yahoo/elide/core/datastore/DataStoreTransaction.java
```diff
@@ -93,15 +94,15 @@ default void preCommit(RequestScope scope) {
      * @param <T> the class to create
      * @return a new instance of type T
      */
-    default <T> T createNewObject(Class<T> entityClass) {
+     default <T> T createNewObject(Type<T> entityClass) {
```
moizarafat, line 97: Indentation is off here.

elide-core/src/main/java/com/yahoo/elide/core/datastore/DataStoreTransaction.java
```diff
@@ -93,15 +94,15 @@ default void preCommit(RequestScope scope) {
...
         T obj;
         try {
             obj = entityClass.newInstance();
         } catch (java.lang.InstantiationException | IllegalAccessException e) {
             obj = null;
         }
         return obj;
-    }
+     }
```
moizarafat, line 105: indentation is off here too.

rishi-aga (dismissed):

PR #1772, opened 2021-01-19
Remove Path while creating Base Urls in Controllers/Resources

moizarafat:
## Description
Remove Path while creating Base Urls in Controllers/Resources

## Motivation and Context
Base URL should just conrain Base URL Components

## How Has This Been Tested?
Existing Tests Pass

## License


aklish (changes requested):
We need to also consider DefaultJSONApiLinks.  There is logic that falls back on the request URI if baseURL is not set.

moizarafat:
> We need to also consider DefaultJSONApiLinks. There is logic that falls back on the request URI if baseURL is not set.

Fixed.

aklish (approved):

PR #1775, opened 2021-01-20
Async GraphQL check

AvaniMakwana:
## Description
Validate if GraphQL is enable before Async query execution. Throw Exception in case GraphQL is disabled and Async query is of type GraphQL.

## Motivation and Context
Validate if GraphQL is enable before Async query execution

## How Has This Been Tested?
Unit test included

## License


elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncQueryHook.java
```diff
@@ -41,6 +41,9 @@ public void validateOptions(AsyncAPI query, RequestScope requestScope) {
         super.validateOptions(query, requestScope);
 
         if (query.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
+            if (!getAsyncExecutorService().isEnableGraphQL()) {
+                throw new InvalidOperationException("GraphQL is disabled. Please enable GraphQL in settings.");
```
aklish, line 45: Can we add one test to ensure this path is reached?

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/DisableGraphQLAsyncTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static io.restassured.RestAssured.given;
+import static org.hamcrest.Matchers.equalTo;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.test.context.ActiveProfiles;
+
+/**
+ * Executes Async tests with Aggregation Store disabled.
+ */
+@ActiveProfiles("disableGraphQL")
```
aklish, line 24: Rather than create an entire application.yaml for this one test, you can set individual settings for tests using code like:

```java
@TestPropertySource(
        properties = {
                "elide.json-api.enableLinks=true"
        }
)
```

elide-standalone/src/test/java/example/ElideStandaloneDisableGraphQLTest.java
```diff
@@ -0,0 +1,207 @@
...
+public class ElideStandaloneDisableGraphQLTest {
+
+    protected ElideStandalone elide;
+
+    @BeforeAll
+    public void init() throws Exception {
+        elide = new ElideStandalone(new ElideStandaloneSettings() {
+
+            @Override
+            public ElideSettings getElideSettings(EntityDictionary dictionary, DataStore dataStore) {
```
aklish, line 50: Can we minimize all of this configuration to as little as possible to simplify the test?

elide-standalone/src/test/java/example/ElideStandaloneDisableGraphQLTest.java
```diff
@@ -0,0 +1,207 @@
...
+                                attributes(
+                                        attr("query", "/post"),
+                                        attr("queryType", "GRAPHQL_V1_0"),
+                                        attr("status", "QUEUED")
+                                        )
+                                )
+                        ).toJSON())
+        .when()
+        .post("/api/v1/asyncQuery");
+//        .then()
```
aklish, line 204: Should this be removed or uncommented?

PR #1776, opened 2021-01-20
Issue698

### LINKED ISSUE #698: DataStoreTransaction Generics
It would be handy if the DataStoreTransaction interface supported generics instead of simply returning, and operating on, Object types.  I've made a number of custom DataStoreTransaction implementations that only operate on specific types.  I'd like to restrict the method signatures to only operate on my expected types, but I can't do that because the interface only applies to the Object type.

aklish:
Resolves #698

## Description
Updates DataStoreTransaction interface to replace Object with Java generics.

## License


elide-core/src/test/java/com/yahoo/elide/core/datastore/inmemory/InMemoryStoreTransactionTest.java
```diff
@@ -142,6 +142,7 @@ public void testFullFilterPredicatePushDown() {
 
         when(wrappedTransaction.supportsFiltering(eq(scope), any(), eq(projection))).thenReturn(DataStoreTransaction.FeatureSupport.FULL);
         when(wrappedTransaction.loadObjects(eq(projection), eq(scope))).thenReturn(books);
+        when(wrappedTransaction.loadObjects(eq(projection), eq(scope))).thenReturn(books);
```
rishi-aga, line 145: Seems duplicate line.

rishi-aga (approved):

PR #1787, opened 2021-01-22
Async refactor: Part 2 - Async Query Classes Refactoring

moizarafat:
## Description
Refactoring AsyncQuery and its supporting classes as per the new interfaces/abstract class.
 
## Motivation and Context
Refactoring for the new contract.

## How Has This Been Tested?
Existing test cases pass.

## License


elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncQueryOperation.java
```diff
@@ -21,4 +31,14 @@
...
     public abstract Integer calculateRecordCount(AsyncQuery queryObj, ElideResponse response);
 
+    /**
+     * Check if Elide Response is NULL.
+     * @param response ElideResponse object.
+     * @throws NoHttpResponseException NoHttpResponseException Exception.
+     */
+    public void nullResponseCheck(ElideResponse response) throws NoHttpResponseException {
+        if (response == null) {
+            throw new NoHttpResponseException("Response for request returned as null");
```
aklish, line 41: We usually don't throw library exceptions this deep in Elide.  What does the server generate here?

elide-async/src/main/java/com/yahoo/elide/async/operation/GraphQLAsyncQueryOperation.java
```diff
@@ -0,0 +1,81 @@
...
+@Slf4j
+public class GraphQLAsyncQueryOperation extends AsyncQueryOperation {
+
+    public GraphQLAsyncQueryOperation(AsyncExecutorService service) {
+        super(service);
+    }
+
+    @Override
+    public AsyncAPIResult execute(AsyncAPI queryObj, RequestScope scope) {
+        ElideResponse response = null;
```
aklish, line 42: This block is repeated for GraphQL and JSON-API.

elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncQueryOperation.java
aklish: Since this is just caught and then rethrown as IllegalStateException, is there any reason not to throw that directly here?    What do you expect the service to return to the end user in this scenario?

elide-async/src/main/java/com/yahoo/elide/async/service/thread/AsyncAPICallable.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.net.URISyntaxException;
+import java.util.concurrent.Callable;
+
+/**
+ * AsyncAPI implementation of Callable for executing the query provided in AsyncQuery.
+ * It will also update the query status and result object at different stages of execution.
+ */
+@Slf4j
+public class AsyncAPICallable implements Callable<AsyncAPIResult> {
```
aklish, line 23: Maybe we should merge this with the Operation (make the operation extend callable) since it doesn't do anything but call the operation.

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncAPIOperation.java
```diff
@@ -7,19 +7,12 @@
...
-public interface AsyncAPIOperation<T extends AsyncAPI> {
-
-    /**
-     * Execute the AsyncAPI request.
-     * @param queryObj AsyncAPI type object.
-     * @param scope RequestScope object.
-     * @return AsyncAPIResult object
-     */
-    public abstract AsyncAPIResult execute(AsyncAPI queryObj, RequestScope scope);
+public interface AsyncAPIOperation<T extends AsyncAPI> extends Callable<AsyncAPIResult> {
```
aklish, line 17: Can we remove this now as it just wraps Callable?

elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncQueryOperation.java
```diff
@@ -6,21 +6,51 @@
...
 
-    public AsyncQueryOperation(AsyncExecutorService service) {
+    public AsyncQueryOperation(AsyncExecutorService service, AsyncAPI queryObj, RequestScope scope) {
         this.service = service;
+        this.queryObj = queryObj;
+        this.scope = scope;
+    }
+
+    @Override
+    public AsyncAPIResult call() throws URISyntaxException, NoHttpResponseException {
```
aklish, line 41: We don't want to propagate these kinds of exceptions up the stack (it propagates a package dependency).

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncAPIHook.java
```diff
@@ -110,4 +110,12 @@ protected void updatePrincipalName(AsyncAPI query, RequestScope requestScope) {
             query.setPrincipalName(principal.getName());
         }
     }
+
+    /**
+     * Get AsyncAPIOperation to use.
```
aklish, line 115: Change the description here to match the return type.

elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncQueryCallableOperation.java
```diff
@@ -0,0 +1,83 @@
...
+    public AsyncQueryCallableOperation(AsyncExecutorService service, AsyncAPI queryObj, RequestScope scope) {
+        this.service = service;
+        this.queryObj = (AsyncQuery) queryObj;
+        this.scope = scope;
+    }
+
+    @Override
+    public AsyncAPIResult call() throws URISyntaxException {
+        ElideResponse response = null;
+        log.debug("AsyncQuery Object from request: {}", this);
```
aklish, line 42: I don't see toString implemented on this class.  I assume this will just log the object reference.

aklish (changes requested):

aklish (approved):

PR #1788, opened 2021-01-23
Async refactor: Part 3 - TableExport Classes Refactoring

moizarafat:
## Description
Refactoring TableExport and its supporting classes as per the new interfaces/abstract class.

## Motivation and Context
Refactoring for the new contract.

## How Has This Been Tested?
Existing Test Cases pass

## License


elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportCallableOperation.java
```diff
@@ -6,51 +6,193 @@
...
+        } catch (MalformedURLException e) {
+            log.debug("Generating download url failure. {}", e.getMessage());
+            throw new IllegalStateException(e);
+        }
+
+        TableExportResult queryResult = new TableExportResult();
+        queryResult.setHttpStatus(200);
+        queryResult.setCompletedOn(new Date());
+        queryResult.setUrl(downloadURL);
+        queryResult.setRecordCount(recordNumber);
```
moizarafat, line 100: This record count does not include the Header. Should we include it?

elide-async/src/main/java/com/yahoo/elide/async/export/formatter/CSVExportFormatter.java
```diff
@@ -64,15 +75,46 @@ public String format(PersistentResource resource, Integer recordNumber) {
...
+        Iterator itr = projection.getAttributes().iterator();
         StringBuilder str = new StringBuilder();
-
-        try {
-            str.append(mapper.writeValueAsString(resource.getObject()));
-        } catch (JsonProcessingException e) {
-            log.error("Exception when converting to JSON {}", e.getMessage());
-            throw new IllegalStateException(e);
+        int columnCount = 0;
+        while (itr.hasNext()) {
```
aklish, line 88: I think this can be written in more readable fashion as a stream:

```java
projection.getAttributes().stream().
   .map(attr -> //convert to name of the attribute)
   .map(//append quotes)
   .collect(Collectors.joining(",")
```

elide-async/src/main/java/com/yahoo/elide/async/export/formatter/JSONExportFormatter.java
```diff
@@ -18,7 +19,7 @@
 @Slf4j
 public class JSONExportFormatter implements TableExportFormatter {
     private static final String COMMA = ",";
-    private ObjectMapper mapper = new ObjectMapper();
+    private static ObjectMapper mapper = new ObjectMapper();
```
aklish, line 22: not sure about this.   Why share this across threads?  Can we avoid the static?

elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -7,44 +7,61 @@
...
     @Override
-    public Callable getOperation(AsyncAPI query, RequestScope requestScope) {
-        // TODO Auto-generated method stub
-        return null;
+    public Callable<AsyncAPIResult> getOperation(AsyncAPI export, RequestScope requestScope) {
+        Callable operation = null;
+        TableExportFormatter formatter = supportedFormatters.get(((TableExport) export).getResultType());
+
+        if (formatter == null) {
+            throw new InvalidOperationException("Formatter unavailable for " + ((TableExport) export).getResultType());
```
aklish, line 56: Rather than duplicate the cast and function call, you can save this as a variable and reuse the variable.

elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -7,44 +7,61 @@
...
+        }
+
+        if (export.getQueryType().equals(QueryType.GRAPHQL_V1_0)) {
+            operation = new GraphQLTableExportCallableOperation(formatter, getAsyncExecutorService(), export,
```
aklish, line 60: This is a really long noun.  I would remove either Callable or Operation from all of these classes.

elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -7,44 +7,61 @@
...
+                    (com.yahoo.elide.core.RequestScope) requestScope);
+        } else {
+            throw new InvalidOperationException(export.getQueryType() + "is not supported");
```
aklish, line 63: I forget why don't we support JSON-API?

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportCallableOperation.java
```diff
@@ -6,51 +6,193 @@
...
+            finalResults = Observable.just(preResult).concatWith(results);
+        } else {
+            finalResults = results;
+        }
+
+        if (postResult != null) {
+            finalResults.concatWith(Observable.just(postResult));
+        }
+
+        storeResults(exportObj, service.getResultStorageEngine(), results);
```
aklish, line 86: Don't we want to store the final results here?

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportCallableOperation.java
```diff
@@ -6,51 +6,193 @@
...
+        // Stitch together Pre-Formatted, Formatted, Post-Formatted results of Formatter in single observable.
+        Observable<String> finalResults = Observable.empty();
+
+        if (preResult != null) {
+            finalResults = Observable.just(preResult).concatWith(results);
+        } else {
+            finalResults = results;
+        }
+
+        if (postResult != null) {
```
aklish, line 82: Maybe make a private function that concats an observable and checks for null.

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportCallableOperation.java
aklish: No.  I don't think so.  But isn't the query result set in some other base class?

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
moizarafat: We will support it. we have not yet written it so till then have included this check.

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportCallableOperation.java
moizarafat: TableExportResult and AsyncQueryResult have different properties in the query result so the operation classes (TableExportCallableOperation and AsyncQueryCallableOperation) set them separately.

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportOperation.java
```diff
@@ -0,0 +1,201 @@
...
+            exportResult.setMessage(e.getMessage());
+        } finally {
+            // Follows same flow as GraphQL. The query may result in failure but request was successfully processed.
+            exportResult.setHttpStatus(200);
+            exportResult.setCompletedOn(new Date());
+            return exportResult;
+        }
+    }
+
+    private Observable<String> concatStringWithObservable(String toConcat, Observable<String> observable,
```
moizarafat, line 97: Observable can not be created on a null object. Can not concat empty Observable with another observable. Hence had to rely on `boolean stringFirst` to determine the order to concat.

elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -7,44 +7,66 @@
...
+        TableExportFormatter formatter = supportedFormatters.get(resultType);
+
+        if (formatter == null) {
+            throw new InvalidOperationException("Formatter unavailable for " + resultType);
+        }
+
+        if (queryType.equals(QueryType.GRAPHQL_V1_0)) {
+            operation = new GraphQLTableExportOperation(formatter, getAsyncExecutorService(), export, scope);
+        } else {
+            // TODO - Support JSONAPI
```
moizarafat, line 67: Will do it in the next phase.

aklish (approved):

PR #1811, opened 2021-02-03
TableExport Spring Controller

moizarafat:
## Description
i) Spring Controller for the export functionality.
ii) Tweaks to the Formatters based on validation issues.

## Motivation and Context
Spring Controller for the export functionality.

## How Has This Been Tested?
Existing and New test cases pass.

## License


elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/AsyncTest.java
```diff
@@ -106,4 +126,158 @@ public void testAsyncApiEndpoint() throws InterruptedException {
             }
         }
     }
+// TODO Uncomment once AsyncExecutorService Singleton is removed. https://github.com/yahoo/elide/issues/1798
```
moizarafat, line 129: Can we have these test cases commented till the Singleton is removed?

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/DisableAggStoreAsyncTest.java
```diff
@@ -5,12 +5,114 @@
...
+import io.restassured.response.Response;
+
+import javax.ws.rs.core.MediaType;
 
 /**
  * Executes Async tests with Aggregation Store disabled.
  */
-@ActiveProfiles("disableAggStore")
-public class DisableAggStoreAsyncTest extends AsyncTest {
+@SqlMergeMode(SqlMergeMode.MergeMode.MERGE)
```
moizarafat, line 35: Removed dependency on separate Profile yamls. And also if we add tableExport test cases to AsyncTest they will not work here.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -1359,7 +1370,7 @@ private Resource toResource(final Supplier<Map<String, Relationship>> relationsh
      *
      * @return Relationship mapping
      */
-    protected Map<String, Relationship> getRelationships() {
+    public Map<String, Relationship> getRelationships() {
```
aklish, line 1373: I'm not sure about how I feel about making more methods here public just for tests.  This is one of the classes that has so many touch points - we really want to keep the contract locked down.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -49,7 +62,9 @@
 @EntityScan(basePackageClasses = AsyncQuery.class)
 @EnableConfigurationProperties(ElideConfigProperties.class)
 @ConditionalOnExpression("${elide.async.enabled:false}")
-public class ElideAsyncConfiguration {
+@EnableAsync
```
aklish, line 65: Why did we add these two annotations?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -64,7 +79,7 @@
     public AsyncExecutorService buildAsyncExecutorService(Elide elide, ElideConfigProperties settings,
             AsyncAPIDAO asyncQueryDao, EntityDictionary dictionary,
             @Autowired(required = false) ResultStorageEngine resultStorageEngine) {
-        AsyncProperties asyncProperties = settings.getAsync();
+    AsyncProperties asyncProperties = settings.getAsync();
```
aklish, line 82: Spacing looks off.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/controllers/ExportController.java
```diff
@@ -0,0 +1,103 @@
...
+                            recordCount.getAndIncrement();
+                        },
+                        error -> {
+                            if (recordCount.get() != 0) {
+                                // Add error message in the attachment as a way to signal errors.
+                                // Attachment download has already started.
+                                // sendError can not be called,
+                                // as it will result in "Cannot call sendError() after the response has been committed".
+                                // This will return 200 status.
+                                outputStream.write(
```
aklish, line 72: What happens if these throw IOExceptions?

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/DynamicConfigTest.java
```diff
@@ -25,9 +25,11 @@
  */
 @SqlMergeMode(SqlMergeMode.MergeMode.MERGE)
 @Sql(executionPhase = Sql.ExecutionPhase.BEFORE_TEST_METHOD,
-        statements = "CREATE TABLE PlayerStats (name varchar(255) not null,"
+        statements = "DROP TABLE PlayerStats IF EXISTS;"
```
aklish, line 28: shouldn't this be CREATE TABLE IF NOT EXISTS instead of doing drops here?

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/AsyncTest.java
```diff
@@ -33,7 +35,25 @@
...
+                + "DELETE FROM Stats; DELETE FROM PlayerStats;")
+@Sql(statements = {
+        "DROP TABLE Stats IF EXISTS;",
+        "CREATE TABLE Stats(id int, measure int, dimension VARCHAR(255));",
+        "INSERT INTO Stats (id, measure, dimension) VALUES\n"
+                + "\t\t(1,100,'Foo'),"
+                + "\t\t(2,150,'Bar');"
+})
+@Sql(statements = {
+        "DROP TABLE PlayerStats IF EXISTS;",
```
aklish, line 48: YOu shouldn't need to add all these DROP/CREATE TABLE statements

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/ControllerTest.java
```diff
@@ -463,4 +466,13 @@ public void graphqlTestForbiddenCreate() {
                 .body("errors.message", contains("CreatePermission Denied"))
                 .statusCode(200);
     }
+
+    // Controller disabled by default.
+    @Test
+    public void exportControllerTest() {
+        when()
+                .get("/export/asyncQueryId")
```
aklish, line 474: I'm not sure this is testing the controller is disabled or just that asyncQueryId doesn't exist as an export object.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
```diff
@@ -153,10 +168,71 @@ public AsyncAPIDAO buildAsyncAPIDAO(Elide elide) {
...
+    }
+
+    /**
+     * Handles Concurrent Calls for the Export controller in the background.
+     * @return CallableProcessingInterceptor
+     */
+    @Bean
+    @ConditionalOnMissingBean
+    @ConditionalOnProperty(prefix = "elide.async.export", name = "enabled", matchIfMissing = false)
+    public CallableProcessingInterceptor callableProcessingInterceptor() {
```
aklish, line 230: What is all this for?

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/export/formatter/JSONExportFormatter.java
```diff
@@ -38,26 +49,49 @@ public String format(PersistentResource resource, Integer recordNumber) {
...
         }
 
         StringBuilder str = new StringBuilder();
-
         try {
-            str.append(mapper.writeValueAsString(resource.getObject()));
+            JsonApiDocument jsonApiDocument = new JsonApiDocument();
+            //TODO Make this a document processor
+            Data<Resource> data = resource == null ? null
+                    : new Data<>(resource.toResource(getRelationships(resource), getAttributes(resource)));
```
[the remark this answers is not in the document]
moizarafat: `resource.toResource()` relies on the provided implementation of the getAttributes method in PersistenResource. It pulls the list of the Attributes from the dictionary which means it will have all the columns and not only the ones requested by the user. It does have some logic based on sparseFields but sparseFields is not populated in case of Graphql requests.

elide-async/src/main/java/com/yahoo/elide/async/export/formatter/JSONExportFormatter.java
```diff
@@ -38,26 +49,49 @@ public String format(PersistentResource resource, Integer recordNumber) {
...
+            JsonNode node = jsonMapper.toJsonObject(jsonApiDocument).get("data").get("attributes");
+            str.append(jsonMapper.writeJsonApiDocument(node));
         } catch (JsonProcessingException e) {
             log.error("Exception when converting to JSON {}", e.getMessage());
             throw new IllegalStateException(e);
         }
         return str.toString();
     }
 
+    protected static Map<String, Object> getAttributes(PersistentResource resource) {
```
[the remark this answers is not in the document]
moizarafat: Same comment as above.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
moizarafat: This was by mistake. fixed

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/ControllerTest.java
moizarafat: This is for disabled controller

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/controllers/ExportController.java
moizarafat: io.reactivex.exceptions.CompositeException will be thrown by the observable. Processing would have stopped anyways.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAsyncConfiguration.java
moizarafat: This is for providing executors and controlling other configs so that controller can run the export in the background instead of blocking the threads. I think we can move it out of auto-configure and provide as a reference in the example repo.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/AsyncTest.java
moizarafat: I have moved these CREATE and DELETE tables into a SQL file and then each test can call it in the `@SQL` Annotation as a script.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/ControllerTest.java
moizarafat: They both send 404 so not sure how to separate it out. In case of controller disabled we will get an `error` message `Not Found`. Does not happen in case of query not found.

aklish: What happens if you post an export request?

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/DisableAggStoreAsyncTest.java
```diff
@@ -5,12 +5,115 @@
...
+)
+@TestPropertySource(
+        properties = {
+                "elide.aggregation-store.enabled=false"
+        }
+)
+public class DisableAggStoreAsyncTest extends IntegrationTest {
+
+    @Test
+    public void testAsyncApiEndpoint() throws InterruptedException {
```
aklish, line 50: What is the goal of this test exactly?  Verify that async works when agg store is disabled?  We should document that.

aklish (changes requested):

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/DisableAggStoreAsyncTest.java
moizarafat: Yes, will update.

PR #1816, opened 2021-02-05
Metadatastore Models Permission

moizarafat:
## Description
Add permissions on Metadatastore models to make them read-only.

## Motivation and Context
Add permissions on Metadatastore models to make them read-only.

## How Has This Been Tested?
Existing tests pass.

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/package-info.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/package-info.java
@@ -1,0 +1,17 @@
+/*
+ * Copyright 2021, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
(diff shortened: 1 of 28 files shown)


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/package-info.java
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright 2021, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+/**
+ * Models Package.
+ */
+@ReadPermission(expression = "Prefab.Role.All")
+@UpdatePermission(expression = "Prefab.Role.None")
```
moizarafat, line 10: We can not test Update/Read/Create cos the `Table.java` class is abstract so the post calls won't work.

aklish (approved):

PR #1821, opened 2021-02-08
Convert user provided alias before using in SQL

rishi-aga:
Convert user provided alias before using in SQL

## License


elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/framework/SQLUnitTest.java
```diff
@@ -535,4 +590,60 @@ protected void testQueryExecution(Query query) {
...
+                        + "GROUP BY \n"
+                        + "    `com_yahoo_elide_datastores_aggregation_example_PlayerStats_XXX`.`overallRating_207658499`, \n"
+                        + "    `com_yahoo_elide_datastores_aggregation_example_PlayerStats_XXX`.`playerLevel_96024484`, \n"
+                        + "    PARSEDATETIME(FORMATDATETIME(`com_yahoo_elide_datastores_aggregation_example_PlayerStats_XXX`.`recordedDate_155839778`, 'yyyy-MM'), 'yyyy-MM') \n"
+                        + "HAVING \n"
+                        + "    AVG(`com_yahoo_elide_datastores_aggregation_example_PlayerStats_XXX`.`highScore`) > :XXX \n"
+                        + "ORDER BY \n"
+                        + "    `com_yahoo_elide_datastores_aggregation_example_PlayerStats_XXX`.`overallRating_207658499` DESC ";
+
+        if (useAliasForOrderByClause) {
```
moizarafat, line 638: May be L-638 and L644-L646 could be 2 separate private methods in SQLUnitTest.java so that they can be used by other tests if needed.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/query/ColumnProjection.java
```diff
@@ -32,6 +33,16 @@ default String getAlias() {
         return getName();
     }
 
+    default String getSQLAlias() {
```
aklish, line 36: We tried to keep the concept of SQL out of the query engine classes like ColumnProjection.  This is why I thought the column ID might work better here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceVisitor.java
```diff
@@ -69,7 +69,7 @@ protected String visitFieldDimension(ColumnProjection dimension) {
 
         //This is a nested query.  Don't do table lookups.
         } else {
-            return getFieldAlias(applyQuotes(tableAliases.peek()), applyQuotes(dimension.getName()));
+            return getFieldAlias(applyQuotes(tableAliases.peek()), applyQuotes(dimension.getSQLAlias()));
```
aklish, line 72: What about metrics?

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceVisitor.java
rishi-aga: this method gets dimensions when columnType is `FIELD` (https://github.com/yahoo/elide/blob/master/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/ColumnVisitor.java#L59)
Metric will  goes to visitFormulaMetric --> visitFormulaColumn --> visitPhysicalReference

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/query/ColumnProjection.java
rishi-aga: renamed to getSafeAlias method

aklish (approved):

PR #1824, opened 2021-02-09
Data Export: JSON API support

AvaniMakwana:
## Description
Adding JSON API Support to data export feature.

## Motivation and Context
Adding JSON API Support to data export feature.

## How Has This Been Tested?
Unit tests included.

## License


elide-async/src/main/java/com/yahoo/elide/async/hooks/TableExportHook.java
```diff
@@ -63,6 +64,8 @@ public void validateOptions(AsyncAPI export, RequestScope requestScope) {
 
         if (queryType.equals(QueryType.GRAPHQL_V1_0)) {
             operation = new GraphQLTableExportOperation(formatter, getAsyncExecutorService(), export, scope);
+        } else if (queryType.equals(QueryType.JSONAPI_V1_0)) {
+            operation = new JSONAPITableExportOperation(formatter, getAsyncExecutorService(), export, scope);
         } else {
             // TODO - Support JSONAPI
```
moizarafat, line 70: We can remove this TODO comment now.

elide-async/src/main/java/com/yahoo/elide/async/operation/JSONAPITableExportOperation.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.Collections;
+import java.util.UUID;
+
+import javax.ws.rs.core.MultivaluedHashMap;
+import javax.ws.rs.core.MultivaluedMap;
+
+@Slf4j
+public class JSONAPITableExportOperation extends TableExportOperation {
+
+    private static final String PATH = "/tableExport";
```
moizarafat, line 34: I don't think we need this variable.

elide-async/src/main/java/com/yahoo/elide/async/operation/JSONAPITableExportOperation.java
```diff
@@ -0,0 +1,79 @@
...
+            AsyncAPI export, RequestScope scope) {
+        super(formatter, service, export, scope);
+    }
+
+    @Override
+    public RequestScope getRequestScope(TableExport export, User user, String apiVersion, DataStoreTransaction tx)
+            throws URISyntaxException {
+        UUID requestId = UUID.fromString(export.getRequestId());
+        MultivaluedMap<String, String> queryParams = getQueryParams(export.getQuery());
+        return new RequestScope("", "", apiVersion, null, tx, user, queryParams,
```
moizarafat, line 46: The 2nd param here should not be blank. We should get the Path from the queryObj.getQuery. Please refer JSONAPIAsyncQueryOperation.

elide-async/src/main/java/com/yahoo/elide/async/operation/JSONAPITableExportOperation.java
```diff
@@ -0,0 +1,79 @@
...
+                Collections.emptyMap(), requestId, getService().getElide().getElideSettings());
+    }
+
+    @Override
+    public EntityProjection getProjection(TableExport export, String apiVersion) throws BadRequestException {
+        EntityProjection projection = null;
+        try {
+            Elide elide = getService().getElide();
+            projection = new EntityProjectionMaker(elide.getElideSettings().getDictionary(),
+                    getScope()).parsePath(PATH);
```
moizarafat, line 56: parsePath should be parsing the query in the queryObj

elide-async/src/main/java/com/yahoo/elide/async/operation/JSONAPITableExportOperation.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * This method parses the url and gets the query params.
+     * And adds them into a MultivaluedMap
+     * @param uri URIBuilder instance
+     * @return MultivaluedMap with query parameters
+     * @throws URISyntaxException
+     */
+    private MultivaluedMap<String, String> getQueryParams(String query) throws URISyntaxException {
```
moizarafat, line 71: Since this is exactly the same as getQueryParams in JSONAPIAsyncQueryOperation, we can make this method as public static in JSONAPIAsyncQueryOperation and reuse here. It works on a String args so we can have it as static,

elide-async/src/main/java/com/yahoo/elide/async/operation/JSONAPITableExportOperation.java
```diff
@@ -0,0 +1,79 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.net.URISyntaxException;
+import java.util.Collections;
+import java.util.UUID;
+
+import javax.ws.rs.core.MultivaluedHashMap;
+import javax.ws.rs.core.MultivaluedMap;
+
```
moizarafat, line 30: Please add a class  level comment.

moizarafat (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/operation/JSONAPITableExportOperation.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    public EntityProjection getProjection(TableExport export, String apiVersion) throws BadRequestException {
+        EntityProjection projection = null;
+        try {
+            URIBuilder uri = new URIBuilder(export.getQuery());
+            Elide elide = getService().getElide();
+            projection = new EntityProjectionMaker(elide.getElideSettings().getDictionary(),
+                    getScope()).parsePath(JSONAPIAsyncQueryOperation.getPath(uri));
+
+        } catch (Exception e) {
```
aklish, line 59: This is too broad.  What exception are you trying to catch here?

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportOperation.java
```diff
@@ -110,9 +111,10 @@ public AsyncAPIResult call() {
      * @param prevScope RequestScope object.
      * @param projection Entity projection.
      * @return Observable PersistentResource
+     * @throws URISyntaxException URISyntaxException from malformed or incorrect URI
```
aklish, line 114: For what URI will this exception be thrown?

elide-async/src/test/java/com/yahoo/elide/async/operation/JsonAPITableExportOperationTest.java
```diff
@@ -0,0 +1,142 @@
...
+        when(asyncExecutorService.getResultStorageEngine()).thenReturn(engine);
+        when(requestScope.getApiVersion()).thenReturn(NO_VERSION);
+        when(requestScope.getUser()).thenReturn(user);
+        when(requestScope.getElideSettings()).thenReturn(elide.getElideSettings());
+        when(requestScope.getBaseUrlEndPoint()).thenReturn("https://elide.io");
+    }
+
+    @Test
+    public void testProcessQuery() throws URISyntaxException, IOException  {
+        dataPrep();
```
aklish, line 87: I can't see the difference between this test setup and the next one.  What makes one succeed where the other fails?

aklish (changes requested):

elide-async/src/test/java/com/yahoo/elide/async/operation/JsonAPITableExportOperationTest.java
AvaniMakwana: there is a difference in Query. /tableExport vs /tableExport1

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportOperation.java
AvaniMakwana: for parsing Queryparam & path from querystring from getRequestScope() in JSONAPITableExportOperation

aklish: Does it make sense for every TableExportOperation to parse URIs?  If not, we shouldn't generalize the exception handling here in this way.

elide-async/src/main/java/com/yahoo/elide/async/operation/JSONAPITableExportOperation.java
```diff
@@ -33,15 +34,19 @@
...
-    public RequestScope getRequestScope(TableExport export, User user, String apiVersion, DataStoreTransaction tx)
-            throws URISyntaxException {
+    public RequestScope getRequestScope(TableExport export, User user, String apiVersion, DataStoreTransaction tx) {
         UUID requestId = UUID.fromString(export.getRequestId());
-        URIBuilder uri = new URIBuilder(export.getQuery());
+        URIBuilder uri;
+        try {
+            uri = new URIBuilder(export.getQuery());
+        } catch (URISyntaxException e) {
+            throw new BadRequestException(e.getMessage());
```
aklish, line 48: IllegalStateException would make more sense here.  I"m not sure this should ever happen (and there would be nothing the client can do to resolve it).

elide-async/src/main/java/com/yahoo/elide/async/operation/JSONAPITableExportOperation.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public EntityProjection getProjection(TableExport export, String apiVersion) throws BadRequestException {
+        EntityProjection projection = null;
+        try {
+            URIBuilder uri = new URIBuilder(export.getQuery());
+            Elide elide = getService().getElide();
+            projection = new EntityProjectionMaker(elide.getElideSettings().getDictionary(),
+                    getScope()).parsePath(JSONAPIAsyncQueryOperation.getPath(uri));
```
aklish, line 62: This is the original scope - don't we want the new one we create in the method above?  The request scope include the transaction (which will be different than the transaction used to run this request).  I'm not sure how this works in practice.

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportOperation.java
```diff
@@ -43,7 +43,7 @@
     @Getter private AsyncExecutorService service;
     private Integer recordNumber = 0;
     private TableExport exportObj;
-    private RequestScope scope;
+    @Getter private RequestScope scope;
```
aklish, line 46: Shouldn't need this I think.

elide-async/src/test/java/com/yahoo/elide/async/operation/JsonAPITableExportOperationTest.java
```diff
@@ -0,0 +1,160 @@
...
+
+import java.io.IOException;
+import java.net.URISyntaxException;
+import java.nio.file.Path;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class JsonAPITableExportOperationTest {
```
aklish, line 45: Do we have an end to end test?  I feel like that transaction bug above ought to be caught somewhere.

elide-async/src/test/java/com/yahoo/elide/async/operation/JsonAPITableExportOperationTest.java
```diff
@@ -0,0 +1,160 @@
...
+        queryObj.setQuery(query);
+        queryObj.setQueryType(QueryType.JSONAPI_V1_0);
+        queryObj.setResultType(ResultType.CSV);
+
+        JSONAPITableExportOperation jsonAPIOperation = new JSONAPITableExportOperation(new JSONExportFormatter(elide), asyncExecutorService,
+                queryObj, requestScope, engine);
+        TableExportResult queryResultObj = (TableExportResult) jsonAPIOperation.call();
+
+        assertEquals(200, queryResultObj.getHttpStatus());
+        assertEquals("EntityProjection generation failure.", queryResultObj.getMessage());
```
aklish, line 139: This is not a very good error message to return to clients.  Let's fix this.

elide-async/src/test/java/com/yahoo/elide/async/operation/JsonAPITableExportOperationTest.java
```diff
@@ -0,0 +1,160 @@
...
+        queryObj.setQueryType(QueryType.JSONAPI_V1_0);
+        queryObj.setResultType(ResultType.CSV);
+
+        JSONAPITableExportOperation jsonAPIOperation = new JSONAPITableExportOperation(new JSONExportFormatter(elide), asyncExecutorService,
+                queryObj, requestScope, engine);
+        TableExportResult queryResultObj = (TableExportResult) jsonAPIOperation.call();
+
+        assertEquals(200, queryResultObj.getHttpStatus());
+        assertTrue("https://elide.io/export/edc4a871-dff2-4054-804e-d80075cf827d".equals(queryResultObj.getUrl().toString()));
+        assertEquals(1, queryResultObj.getRecordCount());
```
aklish, line 101: Should we verify the query result message like we do on other tests?

aklish (changes requested):

elide-async/src/main/java/com/yahoo/elide/async/operation/JSONAPITableExportOperation.java
moizarafat: if the query is malformed, then we can get this error which can be considered a BadRequest.

elide-async/src/test/java/com/yahoo/elide/async/operation/JsonAPITableExportOperationTest.java
moizarafat: As discussed, `EntityProjectionMaker.parse` does not make use of Transaction or requestscope.

aklish (approved):

moizarafat (approved):

aklish (approved):

PR #1833, opened 2021-02-11
Validators for JSON and Graphql Exporter

rishi-aga:
Validators for JSON and Graphql Exporter

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/export/validator/NoRelationshipsProjectionValidator.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/export/validator/NoRelationshipsProjectionValidator.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2021, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.async.export.validator;
+
+import com.yahoo.elide.core.exceptions.BadRequestException;
+import com.yahoo.elide.core.request.EntityProjection;
+
+import java.util.Collection;
+
+/**
+ * Validates none of the projections have relationships.
+ */
+public class NoRelationshipsProjectionValidator implements Validator {
+
+    private static final Validator INSTANCE = new NoRelationshipsProjectionValidator();
+
+    public static Validator getInstance() {
+        return INSTANCE;
+    }
+
+    @Override
+    public void validateProjection(Collection<EntityProjection> projections) {
+        for (EntityProjection projection : projections) {
+            if (!projection.getRelationships().isEmpty()) {
+                throw new BadRequestException(
+                                "Export is not supported for Query that requires traversing Relationships.");
+            }
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


elide-async/src/main/java/com/yahoo/elide/async/export/validator/NoRelationshipsProjectionValidator.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Collection;
+
+/**
+ * Validates none of the projections have relationships.
+ */
+public class NoRelationshipsProjectionValidator implements Validator {
+
+    private static final Validator INSTANCE = new NoRelationshipsProjectionValidator();
+
+    public static Validator getInstance() {
```
aklish, line 20: Let's not use singletons.

elide-async/src/test/java/com/yahoo/elide/async/operation/GraphQLTableExportOperationTest.java
```diff
@@ -138,6 +143,69 @@ public void testProcessBadQuery() throws URISyntaxException, IOException  {
...
+
+        assertEquals(200, queryResultObj.getHttpStatus());
+        assertEquals("Export is only supported for single Query with one root projection.",
+                        queryResultObj.getMessage());
+        assertEquals(null, queryResultObj.getRecordCount());
+        assertEquals(null, queryResultObj.getUrl());
+    }
+
+    @Test
+    public void testProcessMultipleQuery() {
```
aklish, line 189: How is this test different than the one above it?

elide-async/src/main/java/com/yahoo/elide/async/operation/GraphQLTableExportOperation.java
```diff
@@ -41,7 +41,8 @@
 
     public GraphQLTableExportOperation(TableExportFormatter formatter, AsyncExecutorService service,
             AsyncAPI export, RequestScope scope, ResultStorageEngine engine) {
-        super(formatter, service, export, scope, engine);
+        super(formatter, service, export, scope, engine,
+                        Arrays.asList(NoRelationshipsProjectionValidator.getInstance()));
```
aklish, line 45: It seems like both need the same validators.  Should we move the NoRelationshipsProjectionValdator to the super class list?

aklish (changes requested):

elide-async/src/test/java/com/yahoo/elide/async/operation/GraphQLTableExportOperationTest.java
rishi-aga: this one has 2 queries with single projection and other one has single query with more than one projection

elide-async/src/main/java/com/yahoo/elide/async/operation/GraphQLTableExportOperation.java
rishi-aga: Single `SingleRootProjectionValidator` is must as I am taking first one from collection in call method, so kept it in superclass. `NoRelationshipsProjectionValdator` could be optional for any new subclass.

aklish (approved):

PR #1842, opened 2021-02-15
Integration Tests for Table Export

moizarafat:
Resolves #1272

## Description
Integration Tests for Table Export

## How Has This Been Tested?
New and Existing Tests pass

## License


elide-integration-tests/src/test/java/example/ForbiddenBook.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+@Entity
+@Table(name = "forbiddenbook")
+@Include
+@Paginate
+@Audit(action = Audit.Action.CREATE,
+        operation = 10,
+        logStatement = "{0}",
+        logExpressions = {"${forbiddenBook.title}"})
+@ReadPermission(expression = "Prefab.Role.None")
```
aklish, line 28: There is already an entity, `NoReadEntity` for testing this kind of security check.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/framework/AsyncIntegrationTestApplicationResourceConfig.java
```diff
@@ -88,15 +94,15 @@ protected void configure() {
...
                 bind(elide).to(Elide.class).named("elide");
 
                 AsyncAPIDAO asyncAPIDao = new DefaultAsyncAPIDAO(elide.getElideSettings(), elide.getDataStore());
                 bind(asyncAPIDao).to(AsyncAPIDAO.class);
 
-                // TODO Pass resultStorageEngine to TableExportHook
-                ResultStorageEngine resultStorageEngine =
-                        new FileResultStorageEngine(System.getProperty("java.io.tmpDir"));
+                // Create ResultStorageEngine using /tmp
+                ResultStorageEngine resultStorageEngine = new FileResultStorageEngine("/tmp");
```
aklish, line 105: We should use Java libraries here that manage tmp files (rather than using /tmp).  This will also likely fail on windows systems.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1441 @@
...
+                .then()
+                .statusCode(org.apache.http.HttpStatus.SC_OK);
+
+        String expectedResponse = "{\"data\":{\"tableExport\":{\"edges\":[{\"node\":{\"id\":\"edc4a871-dbf2-4054-804e-d80075cf828e\","
+                + "\"query\":\"{\\\"query\\\":\\\"{ book { edges { node { title authors {edges { node { name } } } } } } }\\\", \\\"variables\\\":null}\","
+                + "\"queryType\":\"GRAPHQL_V1_0\",\"status\":\"PROCESSING\",\"resultType\":\"CSV\"}}]}}}";
+        assertEquals(expectedResponse, response.extract().body().asString());
+
+        int i = 0;
+        while (i < 1000) {
```
aklish, line 1172: We need to stop repeating this test structure over and over - and generalize this somehow.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1441 @@
...
+
+        this.port = super.getPort();
+    }
+
+    @Override
+    protected DataStoreTestHarness createHarness() {
+        DataStoreTestHarness dataStoreTestHarness = super.createHarness();
+        return new DataStoreTestHarness() {
+                public DataStore getDataStore() {
+                    return new AsyncDelayDataStore(dataStoreTestHarness.getDataStore(), 5000);
```
aklish, line 134: 5 seconds seems like a long time wait for each test.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1441 @@
...
+    /**
+     * Various tests for a JSONAPI query as a TableExport Request with asyncAfterSeconds value set to 0.
+     * Happy Path Test Scenario 1
+     * @throws InterruptedException InterruptedException
+     * @throws IOException IOException
+     */
+    @Test
+    public void jsonApiHappyPath1() throws InterruptedException, IOException {
+
+        AsyncDelayStoreTransaction.sleep = true;
```
aklish, line 187: This would be significantly cleaner if sleeping (and the duration) was controlled by a request header (which the AsyncDelayStore could access via the request scope).  Can we clean that up?

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1441 @@
...
+        ObjectNode graphqlNode = JsonNodeFactory.instance.objectNode();
+        graphqlNode.put("query", query);
+        if (variables != null) {
+            graphqlNode.set("variables", mapper.valueToTree(variables));
+        }
+        return graphqlNode;
+    }
+
+    private String getStoredFileContents(String id) throws IOException {
+        StringBuilder lines = new StringBuilder();
```
aklish, line 1432: What if we leveraged the result storage engine here?  Then this code doesn't need to know the location of the temp file (see earlier comment).

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1441 @@
...
+                        .body("data.attributes.result.message", nullValue())
+                        .body("data.attributes.result.recordCount", equalTo(3))
+                        .body("data.attributes.result.url",
+                                equalTo("http://localhost:" + port + "/export/edc4a871-dff2-4054-804e-d80075cf830a"))
+                        .body("data.attributes.result.httpStatus", equalTo(200));
+
+                assertEquals("\"title\"\n"
+                        + "\"For Whom the Bell Tolls\"\n"
+                        + "\"Song of Ice and Fire\"\n"
+                        + "\"Ender's Game\"\n", getStoredFileContents("edc4a871-dff2-4054-804e-d80075cf830a"));
```
aklish, line 247: Shouldn't this be downloading the file via the API and using that for comparison?

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
aklish: Scratch that.  I think we can remove this function (see earlier comment).

aklish (changes requested):

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
moizarafat: changing to 1 seconds

elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncQueryOperation.java
```diff
@@ -76,10 +75,9 @@ public void nullResponseCheck(ElideResponse response) {
...
      * Execute the Async Query Request.
      * @param queryObj AsyncAPI type object.
-     * @param user User object.
-     * @param apiVersion Api Version.
+     * @param scope RequestScope.
      * @return response ElideResponse object.
      * @throws URISyntaxException URISyntaxException Exception.
      */
-    public abstract ElideResponse execute(AsyncAPI queryObj, User user, String apiVersion)  throws URISyntaxException;
+    public abstract ElideResponse execute(AsyncAPI queryObj, RequestScope scope)  throws URISyntaxException;
```
moizarafat, line 82: Did this contract change so that we can extract the requestHeaders and use when making calls to elide.get/QueryRunner.run.

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportOperation.java
```diff
@@ -172,13 +170,11 @@ public AsyncAPIResult call() {
...
      * @param exportObj TableExport type object.
-     * @param user User Object.
-     * @param apiVersion API Version.
+     * @param scope RequestScope.
      * @param tx DataStoreTransaction.
      * @return RequestScope Type Object
      */
-    public abstract RequestScope getRequestScope(TableExport exportObj, User user,
-            String apiVersion, DataStoreTransaction tx);
+    public abstract RequestScope getRequestScope(TableExport exportObj, RequestScope scope, DataStoreTransaction tx);
```
moizarafat, line 177: Did this contract change so that we can extract the requestHeaders and use when creating the new RequestScope

elide-async/pom.xml
```diff
@@ -91,6 +91,18 @@
         </dependency>
 
         <!-- Test -->
+        <!-- For Export End-Point Test -->
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
```
aklish, line 97: Should this be marked as test as well?

elide-async/src/main/java/com/yahoo/elide/async/operation/TableExportOperation.java
```diff
@@ -172,13 +170,11 @@ public AsyncAPIResult call() {
     /**
      * Initialize and get RequestScope.
      * @param exportObj TableExport type object.
-     * @param user User Object.
-     * @param apiVersion API Version.
+     * @param scope RequestScope.
```
aklish, line 173: Since this takes a scope and returns a scope - we should add a bit more explanation here in the comments explaining the difference.

elide-integration-tests/src/test/java/com/yahoo/elide/initialization/IntegrationTest.java
```diff
@@ -121,6 +124,9 @@ protected final Server setUpServer() throws Exception {
         servletContextHandler.setContextPath("/");
         server.setHandler(servletContextHandler);
 
+        servletContextHandler.setAttribute(AsyncIntegrationTestApplicationResourceConfig.ASYNC_EXECUTOR_ATTR, Executors.newFixedThreadPool(5));
+        servletContextHandler.setAttribute(AsyncIntegrationTestApplicationResourceConfig.STORAGE_DESTINATION_ATTR, Files.createTempDirectory("asyncIT"));
```
aklish, line 128: Ideally, this would be done in a subclass of IntegrationTest.  Maybe there can be a function that manipulates the servletContextHandler that subclasses can override.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1250 @@
...
+
+        com.yahoo.elide.async.models.TableExport queryObj = new com.yahoo.elide.async.models.TableExport();
+        queryObj.setId(id);
+        queryObj.setQuery(query);
+        queryObj.setResultType(ResultType.CSV);
+        queryObj.setQueryType(QueryType.JSONAPI_V1_0);
+        queryObj.setPrincipalName("owner-user");
+
+        dataStore.populateEntityDictionary(
+                        new EntityDictionary(AsyncIntegrationTestApplicationResourceConfig.MAPPINGS));
```
aklish, line 1135: You create two entity dictionaries in this function.  Let's create just one.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1250 @@
...
+                .body("data.attributes.status", equalTo("COMPLETE"))
+                .body("data.attributes.result.recordCount", nullValue())
+                .body("data.attributes.result.url", nullValue())
+                .body("data.attributes.result.message", equalTo("Unknown collection group"))
+                .body("data.attributes.result.httpStatus", equalTo(200));
+
+    }
+
+    /**
+     * Various tests for an unknown collection (group) that does not exist GrapqhQL query as a TableExport Request.
```
aklish, line 663: Grammar issues with this sentence.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1250 @@
...
+                .contentType(MediaType.APPLICATION_JSON)
+                .accept(MediaType.APPLICATION_JSON)
+                .body(graphQLJsonNode)
+                .post("/graphQL")
+                .then()
+                .statusCode(org.apache.http.HttpStatus.SC_OK);
+
+        String output = response.extract().body().asString();
+        assertEquals(true, output.contains("errors"));
+        assertEquals(true, output.contains("data.resultType"));
```
aklish, line 571: Is this the error?  THis is pretty cryptic.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1250 @@
...
+     * Test for ResultType Set to Invalid.
+     */
+    @Test
+    public void graphQLTestCreateFailOnResultType() {
+
+        TableExport queryObj = new TableExport();
+        queryObj.setId("edc4a871-dff2-4054-804e-d80075cf939e");
+        queryObj.setAsyncAfterSeconds(0);
+        queryObj.setQueryType("GRAPHQL_V1_0");
+        queryObj.setStatus("QUEUED");
```
aklish, line 537: What triggers the error here?  Setting the status?

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
```diff
@@ -0,0 +1,1250 @@
...
+                .contentType(MediaType.APPLICATION_JSON)
+                .accept(MediaType.APPLICATION_JSON)
+                .body(graphQLJsonNode)
+                .post("/graphQL")
+                .then()
+                .statusCode(org.apache.http.HttpStatus.SC_OK);
+
+        String output = response.extract().body().asString();
+        assertEquals(true, output.contains("errors"));
+        assertEquals(true, output.contains("CreatePermission Denied"));
```
aklish, line 524: Why does this fail with create permission?

aklish (changes requested):

elide-async/pom.xml
moizarafat: we can not mark javax.servlet-api as test, cos its needed to for import javax.servlet.http.HttpServletResponse;  in Resource class.

moizarafat: ExportApiEndpoint

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
moizarafat: Its for an unsupported resulttype, XLS.

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
moizarafat: It fails the security check `@SecurityCheck(AsyncAPIStatusQueuedValue.VALUE_IS_QUEUED)`

elide-integration-tests/src/test/java/com/yahoo/elide/async/integration/tests/TableExportIT.java
moizarafat: It's a partial error message. I am just making sure the error message contains resultType to confirm a value of resultType caused this error.

aklish (approved):

PR #1843, opened 2021-02-15
Add 'hasmember' operator support across to-many relationships.

### LINKED ISSUE #1822: Many to Many relation query filtering issue
A use case is explained below, when implemented using Elide/ (may be ORM issue), the filtering is not working as expected.

Scenario is, 
1) Books can be associated with multiple Tags (ex: comic, new_releases, horror etc).
2) Books and Tags relation is many-to-many
3) I need to query the books which are "comic AND new-releases"

The DB design is as follows,
1) Table 1 - book (id, description, ...)
2) Table 2 - tag (id, description, ...)
3) Table 3 - book_tag (book_id, tag_id) (relationship table which holds the foreign key with books and tags tables)

Query formed using elide is as follows
GET /api/v1/books?filter=tags.id==comic;tags.id==newReleases

The resulting query created is
select distinct book0_.id as id1_0_ from book book0_ left outer join book_tag tag1_ on book0_.id=tag1_.book_id left outer join tag tag2_ on tag1_.tag_id=tag2_.id where (lower(tag2_.id) in (lower(?))) and (lower(tag2_.id) in (lower(?))) limit ?

This query will never satisfy, as the SQL search in the same row with different Id.

The actual query to be formed should be based on "groupby() having count() = X"

Question:
1) Can we form a filter condition in Elide to support this ?
2) Is it expected from Elide to have some logic from its side to detect primary key filtering and form the SQL differently ?

aklish:
Resolves #1822

## Description
Add support for 'hasmember' queries that traverse to-many relationships like:
`/item?filter=tags.id=hasmember='new';tags.id=hasmember='sale'`

## Motivation and Context
See #1822 

## How Has This Been Tested?
New operator unit tests, dialect unit tests, and JSON-API integration tests.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
@@ -397,10 +397,8 @@ public class RSQLFilterDialect implements FilterDialect, SubqueryFilterDialect, 
                     if (FilterPredicate.isLastPathElementAssignableFrom(dictionary, path, COLLECTION_TYPE)) {
                         throw new RSQLParseException("Invalid Path: Last Path Element cannot be a collection type");
                     }
-                } else {
-                    if (!FilterPredicate.isLastPathElementAssignableFrom(dictionary, path, COLLECTION_TYPE)) {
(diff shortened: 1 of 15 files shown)


moizarafat (approved):

aklish:
One thing to note - this fix will only be available in Elide 5.  I'm not sure how to back port this in a general way to Elide 4 without breaking interface contracts.

PR #1852, opened 2021-02-20
Move Include to package-info

wcekan:
## Description
Updated the example bean to move Include annotation into package-info.  This will verify we locate the annotation correctly in package-info there is none on the bean.

## Motivation and Context
Elide 5.x had an issue due to a migration bug and adding test to 4.x to make sure it is tested.

## How Has This Been Tested?
Unit testing.

## License


elide-core/src/main/java/com/yahoo/elide/core/datastore/inmemory/HashMapDataStore.java
```diff
@@ -42,7 +46,20 @@ public HashMapDataStore(Set<Package> beanPackages) {
...
             ClassScanner.getAnnotatedClasses(beanPackage, Include.class).stream()
                 .filter(modelClass -> modelClass.getName().startsWith(beanPackage.getName()))
-                .forEach(modelClass -> dataStore.put(modelClass, Collections.synchronizedMap(new LinkedHashMap<>())));
+                .filter(modelClass -> dictionary.getFirstAnnotation(modelClass,
+                        Arrays.asList(Include.class, Exclude.class)) instanceof Include)
+                .forEach(modelClass -> dataStore.put(modelClass,
+                        Collections.synchronizedMap(new LinkedHashMap<>())));
+        }
+
+        for (Package beanPackage : beanPackages) {
```
aklish, line 55: What's the purpose of the second scan?  The classpath scans are pretty expensive (they take multiple seconds).  This could slow down the build quite a bit.

aklish (changes requested):

elide-core/src/main/java/com/yahoo/elide/core/datastore/inmemory/HashMapDataStore.java
wcekan: Used one scan.

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -240,6 +240,16 @@ public EntityBinding(EntityDictionary dictionary, Class<?> cls, String type, Str
      */
     private void bindEntityFields(Class<?> cls, String type, Collection<AccessibleObject> fieldOrMethodList) {
         for (AccessibleObject fieldOrMethod : fieldOrMethodList) {
+            // Special handling for Eclipse. Load missing annotations
+            if (fieldOrMethod.getAnnotations().length == 0 && fieldOrMethod instanceof Method) {
```
aklish, line 244: Can you explain a bit more what this is about?

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
```diff
@@ -390,7 +400,7 @@ private void bindRelation(AccessibleObject fieldOrMethod, String fieldName, Clas
 
         relationshipsDeque.push(fieldName);
         fieldsToValues.put(fieldName, fieldOrMethod);
-        fieldsToTypes.put(fieldName, fieldType);
+        fieldsToTypes.put(fieldName, fieldType == null ? Void.class : fieldType);
```
aklish, line 403: Same here.

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
wcekan: Moved to #1909

elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java
wcekan: I get a NPE when fieldType is null.

wcekan: Is it probably the generics that causes the null fieldType.

https://github.com/yahoo/elide/blob/f720136a3bffa03ee90d1fbe5f52a44d57c3d6dd/elide-example-models/src/main/java/com/yahoo/elide/models/generics/Peon.java#L23

wcekan: That is exactly the problem. RawType of type `T` is null.

https://github.com/yahoo/elide/blob/f720136a3bffa03ee90d1fbe5f52a44d57c3d6dd/elide-core/src/main/java/com/yahoo/elide/core/EntityBinding.java#L484

PR #1877, opened 2021-03-04
Parse Arguments from Filter Expression

rishi-aga:
## Description
Parse Arguments from Filter Expression. (Argument Values must be encoded when passed in filter expression)
If a field exists in both filter & projection then:
a) If arguments are passed in the filter, these arguments override the arguments provided in the projection.
b) If no arguments are passed in the filter, `filter` will use the arguments provided in the projection.

## Motivation and Context
https://github.com/yahoo/elide/discussions/1829

## How Has This Been Tested?
Existing tests pass.
New tests included

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
@@ -46,6 +46,9 @@ import cz.jirutka.rsql.parser.ast.OrNode;
 import cz.jirutka.rsql.parser.ast.RSQLOperators;
 import cz.jirutka.rsql.parser.ast.RSQLVisitor;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
@@ -65,7 +68,8 @@ public class RSQLFilterDialect implements FilterDialect, SubqueryFilterDialect, 
     private static final String SINGLE_PARAMETER_ONLY = "There can only be a single filter query parameter";
     private static final String INVALID_QUERY_PARAMETER = "Invalid query parameter: ";
     private static final Pattern TYPED_FILTER_PATTERN = Pattern.compile("filter\\[([^\\]]+)\\]");
-    private static final Pattern FILTER_ARGUMENTS_PATTERN = Pattern.compile("\\[(.+?)\\]");
+    // field name followed by zero or more square brackets having non-empty argument name and value separated by ':'
+    private static final String FILTER_SELECTOR_REGEX = "(\\w+)(\\[(\\w+?):([^\\]]+?)\\])*$";
     private static final ComparisonOperator INI = new ComparisonOperator("=ini=", true);
     private static final ComparisonOperator NOT_INI = new ComparisonOperator("=outi=", true);
     private static final ComparisonOperator ISNULL_OP = new ComparisonOperator("=isnull=", false);
@@ -309,6 +313,11 @@ public class RSQLFilterDialect implements FilterDialect, SubqueryFilterDialect, 
             Type entityType = rootEntityType;
 
             for (String associationName : associationNames) {
+
+                if (!associationName.matches(FILTER_SELECTOR_REGEX)) {
+                    throw new RSQLParseException("Filter expression is not in expected format at: " + associationName);
+                }
+
                 // if the association name is "id", replaced it with real id field name
                 // id field name can be "id" or other string, but non-id field can't have name "id".
                 if (associationName.equals(REGULAR_ID_NAME)) {
@@ -316,9 +325,16 @@ public class RSQLFilterDialect implements FilterDialect, SubqueryFilterDialect, 
                 }
 
                 Set<Argument> arguments = new HashSet<>();
-                if (associationName.indexOf('[') > 0) {
-                    parseArguments(associationName, arguments);
-                    associationName = associationName.substring(0, associationName.indexOf('['));
+                int argsIndex = associationName.indexOf('[');
+                if (argsIndex > 0) {
+                    try {
+                        parseArguments(associationName.substring(argsIndex), arguments);
+                    } catch (UnsupportedEncodingException | IllegalArgumentException e) {
+                        throw new RSQLParseException(
+                                        String.format("Filter expression is not in expected format at: %s. %s",
+                                                        associationName, e.getMessage()));
+                    }
+                    associationName = associationName.substring(0, argsIndex);
                 }
                 addDefaultArguments(arguments, dictionary.getAttributeArguments(entityType, associationName));
                 String typeName = dictionary.getJsonAliasFor(entityType);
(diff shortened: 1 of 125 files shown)


elide-core/src/test/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialectTest.java
```diff
@@ -512,4 +517,35 @@ public void testMemberOfOperatorOnRelationshipException() throws Exception {
...
+        expr = assertDoesNotThrow(() -> dialect.parse(bookType, Collections.emptySet(),
+                        "genre=in=(sci-fi,action),title[a:][:y]==Hemingway", NO_VERSION));
+        assertEquals("(book.genre IN [sci-fi, action] OR book.title IN [Hemingway])", expr.toString());
+
+        // Missing argument name or value (: missing within [])
+        Exception e = assertThrows(ParseException.class, () -> dialect.parse(bookType, Collections.emptySet(),
+                        "genre=in=(sci-fi,action),title[ab][x:y]==Hemingway", NO_VERSION));
+        assertEquals("Failed to parse arguments from filter expression at: title[ab][x:y]", e.getMessage());
+
+        // Anything after first '[' and not within [] is ignored
```
aklish, line 541: I think we should be strict here and throw an exception (see earlier comment on refining the regex).

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/resolvers/DailyAverageScorePerPeriodResolver.java
```diff
@@ -52,6 +52,7 @@ public QueryPlan resolve(MetricProjection projection) {
                         .expression(projection.getExpression())
                         .columnType(projection.getColumnType())
                         .valueType(projection.getValueType())
+                        .arguments(projection.getArguments())
```
aklish, line 55: Can we try a filter (with arguments) on a query with a two pass aggregation like daily average score and make sure it works?

aklish (changes requested):
Small comments mostly.  However, we need to understand what character limits we have between the square braces and add some tests there as well.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -319,13 +329,45 @@ private Path buildPath(Type rootEntityType, String selector) {
                             String.format("No such association %s for type %s", associationName, typeName));
                 }
 
-                path.add(new Path.PathElement(entityType, fieldType, associationName));
+                path.add(new Path.PathElement(entityType, fieldType, associationName, associationName, arguments));
```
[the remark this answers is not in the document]
rishi-aga: fieldName and alias. (https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/Path.java#L43)

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -319,13 +329,45 @@ private Path buildPath(Type rootEntityType, String selector) {
...
+            Matcher matcher = FILTER_ARGUMENTS_PATTERN.matcher(attributeName);
+            while (matcher.find()) {
+                String[] pair = matcher.group(1).split(":", 2);
+                if (pair.length < 2) {
+                    throw new RSQLParseException(
+                                    "Failed to parse arguments from filter expression at: " + attributeName);
+                }
+                arguments.add(Argument.builder()
+                                .name(pair[0])
+                                .value(pair[1])
```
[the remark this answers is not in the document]
rishi-aga: Argument only has name and value. (https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/request/Argument.java#L18)

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -319,13 +329,45 @@ private Path buildPath(Type rootEntityType, String selector) {
...
+                            .map(arg -> arg.getName())
+                            .collect(Collectors.toSet());
+
+            // Check if there is any argument which has default value but not provided by client, then add it.
+            availableArgTypes.stream()
+                            .filter(argType -> !clientArgNames.contains(argType.getName()))
+                            .filter(argType -> argType.getDefaultValue() != null)
+                            .map(argType -> Argument.builder()
+                                            .name(argType.getName())
+                                            .value(argType.getDefaultValue())
```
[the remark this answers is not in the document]
rishi-aga: Argument only has name and value. (https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/request/Argument.java#L18)

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -319,13 +329,45 @@ private Path buildPath(Type rootEntityType, String selector) {
...
                 entityType = fieldType;
             }
             return new Path(path);
         }
 
+        private void parseArguments(String attributeName, Set<Argument> arguments) {
+            Matcher matcher = FILTER_ARGUMENTS_PATTERN.matcher(attributeName);
+            while (matcher.find()) {
+                String[] pair = matcher.group(1).split(":", 2);
+                if (pair.length < 2) {
```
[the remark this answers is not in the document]
rishi-aga: `.split(":", 2)` wont return more than 2

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -65,7 +68,8 @@
     private static final String SINGLE_PARAMETER_ONLY = "There can only be a single filter query parameter";
     private static final String INVALID_QUERY_PARAMETER = "Invalid query parameter: ";
     private static final Pattern TYPED_FILTER_PATTERN = Pattern.compile("filter\\[([^\\]]+)\\]");
-    private static final Pattern FILTER_ARGUMENTS_PATTERN = Pattern.compile("\\[(.+?)\\]");
+    // field name followed by zero or more square brackets having non-empty argument name and value separated by ':'
```
aklish, line 71: Let's put in an example in the comment like [grain:day][foo:bar] to aid understanding

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -65,7 +68,8 @@
...
+    private static final String FILTER_SELECTOR_REGEX = "(\\w+)(\\[(\\w+?):([^\\]]+?)\\])*$";
```
aklish, line 72: Do we need the ? marks here?

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
```diff
@@ -336,17 +352,16 @@ private Path buildPath(Type rootEntityType, String selector) {
...
         }
 
-        private void parseArguments(String attributeName, Set<Argument> arguments) {
-            Matcher matcher = FILTER_ARGUMENTS_PATTERN.matcher(attributeName);
+        private void parseArguments(String argsString, Set<Argument> arguments) throws UnsupportedEncodingException {
+            if (argsString == null || argsString.isEmpty()) {
+                return;
+            }
+
+            Matcher matcher = Pattern.compile("\\[(\\w+?):([^\\]]+?)\\]").matcher(argsString);
```
aklish, line 360: Can we use the original regex here?

aklish (changes requested):

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
rishi-aga: it will work correctly without `?` also so removing them.

elide-core/src/main/java/com/yahoo/elide/core/filter/dialect/RSQLFilterDialect.java
rishi-aga: it doesn't work here  as the number of `[]` is unknown

aklish (approved):

PR #1882, opened 2021-03-09
Correct exception logging

wcekan:
## Description
Always re-throw or re-interrupt when catching the `InterruptedException`.

When the final argument in a slf4j log message is a throwable, it generates a stack trace and is not one of the formatting arguments.
It appears the code had intended to display only the .toString based on the unused `{}`

> 05:46:51.561 [qtp135639164-244] ERROR c.y.e.a.service.AsyncExecutorService.executeQuery(AsyncExecutorService.java:96) - TimeoutException: {}
java.util.concurrent.TimeoutException: null
	at java.util.concurrent.FutureTask.get(FutureTask.java:205)
	at com.yahoo.elide.async.service.AsyncExecutorService.executeQuery(AsyncExecutorService.java:85)
	at com.yahoo.elide.async.hooks.AsyncAPIHook.executeAsync(AsyncAPIHook.java:100)
	at com.yahoo.elide.async.hooks.AsyncAPIHook.executeHook(AsyncAPIHook.java:67)
	at com.yahoo.elide.async.hooks.AsyncQueryHook.execute(AsyncQueryHook.java:38)
	at com.yahoo.elide.async.hooks.AsyncQueryHook.execute(AsyncQueryHook.java:28)

## Motivation and Context
Incorrect logging

## How Has This Been Tested?
Unit testing

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncAPIUpdateOperation.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncAPIUpdateOperation.java
@@ -40,10 +40,10 @@ public class AsyncAPIUpdateOperation implements Runnable {
             asyncAPIDao.updateAsyncAPIResult(queryResultObj, queryObj.getId(), queryObj.getClass());
 
         } catch (InterruptedException e) {
-            log.error("InterruptedException: {}", e);
+            log.error("InterruptedException: {}", e.toString());
             asyncAPIDao.updateStatus(queryObj.getId(), QueryStatus.FAILURE, queryObj.getClass());
         } catch (Exception e) {
-            log.error("Exception: {}", e);
+            log.error("Exception: {}", e.toString());
             asyncAPIDao.updateStatus(queryObj.getId(), QueryStatus.FAILURE, queryObj.getClass());
         }
     }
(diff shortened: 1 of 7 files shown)


elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncAPIUpdateOperation.java
```diff
@@ -40,10 +40,10 @@ public void run() {
             asyncAPIDao.updateAsyncAPIResult(queryResultObj, queryObj.getId(), queryObj.getClass());
 
         } catch (InterruptedException e) {
-            log.error("InterruptedException: {}", e);
+            log.error("InterruptedException: {}", e.toString());
```
wcekan, line 43: If the stack trace is intended, we need to instead remove the `{}` from the formatting
```java
log.error("InterruptedException", e);
```

moizarafat: stacktrace was not intended.

moizarafat (approved):

elide-async/src/main/java/com/yahoo/elide/async/operation/AsyncAPIUpdateOperation.java
aklish: What is the purpose of:
 Thread.currentThread().interrupt();
?

PR #1894, opened 2021-03-10
Cleanup IDE warnings

wcekan:
## Description
Code cleanup
* Diamonds on CharType
* Remove unnecessary casts
* Remove unused methods
* Precompile replace patterns
* Removed dead code for impossible else

## Motivation and Context
Code quality

## How Has This Been Tested?
Unit testing

## License

### DIFF
--- a/elide-swagger/src/main/java/com/yahoo/elide/swagger/SwaggerBuilder.java
+++ b/elide-swagger/src/main/java/com/yahoo/elide/swagger/SwaggerBuilder.java
@@ -767,12 +767,10 @@ public class SwaggerBuilder {
         for (Type<?> clazz : allClasses) {
             if (clazz instanceof ClassType) {
                 models.putAll(converters.readAll(((ClassType) clazz).getCls()));
-            } else if (clazz instanceof Type) {
+            } else {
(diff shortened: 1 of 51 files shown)


moizarafat (approved):

elide-swagger/src/main/java/com/yahoo/elide/swagger/SwaggerBuilder.java
```diff
@@ -767,12 +767,10 @@ public Swagger build() {
         for (Type<?> clazz : allClasses) {
             if (clazz instanceof ClassType) {
                 models.putAll(converters.readAll(((ClassType) clazz).getCls()));
-            } else if (clazz instanceof Type) {
+            } else {
```
aklish, line 770: Not sure about this removal.  The type here is the elide Type - not the superclass of Class.

aklish (approved):

elide-swagger/src/main/java/com/yahoo/elide/swagger/SwaggerBuilder.java
wcekan: `Type<?> clazz` the expression is always true.

PR #1921, opened 2021-03-17
Agg Store Model Changes for Parameterized Metric Support

moizarafat:
## Description
FunctionArgument model changes as per the RFC: https://github.com/yahoo/elide/discussions/1829
MetricFunction renamed to Function.
Added Function as a property to Table, Column, Dimensions, Timedimension, Metrics.

## Motivation and Context
Parameterized Metrics: https://github.com/yahoo/elide/discussions/1829

## How Has This Been Tested?
Existing and New Test cases pass

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Dimension.java
```diff
@@ -7,15 +7,44 @@
...
     public Dimension(Table table, String fieldName, EntityDictionary dictionary) {
-        super(table, fieldName, dictionary);
+        super(table, fieldName, dictionary, getConstructFunction(table, fieldName, dictionary));
+    }
+
+    protected static ConstructFunction getConstructFunction(Table table, String fieldName,
+            EntityDictionary dictionary) {
+        return () -> {
+            Type<?> tableClass = dictionary.getEntityClass(table.getName(), table.getVersion());
+            DimensionFormula formula = dictionary.getAttributeOrRelationAnnotation(
```
aklish, line 32: Formula is never used here.  We shouldn't need to  look it up.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/models/Dimension.java
```diff
@@ -7,15 +7,44 @@
...
  */
 @Include(rootLevel = false, type = "dimension")
 @EqualsAndHashCode(callSuper = true)
 public class Dimension extends Column {
     public Dimension(Table table, String fieldName, EntityDictionary dictionary) {
-        super(table, fieldName, dictionary);
+        super(table, fieldName, dictionary, getConstructFunction(table, fieldName, dictionary));
+    }
+
+    protected static ConstructFunction getConstructFunction(Table table, String fieldName,
```
aklish, line 28: I don't see a difference here between this construction mechanism and the one in Metric or TimeDimension.  Maybe we can remove this responsibility from these different classes and just construct this in Column?

aklish (changes requested):

aklish (approved):

PR #1954, opened 2021-03-23
Replace all occurrence of '{{}}' within grain expression

rishi-aga:
Replace all occurrence of '{{}}' within grain expression

## License


aklish (approved):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/query/SQLTimeDimensionProjection.java
```diff
@@ -69,7 +69,7 @@ public SQLTimeDimensionProjection(TimeDimension column,
     @Override
     public String toSQL(Queryable source, SQLReferenceTable table) {
         //TODO - We will likely migrate to a templating language when we support parameterized metrics.
-        return grain.getExpression().replaceFirst(TIME_DIMENSION_REPLACEMENT_REGEX,
+        return grain.getExpression().replaceAll(TIME_DIMENSION_REPLACEMENT_REGEX,
```
wcekan, line 72: Should we make this a compiled Pattern?

PR #1957, opened 2021-03-24
Parameterized Metrics: Aggregation Store Model Changes Set 2

moizarafat:
## Description
Continuing with Model changes for Parameterized Metrics.
i) Introduced EntityArguments in EntityDictionary and EntityBinding to capture Table Arguments.
ii) Graqhql Models, EntityProjection etc. capture Entity/Table Arguments and make it query accessible from graphql.
iii) TableMeta, DimensionFormula, MetricFormula now capture the Arguments. 
iv) Table Arguments and Column Arguments are now going to be available to pass thru grapqhl.
v) Miscellanous changes to populate and propagate arguments.

## Motivation and Context
https://github.com/yahoo/elide/discussions/1829

## How Has This Been Tested?
Existing Test pass. New Test cases to be added later.

## License


rishi-aga (dismissed):

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -477,6 +477,27 @@ public GraphQLInputObjectType classToInputObject(Type<?> clazz) {
...
 
+    /**
+     * Build an Argument list object for the given entity
+     * @param entityClass The Entity class to which this attribute belongs to.
+     * @param fetcher The data fetcher to associated with the newly created GraphQL Query Type
+     * @param dictionary The dictionary that contains the runtime type information for the parent class.
+     * @return Newly created GraphQLArgument Collection for the given entity
+     */
+    public List<GraphQLArgument> entityArgumentToQueryObject(Type<?> entityClass,
+                                                                DataFetcher fetcher,
```
aklish, line 488: Fetcher is unused here.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -177,6 +177,7 @@ public GraphQLSchema build() {
                     .argument(pageFirstArgument)
                     .argument(pageOffsetArgument)
                     .argument(buildInputObjectArgument(clazz, true))
+                    .argument(generator.entityArgumentToQueryObject(clazz, dataFetcher, entityDictionary))
```
aklish, line 180: We should do this for relationships as well - not just the root objects.

aklish (changes requested):
Overall changes look good.  There are no tests.  Were you going to add those in another PR?  If not, we need unit tests and a few IT tests for all of these changes.

aklish (approved):

PR #1960, opened 2021-03-25
MaxAsyncAfterSeconds Default Value Assignment

### LINKED ISSUE #1949: Incorrect Error Received: Invalid Async After Seconds during Cancellation of Async
Incorrect Error is being received when cancelling the Async Request

## Expected Behavior
The CANCELLATION should happen without any error.

## Current Behavior
CANCELLATION Throws an Error: ` Invalid Async After Seconds`

## Possible Solution
Currently in ASYNCAPI models we default the AsyncAfterSeconds to 10. So if a user overrides the default time in Application.yaml to anything lower than 10 seconds we get this error. AsyncAfterSeconds is a transient field so during the update the user will not be passing its value again so a default value is used again.

Instead of hardcoding the default value to 10 in the model. we should have a getMethod that checks if the user has set a value in application.yaml. If there is a value set there then use it else use 10.

User has supplied a value at Query Time 
```
--- Yes --> Use That
--- No --> Check for default value in Application.Yaml/Standalone Properties.
           ---> Value Found --> Yes ---> Use that as the default Value
                                         --> No ---> 10 seconds 
```

To make this to work for both spring and standalone, I recommend we add the user-supplied Configuration value to AsyncExecutorService and then retrieve from there in the model.  

This will be a temporary solution till we make all the properties from Application.Yaml/Standalone Properties available in the `ElideSettings` object

1. In `application.yaml` set `maxAsyncAfterSeconds: 1`
2. Run a mutation to cancel an existing query: 
`mutation ($id: ID) {
  asyncQuery(op: UPDATE, data: {id: $id, status: CANCELLED}) {
    edges {
      node {
        id
        status
        __typename
      }
      __typename
    }
    __typename
  }
}`
## Context
A bug raised by Yavin User (Kevin Hinterlong)

moizarafat:
Resolves #1949 

## Description
Fix for "Invalid Max Async After Seconds" error when the configured MaxAsyncAfterSeconds in application.yaml or standalone  settings was less than 10.

## Motivation and Context
Resolves incorrect error message

## How Has This Been Tested?
Existing test case pass.
`
## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncAPIHook.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncAPIHook.java
@@ -9,6 +9,8 @@ import static com.yahoo.elide.annotation.LifeCycleHookBinding.Operation.CREATE;
 import static com.yahoo.elide.annotation.LifeCycleHookBinding.Operation.READ;
 import static com.yahoo.elide.annotation.LifeCycleHookBinding.TransactionPhase.POSTCOMMIT;
 import static com.yahoo.elide.annotation.LifeCycleHookBinding.TransactionPhase.PRESECURITY;
+
+import com.yahoo.elide.ElideSettings;
 import com.yahoo.elide.annotation.LifeCycleHookBinding;
 import com.yahoo.elide.async.models.AsyncAPI;
 import com.yahoo.elide.async.models.AsyncAPIResult;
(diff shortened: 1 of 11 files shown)


rishi-aga (approved):

elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncAPIHook.java
```diff
@@ -58,6 +60,10 @@ protected void validateOptions(AsyncAPI query, RequestScope requestScope) {
      */
     protected void executeHook(LifeCycleHookBinding.Operation operation, LifeCycleHookBinding.TransactionPhase phase,
             AsyncAPI query, RequestScope requestScope, Callable<AsyncAPIResult> queryWorker) {
+        if (query.getAsyncAfterSeconds() == null) {
```
aklish, line 63: Shouldn't this block go inside the block below?

elide-async/src/main/java/com/yahoo/elide/async/hooks/AsyncAPIHook.java
```diff
@@ -58,6 +60,10 @@ protected void validateOptions(AsyncAPI query, RequestScope requestScope) {
...
+            query.setAsyncAfterSeconds(elideSettings.getMaxAsyncAfterSeconds());
+        }
+
```
aklish, line 66: We need a much better explanation (add a lot of details) why the hook is written this way.  It appears wrong to the casual reader.

aklish (changes requested):

PR #1965, opened 2021-03-28
Issue 1724 embedded objects filters

### LINKED ISSUE #1724: Support filters & sorting on embedded object attributes.
Imagine classes like:

```java
class House {
  @Embedded Address address 
}

class Address {
   Integer streetNumber
   String street;
   City city;
   State state;
}```

Elide should support filters and sorting on 'house.address.street' or 'house.address.city.name'

This will require the following changes:
- The entity dictionary will need to understand fields of nested objects that are not themselves entities (similar to how we do schema introspection for GraphQL).
- Filter and Sort paths must allow navigation through both attributes (as well as relationships).
- Filter dialects will need to allow navigation through both attributes (as well as relationships).
- In memory filter and sorting may require changes.
- JPQL query generation may require changes.

### PR DESCRIPTION
**Pull request by donanton-dehipitiarachchi**
Resolves #1724 

## Description
This is an initial version to get review feedback. Only the filtering aspect is implemented. Definitely can further improve and refactor. Used exception handling to branch out to NonEnitityDictionary. Many not be the best approach. Let me know your comments.

## Motivation and Context
Added filtering capability with embedded objects

## How Has This Been Tested?
Added a unit test class  EmbeddedFilterIT in elide-integration-tests module.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/ElideSettingsBuilder.java
+++ b/elide-core/src/main/java/com/yahoo/elide/ElideSettingsBuilder.java
@@ -10,6 +10,7 @@ import com.yahoo.elide.core.audit.AuditLogger;
 import com.yahoo.elide.core.audit.Slf4jLogger;
 import com.yahoo.elide.core.datastore.DataStore;
 import com.yahoo.elide.core.dictionary.EntityDictionary;
+import com.yahoo.elide.core.dictionary.NonEntityDictionary;
 import com.yahoo.elide.core.exceptions.HttpStatus;
 import com.yahoo.elide.core.filter.dialect.RSQLFilterDialect;
 import com.yahoo.elide.core.filter.dialect.graphql.FilterDialect;
@@ -42,6 +43,7 @@ public class ElideSettingsBuilder {
     private AuditLogger auditLogger;
     private JsonApiMapper jsonApiMapper;
     private EntityDictionary entityDictionary = new EntityDictionary(new HashMap<>());
+    private NonEntityDictionary nonEntityDictionary = new NonEntityDictionary();
     private Function<RequestScope, PermissionExecutor> permissionExecutorFunction = ActivePermissionExecutor::new;
     private List<JoinFilterDialect> joinFilterDialects;
     private List<SubqueryFilterDialect> subqueryFilterDialects;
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: elide-core/src/main/java/com/yahoo/elide/ElideSettingsBuilder.java [hunk 1]
```diff
@@ -125,6 +128,11 @@ public ElideSettingsBuilder withEntityDictionary(EntityDictionary entityDictiona
         return this;
     }
 
+    public ElideSettingsBuilder withNonEntityDictionary(NonEntityDictionary nonEntityDictionary) {
```
aklish, line 131: Another possibility here is to embed NonEntityDictionary inside EntityDictionary itself - and add some new methods in EntityDictionary that allow lookups of entities & non-entities and their fields alike. 

Then those methods can be used in contexts where we don't care if the entity is an elide model or not (like path traversal).

I'm not sure what will work better in practice so I would try both and see which one is cleaner.

elide-integration-tests/src/test/java/com/yahoo/elide/tests/EmbeddedFilterIT.java
```diff
@@ -0,0 +1,125 @@
...
+        tx.commit(null);
+        tx.close();
+    }
+
+    @Test
+    public void testJsonApiFilterByNumber() throws JsonProcessingException {
+        JsonNode responseBody = mapper.readTree(
+                given()
+                        .when()
+                        .get("/house?filter=address.number=='1409'")
```
aklish, line 84: The IT tests are run against multiple stores (in-memory, hibernate, and JPA). 

By default (run inside the IDE), they run against the in-memory store.  Do these tests work against the JPA store?

aklish (changes requested):
The build is failing for checkstyle errors.  

You'll want to 'mvn clean install' on the root of the project and fix these plus any other build errors.

elide-core/src/main/java/com/yahoo/elide/ElideSettingsBuilder.java
donanton-dehipitiarachchi: I like this idea. let me try this approach as well.

lamtd81:
Hello, any update on this issue? I am trying to get the embedded objects filtering to work, but it looks like the code from this PR is too out of date to be used in place of version 5.0.6.

Any chance someone is still looking into this? I would gladly help but I'm not sure where to start; I'm not very familiar with GitHub and I can't seem to find a way to merge the master branch into this PR in order to get it running.

aklish:
> Hello, any update on this issue? I am trying to get the embedded objects filtering to work, but it looks like the code from this PR is too out of date to be used in place of version 5.0.6.
> 
> Any chance someone is still looking into this? I would gladly help but I'm not sure where to start; I'm not very familiar with GitHub and I can't seem to find a way to merge the master branch into this PR in order to get it running.

Unfortunately, the work for this PR is not complete, and there is still a lot to do to enable this feature.   It is on my top 5 list of things we need to add to Elide though.  Let me inquire around if anyone is able to shepherd or complete this PR.

aklish:
We are actively working on a new fix for this.

PR #1970, opened 2021-03-28
elide-datastore-aggregation: Fix build warnings

### LINKED ISSUE #1902: Fix elide-datastore-aggregation build warnings:
Fix the following build warnings for the aggregation data store:

```
[WARNING] /sd/workspace/src/github.com/yahoo/elide/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/Time.java:[27,1] Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
```

and 

```
[WARNING] /sd/workspace/src/github.com/yahoo/elide/elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/PlayerStats.java:[36,1] Generating equals/hashCode implementation but without a call to superclass, even though this class does not extend java.lang.Object. If this is intentional, add '@EqualsAndHashCode(callSuper=false)' to your type.
```

moizarafat:
Resolves #1902

## Description
Fix Build warnings with Lombok.

## Motivation and Context
Resolve build warnings for Hacktogether

## How Has This Been Tested?
Existing Tests pass. Build warnings are gone.

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/Time.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/Time.java
@@ -10,7 +10,9 @@ import com.yahoo.elide.core.type.Type;
 import com.yahoo.elide.core.utils.coerce.converters.ElideTypeConverter;
 import com.yahoo.elide.core.utils.coerce.converters.Serde;
 import com.yahoo.elide.datastores.aggregation.metadata.enums.TimeGrain;
-import lombok.Data;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
 
 import java.time.LocalDateTime;
 import java.time.ZoneOffset;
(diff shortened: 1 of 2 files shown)


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/timegrains/Time.java
```diff
@@ -24,7 +26,9 @@
 /**
  * Time date type for all analytic model time dimensions.
  */
-@Data
+@RequiredArgsConstructor
```
moizarafat, line 29: Since we provide our own implementation of Hash, Equals and ToString using `@Data` annotation does not seem fit.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/example/PlayerStats.java
```diff
@@ -33,7 +33,7 @@
  */
 @Include
 @VersionQuery(sql = "SELECT COUNT(*) from playerStats")
-@EqualsAndHashCode
+@EqualsAndHashCode(callSuper = false)
```
moizarafat, line 36: Using `true`, fails the test cases. So setting to `false` which is equivalent to default `skip`.

aklish (approved):

Issue #1892, opened 2021-03-09
Fix elide-model-config build warnings


aklish:
Fix the following build warnings in elide-model-config:

```
[WARNING] /sd/workspace/src/github.com/yahoo/elide/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Dimension.java:[64,21] @Builder will ignore the initializing expression entirely. If you want the initializing expression to serve as default, add @Builder.Default. If it is not supposed to be settable during building, make the field final.
[WARNING] /sd/workspace/src/github.com/yahoo/elide/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Dimension.java:[67,20] @Builder will ignore the initializing expression entirely. If you want the initializing expression to serve as default, add @Builder.Default. If it is not supposed to be settable during building, make the field final.
[WARNING] /sd/workspace/src/github.com/yahoo/elide/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Dimension.java:[80,25] @Builder will ignore the initializing expression entirely. If you want the initializing expression to serve as default, add @Builder.Default. If it is not supposed to be settable during building, make the field final.
[WARNING] /sd/workspace/src/github.com/yahoo/elide/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Dimension.java:[84,25] @Builder will ignore the initializing expression entirely. If you want the initializing expression to serve as default, add @Builder.Default. If it is not supposed to be settable during building, make the field final.
[WARNING] /sd/workspace/src/github.com/yahoo/elide/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Dimension.java:[88,25] @Builder will ignore the initializing expression entirely. If you want the initializing expression to serve as default, add @Builder.Default. If it is not supposed to be settable during building, make the field final.
[WARNING] /sd/workspace/src/github.com/yahoo/elide/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Table.java:[67,21] @Builder will ignore the initializing expression entirely. If you want the initializing expression to serve as default, add @Builder.Default. If it is not supposed to be settable during building, make the field final.
[...]

PR #1971, opened 2021-03-29
Fix elide-model-config build warnings

rishi-aga:
Resolves https://github.com/yahoo/elide/issues/1892

## Description
Fix Build warnings for elide-model-config

## How Has This Been Tested?
Existing tests pass.

## License

### DIFF
--- a/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Argument.java
+++ b/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Argument.java
@@ -48,6 +48,7 @@ public class Argument implements Named {
 
     @JsonProperty("values")
     @JsonDeserialize(as = LinkedHashSet.class)
+    @Builder.Default
     private Set<String> values = new LinkedHashSet<>();
 
     @JsonProperty("tableSource")
--- a/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Dimension.java
+++ b/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Dimension.java
@@ -14,7 +14,6 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
-import lombok.Singular;
 
 import java.util.ArrayList;
 import java.util.LinkedHashSet;
@@ -62,9 +61,11 @@ public class Dimension implements Named {
     private String category;
 
     @JsonProperty("hidden")
+    @Builder.Default
     private Boolean hidden = false;
 
     @JsonProperty("readAccess")
+    @Builder.Default
     private String readAccess = "Prefab.Role.All";
 
     @JsonProperty("definition")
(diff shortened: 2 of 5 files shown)


elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Measure.java
```diff
@@ -74,10 +75,11 @@
 
     @JsonProperty("tags")
     @JsonDeserialize(as = LinkedHashSet.class)
+    @Builder.Default
     private Set<String> tags = new LinkedHashSet<>();
 
     @JsonProperty("arguments")
-    @Singular
+    @Builder.Default
```
wcekan, line 82: Why are we replacing Singular?

rishi-aga: 👍

aklish (approved):

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Table.java
```diff
@@ -68,10 +66,10 @@
     private String dbConnectionName;
 
     @JsonProperty("isFact")
-    private Boolean isFact = true;
+    private Boolean isFact;
```
wcekan, line 69: @aklish are this fields purposely modifiable after building?  Or should they all be final?
```suggestion
    private final Boolean isFact;
```

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Table.java
```diff
@@ -51,7 +50,6 @@
 @Data
```
wcekan, line 50: Use builder for Jackson deserializer
```suggestion
@JsonDeserialize(builder = Table.TableBuilder.class)
@Data
```

wcekan (changes requested):
Refactor builders using JsonDeserialize and remove no arg constructors.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Argument.java
```diff
@@ -33,7 +32,6 @@
 @Data
```
wcekan, line 32:
```suggestion
@JsonDeserialize(builder = Argument.ArgumentBuilder.class)
@Data
```

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Dimension.java
```diff
@@ -78,23 +76,32 @@
...
 
+    public Dimension() {
+        this.hidden = false;
+        this.readAccess = "Prefab.Role.All";
+        this.grains = new ArrayList<>();
+        this.tags = new LinkedHashSet<>();
+        this.values = new LinkedHashSet<>();
+        this.arguments = new ArrayList<>();
+    }
+
```
wcekan, line 104:
```suggestion
```

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Table.java
rishi-aga: I think we cannot declare all fields final, because we have inheritance logic

PR #1979, opened 2021-04-01
Populate table context at service start.

rishi-aga:

## Description
Populate table context at service start.
Formula Validation using Handlebars.

## License


elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java
```diff
@@ -1095,6 +1095,28 @@ public void bindEntity(EntityBinding entityBinding) {
...
     }
 
+    /**
+     * Return a list of field name for given annotations.
+     *
+     * @param entityClass     the entity class
+     * @param annotationClassList given annotation types
+     * @return Set of field names having given annotation.
+     */
+    public Set<String> getFieldNamesWithAnnotations(Type<?> entityClass,
```
aklish, line 1105: Any time we add a method to the dictionary, we should add a corresponding test in EntityDictionaryTest.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/ColumnContext.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Column Context for Handlebars Resolution.
+ */
+@Data
+@AllArgsConstructor
+public class ColumnContext {
+    private final String definition;
+    private final String path;
+    private final Map<String, Object> defaultColumnArgs;
```
aklish, line 26: I believe argument defaults will always be substituted in by the parser (GraphQL in this case), so argument defaults should not need special treatment.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/ColumnContext.java
```diff
@@ -0,0 +1,46 @@
...
+
+    @Override
+    public String toString() {
+
+        String expr = definition;
+        for (String reference : resolveFormulaReferences(expr)) {
+            String referenceWithPath = reference;
+            if (!path.isEmpty()) {
+                if (reference.startsWith(SQL_HELPER_PREFIX)) {
+                    referenceWithPath = reference + " path='" + path + "'";
```
aklish, line 36: Not sure what this replacement is trying to accomplish.  Can you add a comment with the example text that is being swapped in each case?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTableTest.java
```diff
@@ -77,4 +81,28 @@ public void testJoin() {
...
+        Map<String, Object> tableDefaultArgsContext = tableContext.getDefaultTableArgs();
+
+        // For primary table, {{$$table.args.beginDate}} should resolve to "2020-01-01"
+        Map<String, Object> tableArgs =
+                        (Map<String, Object>) ((Map<String, Object>) tableDefaultArgsContext.get("$$table"))
+                                        .get("args");
+        assertEquals(1, tableArgs.size());
+        assertEquals("2020-01-01", tableArgs.get("beginDate"));
+
+        // For 'playerRank' column, {{$$column.args.beginDate}} should resolve to "1"
```
aklish, line 99: replace beginDate with playerRank

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
```diff
@@ -31,18 +32,27 @@
...
 import java.util.stream.Collectors;
 
+import javax.persistence.JoinColumn;
+
 /**
  * Table stores all resolved physical reference and join paths of all columns.
  */
 public class SQLReferenceTable {
     public static final String PERIOD = ".";
+    public static final String DOLLAR = "$$";
```
aklish, line 51: DOLLAR and DOUBLE_DOLLAR are the same.  This looks like a typo.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
```diff
@@ -122,6 +144,7 @@ public void resolveAndStoreAllReferencesAndJoins(Queryable queryable) {
         //References and joins are stored by their source that produces them (rather than the query that asks for them).
         Queryable key = queryable.getSource();
         SQLDialect dialect = queryable.getSource().getConnectionDetails().getDialect();
+        boolean isNested = queryable.isNested();
```
aklish, line 147: This method is getting pretty long.  We might want to break it up.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
```diff
@@ -134,12 +157,39 @@ public void resolveAndStoreAllReferencesAndJoins(Queryable queryable) {
...
+                TableContext tableCtx = new TableContext(key.getAlias(),
+                                                         dialect,
+                                                         prepareColumnArgsMap(table.getArguments(), TBL_PREFIX));
+                globalTablesContext.put(key, tableCtx);
+                Set<Type<?>> visitedTableContext = new HashSet<>();
+                populateTableContext(entityClass, tableCtx, StringUtils.EMPTY, visitedTableContext);
+            } else {
+                TableContext tableCtx = new TableContext(key.getAlias(),
+                                                         dialect,
+                                                         Collections.emptyMap());
```
aklish, line 175: With nesting, arguments maybe required in either the inner or outer queries.  We'll want to preserve them so they are available for substitution at all levels.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
```diff
@@ -134,12 +157,39 @@ public void resolveAndStoreAllReferencesAndJoins(Queryable queryable) {
...
+        }
+
         SQLJoinVisitor joinVisitor = new SQLJoinVisitor(metaDataStore);
 
         queryable.getColumnProjections().forEach(column -> {
-            // validate that there is no reference loop
-            validator.visitColumn(queryable, column);
+            if (!isNested) {
+                // validate that there is no reference loop
+                FormulaValidator validator = new FormulaValidator(globalTablesContext.get(key));
```
aklish, line 190: With optimization, joins can occur in the outer or inner queries - hence a loop can appear in an outer query.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
```diff
@@ -194,6 +244,56 @@ public void resolveAndStoreAllReferencesAndJoins(Queryable queryable) {
...
+                    Set<Type<?>> visitedEntities) {
+
+        Set<Type<?>> entities = new HashSet<Type<?>>(visitedEntities);
+        entities.add(entityClass);
+
+        SQLTable table = (SQLTable) metaDataStore.getTable(entityClass);
+        table.getColumns().forEach(column -> {
+            tableCtx.put(column.getName(), new ColumnContext(column.getExpression(),
+                                                             prefixPath,
+                                                             prepareColumnArgsMap(column.getArguments(), COL_PREFIX)));
```
aklish, line 257: I don't understand the $$column prefix here.  Shouldn't a ColumnContext not care about its handlebars prefix?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/FormulaValidator.java
```diff
@@ -5,95 +5,104 @@
...
-            } else {
-                ColumnProjection referenceColumn = source.getColumnProjection(reference);
-
-                // if the reference is to a logical column, check it
-                if (referenceColumn != null && !reference.equals(column.getName())) {
-                    visitColumn(source, referenceColumn);
+            while (!template.collect(TagType.VAR).isEmpty()) {
+                String resolved = template.apply(tableCtx);
+                current = getHandlebarVariables(resolved);
+                if (!visited.add(current)) {
```
aklish, line 81: I'm not sure this is bulletproof.  What if the references are in a different order in the string?  

A better way to validate might iterate over all the columns in a fact table and resolve them - and if the same context object and key come up twice - throw an exception.  This could be implemented by a subclass of the TableContext method that overrides the get call.

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/HandlebarResolvor.java
```diff
@@ -0,0 +1,128 @@
...
+
+import java.io.IOException;
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Matcher;
+
+public class HandlebarResolvor {
```
moizarafat, line 28: Resolver

elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java
rishi-aga: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/ColumnContext.java
rishi-aga: Removed the `path` property from ColumnDefinition, its working without it as well

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
rishi-aga: Removed it.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/FormulaValidator.java
rishi-aga: Implemented using a different approach. With overriding get method in a subclass, I was facing the issue if the expression is `{{name}} {{name}}` then it's not a cycle but same context object and key come up twice.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
rishi-aga: Added validation for nested queries.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
rishi-aga: 👍

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/ColumnContext.java
rishi-aga: when a logical column refers to another logical column which is not in query, then we will need it

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/FormulaValidator.java
```diff
@@ -5,95 +5,128 @@
...
+
+        Set<String> visited = new HashSet<>(alreadyVisited);
+
+        // Physical References starts with $
+        if (columnName.lastIndexOf('$') == 0) {
+            return;
         }
 
-        Type<?> tableClass = dictionary.getEntityClass(source.getName(), source.getVersion());
+        if (!visited.add(tblCtx.getAlias() + PERIOD + columnName)) {
```
aklish, line 78: if I have a cycle A -> B -> A
won't my aliases be:
A,
A.B
A.B.A
And no cycle found?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
```diff
@@ -194,6 +281,26 @@ public void resolveAndStoreAllReferencesAndJoins(Queryable queryable) {
         return projections;
     }
 
+    private void addJoinContextsforCurrentContext(Queryable rootQueryable, TableContext tableCtx) {
+
+        Type<?> rootEntityClass = dictionary.getEntityClass(rootQueryable.getName(), rootQueryable.getVersion());
+
+        for (String joinField : dictionary.getFieldNamesWithAnnotations(rootEntityClass,
+                        Arrays.asList(Join.class, JoinColumn.class))) {
+            Type<?> joinClass = dictionary.getType(rootEntityClass, joinField);
```
aklish, line 290: For a toMany join, this will return the collection type instead of the join type.  use getParameterizedType instead.

aklish, line 290: Let's use new SQLJoin class instead of going directly to the dictionary here.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/TableContext.java
```diff
@@ -0,0 +1,90 @@
...
+@Getter
+@ToString
+@Builder
+public class TableContext extends HashMap<String, Object> {
+
+    private final String alias;
+    private final SQLDialect dialect;
+    private final Map<String, Object> defaultTableArgs;
+    @Builder.Default
+    private final Map<String, TableContext> joins = new HashMap<>();
```
aklish, line 34: Instead of storing TableContext objects here for joins, I wonder if it would be clearer if we simply store the SQLTable directly in the context (which now also has a map of SQLJoin objects).  We are building a context object anyway each time.  It's not intuitive why the context copies are necessary - and reusing existing metadata concepts will remove this distinction.

Similarly, we might no longer need ColumnContext (because we have a reference to the underlying column).   My point is that we already have metadata objects - it might be clearer if we use them directly here.

While this suggestion may work nicely for tables, it may not work for nested queries.  

It might be good revisit the metadata design a bit here.  Let's discuss.

aklish (changes requested):

PR #1998, opened 2021-04-12
Formula Validation using ExpressionParser

rishi-aga:

## Description
Formula Validation using ExpressionParser.

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/expression/ExpressionParser.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/expression/ExpressionParser.java
@@ -6,6 +6,9 @@
 
 package com.yahoo.elide.datastores.aggregation.queryengines.sql.expression;
 
+import static com.yahoo.elide.core.utils.TypeHelper.nullOrEmpty;
+import static com.yahoo.elide.datastores.aggregation.queryengines.sql.metadata.SQLReferenceTable.PERIOD;
+
 import com.yahoo.elide.core.Path;
 import com.yahoo.elide.core.dictionary.EntityDictionary;
 import com.yahoo.elide.core.type.Type;
@@ -14,10 +17,16 @@ import com.yahoo.elide.datastores.aggregation.metadata.MetaDataStore;
 import com.yahoo.elide.datastores.aggregation.metadata.enums.ColumnType;
 import com.yahoo.elide.datastores.aggregation.query.ColumnProjection;
 import com.yahoo.elide.datastores.aggregation.query.Queryable;
-
+import com.github.jknack.handlebars.EscapingStrategy;
+import com.github.jknack.handlebars.Handlebars;
+import com.github.jknack.handlebars.Helper;
+import com.github.jknack.handlebars.Options;
+import com.github.jknack.handlebars.Template;
 import com.google.common.base.Preconditions;
 
+import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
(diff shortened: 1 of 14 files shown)


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/expression/ExpressionParser.java
```diff
@@ -50,6 +88,17 @@ public ExpressionParser(MetaDataStore store) {
...
 
         for (String referenceName : referenceNames) {
+
+            // Change `sql from='joinName' column='columnName[a1:v1][a2:v2]'` to `joinName.columnName`
+            if (referenceName.startsWith(SQL_HELPER_PREFIX)) {
+                try {
+                    Template template = handlebars.compileInline(toFormulaReference(referenceName));
+                    referenceName = template.apply(Collections.emptyMap());
+                } catch (IOException e) {
+                    // Do Nothing
```
aklish, line 98: Maybe throw an IllegalStateException here?

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
```diff
@@ -69,13 +69,16 @@ protected SQLReferenceTable(MetaDataStore metaDataStore, Set<Queryable> queryabl
...
-        queryables.stream().forEach(queryable -> {
-            Queryable next = queryable;
-            do {
-                resolveAndStoreAllReferencesAndJoins(next);
-                next = next.getSource();
-            } while (next.isNested());
-        });
+        queryables
+           .stream()
+           .filter(queryable -> queryable.isNested() || queryable.isRoot())
```
aklish, line 74: What's this filter for?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/custom/DailyAverageScorePerPeriod.java
```diff
@@ -56,7 +56,7 @@ public QueryPlan resolve(Query query) {
                 .metricProjection(SQLMetricProjection.builder()
                         .alias(getAlias())
                         .name(getName())
-                        .expression(getExpression())
+                        .expression("AVG({{$highScore}})")
```
aklish, line 59: Isn't this the same as before?  Why the change?

aklish (changes requested):

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/custom/DailyAverageScorePerPeriod.java
rishi-aga: added `$`

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/metadata/SQLReferenceTable.java
rishi-aga: Wanted to exclude adding those queryable for Dynamic tables which are already there in Static Table, Static Table's queryable already get the preference.
Expression Parser was failing for the scenario: When it is a Query over SQLTable, then Query's ColumnProjection may reference a logical column which is defined in SQLTable but not in Query. It fails at: https://github.com/yahoo/elide/blob/master/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/expression/ExpressionParser.java#L81

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/query/SQLColumnProjection.java
```diff
@@ -41,6 +41,14 @@ default String toSQL(Queryable source, SQLReferenceTable lookupTable) {
     default boolean canNest(Queryable source, SQLReferenceTable lookupTable) {
         SQLDialect dialect = source.getConnectionDetails().getDialect();
         String sql = toSQL(source.getSource(), lookupTable);
+
+        // TODO: Remove this when switching to TableContext
+        // Resolved Reference would be empty if value is not provided in @MetricFormula as value is optional.
+        // Once we change this to use TableContext, then it should be able to resolve expression directly.
+        if (this instanceof SQLMetricProjection && sql.equals("")) {
```
aklish, line 48: Instead of this hack here, why not just change DailyAverageScorePerPeriod to override canNest to true.

aklish (approved):

PR #2016, opened 2021-04-22
Inmemory security filtercheck

Chandrasekar-Rajasekar:
## Description
Bypass filter expression check if it is executed in datastore. Add visitor that wraps all filter expression check and change its state as it traverses through the expression. 
Add normalization visitor for permission expression. 

## Motivation and Context
FilterCheck that already executed in datastore should not be checked against the records returned by the datastore. With aggregation datastore, this will always evaluate to false if the filter path element is not present in group by dimensions. 

## How Has This Been Tested?
Yet to Test

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/security/checks/DatastoreEvalFilterExpressionCheck.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/security/checks/DatastoreEvalFilterExpressionCheck.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2021, Yahoo Inc.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+
+package com.yahoo.elide.core.security.checks;
+
+import com.yahoo.elide.core.security.ChangeSpec;
+import com.yahoo.elide.core.security.RequestScope;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.Optional;
+
+public class DatastoreEvalFilterExpressionCheck<T> extends OperationCheck<T> {
+    @Getter @Setter
+    private boolean executedInMemory;
+    private boolean negated = false;
+    @Getter
+    private FilterExpressionCheck<T> wrappedFilterExpressionCheck;
+
+    public DatastoreEvalFilterExpressionCheck(FilterExpressionCheck<T> wrappedFilterExpressionCheck) {
+        this.wrappedFilterExpressionCheck = wrappedFilterExpressionCheck;
+        this.executedInMemory = false;
+    }
+
+    @Override
+    public boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
+        if (executedInMemory) {
+            return negated ^ wrappedFilterExpressionCheck.ok(object, requestScope, changeSpec);
+        }
+        return true;
+    }
+
+    public void negate() {
+        negated = !negated;
+    }
+
+    @Override
+    public String toString() {
+        return "DatastoreEvalFilterExpressionCheck{" +
+                "executedInMemory=" + executedInMemory +
+                ", negated=" + negated +
+                ", wrappedFilterExpressionCheck=" + wrappedFilterExpressionCheck.getClass() +
+                '}';
+    }
+}
--- a/elide-core/src/main/java/com/yahoo/elide/core/security/checks/DatastoreEvalFilterExpressionCheck.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/security/checks/DatastoreEvalFilterExpressionCheck.java
@@ -40,10 +40,10 @@ public class DatastoreEvalFilterExpressionCheck<T> extends OperationCheck<T> {
 
     @Override
     public String toString() {
-        return "DatastoreEvalFilterExpressionCheck{" +
-                "executedInMemory=" + executedInMemory +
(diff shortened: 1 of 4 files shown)


elide-core/src/main/java/com/yahoo/elide/core/security/checks/DatastoreEvalFilterExpressionCheck.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+        return true;
+    }
+
+    public void negate() {
+        negated = !negated;
+    }
+
+    @Override
+    public String toString() {
```
aklish, line 42: The class name (DatastoreEvalFilterExpressionCheck) will get logged in the 403 response if the user turns on verbose errors.  That's not what we want in this case.  We need to explore all the different logging paths and make sure there is no regression there.

elide-core/src/main/java/com/yahoo/elide/core/security/checks/DatastoreEvalFilterExpressionCheck.java
```diff
@@ -0,0 +1,49 @@
...
+import com.yahoo.elide.core.security.RequestScope;
+
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.Optional;
+
+public class DatastoreEvalFilterExpressionCheck<T> extends OperationCheck<T> {
+    @Getter @Setter
+    private boolean executedInMemory;
```
aklish, line 19: executedInMemory is past tense which makes it sounds like this has already happened.  I believe the intent here it to mark whether this expression _should_ run in memory when evaluated. 

I would change this to 'executeInMemory'.

elide-core/src/main/java/com/yahoo/elide/core/security/visitors/InmemoryPermissionExpressionVisitor.java
```diff
@@ -0,0 +1,143 @@
...
+            return FALSE_EXPRESSION_PAIR;
+        }
+        if (Objects.equals(expressionPair, FALSE_EXPRESSION_PAIR)) {
+            return TRUE_EXPRESSION_PAIR;
+        }
+
+        if (expressionPair.getKey() instanceof CheckExpression) {
+            CheckExpression checkExpression = (CheckExpression) expressionPair.getKey();
+            if (checkExpression.getCheck() instanceof DatastoreEvalFilterExpressionCheck) {
+                ((DatastoreEvalFilterExpressionCheck) checkExpression.getCheck()).negate();
```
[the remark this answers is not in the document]
aklish: Better idea I think: If you invert DatastoreEvalFilterExpressionCheck to something like:

```java
class WrappedFilterExpressionCheck extends OperationCheck {
  @Getter @Setter
  boolean executedInDatabase;
  FilterExpressionCheck wrapped;
  @Override
  public boolean ok(T object, RequestScope requestScope, Optional<ChangeSpec> changeSpec) {
    if (executedInDatabse) {
       return true;
    }
    return wrapped.ok(object, requestScope, changeSpec);
  }
} 
```

Then you don't need the negate method at all.  When you evaluate the expression in memory, the not operator will do the negation for you.

aklish (changes requested):

elide-core/src/main/java/com/yahoo/elide/core/security/visitors/InmemoryPermissionExpressionVisitor.java
```diff
@@ -0,0 +1,143 @@
...
+import org.apache.commons.lang3.tuple.ImmutablePair;
+import org.apache.commons.lang3.tuple.Pair;
+
+import java.util.Objects;
+import java.util.function.Function;
+
+/**
+ * Traverse the Expression Parser and returns the Permission Expression and boolean Pair.
+ */
+public class InmemoryPermissionExpressionVisitor implements ExpressionVisitor<Pair<Expression, Boolean>> {
```
aklish, line 31: I'm not sure we need this boolean on the right.  Each individual expression should know if it needs to run in memory or not.

aklish (changes requested):

elide-core/src/main/java/com/yahoo/elide/core/security/visitors/ExecuteFilterCheckInMemoryVisitor.java
```diff
@@ -0,0 +1,53 @@
...
+            ((DatastoreEvalFilterExpressionCheck) check).setExecutedInMemory(true);
+        }
+        return checkExpression;
+    }
+
+    @Override
+    public Expression visitAndExpression(AndExpression andExpression) {
+        andExpression.getLeft().accept(this);
+        andExpression.getRight().accept(this);
+        return andExpression;
```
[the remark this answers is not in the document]
Chandrasekar-Rajasekar: Should be the same. This visitor does not modify any object reference. It just sets the in-memory boolean value. So the object references in the expression tree should still be the same.

elide-core/src/main/java/com/yahoo/elide/core/security/visitors/InmemoryPermissionExpressionVisitor.java
Chandrasekar-Rajasekar: If we don't have a boolean, then we have to include the executeInMemory or executeInDatastore variable in Check Expression. Then change it for all types of checks in the Check expression.

PR #2020, opened 2021-04-22
Resolve arguments in FromSubquery

rishi-aga:
Resolve arguments in FromSubquery

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/query/QueryTranslator.java
```diff
@@ -54,13 +58,21 @@ public QueryTranslator(SQLReferenceTable referenceTable, SQLDialect sqlDialect)
...
+        NativeQuery.NativeQueryBuilder builder = null;
 
         if (query.isNested()) {
-            NativeQuery innerQuery = builder.build();
+            NativeQuery innerQuery = query.getSource().accept(this).build();
 
             builder = NativeQuery.builder().fromClause("(" + innerQuery + ") AS "
                     + applyQuotes(query.getSource().getAlias()));
+        } else {
+            String fromClause = getFromClause(query.getSource());
```
aklish, line 69: This else condition belongs in the other visitQueryable function (which handles the root scenario).

aklish (changes requested):

PR #2040, opened 2021-04-29
Test Cases + Enable Namespace in Standalone and Spring

### LINKED ISSUE #2010: RFC: Analytic Model Namespaces
# Overview

It would be convenient to build a hosted Elide service with multiple analytic models published by different teams.  These models need to be grouped together by subject area for better navigation.

Elide should introduce a new Namespace concept for organizing tables.

# HJSON Directory Layout

Namespaces will be created in a new namespace directory:

```
CONFIG_ROOT/
  ├── models/
  |  ├── tables/
  |  |  ├── model1.hjson
  |  |  ├── model2.hjson
  |  ├── namespaces/
  |  |  ├── namespace1.hjson
  |  |  ├── namespace2.hjson
  |  ├── security.hjson
  |  └── variables.hjson
  ├── db/
  |  ├── sql/
  |  |  ├── db1.hjson
  |  ├── variables.hjson
```

Namespaces and namespace files are optional.

# HJSON Configuration

Namespace files will have the following properties:

```
{
  name: MyNamespace
  friendlyName: 'My Namespace'
  readAccess: 'Admin'  #Default table permissions.  Can be overridden in any given table.
  description: 'A description of the MyNamespace namespace'
}
```

Table files will have a new, optional namespace attribute.  The attribute should be set to the name of a namespace.  Each table can only belong to a single namespace.  If no namespace is defined, a table belongs to the 'default' namespace.
```
{ 
  name: PlayerStats
  namespace: MyNamespace
  table: playerStats
  dbConnectionName: Presto Data Source
  ...
}
```

Table joins will also have a new, optional namespace attribute:

```
joins: [
 {
   name: playerCountry
   namespace: Foo
   to: PlayerCountry
   kind: toOne
   type: left
   definition: '{{playerCountry.id}} = {{country_id}}'
  }
]
```

If no namespace attribute is defined, the join refers to the table in the 'default' namespace.

# Metadata

Elide will introduce a new metadata model, Namespace:

```java
@Include(type = "namespace")
class Namespace {
    @Id
    private final String id;

    private final String name;

    private final String friendlyName;

    private final String description;

    @OneToMany
    private Set<Table> tables;
}
```

The Table metadata model will have a new relationship to its Namespace:

```java
@Include(type = "table")
class Table {
    ...

    @OneToOne
    private Namespace namespace;
}
```

The table ID will be modified to include the namespace.

# Default Namespace

The default namespace properties can be changed by creating a HJSON namespace with the name "default" - case insensitive.

moizarafat:
Resolves #2010 

## Description
i) Enable Namespace HJSON Config parsing in Elide Standalone and Spring.
ii) Test cases for Agg Store Changes to add namespace metadata

## Motivation and Context
#2010

## How Has This Been Tested?
Existing and new test cases pass

## License

### DIFF
--- a/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/jsonformats/ElideNamespaceNameFormatAttr.java
+++ b/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/jsonformats/ElideNamespaceNameFormatAttr.java
@@ -26,9 +26,13 @@ public class ElideNamespaceNameFormatAttr extends AbstractFormatAttribute {
 
     public static final String FORMAT_NAME = "elideNamespaceName";
     public static final String FORMAT_KEY = "elideNamespaceName.error.format";
-    public static final String FORMAT_MSG =
+    public static final String FORMAT_ADDITIONAL_KEY = "elideNamespaceName.error.additional";
+    public static final String FORMAT_KEY_MSG =
                     "Name [%s] is not allowed. Name must start with an alphabetic character and can include "
                     + "alaphabets, numbers and '_' only. Name must not be a variant of 'default' with different cases.";
+    public static final String FORMAT_ADDITIONAL_KEY_MSG =
+            "Name [%s] clashes with the 'default' namespace. Either change the case or pick a different namespace "
+            + "name.";
     public static final String DEFAULT_NAME = "default";
 
     public ElideNamespaceNameFormatAttr() {
@@ -45,7 +49,7 @@ public class ElideNamespaceNameFormatAttr extends AbstractFormatAttribute {
         }
 
         if (!input.equals(DEFAULT_NAME) && input.toLowerCase(Locale.ENGLISH).equals(DEFAULT_NAME)) {
-            report.error(newMsg(data, bundle, FORMAT_KEY).putArgument("value", input));
+            report.error(newMsg(data, bundle, FORMAT_ADDITIONAL_KEY).putArgument("value", input));
         }
     }
 }
--- a/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/jsonformats/ElideNamespaceNameFormatAttr.java
+++ b/elide-model-config/src/main/java/com/yahoo/elide/modelconfig/jsonformats/ElideNamespaceNameFormatAttr.java
@@ -29,7 +29,7 @@ public class ElideNamespaceNameFormatAttr extends AbstractFormatAttribute {
     public static final String FORMAT_ADDITIONAL_KEY = "elideNamespaceName.error.additional";
     public static final String FORMAT_KEY_MSG =
                     "Name [%s] is not allowed. Name must start with an alphabetic character and can include "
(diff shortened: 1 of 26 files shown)


elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/NamespaceConfig.java
```diff
@@ -39,19 +41,23 @@
...
-        this.name = name;
-        this.friendlyName = friendlyName;
-        this.description = description;
-        this.apiVersion = apiVersion;
+    /**
+     * Returns description of the namespace object.
+     * If null, returns the name.
+     * @return description
+     */
+    public String getDescription() {
```
moizarafat, line 60: Keeping Same as Table.

aklish (changes requested):

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/jsonformats/ElideNamespaceNameFormatAttr.java
```diff
@@ -0,0 +1,51 @@
...
+ * </p>
+ */
+public class ElideNamespaceNameFormatAttr extends AbstractFormatAttribute {
+    public static final Pattern NAME_FORMAT_REGEX = ElideNameFormatAttr.NAME_FORMAT_REGEX;
+
+    public static final String FORMAT_NAME = "elideNamespaceName";
+    public static final String FORMAT_KEY = "elideNamespaceName.error.format";
+    public static final String FORMAT_MSG =
+                    "Name [%s] is not allowed. Name must start with an alphabetic character and can include "
+                    + "alaphabets, numbers and '_' only. Name must not be a variant of 'default' with different cases.";
```
aklish, line 31: Can these be two different messages for the two different error scenarios?  Also, the second condition is a little hard to follow.  Maybe say something like:  'Name [%s] clashes with the 'default' namespace.  Either change the case or pick a different namespace name.'

elide-model-config/src/test/java/com/yahoo/elide/modelconfig/validator/DynamicConfigValidatorTest.java
```diff
@@ -58,7 +58,7 @@ public void testValidNamespace() throws Exception {
 
         assertEquals("default", child.getNamespace()); //PlayerStatsChild -> no namespace was provided, so defaulted
         assertEquals("PlayerNamespace", parent.getNamespace());
-        assertEquals("DEfault", referred.getNamespace()); // Namespace in HJson "DEfault". Matched case insensitively with "default" namespace
+        assertEquals("default", referred.getNamespace()); // Namespace in HJson "DEfault". Matched case insensitively with "default" namespace
```
aklish: Comment is wrong.

aklish (changes requested):

aklish (approved):

PR #2047, opened 2021-04-30
Add ability to handle conversion of Date instances to ISO8601DateSerde

### LINKED ISSUE #2045: Comparison between java.util.Date & java.sql.Timestamp errors out in specific cases
Inmemory comparison between java.util.Date & java.sql.Timestamp throws an exception when the ISO8601DateSerde is registered to handle these.

## Expected Behavior
The comparison in such cases should not throw an exception

## Current Behavior
Observed in cases where the incoming string filter is converted into Date and compared with a Timestamp, currently the conversion fails since the registered serde does not handle Date to Timestamp conversion.
```java
Convert type 'java.util.Date' value 'Mon Sep 14 22:00:00 EDT 2020' to type 'java.sql.Timestamp'
 Using converter com.yahoo.elide.utils.coerce.CoerceUtil$1@50fa99c0
java.lang.ClassCastException: class java.util.Date cannot be cast to class java.lang.String (java.util.Date and java.lang.String are in module java.base of loader 'bootstrap')
	at com.yahoo.elide.utils.coerce.converters.ISO8601DateSerde.deserialize(ISO8601DateSerde.java:18)
	at com.yahoo.elide.utils.coerce.CoerceUtil$1.convert(CoerceUtil.java:71)
```

blitzmohit:
Resolves #2045

## Description
ISO8601DateSerde can now handle conversion from the different instances of Date class as well.

## Motivation and Context
Details in #2045

## How Has This Been Tested?
Added unit test

## License


elide-core/src/test/java/com/yahoo/elide/core/utils/coerce/converters/ISO8601DateSerdeTest.java
```diff
@@ -66,6 +67,14 @@ public void testSQLTimestampDeserialization() throws Exception {
         assertEquals(expected, actual);
     }
 
+    @Test
+    public void testDateToSQLTimestampDeserialization() throws Exception {
+        ISO8601DateSerde serde =
+                new ISO8601DateSerde("yyyy-MM-dd", TimeZone.getTimeZone("UTC"), java.sql.Timestamp.class);
+        java.sql.Timestamp expected = new Timestamp(0);
+        java.sql.Timestamp actual = (java.sql.Timestamp) serde.deserialize(new Date(0));
```
wcekan, line 75: Timestamp already imported.
```suggestion
        Timestamp expected = new Timestamp(0);
        Timestamp actual = (Timestamp) serde.deserialize(new Date(0));
```

wcekan, line 75: Timestamp already imported.
```suggestion
        ISO8601DateSerde serde =
                new ISO8601DateSerde("yyyy-MM-dd", TimeZone.getTimeZone("UTC"), Timestamp.class);
        Timestamp expected = new Timestamp(0);
        Timestamp actual = (Timestamp) serde.deserialize(new Date(0));
```

elide-core/src/main/java/com/yahoo/elide/core/utils/coerce/converters/ISO8601DateSerde.java
```diff
@@ -15,7 +15,7 @@
 /**
  * Serializes ISO8601 Dates to Strings and vice versa.
  */
-public class ISO8601DateSerde implements Serde<String, Date> {
+public class ISO8601DateSerde implements Serde<Object, Date> {
```
aklish, line 18: Will this require the registration of a new Serde to work around the issue you are seeing or will it work with the default Serde?

blitzmohit: It works with the default serde from what I tested.

What are your thoughts about modifying the Coerce Util lookup to consider the incoming class type as well cause the reason it was failing earlier was being picked up for <Date, Timestamp> when clearly it was registered for <String, Timestamp> since only the targetType is used for deciding what serde to use.

aklish: I like the idea of matching both types.  Serdes were originally only intended to convert from primitive types (that are in request/response payloads) to more complex things.   We use CoerceUtil in many other places though now - so it probably ought to consider both types.

aklish (approved):

aklish (approved):

PR #2050, opened 2021-05-01
Refactored AnyFieldExpression logic

aklish:
Changes AnyField permission logic from Elide 3 semantics to match current documentation.

## License


Chandrasekar-Rajasekar (approved):

elide-core/src/main/java/com/yahoo/elide/core/security/permissions/PermissionExpressionBuilder.java
```diff
@@ -211,19 +211,47 @@ private Expression buildAnyFieldExpression(final PermissionCondition condition,
...
             allFieldsExpression = new OrExpression(allFieldsExpression, fieldExpression);
         }
 
-        return new AnyFieldExpression(condition, entityExpression, allFieldsExpression);
+        if (! fieldExpressionUsed) {
+            //If there are no permissions, allow access...
+            if (entityExpression == null) {
+                return SUCCESSFUL_EXPRESSION;
+            } else {
+                return new AnyFieldExpression(condition, entityExpression);
```
wcekan, line 250: I can fix this in a later PR.
```java
            if (entityExpression == null) {
                return SUCCESSFUL_EXPRESSION;
            }
            return new AnyFieldExpression(condition, entityExpression);
```

PR #2051, opened 2021-05-02
Source sparse fields from entity projection

aklish:
Some logic in Elide that examines JSON-API sparse fields is not applied for GraphQL:

- Determining whether a collection can be paginated.
- Building security filters from only the set of requested fields (not the entire entity model).
.
This PR changes Elide core to leverage the EntityProjection to build the list of requested fields (rather than sourcing from sparse fields from the request header) allowing the same logic to be leveraged for both JSON-API and GraphQL

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -1790,6 +1790,7 @@ public class PersistentResource<T> implements com.yahoo.elide.core.security.Pers
         Set<String> filteredSet = new LinkedHashSet<>();
         for (String field : fields) {
             try {
+
                 if (checkIncludeSparseField(requestScope.getSparseFields(), typeName, field)) {
                     checkFieldAwareReadPermissions(field);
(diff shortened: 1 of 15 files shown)


elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -91,7 +91,7 @@
     private final DataStoreTransaction transaction;
     private final RequestScope requestScope;
     private int hashCode = 0;
-
+    public static final Set<String> ALL_FIELDS = null;
```
wcekan, line 94: While you are updating the code, can you add the `<T>` argument to type definition above?
```java
private final Type<T> type;
```

aklish: Makes sense - but I think that should be a separate PR.  Small PRs are easier to fit in smaller time blocks for review.

Chandrasekar-Rajasekar (approved):

Chandrasekar-Rajasekar (approved):

PR #2067, opened 2021-05-04
Expose query params on security.RequestScope

Brutus5000:
Resolves https://github.com/yahoo/elide/discussions/1985

## Description
This changes exposes the already existing queryParams to the security.RequestScope interface.

## Motivation and Context
LifeCycleHooks only expose the security.RequestScope. This change makes the queryParams visible there as well.

## How Has This Been Tested?
It's implicitly tested as we just extend an already tested interface.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/security/RequestScope.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/security/RequestScope.java
@@ -5,8 +5,8 @@
  */
 package com.yahoo.elide.core.security;
 
-import java.util.Optional;
-import javax.ws.rs.core.MultivaluedMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * The request scope interface passed to checks.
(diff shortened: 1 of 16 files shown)


elide-core/src/main/java/com/yahoo/elide/core/security/RequestScope.java
```diff
@@ -13,4 +16,5 @@
     String getApiVersion();
     String getRequestHeaderByName(String headerName);
     String getBaseUrlEndPoint();
+    Optional<MultivaluedMap<String, String>> getQueryParams();
```
aklish, line 19: Let's make this Map<String, List<String>> so we can remove the dependency on javax.ws.rs.core from this interface.

Also - what about an empty map denoting optional?

Brutus5000: I dropped the Optional throughout the code. Hope that's what you had in mind.

aklish (approved):

PR #2074, opened 2021-05-06
httpclient 4.5.13

wcekan:
## Description
Avoid CVE-2020-13956
```
httpclient-4.5.3.jar (pkg:maven/org.apache.httpcomponents/httpclient@4.5.3, cpe:2.3:a:apache:httpclient:4.5.3:*:*:*:*:*:*:*) : CVE-2020-13956
```

## Motivation and Context
Security

## How Has This Been Tested?
Unit testing

## License


elide-datastore/elide-datastore-aggregation/pom.xml
```diff
@@ -104,6 +104,12 @@
             </exclusions>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
```
aklish, line 108: What requires this?

wcekan: avatica-core

wcekan: changed scope to runtime to match the dependencies of calcite

aklish: We don't use calcite to make HTTP connections though.  Is there any reason we can't simply exclude it?

wcekan: No.  If we mention `calcite` without any mention of `httpclient` the downstream will pull in the `calcite` old version.  They have the option to exclude it, but sadly `maven` does not let us exclude on their behalf.

aklish (approved):

elide-datastore/elide-datastore-aggregation/pom.xml
aklish: Can we add a comment:

Avoid CVE-2020-13956

aklish (changes requested):

aklish (approved):

PR #2086, opened 2021-05-10
Add AggregationStorePermissionExecutor

### LINKED ISSUE #2048: RFC: Aggregation Store Security Rules
# Overview

Analytic queries differ from CRUD queries in some fundamental ways.  The most important is that analytic queries group data by the projected fields.  This means that the results
returned will never hydrate non-projected fields.  

Security rules expressed on filters can include filters on fields that are not projected in the result.  This means those rules can never be evaluted in-memory in the Elide service.

Analytic queries also only support data reads, so Create, Update, Delete, and Transfer permissions do not apply.

# Semantics and Restrictions

Given the differences between CRUD and analytic query behavior, Elide will enforce the following restrictions on model security permissions:
1.  Fields can only be decorated with User checks.
2.  Tables can only be decorated with any combination (AND, OR, NOT) of User Checks and Filter Expression Checks.
3.  Table permissions override namespace permisions.
4.  Field permissions and table permissions are evaluated differently.

## Table Perimssion Rules

A client can see a record in the table if the following rule is true: `(entityRule AND (field1Rule OR field2Rule ... OR fieldNRule)`

## Field Permission Rules

A client can see a field in a record if the field user check evalutes to true.  The absence of a user check implies access.

# Stages of Load

When a load is called, the following security checks occur in Elide:

## Avoiding the Load

The first step Elide takes is to determine if the client can read anything from the collection at all by evaluating user checks alone.
If the user checks fail here, a 403 error is returned and nothing is read.

Elide calls:
```java
permissionExecutor.checkUserPermissions(resourceClass, ReadPermission.class);
```
  Like the `ActivePermissionExecutor`, only the user checks are evaluated using the following formula:

`(entityRule AND (field1Rule OR field2Rule ... OR fieldNRule)`
                      
## Filtering the records in the database
                      
If the load cannot be avoided, Elide generates a security filter expression rule to push to the database by calling:
                      
```java               
permissionExecutor.getReadPermissionFilter(loadClass);
```

The `AggregationStorePermissionExecutor` will only look at the ReadPermission rule on the table (or namespace if table is undefined) and extract the filter
expression (if any) that can be executed.

## Filtering the records in memory

After Elide loads records from the database, it attempts to check if the client can see each one.  This is a two step process.
The first step evalutes a logical disjunction of all the field level permission rules by calling on each record:

```java
permissionExecutor.checkPermission(ReadPermission.class, record);
```

Given the restrictions of the `AggregationStorePermissionExecutor`, no checks are required to prune rows at this stage because:
1. Either the load was aborted because user checks failed.
2. No filtering was done because the user checks passed.
3. Filtering was performed and any rows returned would have passed the filter rules.

This call is effectively a NOOP.

## Verifying client filters for access to fields

The second stage determines if the client filtered on a field they cannot see by calling the following on each field referenced in the filter:

```java
permissionExecutor.checkSpecificFieldPermissions(resource, null, ReadPermission.class, fieldName);
```

Given that:
1. There are no relationships in the Aggregation store (so no relationships to traverse)
2. Fields are restricted to user checks

the above call simply needs to execute a user check on the specific class & field.

## Pruning fields from the response that the client cannot see

The final step Elide takes is to remove field from the model that the client cannot see:

```java
permissionExercutor.checkSpecificFieldPermissions(resource, null, ReadPermission.class, fieldName);
```

This call performs the same action as the prior step.

Chandrasekar-Rajasekar:
Resolves #2048 

## Description
Add AggregationStorePermissionExecutor for all models managed by Aggregation Datastore

## License


elide-core/src/main/java/com/yahoo/elide/core/security/executors/AggregationStorePermissionExecutor.java
```diff
@@ -31,13 +31,39 @@ public AggregationStorePermissionExecutor(RequestScope requestScope) {
...
+     * @param requestedFields the list of requested fields
+     * @param <A>
+     * @return ExpressionResult - result of the above any field expression
+     */
     @Override
     public <A extends Annotation> ExpressionResult checkPermission(Class<A> annotationClass,
                                                                    PersistentResource resource,
                                                                    Set<String> requestedFields) {
-        return ExpressionResult.PASS;
+        Expression expression = expressionBuilder.buildUserCheckAnyExpression(
```
aklish, line 47: Can we add a Preconditions check or maybe return FAIL if annotationClass is anything other than a ReadPermission for all of these checks?

elide-core/src/main/java/com/yahoo/elide/core/security/permissions/PermissionExpressionBuilder.java
```diff
@@ -171,6 +173,40 @@ public PermissionExpressionBuilder(ExpressionResultCache cache, EntityDictionary
...
+                                                                                       Set<String> requestedFields,
+                                                                                       final RequestScope scope) {
+
+        final Function<Check, Expression> leafBuilderFn = (check) ->
+                new CheckExpression(check, null, scope, null, cache);
+
+        ParseTree classPermissions = entityDictionary.getPermissionsForClass(resourceClass, annotationClass);
+        Expression entityExpression = normalizedExpressionFromParseTree(classPermissions, leafBuilderFn);
+
+        Expression anyFiledExpression = buildAnyFieldExpression(
```
aklish, line 199: anyField instead of anyFiled.

aklish (changes requested):

elide-core/src/main/java/com/yahoo/elide/core/security/executors/AggregationStorePermissionExecutor.java
```diff
@@ -0,0 +1,165 @@
...
+            return ExpressionResult.FAIL;
+        }
+
+        Expression expression = expressionBuilder.buildUserCheckAnyExpression(
+                resource.getResourceType(),
+                annotationClass,
+                requestedFields,
+                requestScope);
+
+        return executeExpressions(expression, annotationClass, Expression.EvaluationMode.ALL_CHECKS);
```
aklish, line 58: This should actually call checkPermissions with expressionExecutor set to Optional.empty(). That's how user only checks are evaluated.

aklish: Alternatively, call checkOnlyUserPermissions like elsewhere in this file.

elide-core/src/main/java/com/yahoo/elide/core/security/executors/AggregationStorePermissionExecutor.java
```diff
@@ -0,0 +1,165 @@
...
+                expressionBuilder.buildUserCheckEntityAndAnyFieldExpression(
+                        resourceClass,
+                        annotationClass,
+                        requestedFields,
+                        requestScope);
+
+        return checkOnlyUserPermissions(
+                resourceClass,
+                annotationClass,
+                Optional.empty(),
```
aklish, line 126: I found a bug in the user check cache in Elide 5.  The user check cache will look at either one field or no fields (this case).  However, this check could be called multiple times with different fields.  We really ought to make the cache always map to a set of requested fields.  Optional.empty() isn't right here, but we can't do the right thing until this bug is fixed.

elide-core/src/test/java/com/yahoo/elide/core/security/AggregationStorePermissionExecutorTest.java
```diff
@@ -0,0 +1,191 @@
...
+
+        // evaluated expression = user none -> ForbiddenAccess
+        Assertions.assertThrows(
+                ForbiddenAccessException.class,
+                () -> executor.checkUserPermissions(ClassType.of(Model.class), ReadPermission.class, "metric"));
+
+
+        // evaluated expression = (user none or filter check)
+        Assertions.assertEquals(
+                ExpressionResult.DEFERRED,
```
aklish, line 115: This doesn't seem right to me.  The filter check should never apply at the field level.

aklish: The field level check should not inherit the entity level check here.

elide-core/src/test/java/com/yahoo/elide/core/security/AggregationStorePermissionExecutorTest.java
```diff
@@ -0,0 +1,191 @@
...
+                executor.checkSpecificFieldPermissions(
+                        new PersistentResource(new Model("dim1", 0, 1), "1", scope),
+                        null, ReadPermission.class, "metric2"));
+
+
+        // evaluated expression = (user none or filter check) AND (user all OR user none)
+        Assertions.assertEquals(
+                ExpressionResult.DEFERRED,
```
aklish, line 123: I'm not sure this tests makes sense either.  Let's discuss.  I don't think DEFERRED is a valid return scenario.

elide-core/src/test/java/com/yahoo/elide/core/security/AggregationStorePermissionExecutorTest.java
```diff
@@ -0,0 +1,191 @@
...
+        @ReadPermission(expression = "user none or filter check")
+        class Model1 {
+            String filterDim;
+            long metric;
+        }
+
+        com.yahoo.elide.core.RequestScope scope = bindAndgetRequestScope(Model1.class);
+        PermissionExecutor executor = scope.getPermissionExecutor();
+
+        FilterExpression expression = executor.getReadPermissionFilter(ClassType.of(Model1.class), Collections.emptySet()).orElse(null);
```
aklish, line 150: To make these tests realistic, the set of requested fields should not be empty - but rather set to the all the fields.

aklish (changes requested):

elide-core/src/main/java/com/yahoo/elide/core/security/executors/AggregationStorePermissionExecutor.java
```diff
@@ -0,0 +1,165 @@
...
+     */
+    @Override
+    public <A extends Annotation> ExpressionResult checkPermission(Class<A> annotationClass,
+                                                                   PersistentResource resource,
+                                                                   Set<String> requestedFields) {
+        if (!annotationClass.equals(ReadPermission.class)) {
+            return ExpressionResult.FAIL;
+        }
+
+        Expression expression = expressionBuilder.buildUserCheckAnyExpression(
```
aklish, line 52: I'm not sure this is what you want here from the Permission Builder.  UserCheckAnyExpression will look at a combination of field and entity for each field.

elide-core/src/test/java/com/yahoo/elide/core/security/AggregationStorePermissionExecutorTest.java
```diff
@@ -110,9 +109,9 @@ public long getMetric() {
                 () -> executor.checkUserPermissions(ClassType.of(Model.class), ReadPermission.class, "metric"));
 
 
-        // evaluated expression = (user none or filter check)
+        // evaluated expression = null -> false
         Assertions.assertEquals(
-                ExpressionResult.DEFERRED,
+                ExpressionResult.PASS,
```
aklish, line 114: Shouldn't this be FAIL?

elide-core/src/test/java/com/yahoo/elide/core/security/AggregationStorePermissionExecutorTest.java
```diff
@@ -130,6 +129,13 @@ public long getMetric() {
                         new PersistentResource(new Model("dim1", 0, 1), "1", scope),
                         new HashSet<>(Arrays.asList("filterDim", "metric"))));
 
+        // evaluated expression = (user none OR null)
+        Assertions.assertEquals(
+                ExpressionResult.PASS,
```
aklish, line 134: Shouldn't this be FAIL?

aklish (changes requested):

aklish (approved):

PR #2088, opened 2021-05-12
add validation for security checks in agg model table, namespace and fields

### LINKED ISSUE #2048: RFC: Aggregation Store Security Rules

Chandrasekar-Rajasekar:
Resolves #2048

## Description
Have validations for all security checks in aggregation datastore. 

- Fields can only be decorated with User checks.
- Tables can only be decorated with any combination (AND, OR, NOT) of User Checks and Filter Expression Checks.

## Motivation and Context
Restricting all checks that could run in memory after the records are returned from the aggregation datastore.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java
@@ -196,6 +196,23 @@ public class EntityDictionary {
      * or {@code @ReadPermission(expression="Prefab.Common.UpdateOnCreate")})
      * @param checks a map that links the identifiers used in the permission expression strings
      *               to their implementing classes
+     * @param roleChecks a map that links the user check identifiers used in the permission expression strings
+     *      *               to their User Check object
+     * @param injector a function typically associated with a dependency injection framework that will
+     *                 initialize Elide models.
+     */
+    public EntityDictionary(Map<String, Class<? extends Check>> checks,
+                            Map<String, UserCheck> roleChecks, Injector injector) {
+        this(checks, roleChecks, injector, CoerceUtil::lookup, Collections.emptySet());
+    }
+
+    /**
+     * Instantiate a new EntityDictionary with the provided set of checks and an injection function.
+     * In addition all of the checks * in {@link com.yahoo.elide.core.security.checks.prefab} are mapped
+     * to {@code Prefab.CONTAINER.CHECK} * (e.g. {@code @ReadPermission(expression="Prefab.Role.All")}
+     * or {@code @ReadPermission(expression="Prefab.Common.UpdateOnCreate")})
+     * @param checks a map that links the identifiers used in the permission expression strings
+     *               to their implementing classes
      * @param injector a function typically associated with a dependency injection framework that will
      *                 initialize Elide models.
      * @param entitiesToExclude Set of Models to exclude from Binding.
@@ -215,9 +232,17 @@ public class EntityDictionary {
             Injector injector,
             Function<Class, Serde> serdeLookup,
             Set<Type<?>> entitiesToExclude) {
+        this(checks, null, injector, serdeLookup, entitiesToExclude);
+    }
+
+    public EntityDictionary(Map<String, Class<? extends Check>> checks,
+                            Map<String, UserCheck> roleChecks,
+                            Injector injector,
+                            Function<Class, Serde> serdeLookup,
+                            Set<Type<?>> entitiesToExclude) {
         this.serdeLookup = serdeLookup;
         this.checkNames = Maps.synchronizedBiMap(HashBiMap.create(checks));
-        this.roleChecks = new HashMap<>();
+        this.roleChecks = roleChecks == null ? new HashMap<>() : roleChecks;
         this.apiVersions = new HashSet<>();
         initializeChecks();
         this.injector = injector;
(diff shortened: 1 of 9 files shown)


elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java
```diff
@@ -189,6 +189,23 @@ public EntityDictionary(Map<String, Class<? extends Check>> checks, Injector inj
...
 
+    /**
+     * Instantiate a new EntityDictionary with the provided set of checks and an injection function.
+     * In addition all of the checks * in {@link com.yahoo.elide.core.security.checks.prefab} are mapped
+     * to {@code Prefab.CONTAINER.CHECK} * (e.g. {@code @ReadPermission(expression="Prefab.Role.All")}
+     * or {@code @ReadPermission(expression="Prefab.Common.UpdateOnCreate")})
+     * @param checks a map that links the identifiers used in the permission expression strings
+     *               to their implementing classes
+     * @param roleChecks a map that links the user check identifiers used in the permission expression strings
+     *      *               to their User Check object
```
aklish, line 200: A few typeos:
1. Look like there are asterisks not starting new lines.
2. Change "In addition all of the checks" to "In addition, all of the checks".

elide-integration-tests/src/test/java/com/yahoo/elide/initialization/IntegrationTest.java
```diff
@@ -47,6 +49,7 @@
 
     /* Shared between the test setup code (to insert test data) as well as the Jetty server (to serve test data) */
     protected static DataStoreTestHarness dataStoreHarness;
+    protected static Set<String> userRoles = new HashSet<>();
```
aklish, line 52: Do these have to be static?

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/integration/MetaDataStoreIntegrationTest.java
```diff
@@ -40,6 +40,11 @@
 
 public class MetaDataStoreIntegrationTest extends IntegrationTest {
 
+    public MetaDataStoreIntegrationTest() {
+        super();
+        userRoles.addAll(VALIDATOR.getElideSecurityConfig().getRoles());
```
aklish, line 45: Aggregation data store already has a solution to add role checks to the dictionary.  Rather than create a new mechanism here, we can just copy the existing mechanism for this test file.

aklish (changes requested):

aklish (approved):

PR #2095, opened 2021-05-14
Propogate client query arguments to join tables

rishi-aga:

## Description
Propagate client query arguments to join tables
Refactor: Getting column projection from queryable's source itself, this column projection will have default/available arguments for the column.  So no need to go-to source for default/available arguments again.

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/ColumnContext.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/ColumnContext.java
@@ -211,17 +211,26 @@ public class ColumnContext extends HashMap<String, Object> {
         throw new HandlebarsException(new Throwable("Couldn't find: " + invokedColumnName));
     }
 
-    public static Map<String, Argument> mergedArgumentMap(Map<String, Argument> referencedColumnArgs,
-                                                          Map<String, Argument> callingColumnArgs,
+    /**
+     * Checks if referenced {@link ColumnProjection} or {@link Queryable} arguments are available in either calling
+     * object's arguments or in fixed arguments. If yes, use that value. Fixed arguments gets preference over calling
+     * object's arguments.
+     * @param referencedObjectArgs referenced {@link ColumnProjection} or {@link Queryable} arguments.
+     * @param callingObjectArgs calling {@link ColumnProjection} or {@link Queryable} arguments.
+     * @param fixedArgs Fixed arguments.
+     * @return Available arguments for referenced {@link ColumnProjection} or {@link Queryable}.
+     */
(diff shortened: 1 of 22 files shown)


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/ColumnContext.java
```diff
@@ -97,14 +98,12 @@ public Object get(Object key) {
...
+        ColumnProjection column = this.getQueryable().getSource().getColumnProjection(keyStr);
         if (column != null) {
 
             ColumnProjection newColumn = column.withArguments(
-                            getColumnArgMap(this.getQueryable(),
-                                            column.getName(),
-                                            this.getColumn().getArguments(),
-                                            emptyMap()));
+                            mergedArgumentMap(column.getArguments(),
+                                              this.getColumn().getArguments()));
```
rishi-aga, line 106: Getting column projection from queryable's source itself, this column projection will have default/available arguments for the column.  So no need to go-to source for default/available arguments again.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/metadata/ColumnContext.java
```diff
@@ -233,4 +229,9 @@ private Object resolveSQLHandlebar(final Object context, final Options options)
 
         return columnArgMap;
     }
+
+    public static Map<String, Argument> mergedArgumentMap(Map<String, Argument> referencedColumnArgs,
```
aklish, line 233: javadoc

aklish: These function parameters use terms like column but the function is also leveraged for table arguments.  Let's rename these parameters so they make sense in both contexts.

elide-datastore/elide-datastore-aggregation/src/test/java/com/yahoo/elide/datastores/aggregation/integration/MetaDataStoreIntegrationTest.java
```diff
@@ -200,10 +200,10 @@ public void tableTest() {
         // Verify Table Arguments
         given()
                 .accept("application/vnd.api+json")
-                .get("/table/SalesNamespace_orderDetails?include=arguments")
+                .get("/table/SalesNamespace_orderDetails?include=argumentDefinitions")
```
aklish, line 203: The contract of the API shouldn't change here (should still be argument) even though we changed the model name.  The `@Include` type is still 'argument'.  Not sure why this test is passing.

aklish (changes requested):

aklish (approved):

PR #2097, opened 2021-05-17
Validate Table arguments are provided correctly

rishi-aga:

## Description
HJSON validation for table arguments.

## License


elide-model-config/src/main/java/com/yahoo/elide/modelconfig/model/Named.java
```diff
@@ -31,7 +31,7 @@ default String getGlobalName() {
     default boolean hasName(Collection<? extends Named> collection, String name) {
         return collection
                         .stream()
-                        .map(Named::getName)
+                        .map(Named::getGlobalName)
```
rishi-aga, line 34: this should be global name here. Not related to changes in this PR.

PR #2098, opened 2021-05-17
Prevent nesting for columns that reference joins that contain $$column references

aklish:
Nesting partially expands handlebars references (leaving physical references intact) to create a new outer query expression for a column.   Join expressions cannot be rewritten as part of nesting unfortunately.  Because logical column expressions are removed/expanded during partial expression expansion, a join expression which contains $$column references may not get interpreted correctly and therefore cannot be nested.   

This PR prevents nesting of columns that reference joins that contain $$column handlebars references.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/query/SQLColumnProjection.java
```diff
@@ -80,9 +82,35 @@ default boolean canNest(Queryable source, MetaDataStore metaDataStore) {
...
+
+        //Search for any join expression that contains $$column.  If found, we cannot nest
+        //because rewriting the SQL in the outer expression will lose the context of the calling $$column.
+        JoinExpressionExtractor.JoinResolver columnResolver = (expression, context) -> expression;
+
+        return ! (references.stream().anyMatch(reference -> {
+            Set<String> joinExpressions = reference.accept(new JoinExpressionExtractor(ctx, columnResolver));
+
+            return joinExpressions.stream().anyMatch(expression -> {
+                if (expression.contains(COL_PREFIX)) {
```
rishi-aga, line 108: may be use `$$column.args.` instead of COL_PREFIX which is just `$$column` as we support $$column.expr also.

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/expression/HasColumnArgsVisitor.java
```diff
@@ -0,0 +1,88 @@
...
+                continue;
+            }
+            visitedJoins.add(join);
+
+            boolean matches = parser.parse(table, join.getJoinExpression()).stream().anyMatch(
+                    ref -> ref.accept(this));
+
+            if (matches) {
+                return true;
+            }
```
rishi-aga, line 73:
```suggestion
            return matches;
```

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/query/SQLColumnProjection.java
```diff
@@ -81,9 +84,23 @@ default boolean canNest(Queryable source, MetaDataStore metaDataStore) {
...
+        List<Reference> references = new ExpressionParser(metaDataStore).parse(source, this);
+
+        //Search for any join expression that contains $$column.  If found, we cannot nest
+        //because rewriting the SQL in the outer expression will lose the context of the calling $$column.
+        return references.stream()
+                .map(reference -> reference.accept(
+                        new ReferenceExtractor<JoinReference>(JoinReference.class, metaDataStore, source)))
+                .flatMap(Set::stream)
+                .map(reference -> reference.accept(
+                        new ReferenceExtractor<ColumnArgReference>(ColumnArgReference.class, metaDataStore, source)))
```
rishi-aga, line 100: should the source here be `null`? If it's null, it will start from a join expression and check everything inside the reference tree of the join expression.

aklish: Yes

rishi-aga (approved):

PR #2102, opened 2021-05-18
use AggregationStorePermissionExecutor for Aggregation Store Model

Chandrasekar-Rajasekar:
## Description
Use AggregationStorePermissionExecutor for Aggregation Store Model. 
Fix bugs on AggregationStorePermissionExecutor

## How Has This Been Tested?
IT

## License


elide-core/src/main/java/com/yahoo/elide/core/security/executors/AggregationStorePermissionExecutor.java
```diff
@@ -100,7 +100,7 @@ public AggregationStorePermissionExecutor(RequestScope requestScope) {
                                                                                          ChangeSpec changeSpec,
                                                                                          Class<A> annotationClass,
                                                                                          String field) {
-        throw new UnsupportedOperationException();
+        return checkSpecificFieldPermissions(resource, changeSpec, annotationClass, field);
```
Chandrasekar-Rajasekar, line 103: This method is called a bunch of times. Since we only have user check on fields, I guess it would make sense to do what checkSpecificFieldPermissions in here.

aklish (approved):

PR #2103, opened 2021-05-18
Allow custom serdes to override default ones (#1421).

### LINKED ISSUE #1421: Provide mechanism to override Elide provided Serdes
Elide provides a number of Serdes (type serializer/deserializers) out of the box.   These Serdes are found and registered by scanning for the ElideTypeConverter annotation.

There is no mechanism in Elide to override these default Serdes.

## Expected Behavior

Elide should give precedence to Serdes outside the `com.yahoo.elide` package namespace when registering a new Serde.

edu-chk:
Resolves #1421

## Description
- @ElideTypeConverter annotation removed from a bunch of default serdes
- updated serdes added to the ElideSettingsBuilder
- registering of the 'isoDate' object mapper module has been removed from a JsonApiMapper constructor
in favor of registering it in a common way with all the other serdes during Elide instantiating

## Motivation and Context
To be able to override the default serdes.

## How Has This Been Tested?
A unit test was added for a ElideSettingsBuilder

## License


aklish:
Changes look good, but a few of the GraphQL serde tests are failing:

```
13:55:06 [INFO] Results:
13:55:06 [INFO] 
13:55:06 [ERROR] Failures: 
13:55:06 [ERROR]   GraphQLConversionUtilsTest.testGraphQLConversionUtilsOffsetDateTimeToScalarType:30 expected: not <null>
13:55:06 [ERROR]   GraphQLConversionUtilsTest.testGraphQLConversionUtilsTimeZoneToScalarType:40 expected: not <null>
13:55:06 [INFO] 
13:55:06 [ERROR] Tests run: 130, Failures: 2, Errors: 0, Skipped: 0
```

aklish:
It fails because GraphQL only registered custom scalar types for Serde's registered with ElideTypeConverter:

https://github.com/yahoo/elide/blob/master/elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java#L73-L77

aklish:
Probably need to rewrite that function to look something like:

```java
    private void registerCustomScalars() {

        CoerceUtil.getSerdes().forEach((cls, serde) -> {
            SerdeCoercing serdeCoercing = new SerdeCoercing(ERROR_MESSAGE, serde);

            ElideTypeConverter elideTypeConverter = serde.getClass().getAnnotation(ElideTypeConverter.class);

            if (elideTypeConverter != null) {
                scalarMap.put(ClassType.of(elideTypeConverter.type()), new GraphQLScalarType(elideTypeConverter.name(),
                        elideTypeConverter.description(), serdeCoercing));
            } else {
                scalarMap.put(
                        ClassType.of(cls),
                        new GraphQLScalarType(cls.getSimpleName(), cls.getSimpleName(), serdeCoercing));
            }
        });
    }
```

edu-chk:
> Probably need to rewrite that function to look something like:
> 
> ```java
>     private void registerCustomScalars() {
> 
>         CoerceUtil.getSerdes().forEach((cls, serde) -> {
>             SerdeCoercing serdeCoercing = new SerdeCoercing(ERROR_MESSAGE, serde);
> 
>             ElideTypeConverter elideTypeConverter = serde.getClass().getAnnotation(ElideTypeConverter.class);
> 
>             if (elideTypeConverter != null) {
>                 scalarMap.put(ClassType.of(elideTypeConverter.type()), new GraphQLScalarType(elideTypeConverter.name(),
>                         elideTypeConverter.description(), serdeCoercing));
>             } else {
>                 scalarMap.put(
>                         ClassType.of(cls),
>                         new GraphQLScalarType(cls.getSimpleName(), cls.getSimpleName(), serdeCoercing));
>             }
>         });
>     }
> ```

That is exactly what I've prepared :D
I just had some doubts regarding why it was done in a way so it depends on serdes have the @ElideTypeConverter annotation.

aklish (approved):
Awesome

PR #2105, opened 2021-05-18
Parse table and column arguments in Expression parser

rishi-aga:
## Description
Parse table and column arguments in the Expression parser

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/QueryEngine.java
```diff
@@ -168,6 +168,14 @@ protected void populateMetaData(MetaDataStore metaDataStore) {
...
         //Populate table sources.
         metaDataStore.getTables().forEach(table -> {
+            table.getArgumentDefinitions().forEach(argument -> {
+                argument.setTableSource(TableSource.fromDefinition(
+                        argument.getTableSourceDefinition(),
+                        table.getVersion(),
+                        metaDataStore
+                ));
+            });
+
```
rishi-aga, line 178: this code was missing.
Not related to this PR.

aklish (approved):

PR #2144, opened 2021-06-10
use column name to extract cache key in agg datastore

Chandrasekar-Rajasekar:
## Description
Using an alias as a key could be misleading since users can use the same alias for different columns across different requests.  Use name instead. 

## License


rishi-aga (approved):

aklish:
I'm a little confused why I don't see the latest commit in the changes.

aklish (approved):

PR #2169, opened 2021-06-18
Allow multivalue params and fix header sanitization

### LINKED ISSUE #2166: JsonApiController swallows repeated params
The Elide RequestScope has a field [MultivaluedMap<String, String> queryParams](https://github.com/yahoo/elide/blob/309f35b792124bfbef9cc0bb504b49d2f49f425a/elide-core/src/main/java/com/yahoo/elide/core/RequestScope.java#L58) .
This indicates that we can have multiple parameters such as `a=1&a=2`.

However, the Spring Starter JsonApiController swallows everything except the first occurence of the parameter. I assume it's due to the param type in the controller being [Map<String, String>](https://github.com/yahoo/elide/blob/309f35b792124bfbef9cc0bb504b49d2f49f425a/elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/controllers/JsonApiController.java#L69) rather than Map<String, List\<String>> or MultivaluedMap

So which behavior is the intended one? Before I start fixing the JsonApiController I'd like to have confirmation that you want to support multiple parameters with the same key.

Brutus5000:
Resolves #2166

## Description
1. The JsonApiController now uses a MultiValueMap to accept multiple values. Since Elide requires MultivaluedHashMap which is not supported by Spring I also added a convert method.
2. Header cleaning now takes into account that headers are to be treated as lowercase. Since we don't have a guarantee that all webservers out there will properly convert headers, I force them into a lowercased map.

## Motivation and Context
1) was a bug I encountered when writing a custom hook. The hooks give multivalue maps but in my integration tests it didn't work.
2) I encountered it when writing integration tests for 1)

## How Has This Been Tested?
There are new integration tests for each JsonApiController method verifying the passthrough of the headers and the correct removal of the auth headers.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/jsonapi/resources/JsonApiEndpoint.java
+++ b/elide-core/src/main/java/com/yahoo/elide/jsonapi/resources/JsonApiEndpoint.java
@@ -73,7 +73,8 @@ public class JsonApiEndpoint {
         String jsonapiDocument) {
         MultivaluedMap<String, String> queryParams = uriInfo.getQueryParameters();
         String apiVersion = HeaderUtils.resolveApiVersion(headers.getRequestHeaders());
-        Map<String, List<String>> requestHeaders = HeaderUtils.removeAuthHeaders(headers.getRequestHeaders());
+        Map<String, List<String>> requestHeaders =
+                HeaderUtils.lowercaseAndRemoveAuthHeaders(headers.getRequestHeaders());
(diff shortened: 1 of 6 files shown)


aklish (approved):

aklish:
Build is failing for check styles.

Brutus5000:
Fixed now

aklish (approved):

PR #2179, opened 2021-06-24
Make AS before table alias optional, i.e. dialect based

rishi-aga:
Make AS before table alias optional, i.e. dialect based.
Oracle doesn’t support `AS` in FROM clause and currently, we generate SQL like `table_name AS table_alias` so it doesn't work. Oracle supports `AS` in the SELECT clause so no issues with `column_name AS column_alias`. 

## License


aklish (dismissed):
Looks good.  Is this for Oracle or Vertica?

rishi-aga:
Oracle

Chandrasekar-Rajasekar (approved):

PR #2205, opened 2021-07-22
Add ability to map unknown exceptions to custom exceptions

### LINKED ISSUE #2204: Allow mapping of Hibernate constraint violations
Hibernate generates very generic exceptions when you hit constraints i.e. unique column constraint. This result in very generic error response in Elide.

## Desired Behavior
I want to be able to swap the nested ConstraintViolation with a proper business exception that returns proper error codes etc.

## Current Behavior
Statuscode 423 Locked is returned with the generic text "could not execute statement".

## Possible Solution
I can give Elide a Map<String, Function<ConstraintViolation, Throwable>> where the keys are the names of the constraint hit.
If a ConstraintViolation occurs, Elide will lookup the constraint name and if it finds one, will map the exception by calling the function from the map value.
I would implement this error mapping in the `flush` method of AbstractJpaTransaction. Not sure yet how to pass the map in there.

This would be a JpaDatastore only solution. Ideas for more general solutions are welcome.

## Context
I'm trying to improve error handling for our frontend people.

Brutus5000:
Resolves #2204

## Description
ElideSettings now take an (optional) ErrorMapper. This is a functional interface that can map any Exception into a different CustomErrorException.

When handling a request now, Elide runs throw the previous error handling up to the point where it is considered an unknown exception. In that case, the unknown exception is passed to the error mapper, which can map it into a CustomErrorException. If the mapping is successful, we build an error response out of it, otherwise we rethrow the untouched exception (just as before).

## Motivation and Context
While Elide maps javax.validation.ConstraintValidationException into proper error messages, it fails to do the same for org.hibernate.ConstraintViolationException as the latter ones might only refer to database constraint names, which have no context.
With a custom mapping a developer can capture these exceptions, read them out and turn them into a more meaningful exception (e.g. a tablename_unique_name constraint becomes a UsernameNotUniqueException extending CustomErrorException).

## How Has This Been Tested?
Not yet. This is only RFC so far.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -14,6 +14,7 @@ import com.yahoo.elide.core.datastore.inmemory.InMemoryDataStore;
 import com.yahoo.elide.core.dictionary.Injector;
 import com.yahoo.elide.core.exceptions.BadRequestException;
 import com.yahoo.elide.core.exceptions.CustomErrorException;
+import com.yahoo.elide.core.exceptions.ErrorMapper;
 import com.yahoo.elide.core.exceptions.ErrorObjects;
 import com.yahoo.elide.core.exceptions.ForbiddenAccessException;
 import com.yahoo.elide.core.exceptions.HttpStatus;
@@ -21,7 +22,6 @@ import com.yahoo.elide.core.exceptions.HttpStatusException;
 import com.yahoo.elide.core.exceptions.InternalServerErrorException;
 import com.yahoo.elide.core.exceptions.InvalidURLException;
 import com.yahoo.elide.core.exceptions.JsonPatchExtensionException;
-import com.yahoo.elide.core.exceptions.TimeoutException;
 import com.yahoo.elide.core.exceptions.TransactionException;
 import com.yahoo.elide.core.security.User;
 import com.yahoo.elide.core.utils.ClassScanner;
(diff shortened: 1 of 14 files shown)


aklish:
This looks good.  Will there be a default mapper and what will it do?

Brutus5000:
No, by default the mapper is null, and if it's null no mapping takes place - in that case you have the same behavior as before this change.

aklish:
This looks good.  In the past, we have made these kinds of interfaces beans for Spring boot and elide standalone.   

Testing though is all that's needed for merging.

Brutus5000:
FYI: There is currently no activity due to lack of time, but I for sure plan on adding tests and finish this PR in the next few months.

aklish:
> FYI: There is currently no activity due to lack of time, but I for sure plan on adding tests and finish this PR in the next few months.

Sounds good.  Thanks for the update.

Brutus5000:
@aklish I needed to do some more refactorings in order to get it right. And I also added tests now. Do you think it is sufficient to just test one method, e.g. POST, and leave out the other http methods? (They all end up in the same handleRequest error handling anyway)

aklish:
@Brutus5000 One POST test should be fine.

Brutus5000:
OK, so from my side it's done, but the CI checks seem to be frozen.

aklish:
> OK, so from my side it's done, but the CI checks seem to be frozen.

Reviewing now.  CI from non-writers requires a manually start in screwdriver to kickoff the build.

aklish:
Code changes look good - no comments on what was changed. 

To make this feature ubiquitous for Elide, two things need to be added:

1. Make the same changes made in Elide.java to QueryRunner.java for GraphQL.
2. Make similar changes made in ElideAutoConfiguration to ElideStandaloneSettings for ElideStandalone.

Brutus5000:
@aklish I added the changes for StandaloneElide and QueryRunner.
For standalone I didn't add particular checks, but I added an "empty" error mapper to the test configuration.
For GraphQL QueryRunner all I can do it verify if the mapping is called. The test setup doesn't offer much more, as this part doesn't seem to be covered by unit tests.

aklish (approved):

PR #2219, opened 2021-07-27
Resolves #2215

### LINKED ISSUE #2215: Caching of ClassType instead of instantiating each time?
I'm currently trying to find causes for high GC activity when moving from getter-annotations to field level annotations. I do **not** think this issue is related, but it is a potential finding nevertheless.

https://github.com/yahoo/elide/blob/7edc5d2b329f99f6d14a44f9bd8969d7c9444576/elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java#L2003-L2007

## Expected Behavior

Elide doesn't recreate reusable stateless objects thousands of times for throwaway usage.

## Current Behavior

In my application EntityDictionary.getType creates a new ClassType on every call (object seems never to be instanceof Dynamic). If I'm running this on my main entity for a single id without any includes it is called ~300 times. This scales linearly with a page size. On 10 items on a page it's 3000 times, on 100 times it's 30000 times and so on (I have users scraping max page size of 10k, which causes then 3 million new ClassTypes for a single call).

## Possible Solution

The amount of possible "inputs" from object.getClass() is heavily restricted by the entities used in Elide.
Would it be beneficial to store the ClassTypes in a `HashMap<Class, ClassType>` and look them up every time instead of creating a new ClassType every time?

## Your Environment

* Elide version used: 5.0.3
* Environment name and version (Java 1.8.0_152): JDK 11
* Operating System and version: Arch Linux
* Link to your project: https://github.com/FAForever/faf-java-api/

aklish:
Resolves #2215

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java
@@ -106,7 +106,7 @@ public class EntityDictionary {
 
     public static final String ELIDE_PACKAGE_PREFIX = "com.yahoo.elide";
     public static final String NO_VERSION = "";
-    private static final Map<Class<?>, Type<?>> TYPE_MAP = new HashMap<>();
+    private static final Map<Class<?>, Type<?>> TYPE_MAP = new ConcurrentHashMap<>();
 
     protected final ConcurrentHashMap<Pair<String, String>, Type<?>> bindJsonApiToEntity = new ConcurrentHashMap<>();
     protected final ConcurrentHashMap<Type<?>, EntityBinding> entityBindings = new ConcurrentHashMap<>();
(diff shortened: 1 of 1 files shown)


moizarafat (approved):

Brutus5000:
I noticed you are not using a ConcurrentHashMap. Not sure if this is required here?
Here is a similar solution I coded yesterday but was not able to test yet:

```java
    private static final ConcurrentHashMap<Class<?>, ClassType<?>> CLASS_TYPE_CACHE = new ConcurrentHashMap<>();

    [...]

    public static <T> Type<T> getType(T object) {
        if (object instanceof Dynamic) {
            return ((Dynamic) object).getType();
        } else {
            ClassType<T> classType = (ClassType<T>) CLASS_TYPE_CACHE.computeIfAbsent(
                    object.getClass(),
                    x -> new ClassType<>(object.getClass())
            );
            return classType;
        }
    }
 ```

**Comment by aklish** [comment]
@Brutus5000 Good catch.  Will incorporate your changes.

File: elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java [hunk 1]
```diff
@@ -2001,9 +2002,16 @@ public static String getEntityName(Type<?> modelClass) {
...
     public static <T> Type<T> getType(T object) {
-        return object instanceof Dynamic
-                ? ((Dynamic) object).getType()
-                : new ClassType(object.getClass());
+        if (object instanceof Dynamic) {
+            return ((Dynamic) object).getType();
+        } else {
+            ClassType<T> classType = (ClassType<T>) TYPE_MAP.computeIfAbsent(
+                    object.getClass(),
+                    x -> new ClassType<>(object.getClass())
```
Brutus5000, line 2010: Sorry that lambda was not very smart from me because I was checking with the cast first. This should be ClassType::new instead (since `x == object.getClass()` anyway)

moizarafat (approved):

PR #2222, opened 2021-07-28
Use include to set model description

aklish:
## License

### DIFF
--- a/elide-swagger/src/main/java/com/yahoo/elide/swagger/JsonApiModelResolver.java
+++ b/elide-swagger/src/main/java/com/yahoo/elide/swagger/JsonApiModelResolver.java
@@ -7,6 +7,7 @@ package com.yahoo.elide.swagger;
 
 import com.yahoo.elide.annotation.CreatePermission;
 import com.yahoo.elide.annotation.DeletePermission;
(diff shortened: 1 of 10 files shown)


elide-swagger/src/main/java/com/yahoo/elide/swagger/JsonApiModelResolver.java
```diff
@@ -153,7 +153,14 @@ private ApiModel getApiModel(Type<?> clazz) {
     private String getModelDescription(Type<?> clazz) {
         ApiModel model = getApiModel(clazz);
         if (model == null) {
-            return null;
+
+            String description = EntityDictionary.getEntityDescription(clazz);
+
+            if (description == null || description.isEmpty()) {
```
moizarafat, line 159: May be we can use StringUtils here similar to other classes? https://github.com/yahoo/elide/blob/1a106900807aeab8e12d913fa915c0ce6e2a29d3/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/expression/ExpressionParser.java#L58

moizarafat (approved):

PR #2224, opened 2021-07-29
Added PREFLUSH phase for lifecycle hooks

### LINKED ISSUE #1397: Elide 5.0 - 2 SQLs fired when creating an entity instance.
Currently while creating an entity instance with a pre-commit hook, it runs an insert SQL and an update SQL. 

Detail discussion on Spectrum. https://spectrum.chat/elide/general/life-cycle-hook-execution~5982e92a-a84f-4b64-877a-083eb59cd53e 

## Expected Behavior
Instead of the 2 SQLs, we should fire 1 SQL.

-->

## Current Behavior
2 SQLs are fired, 1 insert and 1 update, because we run a flush before commit.

-->

## Possible Solution
Introduce a pre-flush hook for updating the entities before the flush.

1. https://spectrum.chat/elide/general/life-cycle-hook-execution~5982e92a-a84f-4b64-877a-083eb59cd53e

## Context
https://spectrum.chat/elide/general/life-cycle-hook-execution~5982e92a-a84f-4b64-877a-083eb59cd53e

## Your Environment
https://spectrum.chat/elide/general/life-cycle-hook-execution~5982e92a-a84f-4b64-877a-083eb59cd53e

aklish:
Resolves #1397 

This phase is also useful if you want to clean up references prior to a model deletion in a lifecycle hook - but don't want it to happen prior to security checks.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/Elide.java
+++ b/elide-core/src/main/java/com/yahoo/elide/Elide.java
@@ -463,6 +463,7 @@ public class Elide {
             if (!isReadOnly) {
                 requestScope.saveOrCreateObjects();
             }
+            requestScope.runQueuedPreFlushTriggers();
             tx.flush(requestScope);
 
             requestScope.runQueuedPreCommitTriggers();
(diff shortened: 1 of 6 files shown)


thaingo:
> This phase is also useful if you want to clean up references prior to a model deletion in a lifecycle hook - but don't want it to happen prior to security checks.

@aklish Could you please give more detail of this? It would be good if there is an example. Thanks.

PR #2273, opened 2021-08-25
Issue1724

### LINKED ISSUE #1724: Support filters & sorting on embedded object attributes.

aklish:
Resolves #1724 

- Expands entity dictionary to build bindings for complex attribute types.
- Changes JPQL generation to support navigation through complex attribute hierarchies.
- Unit & IT tests.

Limitations:
 - Does not support collections or maps of complex attributes.

## License


lamtd81:
Hi @aklish,

Many thanks for the fix, however I haven't been able to get it working with a `JPADataStore` :(

In the test case `FilterIT.testRootNestedComplexAttribute`, the `@Embeddable` class `Price` is added to the `entityBindings` property of the `EntityDictionary`. This is done in the `HashMapDataStore.populateEntityDictionary()` method, which doesn't check whether is class is annotated as an `@Entity` or not before adding it.

In the `JpaDataStore.populateEntityDictionary` method, however, it is calling `EntityDictionary.lookupEntityClass()` in order to make sure the class is annotated as an `@Entity` before binding it. 

Because the `@Embeddable` class isn't an `@Entity`, it never gets added to the `entityBindings` and ends up raising the following exception : `java.lang.IllegalArgumentException: Unbound Entity ClassType{cls=class ...}`

Is there any workaround to this? It doesn't look like we can simply remove the lookupEntityClass check.

aklish:
Might be related to #2285.  We saw a similar issue with patching a complex attribute.

lamtd81:
Indeed, I tried it with version 5.0.12-SNAPSHOT and it works fine now! I will wait for the next release then, thanks.

PR #2283, opened 2021-08-30
Issue 2277

### LINKED ISSUE #2277: Strange result when doing Update on embedded object of a model
## Current Behavior

When updating a field of `embedded` object of an entity model, values of remaining fields of the embedded object are set to `null`.

## Expected Behavior

Value of interested field is updated as expected but values of remaining fields of `embedded` object should not be set to `null`.

1. Create a **Sample** instance
```
POST /api/v1/sample HTTP/1.1
Host: localhost:8080
Accept: application/vnd.api+json
Content-Type: application/vnd.api+json
Content-Length: 219

{
    "data": {
        "type": "sample",
        "attributes": {
            "owner": "owner1",
            "address": {
                "city": "HCM",
                "country": "VN"
            }
        }
    }
}
```

2. Fetch created **Sample** instance
```
GET /api/v1/sample HTTP/1.1
Host: localhost:8080
Accept: application/vnd.api+json
Content-Type: application/vnd.api+json
```

3. Update the **Sample** instance
```
PATCH /api/v1/sample/1 HTTP/1.1
Host: localhost:8080
Accept: application/vnd.api+json
Content-Type: application/vnd.api+json
x-owner: owner1
Content-Length: 169

{
    "data": {
        "type": "sample",
        "id": 1,
        "attributes": {
            "address": {
                "city": "Saigon"
            }
        }
    }
}
```

4. Fetch the **Sample** instance again and will see that field `country` is `null` now.

## Your Environment

* Elide version used: 5.0.11
* Environment name and version (Java 1.8.0_152): JDK 15
* Operating System and version: MacOS
* Link to your project: [demo](https://github.com/thaingo/elide-issue-cont)

thaingo:
Resolves #2277

## Description

There are two ways to interpret how updates to complex attribute should be handled. The purpose of this PR is to follow the JSON API spec as explained [in section 7.2.1](https://jsonapi.org/format/#crud-updating-resource-attributes)
## Motivation and Context

#2277 
## How Has This Been Tested?

New unit test and integration test.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -455,13 +455,17 @@ public class PersistentResource<T> implements com.yahoo.elide.core.security.Pers
                                         RequestScope scope) {
         for (String field : updateValue.keySet()) {
             final Object newValue = updateValue.get(field);
-            if (dictionary.isComplexAttribute(ClassType.of(currentValue.getClass()), field)) {
-                final Object newOriginal = dictionary.getValue(currentValue, field, scope);
-                this.updateComplexAttribute(dictionary, (Map<String, Object>) newValue, newOriginal, scope);
-            } else {
-                final Object coercedNewValue =
-                        dictionary.coerce(currentValue, newValue, field, dictionary.getType(currentValue, field));
-                dictionary.setValue(currentValue, field, coercedNewValue);
+            final Object coercedNewValue =
+                    dictionary.coerce(currentValue, newValue, field, dictionary.getType(currentValue, field));
+            final Object newOriginal = dictionary.getValue(currentValue, field, scope);
+            if (!Objects.equals(newOriginal, coercedNewValue)) {
+                if (newOriginal == null
+                        || coercedNewValue == null
+                        || !dictionary.isComplexAttribute(ClassType.of(currentValue.getClass()), field)) {
+                    dictionary.setValue(currentValue, field, coercedNewValue);
+                } else {
+                    this.updateComplexAttribute(dictionary, (Map<String, Object>) newValue, newOriginal, scope);
+                }
             }
         }
     }
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -427,11 +427,7 @@ public class PersistentResource<T> implements com.yahoo.elide.core.security.Pers
         checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, newVal, val);
         if (!Objects.equals(val, newVal)) {
             if (dictionary.isComplexAttribute(EntityDictionary.getType(obj), fieldName)) {
-                if (newVal instanceof Map) {
-                    this.updateComplexAttribute(dictionary, fieldName, newVal, val, requestScope);
-                } else {
-                    this.setValueChecked(fieldName, newVal);
-                }
+                this.updateComplexAttribute(dictionary, fieldName, newVal, val, requestScope);
             } else {
                 this.setValueChecked(fieldName, newVal);
             }
@@ -453,22 +449,29 @@ public class PersistentResource<T> implements com.yahoo.elide.core.security.Pers
 
     private void updateComplexAttribute(EntityDictionary dictionary,
                                         String fieldName,
-                                        Object update,
-                                        Object original,
+                                        Object updateValue,
+                                        Object currentValue,
                                         RequestScope scope) {
-        final Map<String, Object> updateMap = (Map<String, Object>) update;
-        for (String field : updateMap.keySet()) {
-            final Object newValue = updateMap.get(field);
-            if (original != null) {
-                if (dictionary.isComplexAttribute(ClassType.of(original.getClass()), field)) {
-                    final Object newOriginal = dictionary.getValue(original, field, scope);
-                    updateComplexAttribute(dictionary, field, newValue, newOriginal, scope);
+        if (updateValue instanceof Map) {
+            final Map<String, Object> updateValueMap = (Map<String, Object>) updateValue;
+            final Map<String, Object> currentValueMap = (Map<String, Object>) currentValue;
+            for (String field : updateValueMap.keySet()) {
+                final Object newValue = updateValueMap.get(field);
+                if (currentValue != null) {
+                    if (dictionary.isComplexAttribute(ClassType.of(currentValue.getClass()), field)) {
+                        final Object newOriginal = dictionary.getValue(currentValue, field, scope);
+                        updateComplexAttribute(dictionary, field, newValue, newOriginal, scope);
+                    } else {
+                        if (newValue != currentValueMap.get(field)) {
+                            dictionary.setValue(currentValue, field, newValue);
+                        }
+                    }
                 } else {
-                    dictionary.setValue(original, field, newValue);
+                    this.setValueChecked(fieldName, updateValue);
                 }
-            } else {
-                this.setValueChecked(fieldName, update);
             }
+        } else {
+            this.setValueChecked(fieldName, updateValue);
         }
     }
 
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -583,7 +583,11 @@ public class PersistentResource<T> implements com.yahoo.elide.core.security.Pers
                     || !dictionary.isComplexAttribute(EntityDictionary.getType(obj), fieldName)) {
                 this.setValueChecked(fieldName, coercedNewValue);
             } else {
-                this.updateComplexAttribute(dictionary, (Map<String, Object>) newVal, val, requestScope);
+                if (newVal instanceof Map) {
+                    this.updateComplexAttribute(dictionary, (Map<String, Object>) newVal, val, requestScope);
+                } else {
+                    this.setValueChecked(fieldName, coercedNewValue);
+                }
             }
             this.markDirty();
             //Hooks for customize logic for setAttribute/Relation
@@ -616,7 +620,11 @@ public class PersistentResource<T> implements com.yahoo.elide.core.security.Pers
                         || !dictionary.isComplexAttribute(ClassType.of(currentValue.getClass()), field)) {
                     dictionary.setValue(currentValue, field, coercedNewValue);
                 } else {
-                    this.updateComplexAttribute(dictionary, (Map<String, Object>) newValue, newOriginal, scope);
+                    if (newValue instanceof Map) {
+                        this.updateComplexAttribute(dictionary, (Map<String, Object>) newValue, newOriginal, scope);
+                    } else {
+                        dictionary.setValue(currentValue, field, coercedNewValue);
+                    }
                 }
             }
         }
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -40,6 +40,7 @@ import com.yahoo.elide.core.request.Sorting;
 import com.yahoo.elide.core.security.ChangeSpec;
 import com.yahoo.elide.core.security.permissions.ExpressionResult;
 import com.yahoo.elide.core.security.visitors.CanPaginateVisitor;
+import com.yahoo.elide.core.type.ClassType;
 import com.yahoo.elide.core.type.Type;
 import com.yahoo.elide.core.utils.TypeHelper;
 import com.yahoo.elide.core.utils.coerce.CoerceUtil;
@@ -426,7 +427,11 @@ public class PersistentResource<T> implements com.yahoo.elide.core.security.Pers
         Object val = getValueUnchecked(fieldName);
         checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, newVal, val);
         if (!Objects.equals(val, newVal)) {
-            this.setValueChecked(fieldName, newVal);
+            if (dictionary.isComplexAttribute(EntityDictionary.getType(obj), fieldName)) {
+                this.updateComplexAttribute(dictionary, fieldName, newVal, val, requestScope);
+            } else {
+                this.setValueChecked(fieldName, newVal);
+            }
             this.markDirty();
             //Hooks for customize logic for setAttribute/Relation
             if (dictionary.isAttribute(EntityDictionary.getType(obj), fieldName)) {
@@ -441,6 +446,27 @@ public class PersistentResource<T> implements com.yahoo.elide.core.security.Pers
             return true;
         }
         return false;
+    }
+
+    private void updateComplexAttribute(EntityDictionary dictionary,
+                                        String fieldName,
+                                        Object update,
+                                        Object original,
+                                        RequestScope scope) {
+        final Map<String, Object> updateToMap = TypeHelper.toMap(update);
+        for (String field : updateToMap.keySet()) {
+            final Object newValue = updateToMap.get(field);
+            if (original != null) {
+                if (dictionary.isComplexAttribute(ClassType.of(original.getClass()), field)) {
+                    final Object newOriginal = dictionary.getValue(original, field, scope);
+                    updateComplexAttribute(dictionary, field, newValue, newOriginal, scope);
+                } else {
+                    dictionary.setValue(original, field, newValue);
+                }
+            } else {
+                this.setValueChecked(fieldName, update);
+            }
+        }
     }
 
     /**
@@ -1495,14 +1521,7 @@ public class PersistentResource<T> implements com.yahoo.elide.core.security.Pers
         // should be explicitly encapsulated here, not there.
         checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, newValue, existingValue);
 
-        final Object mergedValue;
-        if (existingValue != null) {
-            mergedValue = TypeHelper.merge(existingValue, newValue);
-        } else {
-            mergedValue = newValue;
-        }
-
-        setValue(fieldName, mergedValue);
+        setValue(fieldName, newValue);
     }
 
     /**
--- a/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
@@ -42,7 +42,6 @@ import com.yahoo.elide.core.security.permissions.ExpressionResult;
 import com.yahoo.elide.core.security.visitors.CanPaginateVisitor;
 import com.yahoo.elide.core.type.ClassType;
 import com.yahoo.elide.core.type.Type;
-import com.yahoo.elide.core.utils.TypeHelper;
 import com.yahoo.elide.core.utils.coerce.CoerceUtil;
 import com.yahoo.elide.jsonapi.models.Data;
 import com.yahoo.elide.jsonapi.models.Relationship;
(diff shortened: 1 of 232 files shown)


elide-core/src/main/java/com/yahoo/elide/core/utils/TypeHelper.java
```diff
@@ -156,4 +158,23 @@ public static String getFieldAlias(String tableAlias, String fieldName) {
...
+        final ClassType<?> origClz = ClassType.of(original.getClass());
+        Object returnObject = null;
+        try {
+            returnObject = origClz.newInstance();
+            final Field[] fields = origClz.getDeclaredFields();
+            for (Field field : fields) {
+                field.setAccessible(true);
+                final Object origValue = field.get(original);
+                final Object updateValue = field.get(update);
+                Object value = (updateValue != null && updateValue != origValue) ? updateValue : origValue;
```
aklish, line 172: Thought more about this approach.  I think the issue here is what if the client actually set null in the provided attribute field (what if they wanted to null out a field).  I don't think we can do the merge on the actual objects (since we lose important information about which fields the client actually sent along in the request).

thaingo: I have tried your recommendation and faced several issues. Probably, I do not understand your recommendation entirely. Lets clarify one thing I am stuck with:

What is exactly `original` parameter in the function `updateComplexAttribute(EntityDictionary dictionary, Map updated, Object original, RequestScope scope)`?  If it is [original value:]
(https://github.com/yahoo/elide/blob/3a19d8767497ce15d377ed7f4991c86a6732cefd/elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java#L424)
then issue happens with this `dictionary.isComplexAttribute(original.class, field)`.

So I need your further advise/support if you want me to follow your recommendation. I pushed the change in case you want to look into it.

aklish: Sorry for the confusion.  Original is the current value for the attribute being modified.  Maybe we should call it currentValue or something instead.  The link you provided in the above description is correct.  What is the issue that happens when you call dictionary.isComplexAttribute?

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -441,6 +448,27 @@ public boolean updateAttribute(String fieldName, Object newVal) {
...
+                                        RequestScope scope) {
+        final Map<String, Object> updateToMap = TypeHelper.toMap(update);
+        for (String field : updateToMap.keySet()) {
+            final Object newValue = updateToMap.get(field);
+            if (original != null) {
+                if (dictionary.isComplexAttribute(ClassType.of(original.getClass()), field)) {
+                    final Object newOriginal = dictionary.getValue(original, field, scope);
+                    updateComplexAttribute(dictionary, field, newValue, newOriginal, scope);
+                } else {
+                    dictionary.setValue(original, field, newValue);
```
aklish, line 464: We'll need to call setValueChecked more or less everywhere.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -441,6 +448,27 @@ public boolean updateAttribute(String fieldName, Object newVal) {
         return false;
     }
 
+    private void updateComplexAttribute(EntityDictionary dictionary,
+                                        String fieldName,
+                                        Object update,
+                                        Object original,
+                                        RequestScope scope) {
+        final Map<String, Object> updateToMap = TypeHelper.toMap(update);
```
aklish, line 456: You can assume update will be Map or you can just call dictionary.setValueChecked.  You don't need to convert it.

elide-core/src/main/java/com/yahoo/elide/core/utils/TypeHelper.java
thaingo: It is `java.lang.IllegalArgumentException: Unbound Entity ClassType{cls=class XYZ} ` because we have passed in a original value of the complex attribute into the call while it is expecting an elide model.

aklish: Elide now binds both Elide models and complex attributes in the dictionary.

dictionary.isComplexAttribute should be safe to call on any Elide model or any complex attribute.  What class is it being called on when it barfs?

aklish: We can chat about it on discord if that is simpler.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
thaingo: There is confusion here and I need you help further clarify:
1. There is no such method `dictionary.setValueChecked`.
2. I have to convert the `update` object to `Map` since we don't want to lose information about which fields client actually sent along in request [as you recommended earlier](https://github.com/yahoo/elide/pull/2283/files#r698722236)
I cannot cast the `update` object to `Map` either due to `ClassCastException`.

aklish: 1. look in PersistentResource - not dictionary.
2. update _is_ a Map.  Just check its type (instanceof).  If it's not a map, you can fallback to just setting the value directly.

elide-core/src/main/java/com/yahoo/elide/core/utils/TypeHelper.java
thaingo: Provided more info of the stack trace on discord. 
Basically, I test the demo project in #2283 with the change in this PR.
and class `Address` (embeddable class) is being called in this case.

aklish: Hmm.  I don't see this issue.  I added an Address type into the sample project and the dictionary finds it.  Can you publish a failing test on this branch or in a branch on the example project?

thaingo: Sorry, no such issue! Some how my IDE failed to use latest version of the local maven dependency.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -441,6 +447,34 @@ public boolean updateAttribute(String fieldName, Object newVal) {
         return false;
     }
 
+    private void updateComplexAttribute(EntityDictionary dictionary,
+                                        String fieldName,
+                                        Object updateValue,
+                                        Object currentValue,
+                                        RequestScope scope) {
+        if (updateValue instanceof Map) {
```
thaingo, line 455: I might mistake somewhere but it looks like `updateValue` is not a `Map` in case of complex attribute. Debugged several times and saw that request updating value for a complex attribute goes directly to else branch. Could you look into this and advice?

In case you need, I made an integration test [here](https://github.com/thaingo/elide-issue-cont/blob/0ff2347b66340a60f8ee74a6542ef48993aa50c1/src/test/java/example/SampleTest.java#L94)

aklish: I was able to run your test.  The reason it is not a map is because of this line:

https://github.com/yahoo/elide/pull/2283/files#diff-a9c0257c4c9d6efc89d5fdbd951efcba0266b3965e801fe378e43690f4294416L423

This line coerces the map into an Address.  We need to preserve the map and pass that to updateComplexAttribute.

aklish: Fiddling with it.  Haven't tested this yet, but this is how I would change updateAttribute:

```java
    public boolean updateAttribute(String fieldName, Object newVal) {
        Type<?> fieldClass = dictionary.getType(getResourceType(), fieldName);
        Object coercedValue =  dictionary.coerce(obj, newVal, fieldName, fieldClass);
        Object val = getValueUnchecked(fieldName);
        checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, coercedValue, val);
        if (!Objects.equals(val, coercedValue)) {
            if (val == null
                    || coercedValue == null
                    || ! dictionary.isComplexAttribute(EntityDictionary.getType(obj), fieldName)) {
                this.setValueChecked(fieldName, newVal);
            } else {
                this.updateComplexAttribute(dictionary, fieldName, (Map<String, Object>) newVal, val, requestScope);
            }
            this.markDirty();
            //Hooks for customize logic for setAttribute/Relation
            if (dictionary.isAttribute(EntityDictionary.getType(obj), fieldName)) {
                transaction.setAttribute(obj, Attribute.builder()
                        .name(fieldName)
                        .type(fieldClass)
                        .argument(Argument.builder()
                                .name("_UNUSED_")
                                .value(newVal).build())
                        .build(), requestScope);
            }
            return true;
        }
        return false;
    }
```

aklish: Again - haven't tested, but the updateComplexAttribute might look like:

```java
    private void updateComplexAttribute(EntityDictionary dictionary,
                                        Map<String, Object> updateValue,
                                        Object currentValue,
                                        RequestScope scope) {
        for (String field : updateValue.keySet()) {
            final Object newValue = updateValue.get(field);
            if (dictionary.isComplexAttribute(ClassType.of(currentValue.getClass()), field)) {
                final Object newOriginal = dictionary.getValue(currentValue, field, scope);
                updateComplexAttribute(dictionary, (Map<String, Object>) newValue, newOriginal, scope);
            } else {
                Object coercedValue = dictionary.coerce(currentValue, newValue, field, dictionary.getType(currentValue, field));
                dictionary.setValue(currentValue, field, coercedValue);
            }
        }
    }
```

aklish: The first function would also need to check a feature flag in requestScope.getElideSettings before doing this operation.

thaingo: Thanks for the guidance. 
Made the change and now facing this issue:
```
java.lang.IllegalArgumentException: Unbound Entity ClassType{cls=class example.models.Address}
	at com.yahoo.elide.core.dictionary.EntityDictionary.lookupEntityClass(EntityDictionary.java:1392)
	at com.yahoo.elide.core.dictionary.EntityDictionary.getEntityBinding(EntityDictionary.java:371)
	at com.yahoo.elide.core.dictionary.EntityDictionary.isComplexAttribute(EntityDictionary.java:2192)
	at com.yahoo.elide.core.PersistentResource.updateComplexAttribute(PersistentResource.java:458)
	at com.yahoo.elide.core.PersistentResource.updateAttribute(PersistentResource.java:434)
	at com.yahoo.elide.jsonapi.parser.state.RecordTerminalState.patch(RecordTerminalState.java:96)
	at com.yahoo.elide.jsonapi.parser.state.RecordTerminalState.handlePatch(RecordTerminalState.java:77)
	at com.yahoo.elide.jsonapi.parser.state.StateContext.handlePatch(StateContext.java:112)
```

Reproduce using [this integration test](https://github.com/thaingo/elide-issue-cont/blob/0ff2347b66340a60f8ee74a6542ef48993aa50c1/src/test/java/example/SampleTest.java#L94)

Lets get this working correctly first, I'll handle Potential NPE and feature flag a bit later.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -420,11 +422,17 @@ public boolean matchesId(String checkId) {
...
         Object val = getValueUnchecked(fieldName);
-        checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, newVal, val);
-        if (!Objects.equals(val, newVal)) {
-            this.setValueChecked(fieldName, newVal);
+        checkFieldAwareDeferPermissions(UpdatePermission.class, fieldName, coercedNewValue, val);
+        if (!Objects.equals(val, coercedNewValue)) {
+            if (val == null
+                    || coercedNewValue == null
+                    || !dictionary.isComplexAttribute(EntityDictionary.getType(obj), fieldName)) {
+                this.setValueChecked(fieldName, newVal);
```
thaingo, line 432: Looked into functions: `setValueChecked`, `checkFieldAwareDeferPermissions`, `setValue` and `dictionary.setValue`, I believe that any update/new value for attribute should be coerced before it is passed in to these functions. Am I correct?

aklish: Yes.  That is correct.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
aklish: I found the issue in the multiplex store.  Here is a patch to workaround.  I'll submit a small PR to fix:

diff --git a/elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexManager.java b/elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexManager.java
index f1d4f526f..0be30b3ca 100644
--- a/elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexManager.java
+++ b/elide-datastore/elide-datastore-multiplex/src/main/java/com/yahoo/elide/datastores/multiplex/MultiplexManager.java
@@ -65,7 +65,7 @@ public final class MultiplexManager implements DataStore {
             );
 
             dataStore.populateEntityDictionary(subordinateDictionary);
-            for (EntityBinding binding : subordinateDictionary.getBindings()) {
+            for (EntityBinding binding : subordinateDictionary.getBindings(false)) {
                 // route class to this database manager
                 this.dataStoreMap.put(binding.entityClass, dataStore);

aklish: Fixed here:
https://github.com/yahoo/elide/pull/2285

FYI, your test passes now with this fix and the branch in its current state.

aklish: If we target this PR against elide-6.x, we won't need to add a feature flag for the API change.  Elide 6 is primarily just a bump to Java 11 (no new features) and remove some deprecated APIs - so it won't linger like Elide 5 did.

aklish:
Overall changes look good.  A few comments on checkstyle errors and a few more tests.

Also, can you rebase this off elide-6.x?  elide-6.x will become master in a few days, but this kind of change can't be pushed to the 5.x line without a feature flag.

thaingo:
> Overall changes look good. A few comments on checkstyle errors and a few more tests.
> 
> Also, can you rebase this off elide-6.x? elide-6.x will become master in a few days, but this kind of change can't be pushed to the 5.x line without a feature flag.

Will do soon.

Also, I am worried about the change I am making since I am quite new to the code base. Looks like adding some more of integration tests is the only way to ensure I don't break anything. Your thoughts?

aklish (commented):
> > Overall changes look good. A few comments on checkstyle errors and a few more tests.
> > Also, can you rebase this off elide-6.x? elide-6.x will become master in a few days, but this kind of change can't be pushed to the 5.x line without a feature flag.
> 
> Will do soon.
> 
> Also, I am worried about the change I am making since I am quite new to the code base. Looks like adding some more of integration tests is the only way to ensure I don't break anything. Your thoughts?

I would add the new tests I called out in PersistentResourceTest (rather than make new IT tests).

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
```diff
@@ -76,6 +72,12 @@
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import static com.yahoo.elide.annotation.LifeCycleHookBinding.Operation.CREATE;
```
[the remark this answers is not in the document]
thaingo: I already fixed all CheckStyle errors today with this [commit](https://github.com/yahoo/elide/pull/2283/commits/ab241565a2714b14ee040349135d00d68b72288b)

Do let me know if there is still any CheckStyle error.

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
```diff
@@ -1583,6 +1585,23 @@ public void testUpdateAttributeSuccess() {
         verify(tx, times(1)).save(parent, goodScope);
     }
 
+    @Test
+    public void testUpdateComplexAttributeSuccess() {
```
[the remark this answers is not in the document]
thaingo: Good. Will add these unit tests.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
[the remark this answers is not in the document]
aklish: No checkstyle errors, but other IT tests are failing because of the change to the data model.  Check the screwdriver build:

https://cd.screwdriver.cd/pipelines/6103/builds/733421/steps/build

I recommend building all of elide locally.

elide-core/src/test/java/com/yahoo/elide/core/PersistentResourceTest.java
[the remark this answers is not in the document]
thaingo: all newly added unit tests are passed but my confidence is still low. Will add a few more integration tests and we'll see.

elide-core/src/main/java/com/yahoo/elide/core/PersistentResource.java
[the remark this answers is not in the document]
thaingo: 1. Fixed a few failed tests caused by the changes. Also, found and fixed a corner case with [this commit](https://github.com/yahoo/elide/pull/2283/commits/9bff2c60f7d41d16ceff8b550800b6cfae7fc5b6)

2. Tried building all of elide locally and found a few existing but failed tests (both unit and IT tests) that are not caused by the changes. Fixed one of them like [this](https://github.com/yahoo/elide/pull/2283/commits/4859765ac90f3ad1bbfe4a0995c49dcc67e77842)
but want to double check if you are aware of them, for example [this one](https://github.com/yahoo/elide/blob/master/elide-integration-tests/src/test/java/com/yahoo/elide/tests/UserTypeIT.java)
This surprised me by the way.

3. We want to rebase this onto the elide-6.x but my fork does not have that branch yet. To minimize work, I want to recreate a fork, rebase and then open a new PR for the changes. Your thoughts?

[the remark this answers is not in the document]
aklish: > 1. Fixed a few failed tests caused by the changes. Also, found and fixed a corner case with [this commit](https://github.com/yahoo/elide/pull/2283/commits/9bff2c60f7d41d16ceff8b550800b6cfae7fc5b6)

When is newVal not a map? 

> 2. Tried building all of elide locally and found a few existing but failed tests (both unit and IT tests) that are not caused by the changes. Fixed one of them like [this](https://github.com/yahoo/elide/pull/2283/commits/4859765ac90f3ad1bbfe4a0995c49dcc67e77842)

Makes sense.

>    but want to double check if you are aware of them, for example [this one](https://github.com/yahoo/elide/blob/master/elide-integration-tests/src/test/java/com/yahoo/elide/tests/UserTypeIT.java)

I would need to see the test failure and why it is failing (unfortunately the screwdriver build is not building the branch and so it never gets here).

>    This surprised me by the way.
> 3. We want to rebase this onto the elide-6.x but my fork does not have that branch yet. To minimize work, I want to recreate a fork, rebase and then open a new PR for the changes. Your thoughts?

I think that sounds like a good plan.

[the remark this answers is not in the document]
aklish: That's because this particular model has an attribute of type Object.

private Object obj = "foo";

which is really of type String.  Elide thinks it is a complex attribute because of its type.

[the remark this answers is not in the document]
aklish: Here is the GraphQL query:

https://github.com/yahoo/elide/blob/master/elide-graphql/src/test/resources/graphql/responses/upsert/rootSingleWithId.json

So yes, it appears it is possible to be a non-map.

aklish:
Closing for duplicate in #2301

PR #2408, opened 2021-11-10
Async Redesign for simpler background thread Model

moizarafat:

## License


elide-async/src/main/java/com/yahoo/elide/async/service/thread/AsyncAPIBackgroundRunnable.java
```diff
@@ -0,0 +1,250 @@
...
+        return projectionInfo.getProjections().values();
+    }
+
+    private Observable<PersistentResource> export(TableExport exportObj, RequestScope scope,
+            EntityProjection projection) {
+        Observable<PersistentResource> results = Observable.empty();
+
+        UUID requestId = UUID.fromString(exportObj.getRequestId());
+
+        try {
```
aklish, line 215: This can be cleaned up since we are restricting async to reads.  We don't need to call precommit, or run any lifecycle hooks for example.

aklish (changes requested):

PR #2409, opened 2021-11-12
Refresh scope beans

### LINKED ISSUE #2264: Feature request: Aggregation store - reload models in Hjson config files at runtime without restarting application
Currently any model change in Hjson config files does require an application restart so that the change will be effective. Possibility to reload models in Hjson config files at runtime would make analytic features more appealing and convenient for Elide users.

aklish:
Resolves #2264 
## Description

Makes spring controllers and Elide beans `@RefreshScope` so they will automatically reload when a spring boot based Elide service (with actuators enabled) is refreshed (through the actuator /refresh endpoint).

## Motivation and Context
To enable an HJSON editor that can save and automatically update dynamic models.

## How Has This Been Tested?
Existing tests plus manual tests.

## License


elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/HeaderCleansingTest.java
```diff
@@ -0,0 +1,238 @@
+/*
+ * Copyright 2019, Yahoo Inc.
```
jkusa, line 2:
```suggestion
 * Copyright 2021, Yahoo Inc.
```

jkusa (approved):

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/config/ElideAutoConfiguration.java
```diff
@@ -127,17 +140,34 @@ public DataSourceConfiguration getDataSourceConfiguration() {
...
      * Creates the Elide instance with standard settings.
      * @param dictionary Stores the static metadata about Elide models.
      * @param dataStore The persistence store.
+     * @param transactionRegistry Global transaction registry.
      * @param settings Elide settings.
      * @return A new elide instance.
      */
     @Bean
     @ConditionalOnMissingBean
+    @Scope(SCOPE_PROTOTYPE)
```
thaingo, line 167: Should we clean up prototype scoped objects? It is recommended [here](https://docs.spring.io/spring-framework/docs/3.0.0.M3/reference/html/ch04s04.html#beans-factory-scopes-prototype)

aklish: Good call, but I think I might fold these beans into the the Refresh versions.  I think it is confusing having both.  They are essentially private.

aklish: I did a small refactor.  For the beans that are still declared PROTOTYPE scope, I don't believe any cleanup is required as they don't hold onto connections or sessions.  The only thing which manages connections are transactions.  We expect transactions to complete (or timeout) after which their resources should be cleaned up.  The controllers are also RefreshScope - so they should be garbage collected along with their associated references to any prototype beans.

moizarafat (approved):

PR #2418, opened 2021-11-17
HJSON Configuration Models and DataStore

aklish:
## Description
PR to provide a custom data store to read and write HJSON configuration files as Elide models.

## Motivation and Context
This in conjunction with hot reload will support the creation of a HJSON editor.

## How Has This Been Tested?
New config validation tests and data store tests.  Another PR will integrate with Spring & Standalone and add IT tests.

## License


elide-model-config/src/main/java/com/yahoo/elide/modelconfig/store/ConfigDataStoreTransaction.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Elide DataStoreTransaction which loads/persists HJSON configuration files as Elide models.
+ */
+public class ConfigDataStoreTransaction implements DataStoreTransaction {
+
+
```
moizarafat, line 37: please remove extra line

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/store/ConfigDataStoreTransaction.java
```diff
@@ -0,0 +1,211 @@
...
+        }
+    }
+
+    @Override
+    public void commit(RequestScope scope) {
+        boolean validateOnly = scope.getRequestHeaderByName(VALIDATE_ONLY_HEADER) != null;
+
+        if (! validateOnly) {
+            for (Runnable runnable : todo) {
+                runnable.run();
```
moizarafat, line 112: Do we need some rollback mechanism here. For example while saving 5 files, it was able to save only 2 and 3rd failed. Will it leave the Agg store in a bad state.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/io/FileLoader.java
```diff
@@ -0,0 +1,168 @@
...
+     * @throws IOException If something goes boom.
+     */
+    public Map<String, ConfigFile> loadResources() throws IOException {
+        Map<String, ConfigFile> resourceMap = new LinkedHashMap<>();
+        int configDirURILength = resolver.getResources(this.rootURL)[0].getURI().toString().length();
+
+        Resource[] hjsonResources = resolver.getResources(this.rootURL + HJSON_EXTN);
+        for (Resource resource : hjsonResources) {
+            if (! resource.exists()) {
+                continue;
```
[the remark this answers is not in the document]
aklish: I added better error logging in this case.

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/store/ConfigDataStoreTransaction.java
aklish: Done

elide-model-config/src/main/java/com/yahoo/elide/modelconfig/store/ConfigDataStoreTransaction.java
aklish: I thought about this but took a different route.  There are two kinds of errors - recoverable and unrecoverable.  Unrecoverable would be things like disk corruption or file system being full.  Rollback is not likely to succeed in these cases.  Recoverable would be things like permission errors.

Instead of implementing rollback (which could be messy/complicated), I added extra validation around permission errors.

moizarafat (approved):

PR #2464, opened 2022-01-03
Use GraphQLBigDecimal for BigDecimal Conversions in GraphQLConversionUtils

moizarafat:
## Description
Uses JavaPrimitives.GraphQLBigDecimal for Converting BigDecimal Values to Graphql Input.

## Motivation and Context
BigDecimal Values were converted to Float resulting in precision loss and returning values with Exponents.

## How Has This Been Tested?
Existing test cases passed with updates.

## License

### DIFF
--- a/elide-graphql/src/test/resources/graphql/requests/update/updateComplexAttributeMap.graphql
+++ b/elide-graphql/src/test/resources/graphql/requests/update/updateComplexAttributeMap.graphql
@@ -3,8 +3,8 @@ mutation {
   data: {
          id: "1",
          priceRevisions: [
-            { key: "2016-01-02T00:00Z", value: { units: 126, currency: { currencyCode: "USD" }}},
-            { key: "2016-01-01T00:00Z", value: { units: 125}}
+            { key: "2016-01-02T00:00Z", value: { units: 490056468, currency: { currencyCode: "USD" }}},
+            { key: "2016-01-01T00:00Z", value: { units: 125.7}}
          ]
       }) {
     edges {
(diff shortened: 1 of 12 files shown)


Chandrasekar-Rajasekar (approved):

elide-graphql/src/test/resources/graphql/responses/update/updateComplexAttributeMap.json
```diff
@@ -8,7 +8,7 @@
             {
               "key": "2016-01-02T00:00Z",
               "value": {
-                "units": 126.0,
+                "units": 490056468,
```
aklish, line 11: Why did this value change?

elide-graphql/src/test/resources/graphql/requests/update/updateComplexAttributeMap.graphql
```diff
@@ -3,8 +3,8 @@ mutation {
   data: {
          id: "1",
          priceRevisions: [
-            { key: "2016-01-02T00:00Z", value: { units: 126, currency: { currencyCode: "USD" }}},
-            { key: "2016-01-01T00:00Z", value: { units: 125}}
+            { key: "2016-01-02T00:00Z", value: { units: 490056468, currency: { currencyCode: "USD" }}},
```
aklish, line 6: Why did this value change?

elide-graphql/src/test/resources/graphql/responses/update/updateComplexAttributeMap.json
moizarafat: Wanted to put in some big value to test the change. Prior to this change, this big value will be returned something as 4.9E. With this new change it does not return exponential value and return the big value as is.

aklish (approved):

PR #2466, opened 2022-01-03
use query with filter to populate query parameters in pagination

### LINKED ISSUE #2465: Pagination queries are not working for metric with custom two pass aggregation query.
Elide's GraphQL`pageInfo` queries are not working for Two-Pass Aggregation With Nested Query Structure

## Expected Behavior
Setup - Table that has two-pass aggregation metrics formed using Custom Metric Projection Maker. 
The pagination GraphQL snippet `pageInfo { startCursor endCursor totalRecords }` on these tables is not substituting variables for pageTotal.

## Current Behavior
Error is thrown in H2Agg datastore. Verified the error in Vertica too. 
```
org.h2.jdbc.JdbcSQLDataException: Parameter &#34;#1&#34; is not set;
```

Ref - [SQLQueryEngine](https://github.com/yahoo/elide/blob/master/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java)
For Two-pass aggregation metrics, `expandMetricQueryPlans()` creates a nested query with some or all of the filters pushed to the inner query. `getPageTotal()` method use outer queries filter to populate query parameters in ` supplyFilterQueryParameters()` which throw the error. 

-->

## Possible Solution
We pass the Original Query with Filters in [supplyFilterQueryParameters()](https://github.com/yahoo/elide/blob/master/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java#L287) method to populate query parameters. 

1. Two-Pass Aggregation that makes elide to form nested query can be found [here](https://git.vzbuilders.com/mantis/drax/blob/master/src/main/java/com/yahoo/mantis/metricformula/TwoPassBrandAggregationProjection.java). 
2. Create a table with a metric using the above metric maker. 
3. Query the two-pass agg metric with pagination

## Context
Could not use nested metrics in production

## Your Environment
* Elide version used: 6.0.4

Chandrasekar-Rajasekar:
Resolves #2465

## Description
This uses the original query with filters to populate pagination query filter parameters. 

## How Has This Been Tested?
Unit test

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/SQLQueryEngine.java
@@ -269,7 +269,7 @@ public class SQLQueryEngine extends QueryEngine {
         return resultBuilder.build();
     }
 
-    private long getPageTotal(Query expandedQuery, NativeQuery sql, Query query, SqlTransaction sqlTransaction) {
+    private long getPageTotal(Query expandedQuery, NativeQuery sql, Query clientQuery, SqlTransaction sqlTransaction) {
         ConnectionDetails details = expandedQuery.getConnectionDetails();
         DataSource dataSource = details.getDataSource();
         SQLDialect dialect = details.getDialect();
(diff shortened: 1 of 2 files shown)


rishi-aga (dismissed):
👍

moizarafat (approved):

aklish (approved):

PR #2471, opened 2022-01-04
Disallow ConfigFile path to be changed in ConfigStore

aklish:
ConfigFile path cannot be changed as the ID is derived from it.  It is immutable.  This PR explicitly disallows changes.

## License


elide-model-config/src/main/java/com/yahoo/elide/modelconfig/store/models/ConfigFile.java
```diff
@@ -51,6 +51,7 @@
     @GeneratedValue
     private String id; //Base64 encoded path-version
 
+    @UpdatePermission(expression = "NONE")
```
jkusa, line 54: Can we make this string type safe?

jkusa (approved):
Just the one comment, not a blocker though.

jkusa (approved):

PR #2494, opened 2022-01-12
Aggregation Store: Fix filter by alias with parameterized metric.

aklish:
## Description

There was a bug where the aggregation store was rejecting queries (as invalid) because it could not pair up a measure in the query (with an alias) with a filter predicate on the same measure (with or without the alias). 

## License

### DIFF
--- a/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/DefaultQueryValidator.java
+++ b/elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/DefaultQueryValidator.java
@@ -69,8 +69,8 @@ public class DefaultQueryValidator implements QueryValidator {
 
                 Predicate<ColumnProjection> filterByNameAndArgs =
                         (column) -> (column.getAlias().equals(projection.getAlias())
-                                || column.getName().equals(projection.getName())
-                                && column.getArguments().equals(projection.getArguments()));
+                                || column.getName().equals(projection.getName()))
+                                && column.getArguments().equals(projection.getArguments());
 
                 //Query by (alias or name) and arguments.   The filter may or may not be using the alias.
                 if (query.getColumnProjection(filterByNameAndArgs) == null) {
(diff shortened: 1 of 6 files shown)


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/EntityProjectionTranslator.java
```diff
@@ -202,32 +197,4 @@ private void addHavingMetrics() {
                         getArgumentMapFromArgumentSet(attribute.getArguments())))
                 .collect(Collectors.toList());
     }
-
-    /**
-     * Gets relationship names from {@link EntityProjection}.
-     * @return relationships list of {@link Relationship} names
-     */
-    private Set<String> getRelationships() {
```
aklish, line 210: Removing dead code.

Chandrasekar-Rajasekar (dismissed):

elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/DefaultQueryValidator.java
```diff
@@ -65,10 +66,21 @@ public void validateHavingClause(Query query) {
             validatePredicate(query, predicate);
 
             extractFilterProjections(query, havingClause).stream().forEach(projection -> {
-                if (query.getColumnProjection(projection.getAlias(), projection.getArguments()) == null) {
+
+                Predicate<ColumnProjection> filterByNameAndArgs =
+                        (column) -> (column.getAlias().equals(projection.getAlias())
```
moizarafat, line 71: should there be an additional ( ) around the Left side and Right side expression of || to make sure that condition is true before &&

aklish: Yes - this is a bug (and why the build failed).

Chandrasekar-Rajasekar (approved):

moizarafat (approved):

PR #2500, opened 2022-01-13
make getColumnProjection() method use different name for column Id for different source when alias are involved

Chandrasekar-Rajasekar:
## Description
SQLColumnProjection and SQLMetricProjection nest method identify a column with either getName or getAlias depending on the source. If the source is SQLTable which has no user-supplied alias, we use getName. Else we use getAlias. 

## Motivation and Context
With the Current behavior of using an alias for column id, the metric column set with projected=false, when the source is an instance of SQLTable. This happens if the queryPlan of the simple metric is merged with the queryPlan of 2-pass aggregation metrics. 

## How Has This Been Tested?
Integration Test

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/queryengines/sql/query/SQLColumnProjection.java
```diff
@@ -111,7 +113,8 @@ default boolean canNest(Queryable source, MetaDataStore metaDataStore) {
 
         boolean requiresJoin = requiresJoin(references);
 
-        boolean inProjection = source.getColumnProjection(getAlias(), getArguments(), true) != null;
+        String columnId = (source instanceof SQLTable) ? getName() : getAlias();
```
aklish, line 116: You can check if the source is the root instead of checking using the instance of operator here.

aklish (changes requested):

aklish (approved):

PR #2526, opened 2022-01-26
Support filters on query plans

aklish:
## Description
 * Adds WHERE filter clause support for Aggregation Store query plans.
 * Creates a new interface QueryPlanMerger which is responsible for merging two plans together.
 * Creates a Default implementation of the merger which performs basic merging.

## License


elide-datastore/elide-datastore-aggregation/src/main/java/com/yahoo/elide/datastores/aggregation/query/DefaultQueryPlanMerger.java
```diff
@@ -0,0 +1,146 @@
...
+    public QueryPlan merge(QueryPlan a, QueryPlan b) {
+        Preconditions.checkNotNull(a);
+        Preconditions.checkNotNull(b);
+
+        a = nestToDepth(a, b.nestDepth());
+        b = nestToDepth(b, a.nestDepth());
+
+        assert (a.isNested() || a.getSource().equals(b.getSource()));
+
+        Set<MetricProjection> metrics = Streams.concat(b.getMetricProjections().stream(),
```
michael-mclawhorn: break out into 'protected mergetFoo(QP a, QP b)` for each of the terms, including where Filters/

michael-mclawhorn (approved):

PR #2535, opened 2022-02-15
Ignoring Elide models embedded in complex attributes

aklish:
## Description

Ignored Elide models that are embedded in complex attributes.  Currently, Elide has strange behavior where it sometimes will bind the model as an attribute and later ignore it as an exposed entity.

## License


aklish:
Fixes #2536

rishi-aga (approved):

PR #2649, opened 2022-05-12
Update Chinese README

QubitPi:
## Motivation and Context

We are considering Elide as part of our important data management store infrastructure and I'd like to re-gain Elide knowledge and keep up with it; hence I updated README translation after reading it

## License


README.md
```diff
@@ -1,85 +1,94 @@
 # Elide
```
moizarafat, line 1: I think you meant to update the README.MD file under translations/zh instead of this one.

PR #2824, opened 2022-10-22
Adding support for populating Meta object in JSON-API

### LINKED ISSUE #2820: Make it possible to fill the meta block in jsonApiDocument
### Custom information in Meta Object of JSON API Responses

We want to be able to add custom fields to meta blocks in JSON API responses. This should be allowed according to the standard.

Currently, there does not seem to be a way to do that. 

I tried doing it inside the DataStoreTransaction, since the RequestScope contains the JsonApiDocument.
However, it does not seem to be used in the CollectionTerminalState, but rather creates a brand new one and fills it. 

Is there any way for us to change this is this a problem/ feature request in elide?

aklish:
Resolves #2820

## Description
Adds support for setting the 'meta' field in JSON-API for both collections and for individual objects.   There is a new interface, WithMetadata, that allows an object model or the RequestScope to provide metadata that is serialized with JSON-API responses.
 
A new document processor adds any fields set to the appropriate Meta object in the response.

## License


aklish:
This needs tests. This PR is just to demonstrate a possible solution.

Also, there is no good way for a developer to set Meta for individual objects currently.  We'd likely have to change the DataStoreTransaction API to allow for that.

aklish:
@jkusa @moizarafat Thoughts on this approach?

aklish:
Doc PR here:

https://github.com/yahoo/elide-doc/pull/187

elide-core/src/main/java/com/yahoo/elide/jsonapi/document/processors/WithMetadata.java
```diff
@@ -0,0 +1,36 @@
...
+ * The class carries metadata fields.
+ */
+public interface WithMetadata {
+
+    /**
+     * Sets a metadata property for this request.
+     * @param property
+     * @param value
+     */
+    void setMetadataField(String property, Object value);
```
jkusa, line 22: Is this required? Doesn't look like it is used internally? Is it just in the case for a custom datastore? Should we add a `noop` default interface implementation?

jkusa (approved):

PR #2979, opened 2023-05-08
Add Atomic Operations Support

### LINKED ISSUE #2915: JSON API Atomic Operations support
Are there any plans on implementing the [Atomic Operations Extension](https://jsonapi.org/ext/atomic/) for JSON API?

It seems to be the replacement of the deprecated [JSON Patch Extension](https://jsonapi.org/extensions/jsonpatch/).

justin-tay:
Resolves #2915

## Description
Adds support for JSON API Atomic Operations.

## How Has This Been Tested?
Added the unit and integration tests

## License


aklish:
Nice.  Will take a look soon!

elide-core/src/main/java/com/yahoo/elide/jsonapi/extensions/JsonApiAtomicOperations.java
```diff
@@ -0,0 +1,563 @@
...
+    protected String getFullPath(Ref ref, Operation operation) {
+        if (ref != null) {
+            StringBuilder fullPathBuilder = new StringBuilder();
+            if (ref.getType() == null) {
+                throw new InvalidEntityBodyException(
+                        "Atomic Operations extension requires ref type to be specified.");
+            }
+            fullPathBuilder.append(ref.getType());
+
+            // Only relationship operations or resource update operation should have the id
```
aklish, line 191: Is this true?  What about remove?  That looks like it can also have an ID.

elide-core/src/main/java/com/yahoo/elide/jsonapi/extensions/JsonApiAtomicOperations.java
```diff
@@ -0,0 +1,563 @@
...
+     * @return the ref
+     */
+    private Ref inferRef(JsonApiMapper mapper, Operation operation) {
+        // Attempt to infer the ref from the data
+        if (operation.getData() != null && !operation.getData().isArray()) {
+            try {
+                Resource resource = mapper.forAtomicOperations()
+                        .readResource(operation.getData());
+                if (resource.getType() != null) {
+                    if (resource.getAttributes() == null || resource.getAttributes().isEmpty()) {
```
aklish, line 293: Why do we care about attributes being null/empty here?

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/ControllerTest.java
```diff
@@ -221,26 +231,499 @@ public void jsonForbiddenApiPatchTest() {
...
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .extract();
+        Map<String, Object> groupAttributes = response.path("'atomic:results'[1].data.attributes");
+        assertEquals("Foo2", groupAttributes.get("commonName"));
+        assertEquals("Updated Description", groupAttributes.get("description"));
+
+        Map<String, Object> productAttributes = response.path("'atomic:results'[3].data.attributes");
+        assertEquals("Product1", productAttributes.get("commonName"));
+        assertEquals("Product1 Description", productAttributes.get("description"));
```
aklish, line 372: There are a number of patch extenstion requests in ResourceIT (you can search for JSONAPI_CONTENT_TYPE_WITH_JSON_PATCH_EXTENSION).  They also verify that a subsequent GET request has the correct values.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/ControllerTest.java
```diff
@@ -221,26 +231,499 @@ public void jsonForbiddenApiPatchTest() {
...
+                )
+                .when()
+                .post("/json/operations")
+                .then()
+                .statusCode(HttpStatus.SC_OK)
+                .extract();
+        Map<String, Object> attributes = response.path("'atomic:results'[1].data.attributes");
+        assertEquals("Foo2", attributes.get("commonName"));
+        assertEquals("Updated Description", attributes.get("description"));
+
```
aklish, line 450: Same comment.  Do another round trip request and verify the response matches what we expect (to confirm the database was updated correctly).

elide-core/src/main/java/com/yahoo/elide/jsonapi/extensions/JsonApiAtomicOperations.java
justin-tay: For some reason I recall that the delete needed to be performed on the collection endpoint and needed to have data specified with the type and id. So later on if the data was omitted I will attempt to generate this when specified from the ref.

```java
                    // If data is not provided, but it is to remove a resource the data is optional
                    // and should be generated from the ref
                    if ((data == null || JsonNodeType.NULL.equals(data.getNodeType()))
                            && OperationCode.REMOVE.equals(operation.getOperationCode())) {
                        data = requestScope.getMapper().getObjectMapper().createObjectNode().put("type", ref.getType())
                                .put("id", ref.getId() != null ? ref.getId() : ref.getLid());
                    }
```

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/ControllerTest.java
```diff
@@ -221,26 +231,499 @@ public void jsonForbiddenApiPatchTest() {
...
+                                                type("product"),
+                                                id("com.example.operations.product1"),
+                                                attributes(
+                                                        attr("commonName", "Product1")
+                                                )
+                                        ))
+                                ),
+                                atomicOperation(AtomicOperationCode.add,
+                                        ref(
+                                                type("group/com.example.operationsrel1/products"),
```
aklish, line 517: Is this a valid way of referencing type?  This looks more like a href.

elide-core/src/main/java/com/yahoo/elide/jsonapi/extensions/JsonApiAtomicOperations.java
justin-tay: This is to tell whether it was an relationship operation. If it is a relationship operation and the attributes would be null or empty.

aklish:
One thing that's a little vague in the spec is the use of LID.  I assume it means that an ID in a create op is referenced in LID in subsequent ops.  We might also want to test that scenario to make sure it works as intended.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/ControllerTest.java
justin-tay: I'm actually not sure. I thought the intent was that the ref could always be expressed as a href and vice-versa. It's not really that clear from the spec why there were 2 ways to express the target of the operation.

aklish: My take is that some/most JSON-API APIs only expose models at the root of the URL path (/books, /invoices).  Then the REF objects can be used to navigate to most things.  

The type of the REF then is really the root collection.

elide-core/src/main/java/com/yahoo/elide/jsonapi/extensions/JsonApiAtomicOperations.java
aklish: The spec says if the operations is targeting a resource, then ref/href MUST be used.  The same goes for deleting a resource (non-relationship).  Should this be an error instead (instead of inferring the ref)?

elide-core/src/main/java/com/yahoo/elide/jsonapi/extensions/JsonApiAtomicOperations.java
aklish: The JSON API core spec has this example of a DELETE with the ID in the path:
```
DELETE /photos/1 HTTP/1.1
Accept: application/vnd.api+json
```

The operations spec has this example of specifying the ID in the operation:

```json
{
  "atomic:operations": [{
    "op": "remove",
    "ref": {
      "type": "articles",
      "id": "13"
    }
  }]
}
```

handleDeleteOp does not append the ID to the fullPath if data is missing.  However, if data is present it appends the ID.  Not sure what is more common, but the path will be different depending on the presence of data.  The spec example shows data missing.  This looks to me like it might not work.

elide-core/src/main/java/com/yahoo/elide/jsonapi/extensions/JsonApiAtomicOperations.java
justin-tay: After looking at the spec again it looks like you are correct that for delete resource scenario the ref/href is mandatory. I have updated the code to only infer the ref for the add and update resource scenario where the ref/href is optional. The code thus will not attempt to infer for the delete resource and any relationship operations scenario.

elide-core/src/main/java/com/yahoo/elide/jsonapi/extensions/JsonApiAtomicOperations.java
justin-tay: You're correct I'm not sure why I ended up not specifying the id in the ref for the remove operation. I shall change it to use the id in the remove operation and not generate the data.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/ControllerTest.java
justin-tay: I have updated the test cases.

elide-spring/elide-spring-boot-autoconfigure/src/test/java/example/tests/ControllerTest.java
justin-tay: I have updated the test case to use the href instead of the ref.

justin-tay:
Have added the test cases for LID. Initially I thought that LID was specific to the atomic operations spec but I see it's actually part of the base 1.1 spec.

I think I should have addressed all the outstanding issues.

PR #3010, opened 2023-06-12
Remove singleton DefaultClassScanner accessor

justin-tay:
## Description
This removes the `DefaultClassScanner.getInstance` method.

## Motivation and Context
Previously this method was required because creating a new DefaultClassScanner would perform a class path scan to find classes with the annotations and this caused tests to take a long time as it needed to scan each time a DefaultClassScanner was instantiated. This is no longer needed because the a static class cache is being maintained that is initialised when the first DefaultClassScanner is initialised.

## How Has This Been Tested?

## License


aklish (approved):

aklish:
Needs rebase.

justin-tay:
Did the rebase.

PR #3015, opened 2023-06-16
Add support for `isNull` predicate to evaluate collections

justin-tay:
## Description
This adds support for `isNull` predicate to evaluate collections that a null value is present.

## Motivation and Context
Previously the `isNull` predicate would always return false for collections as the it checks that the collection itself is null. This is modified such that if the collection contains a null value it will return true.

## How Has This Been Tested?
A unit test has been added.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
@@ -450,7 +450,19 @@ public enum Operator {
     //
     // Null checking
     private static <T> Predicate<T> isNull(Path fieldPath, RequestScope requestScope) {
-        return (T entity) -> getFieldValue(entity, fieldPath, requestScope) == null;
+        return (T entity) -> {
+            Object val = getFieldValue(entity, fieldPath, requestScope, true);
+            if (val == null) {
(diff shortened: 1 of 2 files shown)


elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -450,7 +450,19 @@ private static <T> Predicate<T> notinfix(Path fieldPath, List<Object> values,
...
     // Null checking
     private static <T> Predicate<T> isNull(Path fieldPath, RequestScope requestScope) {
-        return (T entity) -> getFieldValue(entity, fieldPath, requestScope) == null;
+        return (T entity) -> {
+            Object val = getFieldValue(entity, fieldPath, requestScope, true);
+            if (val == null) {
+                return true;
+            } else if (val instanceof Collection<?> collection) {
+                for (Object item : collection) {
+                    if (item == null) {
```
aklish, line 459: This feels more like hasNull or hasMember(null) instead of isNull.   The behavior might be unexpected.  Can we accomplish the same thing with hasMember(null)?

justin-tay:
Superceded by:
- #3016

PR #3018, opened 2023-06-22
Filter operations on OpenAPI and GraphQL Schema

### LINKED ISSUE #2641: Option to expose specific CRUD methods based on a config
Not all APIs require a create/update/delete method for example when static data is loaded via backend jobs etc and require only exposing query functionality.

It would be handy to have an elide wide setting that can be overridden by model annotation to specify which Operations are exposed and Swagger and GraphQL UIs take that into account. 

At the moment, I am achieving this via Checks but its not ideal as Update/Delete operations still show up in the Swagger/GraphQL UIs and caller gets the view of capabilities which we don't really support.

justin-tay:
Resolves #2641

I wasn't sure what to use to determine whether or not to expose operations by id for OpenAPI so it's currently using Permission annotations on the id which is rather odd.

## Description
- For OpenAPI document filter paths based on Permission annotations on the entity / id
- For GraphQL schema add per entity RelationshipOp enum with the valid values
- Update TableType to only expose the get collection endpoint on OpenAPI

## Motivation and Context
More accurate documentation on what operations are supported on the backend.

## How Has This Been Tested?
Added unit tests.

## License


elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -150,12 +152,63 @@ public ModelBuilder(EntityDictionary entityDictionary,
...
+        String entityName = entityDictionary.getJsonAliasFor(clazz);
+        String postfix = entityName.substring(0, 1).toUpperCase(Locale.ENGLISH) + entityName.substring(1);
+        GraphQLEnumType relationshipOp = generator.buildClassToEnumType(ClassType.of(RelationshipOp.class),
+                name -> name + postfix, e -> {
+                    RelationshipOp op = RelationshipOp.valueOf(e.name());
+                    switch (op) {
+                    case FETCH:
+                        return canRead(clazz);
+                    case DELETE:
+                        return canDelete(clazz);
```
aklish, line 177: This would require UpdatePermission - not Delete.

elide-core/src/main/java/com/yahoo/elide/annotation/DeletePermission.java
```diff
@@ -16,7 +18,7 @@
 /**
  * Assign custom Delete permission checks.
  */
-@Target({TYPE, PACKAGE})
+@Target({TYPE, PACKAGE, FIELD, METHOD})
```
aklish, line 21: There is no way to Delete a field or method in JSON-API.  You can update relationships (by deleting elements) but that is an UpdatePermission.  I recommend we continue to use UpdatePermission for relationship management.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
```diff
@@ -133,19 +135,31 @@ public GraphQLEnumType classToEnumType(Type<?> enumClazz) {
...
-            builder.value(value.toString(), value);
+        for (Enum<?> value : values) {
+            if (filter.test(value)) {
+                builder.value(value.toString(), value);
+            }
         }
 
         GraphQLEnumType enumResult = builder.build();
-
-        enumConversions.put(enumClazz, enumResult);
```
aklish, line 147: Why did we remove this line?

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -150,12 +152,63 @@ public ModelBuilder(EntityDictionary entityDictionary,
...
         this.excludedEntities = excludedEntities;
     }
 
+    public GraphQLArgument getRelationshipOp(Type<?> clazz) {
+        GraphQLArgument existing = relationshipOpArgument.get(clazz);
+        if (existing != null) {
+            return existing;
+        }
+
+        String entityName = entityDictionary.getJsonAliasFor(clazz);
```
aklish, line 168: I'm really confused what this code is doing.  It looks like it is changing the value of relationship op to include the entity name.  I'm not following the intent here, but we don't want to change the api contract.   I also feel like we really need two distinct methods here - one to get the relationship ops for a root level entity and one to get the relationship ops for a relationship.

aklish (changes requested):

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
justin-tay: Changed to UpdatePermission

elide-core/src/main/java/com/yahoo/elide/annotation/DeletePermission.java
justin-tay: Removed these changes. This was actually done so that the DeletePermission could be applied to the id field to filter off the operations by id on OpenAPI which I think was requested for HJSON as I think those endpoints to query by id don't work. I have changed it to use the Exclude annotation instead.

elide-graphql/src/main/java/com/yahoo/elide/graphql/GraphQLConversionUtils.java
justin-tay: Agree this method is a bit confusing. From what I gather it's not possible to set a graphql enum and indicate that only a subset of values are acceptable. Instead I needed to create a graphql enum per entity. eg. instead of just ElideRelationshipOp I needed to create one per entity, eg. ElideRelationshipOpDownloads, ElideRelationshipOpGroup, ElideRelationshipOpProduct and then indicate the acceptable values for each. In this case the enums needed to be deduplicated based on the graphql enum name and not the enum type. I have changed this to create a specific method for this use case instead of trying to modify the existing method to cater to this.

justin-tay:
relationshipop

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
justin-tay: I'm actually not that clear on the logic for determining the relationship ops from the permissions for graphql. Is there a difference in allowable values in the relationship ops for the root entity vs the relationship?

aklish: Permissions are not as straightforward given that graphql mutations behave more like patch extension requests in JSON-API.  Whether you can perform the operation or not depends not just on the operation being performed but on the data passed along with the operation.

Also some operations like UPSERT and REPLACE are really doing different things depending on the data.

Root entities:

- UPSERT - CreatePermission OR UpdatePermission (depending on whether the object already exists or not)
- DELETE - DeletePermission
- REMOVE - DeletePermission
- REPLACE - CreatePermission OR UpdatePermission OR DeletePermission
- FETCH - ReadPermission
- UPDATE - UpdatePermission

Relationships:
- FETCH - ReadPermission (on the field)
- REMOVE - UpdatePermission (on the field)
- DELETE - DeletePermission (on the type being deleted)
- UPSERT - CreatePermission (on the type being created) OR UpdatePermission(on the field)
- REPLACE - CreatePermission OR UpdatePermission (field and type being manipulated) OR DeletePermission
- UPDATE - UpdatePermission (on the field)

We can problem simplify this though to the following:
Root entities:

- UPSERT - CreatePermission
- DELETE - DeletePermission
- REMOVE - DeletePermission
- REPLACE - Too complex to put a restriction on.  Just always allow it.
- FETCH - ReadPermission
- UPDATE - UpdatePermission

Relationships:
- FETCH - ReadPermission (on the field)
- REMOVE - UpdatePermission (on the field)
- DELETE - DeletePermission (on the type being deleted)
- UPSERT - UpdatePermission (on the field)
- REPLACE - UpdatePermission (on the field)
- UPDATE - UpdatePermission (on the field)

Alternatively, we don't solve for GraphQL at all this way.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
```diff
@@ -150,12 +152,63 @@ public ModelBuilder(EntityDictionary entityDictionary,
...
+        String postfix = entityName.substring(0, 1).toUpperCase(Locale.ENGLISH) + entityName.substring(1);
+        GraphQLEnumType relationshipOp = generator.classToNamedEnumType(ClassType.of(RelationshipOp.class),
+                name -> name + postfix, e -> {
+                    RelationshipOp op = RelationshipOp.valueOf(e.name());
+                    switch (op) {
+                    case FETCH:
+                        return canRead(clazz);
+                    case DELETE:
+                        return canUpdate(clazz);
```
aklish, line 177: I think this is actually canDelete as it deletes the underlying objects as opposed to simply removing them from the relationship.

elide-graphql/src/main/java/com/yahoo/elide/graphql/QueryRunner.java
```diff
@@ -351,6 +353,36 @@ private ElideResponse executeGraphQLRequest(String baseUrlEndPoint, ObjectMapper
...
+    private List<GraphQLError> mapErrors(List<GraphQLError> errors) {
+        List<GraphQLError> result = new ArrayList<>(errors.size());
+        for (GraphQLError error : errors) {
+            if (error instanceof ValidationError validationError
+                    && ValidationErrorType.WrongType.equals(validationError.getValidationErrorType())) {
+                if (validationError.getDescription().contains("ElideRelationshipOp")) {
+                    String queryPath = String.join(" ", validationError.getQueryPath());
+                    result.add(ValidationError.newValidationError()
+                            .description(
+                                    "Invalid operation: " + queryPath + " is read only.")
```
aklish, line 371: Under what circumstances does this error get generated?  How do we know it is an attempt to write something that is read only?

elide-swagger/src/main/java/com/yahoo/elide/swagger/OpenApiBuilder.java
```diff
@@ -248,28 +254,46 @@ public PathItem getRelationshipPath() {
             RelationshipType relationshipType = dictionary.getRelationshipType(parentClass, name);
 
             if (relationshipType.isToMany()) {
-                path.get(new Operation().tags(getTags()).description("Returns the relationship identifiers for " + name)
-                        .responses(new ApiResponses().addApiResponse("200", okPluralResponse)));
+                if (canCreate(type)) {
```
aklish, line 257: Similar to GraphQL, when you are POSTING to a relationship, we are not creating the type but just updating the relationship (which is Update permission).  All of the relationship edits are effectively Updates.

elide-graphql/src/main/java/com/yahoo/elide/graphql/QueryRunner.java
justin-tay: This gets generated in the event an invalid value is used that couldn't map to the GraphQL enum.

I was trying to get the pre-existing test in `NoCacheAggregationDataStoreIntegrationTest` to pass but I guess I should change the expectation instead.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
justin-tay: I think it makes sense to at least attempt to filter the operations as that is better than nothing. I retained some criteria of restricting REPLACE as I think the read only use case should be relatively common and exposing REPLACE is misleading.

elide-graphql/src/main/java/com/yahoo/elide/graphql/ModelBuilder.java
justin-tay: Have changed it to canDelete.

elide-swagger/src/main/java/com/yahoo/elide/swagger/OpenApiBuilder.java
justin-tay: Have made the changes.

aklish (approved):

PR #3024, opened 2023-07-07
Calcluate duplicate paths correctly (#3022)

### LINKED ISSUE #3022: OpenAPIBuilder removes paths for relationships that are 3 levels deep. (7.0.0pr5)
I have a simple graph a->b->c
When I enable the OAS generation the generated spec is including models for a, b & c but only APIs for a & b.

The problem is the pruning of duplicates:
https://github.com/yahoo/elide/blob/7.0.0-pr5/elide-swagger/src/main/java/com/yahoo/elide/swagger/OpenApiBuilder.java#L773

I'm not sure I understand what duplicates it's pruning but the logic seems to be grouping the paths by root  type (a), then iterating over them looking for the shorter path. The paths are:

1. a/{aId}/b/{bId}
2. a/{aId}/b/{bId}/c/{cId}
3. a/{aId}

(this is the order I see)

So on the first pass through path is a/{aId}/b/{bId} and nothing gets added to toRemove (because the only path shorter than it has no lineage).

Second time through is a different story:
path is a/{aId}/b/{bId}/c/{cId} 
compare is  a/{aId}/b/{bId} 

compare.lineage.isEmpty() is false
path == compare is false

compare.shorterThan(path) is true ... so it removes a/{aId}/b/{bId}/c/{cId}......meaning there are no APIs generated for c in the spec.

Not sure I understand what's being attempted by this loop but something definitely looks wrong.

---

After sleeping on it - is the goal to find the shortest path through the graph _to that node_? At the moment it's just finding the shortest path, as well has checking if the path is shorter it should be checking if the final nodes in the path are the same, right?

---

Create a PR with the proposed fix:
https://github.com/yahoo/elide/pull/3024

bladedancer:
Resolves #3022

## Description
The code was not taking into account the node type when determine the shortest path so it was comparing apples and oranges. Rather than group by root type  group by leaf type and name and calculate the shortest path to that named type across all roots.

Also the determination of shortest path was string length not the number of hops. Changed that to be the number of hops so that the naming of the relationships doesn't affect the length.

## Motivation and Context
The swagger generation only included nodes one level deep (the nodes with the shortest path) and omitted deeper nodes. For asimple graph a->b->c the swagger would only include a (the root node) and b (the non-root node with the shortest path). It would omit c.

## How Has This Been Tested?
The unit tests have been updated to cover this scenario.

## License

### DIFF
--- a/elide-swagger/src/main/java/com/yahoo/elide/swagger/OpenApiBuilder.java
+++ b/elide-swagger/src/main/java/com/yahoo/elide/swagger/OpenApiBuilder.java
@@ -8,7 +8,7 @@ package com.yahoo.elide.swagger;
 import static com.yahoo.elide.Elide.JSONAPI_CONTENT_TYPE;
 import static com.yahoo.elide.core.dictionary.EntityDictionary.NO_VERSION;
 
-import com.yahoo.elide.core.Path;
+
 import com.yahoo.elide.core.dictionary.EntityDictionary;
 import com.yahoo.elide.core.dictionary.RelationshipType;
 import com.yahoo.elide.core.filter.Operator;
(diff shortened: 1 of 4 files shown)


aklish (approved):
Nice!

aklish:
Can you rebase this on master?

bladedancer:
@aklish - no problem, rebased, only minor diff in the test assertion counts due to the publisher changes. Let me know if anything else needed.

aklish (approved):

PR #3026, opened 2023-07-07
Elide 3025: Generate the correct schema in the OAS spec for embedded complex attributes. 

### LINKED ISSUE #3025: Generated OAS incorrectly sets objects as nodes
I have an entity that contains a JSONB field - e.g. CORS configuration. This is deserialized as a strongly typed object:

```java
@Include(rootLevel = false, name = "example")
public class Example {
...
    @Type(JsonType.class)
    private CORS cors;
...
}
```

Nothing special about CORS, it's just a pojo - it _does not_ have an include annotation.

## Expected Behavior
I'd expect the generated OAS to at worst refer to this as an Object and ideally generate a schema for it.

## Current Behavior
It's generating the object as if it was a node in the graph.

```java
    CORS:
      type: object
      properties:
        type:
          type: string
          description: ''
        id:
          type: string
        attributes:
          required: []
          type: object
          properties:
            accessControlAllowCredentials:
              type: boolean
              readOnly: false
              writeOnly: false
              example: null
            accessControlAllowHeaders:
              type: array
              readOnly: false
              writeOnly: false
              items:
                type: object
              example: null
            accessControlAllowMethods:
              type: array
              readOnly: false
              writeOnly: false
              items:
                type: object
              example: null
            accessControlAllowOrigin:
              type: array
              readOnly: false
              writeOnly: false
              items:
                type: object
              example: null
            accessControlExposeHeaders:
              type: array
              readOnly: false
              writeOnly: false
              items:
                type: object
              example: null
            accessControlMaxAge:
              type: integer
              format: int32
              readOnly: false
              writeOnly: false
              example: null
        relationships:
          required: []
          type: object
      description: 'Update Permissions : (PreAuthorize.Update AND has permission to lock Project Artifact)'
```

## Possible Solution
Digging into it - the CORS _is_ in the EntityBindings that's why getting the alias succeeds and the object is treated as a resource:

https://github.com/yahoo/elide/blob/master/elide-swagger/src/main/java/com/yahoo/elide/swagger/converter/JsonApiModelResolver.java#L78

A pojo that doesn't have an @Include or @Entity annotation being in the EntityDictionary seems odd but it is intended - EntityDictionary.discoverEmbeddedTypeBindings. 

I presume the fix needs to be that complex attribute types are skipped by JsonApiModelResolver.

bladedancer:
Resolves #3025

## Description
Embedded complex attributes (pojos) may be bindable and appear in the EntityDictionary (as per https://github.com/yahoo/elide/blob/master/elide-core/src/main/java/com/yahoo/elide/core/dictionary/EntityDictionary.java#L1968). This means that the OAS generation sees them as Entities rather than complex attributes and generates incorrect schema.

The change is to check if the entity dictionary contains the type and if it does then check if it is an elideModel. If it's not an elide model then let swagger handle the schema generation.

## Motivation and Context
The generate OAS specification was wrong.

## How Has This Been Tested?
Added new unit test to cover this case.

## License

### DIFF
--- a/elide-swagger/src/main/java/com/yahoo/elide/swagger/OpenApiBuilder.java
+++ b/elide-swagger/src/main/java/com/yahoo/elide/swagger/OpenApiBuilder.java
@@ -51,6 +51,7 @@ import io.swagger.v3.oas.models.responses.ApiResponses;
 import io.swagger.v3.oas.models.tags.Tag;
 
 import lombok.Getter;
+import org.apache.commons.lang3.tuple.Pair;
 
 import java.lang.annotation.Annotation;
 import java.util.ArrayDeque;
--- a/elide-swagger/src/main/java/com/yahoo/elide/swagger/OpenApiBuilder.java
+++ b/elide-swagger/src/main/java/com/yahoo/elide/swagger/OpenApiBuilder.java
@@ -8,7 +8,7 @@ package com.yahoo.elide.swagger;
 import static com.yahoo.elide.Elide.JSONAPI_CONTENT_TYPE;
 import static com.yahoo.elide.core.dictionary.EntityDictionary.NO_VERSION;
 
-import com.yahoo.elide.core.Path;
+
 import com.yahoo.elide.core.dictionary.EntityDictionary;
 import com.yahoo.elide.core.dictionary.RelationshipType;
 import com.yahoo.elide.core.filter.Operator;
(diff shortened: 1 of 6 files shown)


aklish (approved):

aklish:
Can you rebase this on master?

bladedancer:
@aklish - rebased, no real changes, the conflicts were because I based this branch off the fix for https://github.com/yahoo/elide/pull/3024 which had some conflicts. Let me know if there's anything else.

aklish (approved):

PR #3048, opened 2023-08-27
Adds the subsetOf and supersetOf predicates

justin-tay:
## Description
Adds subsetof and supersetof filter operators.

## Motivation and Context
Needed to allow filtering based on subset of and superset of criteria.

## How Has This Been Tested?
Added the tests.

## License


elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -536,6 +565,58 @@ private static <T> Predicate<T> hasMember(Path fieldPath, List<Object> values, R
...
+        return (T entity) -> {
+            Type<?> valueClass = fieldPath.lastElement().get().getFieldType();
+
+            Object leftHandSide = getFieldValue(entity, fieldPath, requestScope);
+
+            BiPredicate<Object, Object> predicate = (a, b) -> a.equals(b);
+
+            List<?> rightHandSide = values.stream().map(value -> CoerceUtil.coerce(value, valueClass)).toList();
+
+            if (leftHandSide instanceof Collection<?> collection && !valueClass.isAssignableFrom(COLLECTION_TYPE)) {
```
aklish, line 578: How is this if condition possible?

justin-tay: https://github.com/yahoo/elide/blob/6e2294ecb3ff076d9d9b831a6be8ac092cd12412/elide-core/src/main/java/com/yahoo/elide/core/filter/predicates/SubsetOfPredicate.java#L40-L42

For instance the inputs could be 

```java
new SubsetOfPredicate(authorAwardsElement, "Bookery prize", "Test");
```

In which case the `leftHandSide` is a collection and the value class is `String` (which isn't a `Collection`).

aklish: The leftHandSide of a predicate author.awards should be a list - not a string.  I imagine Elide would not make a pathElement like the one above based on the model author.awards.

justin-tay: Sorry, I think my example in the explanation wasn't correct.

I think this seems to be related to how the `getFieldValue` method works with respect to a to many relationship.

The integration tests that cover these conditions is in `com.yahoo.elide.tests.FilterIT` in the to many relationship tests, for instance `testRootSubsetOfToManyRelationship`.

The logic is also similar to the existing logic at

https://github.com/yahoo/elide/blob/68ddf67f2d793bd64f195fd63d6fcf045a98aea4/elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java#L601-L617

aklish: You are correct.  When traversing a toMany relationship, it will explode the leftHandSide into a list, even if the last element of the predicate is a scalar type.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -536,6 +565,58 @@ private static <T> Predicate<T> hasMember(Path fieldPath, List<Object> values, R
...
+                for (Object right : rightHandSide) {
+                    if (!collection.stream().anyMatch(object -> predicate.test(right, object))) {
+                        return false;
+                    }
+                }
+            } else {
+                for (Object right : rightHandSide) {
+                    if (!predicate.test(right, leftHandSide)) {
```
aklish, line 612: What about the case where both left hand side is a collection and the valueClass is a collection.  The predicate here is testing equality of the collection in that case with a member on the right hand side.

justin-tay: After looking at it again, I'm actually not sure if the `valueClass` can ever be assignable from a collection type.

In the `testSubsetOfOnAttributes` test in `com.yahoo.elide.tests.FilterIT` filtering on book.awards the `valueClass` is a string type.

```java
public class Book extends BaseId {
    ...
    private Collection<String> awards = new ArrayList<>();
}
```

For it to be a collection type, it would look something like
```java
public class Book extends BaseId {
    ...
    private Collection<Collection<String>> collection = new ArrayList<>();
}
```

I guess the logic is a little misleading as the else condition only happens when `leftHandSide` is null as `leftHandSide instanceof null` is false. I shall fix the logic to make it clearer.

elide-core/src/main/java/com/yahoo/elide/core/filter/Operator.java
```diff
@@ -536,6 +565,60 @@ private static <T> Predicate<T> hasMember(Path fieldPath, List<Object> values, R
...
+        return (T entity) -> {
+            Type<?> valueClass = fieldPath.lastElement().get().getFieldType();
+
+            Object leftHandSide = getFieldValue(entity, fieldPath, requestScope);
+
+            BiPredicate<Object, Object> predicate = (a, b) -> a.equals(b);
+
+            List<?> rightHandSide = values.stream().map(value -> CoerceUtil.coerce(value, valueClass)).toList();
+
+            if (leftHandSide instanceof Collection<?> collection && !valueClass.isAssignableFrom(COLLECTION_TYPE)) {
```
aklish, line 605: Doesn't the case where leftHandSide is instanceof collection && valueClass **_is_** assignable to collection still need to be handled?

PR #3059, opened 2023-08-31
Add api version handling to controllers

justin-tay:
It was a bit difficult to have the settings changes as the next PR so I will be having this one first.

Tentatively I'll split the next few as the following
- Refactor map parameters
- Refactor request scope
- Refactor JSON API
- Refactor settings and config
- Add api version handling for graphql subscriptions

## Description
Adds api version handling to the json api and graphql controllers.

## How Has This Been Tested?

## License

### DIFF
--- a/elide-async/src/main/java/com/yahoo/elide/async/operation/GraphQLAsyncQueryOperation.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/operation/GraphQLAsyncQueryOperation.java
@@ -39,7 +39,8 @@ public class GraphQLAsyncQueryOperation extends AsyncQueryOperation {
         }
         UUID requestUUID = UUID.fromString(queryObj.getRequestId());
         //TODO - we need to add the baseUrlEndpoint to the queryObject.
-        ElideResponse response = runner.run("", queryObj.getQuery(), user, requestUUID, scope.getRequestHeaders());
+        ElideResponse response = runner.run(scope.getBaseUrlEndPoint(), queryObj.getQuery(), user, requestUUID,
+                scope.getRequestHeaders());
         log.debug("GRAPHQL_V1_0 getResponseCode: {}, GRAPHQL_V1_0 getBody: {}",
                 response.getResponseCode(), response.getBody());
         return response;
--- a/elide-async/src/main/java/com/yahoo/elide/async/operation/GraphQLAsyncQueryOperation.java
+++ b/elide-async/src/main/java/com/yahoo/elide/async/operation/GraphQLAsyncQueryOperation.java
@@ -38,7 +38,6 @@ public class GraphQLAsyncQueryOperation extends AsyncQueryOperation {
             throw new InvalidOperationException("Invalid API Version");
         }
         UUID requestUUID = UUID.fromString(queryObj.getRequestId());
-        //TODO - we need to add the baseUrlEndpoint to the queryObject.
         ElideResponse response = runner.run(scope.getBaseUrlEndPoint(), queryObj.getQuery(), user, requestUUID,
                 scope.getRequestHeaders());
         log.debug("GRAPHQL_V1_0 getResponseCode: {}, GRAPHQL_V1_0 getBody: {}",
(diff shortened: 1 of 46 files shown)


elide-async/src/main/java/com/yahoo/elide/async/operation/GraphQLAsyncQueryOperation.java
```diff
@@ -39,7 +39,8 @@ public ElideResponse execute(AsyncApi queryObj, RequestScope scope) throws URISy
         }
         UUID requestUUID = UUID.fromString(queryObj.getRequestId());
         //TODO - we need to add the baseUrlEndpoint to the queryObject.
-        ElideResponse response = runner.run("", queryObj.getQuery(), user, requestUUID, scope.getRequestHeaders());
+        ElideResponse response = runner.run(scope.getBaseUrlEndPoint(), queryObj.getQuery(), user, requestUUID,
```
aklish, line 42: Can remove the TODO above.

elide-core/src/main/java/com/yahoo/elide/jsonapi/links/DefaultJsonApiLinks.java
```diff
@@ -52,16 +52,10 @@ public Map<String, String> getRelationshipLinks(PersistentResource resource, Str
...
-                String jsonApiPath = resource.getRequestScope().getElideSettings().getJsonApiPath();
-                if (StringUtils.isNotEmpty(jsonApiPath)) {
-                    result.append(jsonApiPath);
-                }
-                result.append("/");
-            }
-        } else {
+        if (resource.getRequestScope().getBaseUrlEndPoint() != null) {
+            result.append(resource.getRequestScope().getBaseUrlEndPoint());
+            result.append("/");
```
aklish, line 57: Did we get rid of the json api path here because base URL now includes it?

elide-async/src/main/java/com/yahoo/elide/async/operation/GraphQLAsyncQueryOperation.java
justin-tay: Removed.

elide-core/src/main/java/com/yahoo/elide/jsonapi/links/DefaultJsonApiLinks.java
justin-tay: Yes the base url now includes it.

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/controllers/ApiDocsController.java
```diff
@@ -176,4 +190,34 @@ public ResponseEntity<String> call() throws Exception {
...
+
+        if (StringUtils.isEmpty(baseUrl)) {
+            baseUrl = ServletUriComponentsBuilder.fromCurrentContextPath().build().toUriString();
+        }
+
+        if (prefix.length() > 1) {
+            if (baseUrl.endsWith("/")) {
+                baseUrl = baseUrl.substring(0, baseUrl.length() - 1) + prefix;
+            } else {
+                baseUrl = baseUrl + prefix;
```
aklish, line 217: I'm not quite following these two branches.  Can you add a comment what you expect to be parsing here and the difference between prefix and baseUrl?

justin-tay: Have added the comment. The first branch is just to remove trailing / if present on the base url. 

The base url here is the application base url which can be specified using `elide.base-url`. The prefix is configured using `elide.api-docs.path.` The method is to determine the base url of the api docs controller eg. http://www.example.org/api-docs where the application base url is `http://www.example.org` and the path for api docs is `/api-docs`.

elide-core/src/main/java/com/yahoo/elide/jsonapi/links/DefaultJsonApiLinks.java
aklish: Looked briefly at the code.  It looks to me like the request scope base URl is populated by the controller - but the controller just pulls the baseUrl from the settings object (and does not include the JSON API or GraphQL path here).  What am I missing?

elide-spring/elide-spring-boot-autoconfigure/src/main/java/com/yahoo/elide/spring/controllers/ApiDocsController.java
aklish: 👍

aklish (approved):

aklish:
Are there more PRs needed for API versioning before we cut Elide 7 or is the last one?

PR #3068, opened 2023-09-09
Refactor request scope

justin-tay:
## Description
Refactors the request scope to use route.

## How Has This Been Tested?

## License


elide-core/src/main/java/com/yahoo/elide/jsonapi/extensions/JsonApiAtomicOperationsRequestScope.java
```diff
@@ -67,7 +57,6 @@ public JsonApiAtomicOperationsRequestScope(
      */
     public JsonApiAtomicOperationsRequestScope(String path, JsonApiDocument jsonApiDocument,
             JsonApiAtomicOperationsRequestScope scope) {
-        super(path, scope.getApiVersion(), jsonApiDocument, scope);
-        this.setEntityProjection(new EntityProjectionMaker(dictionary, this).parsePath(path));
```
aklish, line 71: Where does this happen in the new code?

justin-tay: The `JsonApiAtomicOperationsRequestScope` is now extending off `JsonApiRequestScope` so I think it's being set in the `JsonApiRequestScope` in line 121 when `super` gets called.

aklish (approved):

PR #3076, opened 2023-10-02
Refactor settings and config

justin-tay:
## Description
This refactors the settings and configuration to use a builder pattern.

After this is a PR for subscriptions to complete the API version handling and after that I'll follow up with the documentation updates for the settings changes and a series of PRs to refactor the error handling
- Add api version handling for graphql subscriptions

## Motivation and Context
This allows easier customizations of default configuration. As the developer is able to consume a preconfigured builder to customize instead of customizing the settings from scratch.

## How Has This Been Tested?
Existing tests continue to pass and new tests added.

## License


elide-swagger/src/main/java/com/yahoo/elide/swagger/resources/ApiDocsEndpoint.java
```diff
@@ -119,6 +119,15 @@ public Response listJson(
         }
     }
 
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
```
aklish, line 123: Will these ever get invoked given that the other method matches all paths (@Path("{path:.*}")) including the empty path?

justin-tay: Yes. From what I remember in certain JAX-RS setups for `@Path("{path:.*}")` where the path is empty, this will only match if there is a trailing `/`. So this will get invoked for cases where the path is empty and there is no trailing `/`.

PR #3085, opened 2023-10-09
Support api versioning for subscriptions

justin-tay:
## Description
Adds support for GraphQL subscriptions to handle API versioning.

## Motivation and Context
GraphQL subscriptions couldn't easily handle API versioning.

## How Has This Been Tested?
Added the appropriated tests and manually with modified example applications.

## License

### DIFF
--- a/elide-datastore/elide-datastore-jms/src/main/java/com/yahoo/elide/datastores/jms/websocket/SubscriptionWebSocketConfigurator.java
+++ b/elide-datastore/elide-datastore-jms/src/main/java/com/yahoo/elide/datastores/jms/websocket/SubscriptionWebSocketConfigurator.java
@@ -14,8 +14,10 @@ import com.yahoo.elide.core.audit.AuditLogger;
 import com.yahoo.elide.core.audit.Slf4jLogger;
 import com.yahoo.elide.core.datastore.DataStore;
 import com.yahoo.elide.core.dictionary.EntityDictionary;
+import com.yahoo.elide.core.dictionary.Injector;
 import com.yahoo.elide.core.exceptions.ErrorMapper;
 import com.yahoo.elide.core.filter.dialect.RSQLFilterDialect;
+import com.yahoo.elide.core.request.route.RouteResolver;
 import com.yahoo.elide.datastores.jms.JMSDataStore;
 import com.yahoo.elide.graphql.ExecutionResultDeserializer;
 import com.yahoo.elide.graphql.ExecutionResultSerializer;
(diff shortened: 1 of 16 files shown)


elide-datastore/elide-datastore-jms/src/main/java/com/yahoo/elide/datastores/jms/websocket/SubscriptionWebSocketConfigurator.java
```diff
@@ -81,22 +89,68 @@ public class SubscriptionWebSocketConfigurator extends ServerEndpointConfig.Conf
...
+        Object httpSession = request.getHttpSession();
+        if (httpSession != null)
+        {
+            userProperties.put("session", httpSession);
+        }
+
+        Principal userPrincipal = request.getUserPrincipal();
+        if (userPrincipal != null)
+        {
+            userProperties.put("userPrincipal", userPrincipal);
```
aklish, line 150: Is there some plan later to potentially use this principal to construct the Elide user object?

elide-graphql/src/main/java/com/yahoo/elide/graphql/subscriptions/websocket/SubscriptionWebSocket.java
```diff
@@ -231,21 +258,52 @@ private SessionHandler findSession(Session wrappedSession) {
...
+        Map<String, List<String>> headers = new LinkedHashMap<>(session.getRequestParameterMap());
+        // Get the properties from the handshake set in SubscriptionWebSocketConfigurator
+        if (session.getUserProperties().get("headers") instanceof Map handshakeHeaders) {
+            headers.putAll(handshakeHeaders);
+        }
+
+        Route route = routeResolver.resolve(MediaType.APPLICATION_JSON, baseUrl, path, headers,
+                session.getRequestParameterMap());
+
+        String apiVersion = route.getApiVersion();
```
aklish, line 284: The route resolver here that resolved the apiVersion from parameters looks at a different parameter name than the old code.  The old code looked for ApiVersion in the parameters and the new code looks for "v" in the ParameterRouteResolver.

aklish (changes requested):

elide-datastore/elide-datastore-jms/src/main/java/com/yahoo/elide/datastores/jms/websocket/SubscriptionWebSocketConfigurator.java
justin-tay: This user principal here should be the same one that can be accessed from the `jakarta.websocket.Session.getUserPrincipal` method.

The default `SubscriptionWebSocket.UserFactory` already creates a user using `session -> new User(session.getUserPrincipal());`

elide-graphql/src/main/java/com/yahoo/elide/graphql/subscriptions/websocket/SubscriptionWebSocket.java
justin-tay: The `ApiVersion` that it looked for previously was more in line with what was looked for from the header. In the new code this is catered by initializing the header values from the request parameter map in the `createSessionHandler` using `Map<String, List<String>> headers = new LinkedHashMap<>(session.getRequestParameterMap());` in addition to the other values found as HTTP headers. So the `HeaderRouteResolver` should resolve the API Version if configured with `ApiVersion`.

aklish (approved):

PR #3097, opened 2023-10-19
Support second level cache

### LINKED ISSUE #3094: Hibernate L2 cache
# Caching for data get by id
I added L2 cache hibernate (EHCache 3 and Hibernate 6)
https://howtodoinjava.com/hibernate/configuring-ehcache-3-with-hibernate-6/

## Expected Behavior
- L2 cache working for get by id url
- example: GET /v1/entity/entityId

## Current Behavior
- L2 cache not working for get by id url
- example: GET /v1/entity/entityId
  - it always hits the db

justin-tay:
Resolves #3094

## Description
This allows the use of the second level cache.

This was done by modifying the `RootCollectionFetchQueryBuilder` to use the `EntityManager.find` method when querying for an entity by id instead of using a JPQL query. This is the most straightforward way to use the second level cache for querying and doesn't rely on other configuration such as enabling the query cache.

For a JPQL query to use the second level cache the query cache needs to be enabled and used, and also has other performance implications. It is currently not possible to enable the query cache in Elide as this requires the `org.hibernate.cacheable` query hint to be set on the query and there is currently no means to set the query hint in Elide. Also the query cache is not used when `scroll` is used and can only be used for `list`.

## Motivation and Context
This allows the use of the second level cache to possibly improve performance.

## How Has This Been Tested?
Added the relevant unit and integration tests.

## License

### DIFF
--- a/elide-datastore/elide-datastore-jpa/src/test/java/com/yahoo/elide/datastores/jpa/LegacyNPlusOneIT.java
+++ b/elide-datastore/elide-datastore-jpa/src/test/java/com/yahoo/elide/datastores/jpa/LegacyNPlusOneIT.java
@@ -171,7 +171,7 @@ public class LegacyNPlusOneIT extends JPQLIntegrationTest {
                 .statusCode(HttpStatus.SC_OK);
 
         verifyLoggingStatements(
-                "SELECT example_Book FROM example.Book AS example_Book WHERE example_Book.id IN (:XXX)",
+        //      "SELECT example_Book FROM example.Book AS example_Book WHERE example_Book.id IN (:XXX)",
                 "SELECT example_Author FROM example.Book example_Book__fetch JOIN example_Book__fetch.authors example_Author WHERE example_Author.id IN (:XXX) AND example_Book__fetch=:XXX",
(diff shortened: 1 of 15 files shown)


elide-datastore/elide-datastore-jpa/src/test/java/com/yahoo/elide/datastores/jpa/LegacyNPlusOneIT.java
```diff
@@ -171,7 +171,7 @@ public void testSingleElementSubcollectionWithFilter() {
                 .statusCode(HttpStatus.SC_OK);
 
         verifyLoggingStatements(
-                "SELECT example_Book FROM example.Book AS example_Book WHERE example_Book.id IN (:XXX)",
+        //      "SELECT example_Book FROM example.Book AS example_Book WHERE example_Book.id IN (:XXX)",
```
aklish, line 174: If we are not leveraging JPQL, we still probably want to log something here when we do a em.find.

justin-tay: Made the changes to log.

aklish (approved):

PR #3100, opened 2023-10-25
Add @key support for GraphQL Federation

### LINKED ISSUE #2638: Support for GraphQL Federation
It would be very useful to provide support for Graphql federation so Elide can be adopted into a microservices architecture. Any plans to support that?

justin-tay:
Resolves #2638

## Description
This adds support for the `@key` directive for Apollo Federation.

## Motivation and Context
This is required for to resolve from a different subgraph onto the Elide subgraph.

## How Has This Been Tested?

This was also tested using the following example
- https://github.com/justin-tay/federation-jvm-elide-example
- https://github.com/justin-tay/elide-spring-boot-example/tree/apiversion_elide-7.x

## License

### DIFF
--- a/elide-graphql/src/main/java/com/yahoo/elide/graphql/federation/EntitiesDataFetcher.java
+++ b/elide-graphql/src/main/java/com/yahoo/elide/graphql/federation/EntitiesDataFetcher.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2023, the original author or authors.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.graphql.federation;
+
+import com.yahoo.elide.core.PersistentResource;
+import com.yahoo.elide.core.exceptions.BadRequestException;
+import com.yahoo.elide.core.request.EntityProjection;
+import com.yahoo.elide.graphql.GraphQLRequestScope;
+import com.yahoo.elide.graphql.KeyWord;
+import com.yahoo.elide.graphql.containers.NodeContainer;
+
+import com.apollographql.federation.graphqljava._Entity;
+
+import org.apache.commons.lang3.StringUtils;
+
+import graphql.schema.DataFetcher;
+import graphql.schema.DataFetchingEnvironment;
+import io.reactivex.Observable;
+
+import java.util.ArrayList;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+/**
+ * Entities Data Fetcher for Apollo Federation.
+ */
+public class EntitiesDataFetcher implements DataFetcher<List<NodeContainer>> {
+
+    @Override
+    public List<NodeContainer> get(DataFetchingEnvironment environment) throws Exception {
+        List<Map<String, Object>> representations = environment.getArgument(_Entity.argumentName);
+        List<String> ids = representations.stream() // only supports the single id key field
+            .map(representation -> {
+                    String idKey = representation.keySet()
+                            .stream()
+                            .filter(key -> !KeyWord.TYPENAME.getName().equals(key))
+                            .findFirst()
+                            .get();
+                    return (String) representation.get(idKey);
+                })
+            .toList();
+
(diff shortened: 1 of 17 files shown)


elide-graphql/src/main/java/com/yahoo/elide/graphql/federation/EntitiesDataFetcher.java
```diff
@@ -0,0 +1,82 @@
...
+                    return (String) representation.get(idKey);
+                })
+            .toList();
+
+        String entityName = StringUtils.uncapitalize(representations.get(0).get(KeyWord.TYPENAME.getName()).toString());
+
+        GraphQLRequestScope requestScope = environment.getLocalContext();
+        EntityProjection projection = requestScope
+                .getProjectionInfo()
+                .getProjection(null, entityName);
```
aklish, line 54: What if the same entity is requested multiple times with different aliases or with an alias in general?

elide-graphql/src/main/java/com/yahoo/elide/graphql/federation/EntitiesDataFetcher.java
```diff
@@ -0,0 +1,82 @@
...
+            /* handle empty list of ids */
+            if (idList.isEmpty()) {
+                throw new BadRequestException("Empty list passed to ids");
+            }
+
+            return PersistentResource.loadRecords(projection, idList, requestScope);
+        }).orElseGet(() -> PersistentResource.loadRecords(projection, new ArrayList<>(), requestScope));
+
+
+        // Ignore errors as potentially an id on a subgraph no longer exists here
```
aklish, line 67: Is this the behavior defined in the spec?

elide-graphql/src/main/java/com/yahoo/elide/graphql/federation/EntitiesDataFetcher.java
```diff
@@ -0,0 +1,82 @@
...
+        Set<PersistentResource> results = records.onExceptionResumeNext(Observable.empty())
+                .toList(LinkedHashSet::new)
+                .blockingGet();
+
+        // Return node containers in order of the ids from the representations
```
aklish, line 72: Is this behavior defined in the spec?  If the list is large, this could be expensive.

elide-graphql/src/main/java/com/yahoo/elide/graphql/federation/EntitiesDataFetcher.java
justin-tay: The alias isn't part of the representations argument. If different aliases are used in the same query it is called for as many times.

For instance for this query on the supergraph

```
query {
  a:groupReviews {
    id
    stars
    text
    group {
      name
      commonName
    }
  }
  b:groupReviews {
    id
    stars
    text
    group {
      name
      commonName
    }
  }
}
```

The following query gets run twice on the subgraph

```json
{
  "query": "query($representations:[_Any!]!){_entities(representations:$representations){...on Group{commonName}}}",
  "variables": {
    "representations": [
      {
        "__typename": "Group",
        "name": "com.yahoo.elide"
      },
      {
        "__typename": "Group",
        "name": "com.example.repository"
      },
      {
        "__typename": "Group",
        "name": "com.test"
      }
    ]
  }
}
```

elide-graphql/src/main/java/com/yahoo/elide/graphql/federation/EntitiesDataFetcher.java
justin-tay: In https://www.apollographql.com/docs/federation/subgraph-spec/#understanding-query_entities

> The Query._entities field must return a list of entity objects that correspond to the provided representations, in the exact same order. **Entries in the list can be null if no entity exists for a provided representation.**

elide-graphql/src/main/java/com/yahoo/elide/graphql/federation/EntitiesDataFetcher.java
justin-tay: In https://www.apollographql.com/docs/federation/subgraph-spec/#understanding-query_entities

> The Query._entities field **must return a list of entity objects that correspond to the provided representations, in the exact same order**. Entries in the list can be null if no entity exists for a provided representation.

PR #3129, opened 2023-12-14
Refactor table export

justin-tay:
Resolves #2978

## Description
- `ResultType` was changed from an `enum` to a `String` to allow custom result types
- `ResultTypeFileExtensionMapper` was added to allow customization of the mapping of the result type to the file extension
- `TableExportFormatter` was changed to become a factory to create a `ResourceWriter` to generate the result output. This allows state to be maintained during the generation within the `ResourceWriter` instance.
- The output is generated to an `OutputStream` instead of a `Observable<String>` in order to allow generation of binary output formats, for instance XLSX.
- The `ResourceStorageEngine` implementations were modified to store results given a `Consumer<OutputStream>` instead of an `Observable<String>` and the implementations changed to ensure that they could store binary data.
- A XLSX table export formatter was added.
- Fixed CSV generation of nested attributes. Json2Flat was removed as it cannot generate the correct output on a row-by-row basis.
- `TableExportFormattersBuilder` and `TableExportFormattersBuilderCustomizer` added to allow easier configuration of the `TableExportFormatters`.

## Motivation and Context
- Allows custom result types
- Allows binary output
- Allows easier configuration of additional table export formatters

## How Has This Been Tested?
Appropriate unit and integration tests were added

## License


aklish (approved):

aklish:
There are some failing tests.

justin-tay:
Fixed the tests. The line separator used for CSV is no longer system dependent and is fixed as `CRLF` in accordance to [RFC4180](https://datatracker.ietf.org/doc/html/rfc4180#section-2).

PR #3190, opened 2024-03-26
ELIDE-3132: Maintain hook ordering.

### LINKED ISSUE #3132: Lifecycle hook ordering is not deterministic in v7
I've a number of hooks attached to my entites for the same phase/operation. It allows us to split the business logic cleanly.
In v7.0.0pr6 we were seeing the ordering of hook invocations honoring the annotation order.....now it's not.

Based on the limited testing I've done - in EntityBinding fieldTriggers (and classTriggers) are HashSetValuedHashMap.....to my knowledge this doesn't preserve the insertion order.  

## Expected Behavior
The hooks should guarantee the execution order.

## Possible Solution
Could use an ArrayListValuedHashMap - it'll preserve the order.

It's very hit and miss to reproduce. If needed I imagine I could do it by faking the hashcodes.

## Context
The hooks are chaining business logic. The first one populates default values if not set. The later one does validation. If they happen out of order the validation fails.

## Your Environment
* Elide version used: 7.0.0
* Environment name and version: openjdk 17.0.6 2023-01-17 LTS
* Operating System and version: Ubuntu 22

bladedancer:
Resolves #3132

## Description
Ordering is not maintained when declaring lifecyclehook annotations. 

## Motivation and Context
I've a number of hooks attached to my entities for the same phase/operation. It allows us to split the business logic cleanly. The ordering of the hooks is non-deterministic (depends on hash order).

Based on the limited testing I've done - in EntityBinding fieldTriggers (and classTriggers) are HashSetValuedHashMap.....to my knowledge this doesn't preserve the insertion order.

## How Has This Been Tested?
Added unit tests.

## License


aklish (dismissed):

aklish:
There are checkstyle errors in the build.

bladedancer:
> There are checkstyle errors in the build.

Sorry about that - I never ran it and my defaults kicked in. Should be clean now. Let me know.

aklish (approved):

PR #3191, opened 2024-04-01
Entity scan on non-existing data model package silences error

Doom9527:
Resolves#3144

## Description
This PR primarily addresses the scenario where Elide is configured with a non-existent data model package. With the updated implementation, Elide now fails to start gracefully and throws a runtime exception when encountering an invalid data model package during scanning.
In the `combineModelEntities` method, checks are added to verify whether the data model package path exists and is indeed a directory; if the specified `modelPackageName` doesn't correspond to an existing directory, an exception will be thrown, informing the user that the data model package is missing.
Similarly, the `getAllEntities` method has also been augmented with a similar check to ensure the input package name points to an existing directory.

## Motivation and Context
This change resolves the issue where Elide would still start successfully despite scanning for a nonexistent data model package. By catching and reporting such errors at initialization, it significantly improves debugging efficiency and user experience by preventing users from seeing a seemingly happily running Elide with an empty database.

## How Has This Been Tested?
Please review my PR request and if possible, I will write the testing class again

## License


Doom9527:
After further research, I found that it would be better to add a check for non null judgments on the Set collection in the `getAnnotatedClasses` of the `DefaultClassScanner`, which can solve the problem at the source rather than making modifications in the Standalone project

PR #3197, opened 2024-04-04
Entity scan on non-existing data model package silences error

### LINKED ISSUE #3144: Entity scan on non-existing data model package silences error
## Expected Behavior

In order for quicker debugging, when Elide is scanning an non-existing data model package, Elide service is better with not starting successfully but through an Exception instead. 

## Current Behavior

When we configure Elide by binding a non-existing data model package, Elide started up successfully without any errors.

This leads to a confusing user who sees both a happy-running Elide but an empty database

## Possible Solution

I think it might be a simple check during scanning phase if Elide team merits such proposed change

Doom9527:
Resolves #3144

## Description
This PR primarily addresses the scenario where `Elide` is configured with a non-existent data model package. With the updated implementation, `Elide` now fails to start gracefully and throws a runtime exception when encountering an invalid data model package during scanning.

## Motivation and Context
In the `Standalone` or other projects which import `Elide`, the `getAnnotatedClasses` of the `DefaultClassScanner` under the `Elide` is called to scan the package. I add a judgment on the Set, and throw an exception if it is empty.

## How Has This Been Tested?
If a non-existent package name is provided, an exception is thrown as expected.

## License


.github/workflows/ci-cd.yml
```diff
@@ -0,0 +1,26 @@
+name: Elide CI/CD
```
aklish, line 1: Elide already has Screwdriver CI/CD for unit and integration tests which cost the company nothing.

screwdriver.yaml
aklish, line 1: We cannot delete this file.  There have also been hacking attempts of people putting scripts into our CI/CD so only committers can run PR CI/CD to avoid build credentials getting stolen.

aklish (changes requested):
Code looks good and will approve when screwdriver yaml is added back and github workflows (which are not free) are removed.

Doom9527:
Yes, I have added screwdriver yaml back and removed github workflows.

aklish (approved):

PR #3212, opened 2024-05-09
Migrate to networknt json schema validator implementation

justin-tay:
## Description
Migrates the json-schema-validator implementation from `com.github.java-json-tools` to the `com.networknt` implementation.

* Schemas were migrated from the Draft 4 dialect to the Draft 2020-12 dialect.
* The existing custom keywords and formats that can be implemented using the standard keywords and formats in the Draft 2020-12 dialect were removed. Only the custom `elideRSQLFilter` format was retained.

## Motivation and Context
The current json-schema-validator implementation from `com.github.java-json-tools` does not currently seem to be maintained with the last release in 2020.

## How Has This Been Tested?
Existing tests were modified as the validation messages generated for the different implementations are different.

## License


aklish:
We'll have to do a minor release for this given it breaks without a corresponding schema change.

aklish (approved):

PR #3255, opened 2024-07-15
Optimize required dependencies for modules

justin-tay:
## Description
The following are the summary of the main changes. These changes don't affect users using `elide-spring-boot-starter` as that generally pulls all the modules.

**elide-async**
* Sets `elide-graphql` optional to allow users that just wish to use JSON-API. Users that use GraphQL would have included the `elide-graphql` dependency explicitly anyway.

**elide-datastore-aggregation**
* Sets `elide-graphql` and `elide-datastore-multiplex` to test scope as they are only used during testing.

**elide-graphql**
* Sets `jakarta.websocket-api`, `jakarta.websocket-client-api` and `jakarta.jms-api` to optional for users that wish to use GraphQL but not the subscription support.

**elide-spring-boot-autoconfigure**
* Adds a `HttpServletRequestUser` to be used in place of `AuthenticationUser` to remove a required dependency on Spring Security. When Spring Security is used it will by default wrap the `HttpServletRequest` using the `org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestWrapper` which will return Spring's `Authentication` via `getUserPrincipal()` and delegate accordingly when `isUserInRole()` is called. This is the equivalent of what `AuthenticationUser` is doing. Using the `HttpServletRequest` also allows integration with the security provided by the container or custom schemes where a filter is used to wrap the `HttpServletRequest`.

## Motivation and Context
The existing `elide-spring-boot-starter` is very good for pulling most of the dependencies needed to try out all of Elide's features but when it comes to picking and choosing modules to be deployed in production it is not ideal to exclude too many dependencies.

These changes makes it easier to import `elide-bom` and then include the individual modules that won't pull in optional dependencies unnecessarily.

## How Has This Been Tested?
This was tested by modifying the `elide-spring-boot-example` and testing that the modules that should work are working and that the autoconfiguration appropriately backs off and doesn't throw exceptions when modules are not included.

## License

### DIFF
--- a/elide-async/pom.xml
+++ b/elide-async/pom.xml
@@ -46,8 +46,15 @@
     <dependencies>
         <dependency>
             <groupId>com.yahoo.elide</groupId>
+            <artifactId>elide-core</artifactId>
+            <version>7.1.1-SNAPSHOT</version>
+        </dependency>
+
+        <dependency>
+            <groupId>com.yahoo.elide</groupId>
             <artifactId>elide-graphql</artifactId>
             <version>7.1.1-SNAPSHOT</version>
+            <scope>optional</scope>
         </dependency>
 
         <dependency>
--- a/elide-core/src/main/java/com/yahoo/elide/core/security/User.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/security/User.java
@@ -5,25 +5,76 @@
  */
 package com.yahoo.elide.core.security;
 
-import lombok.Getter;
-
 import java.security.Principal;
+import java.util.Objects;
 
 /**
- * Wrapper for opaque user passed in every request.
+ * The user associated with the request.
  */
 public class User {
-    @Getter private final Principal principal;
+    private final Principal principal;
 
+    /**
+     * Constructor.
+     *
+     * @param principal the authenticated user
+     */
     public User(Principal principal) {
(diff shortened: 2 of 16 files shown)


aklish (dismissed):

justin-tay:
Have rebased this on master and tested it.

miks:
@justin-tay Is there any way to still use AuthenticationUser with this update?

justin-tay:
No it's not possible. Why do you specifically need `AuthenticationUser`? In general you should just be working with the `User` interface. If you need `Authentication` this can be obtained from `User.getPrincipal()` as before.

miks:
1. So, can the AuthenticationUser class actually be deleted from Elide?
2. Until updating to 7.1.1, I didn't even know that the AuthenticationUser class existed. Now, while debugging why my app broke after updating from 7.1.0 to 7.1.1 (which, according to semantic versioning, should have been a bugfix), I discovered that AuthenticationUser has a different role-checking mechanism than org.springframework.security.web.servletapi.SecurityContextHolderAwareRequestWrapper (which now adds a role prefix). Also, my custom export creation is now broken—securityContextHolderStrategy.getContext().getAuthentication() returns null when evaluating queries within the export context.

PR #3261, opened 2024-07-27
Add support for obfuscated id or alternate entity id

### LINKED ISSUE #522: Support for Encrypted ID's
We should add support for encrypted and/or obfuscated id's. This could be enabled through the builder and allow the users to provide an obfuscation/encryption mechanism. Similarly, we should provide a sane, efficient, default implementation.

justin-tay:
Resolves #522

## Description
This adds support for 2 different ways of not exposing the primary key of the entity if the primary key undesirably leaks information.
* Allows registering a `IdObfuscator` that can obfuscate the id. Typically an encryption algorithm is used.
* Allows designating another field or property as the `@EntityId` that will be used to look up the record instead of the `@Id`.

Two alternate means are provided to let users choose which best suits their needs. 

The `IdObfuscator` means that no additional column is required but some computation is required to always required to convert the ids. These ids will not be visible on the database level. Also sorting will typically take place on the underlying primary key and not the obfuscated value. If encryption is used for the obfuscation the user will need to safeguard those keys appropriately, also any changes to those keys or algorithm may mean that the ids will also correspondingly change.

Using the `@EntityId` allows the entity to have a totally separate column to identify the record that is not linked to the primary key in any way. A user could for instance choose Cuid2 as the entity id for its records. Sorting on id will sort on the entity id as this is a real column.

## Motivation and Context
The primary key may undesirably leak information, however at the same time using a random value as the primary key may affect performance.

## How Has This Been Tested?

## License

### DIFF
--- a/elide-core/src/test/java/com/yahoo/elide/core/security/obfuscation/AesBytesEncryptor.java
+++ b/elide-core/src/test/java/com/yahoo/elide/core/security/obfuscation/AesBytesEncryptor.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright 2024, the original author or authors.
+ * Licensed under the Apache License, Version 2.0
+ * See LICENSE file in project root for terms.
+ */
+package com.yahoo.elide.core.security.obfuscation;
+
+import java.security.spec.KeySpec;
+
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+import javax.crypto.SecretKeyFactory;
+import javax.crypto.spec.PBEKeySpec;
+import javax.crypto.spec.SecretKeySpec;
+
+public class AesBytesEncryptor {
+    private final SecretKey secretKey;
+
+    public AesBytesEncryptor(SecretKey secretKey) {
+        this.secretKey = secretKey;
+    }
+
+    public AesBytesEncryptor(String password, String salt) {
+        this(from(password, salt));
+    }
+
+    static SecretKey from(String password, String salt) {
+        try {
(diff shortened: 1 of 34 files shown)


elide-core/src/test/java/com/yahoo/elide/core/security/obfuscation/AesBytesEncryptor.java
```diff
@@ -0,0 +1,60 @@
...
+            return new SecretKeySpec(factory.generateSecret(spec).getEncoded(), "AES");
+        } catch (Exception e) {
+            throw new IllegalArgumentException(e);
+        }
+    }
+
+    byte[] encrypt(byte[] byteArray) {
+        try {
+            byte[] key = secretKey.getEncoded();
+            SecretKeySpec secretKeySpec = new SecretKeySpec(key, "AES");
```
aklish, line 40: Does this encryption include randomization or padding?  The ID should be stable so we should hash instead of encrypt.

justin-tay: This class is just for testing purposes. I felt it was best that users used a dedicated crypto library eg. https://docs.spring.io/spring-security/site/docs/current/api/org/springframework/security/crypto/encrypt/BytesEncryptor.html which would provide more options and this approach is less likely to introduce a CVE into the Elide library.

The FunctionIdObfuscator should make it easy to integrate with other libraries.

aklish (approved):

PR #3303, opened 2024-11-25
Add support for extension property of OpenAPI 3_0 (#3301)

### LINKED ISSUE #3301: Add support for extension property of OpenAPI 3_0
Currently, OpenAPI swagger documentation supports the definition of extension properties in the @Schema annotation, it is built like a Map<String, Object> and generally allows the OpenAPI user to define customs properties on the documentation. 

## Expected Behavior

We could use this feature to address which of our attributes is a localized attribute by simply defining this annotation:
```java
@Schema(extensions = {
            @Extension(properties = {
                    @ExtensionProperty(name = "localized", value = true)
            })
    })
```
Multiple extentionProperties can be defined in the same Schema. allowing us to enhance the OpenAPI doc further.
 It could also be useful to define the types of relationships in the OpenAPI documentation with something like this:
```java
@Schema(extensions = {
            @Extension(properties = {
                    @ExtensionProperty(name = "relationType", value = "oneToOne")
            })
    })
```
the output of the schema above for a normal attribute will be something like this:
```json
"properties": {
    "name": {
        "type": "string",
         "readOnly": false,
         "writeOnly": false,
         "localized": true,
         "example": null
    },
   ...
}
```

## Current Behavior

Currently this feature is not supported on Elide meaning that the openAPI doc ignores this property inside the schema annotation

## Possible Solution

A possible solution that can be done is to add inside the JsonApiModelResolver.class a method like this:
```java
    private Map<String, Object> getExtensions(Type<?> clazz, String fieldName) {
        io.swagger.v3.oas.annotations.media.Schema property = getSchema(clazz, fieldName);
        return property == null ? Map.of()
                : Arrays.stream(property.extensions())
                        .flatMap(extension -> Arrays.stream(extension.properties()))
                        .collect(Collectors.toMap(ExtensionProperty::name, ExtensionProperty::value));
}
```

Then call it inside the processAttribute and  processRelationship method by calling the Schema<?>.setExtensions method like this:
```java
attribute.setExtensions(getExtensions(clazzType, attributeName));
```

## Context

The reason I'm opening this feature is that I need to send more information to the FrontEnd point of my applications through OpenAPI to handle various situations, such as localized properties and others. By adding the extension property we are able to send all the necessary information and many more, leading to a more customizable experience for OpenAPI definition.

AleCamerini:
Resolves #3301

## Description

Currently, OpenAPI swagger documentation supports the definition of extension properties in the @Schema annotation, it is built like a Map<String, Object> and generally allows the OpenAPI user to define customs properties on the documentation.

## Motivation and Context

The reason I'm opening this feature is that I need to send more information to the FrontEnd point of my applications through OpenAPI to handle various situations, such as localized properties and others. By adding the extension property we are able to send all the necessary information and many more, leading to a more customizable experience for OpenAPI definition.

## How Has This Been Tested?

Aside from testing these changes by building some unit tests, I also applied the changes to my project and checked the openAPI documentation.
## License

### DIFF
--- a/elide-swagger/src/main/java/com/yahoo/elide/swagger/converter/JsonApiModelResolver.java
+++ b/elide-swagger/src/main/java/com/yahoo/elide/swagger/converter/JsonApiModelResolver.java
@@ -25,6 +25,7 @@ import io.swagger.v3.core.converter.ModelConverter;
 import io.swagger.v3.core.converter.ModelConverterContext;
 import io.swagger.v3.core.jackson.ModelResolver;
 import io.swagger.v3.core.util.Json;
+import io.swagger.v3.oas.annotations.extensions.ExtensionProperty;
 import io.swagger.v3.oas.annotations.media.Schema.AccessMode;
 import io.swagger.v3.oas.annotations.media.Schema.RequiredMode;
 import io.swagger.v3.oas.models.media.Schema;
@@ -35,6 +36,7 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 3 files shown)


aklish (approved):

aklish:
Nice work.  Running the CI build now.

PR #3313, opened 2024-12-07
Upgrade Spring Boot from 3.3.3 to 3.4.0

justin-tay:
Resolves #3305 

## Description
Upgrades Spring Boot from 3.3.3 to 3.4.0. Also upgrades HttpClient5 which was causing the test failure.

## How Has This Been Tested?

## License


aklish (approved):

bladedancer:
This probably won't be actionable but it may be of interest. I tried an upgrade to 6.4 using 7.1.4 and it seemed ok but we hit a subtle problem with atomic operations.
I was seeing an issue that was triggered by removeOrphans check. It was failing with a transient object exception. 

Plan to try this again in the new year and if reproducible I'll share additional detail.

PR #3397, opened 2025-11-28
Include number, temporal, date and enum in OpenAPI sortable fields

### LINKED ISSUE #3396: Some fields don't show up in sort of Operations in Swagger
When I use Elide in Spring Boot and use types like Integer, BigDecimal, OffsetDateTime, ZonedDateTime, LocalDateTime, LocalDate, LocalTime, Enum as String and some others, they don't show up in Swagger UI in sort array of the operation, although sorting these fields is possible. The issue is only with visualization.

justin-tay:
Resolves #3396 

## Description
Updated `OpenApiBuilder.getSortParameter()` to consider attributes that are assignable from `Number`, `Temporal`, `Enum`, and `Date` as sortable.

## Motivation and Context
Fixes the issue where sortable parameters are not in the OpenAPI document.

## How Has This Been Tested?
Added test case to ensure that the OpenAPI document generated now list such parameters as sortable.

## License

### DIFF
--- a/elide-core/src/main/java/com/yahoo/elide/core/type/ClassType.java
+++ b/elide-core/src/main/java/com/yahoo/elide/core/type/ClassType.java
@@ -13,6 +13,7 @@ import lombok.Getter;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
 import java.math.BigDecimal;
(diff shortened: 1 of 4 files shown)


Sovereignsun:
In what version will the changes be present? What's the nearest version?

